PR #3, opened 2026-05-24
chore: beta-readiness cleanup

antoinedc:
Prep work to ship bui to external beta testers who bring their own remote tmux server. Audit + decisions + implementation in one pass — see commit body for full breakdown.

## Highlights

**Blocker fixes**
- `src/renderer/api/httpApi.ts:29` — replace hardcoded \`157.90.224.92:8787\` fallback with fail-fast throw. The old fallback would have silently pointed a stranger's web/mobile client at the dev box.
- `src/main/index.ts:480` — gate screenshot detector on \`process.platform === \"darwin\"\`. The 500ms clipboard poller + \`fs.watch ~/Desktop\` were Mac-only by design but ran on every platform.

**New setup wizard** (replaces all the manual SSH/tmux/opencode onboarding steps)
- \`src/main/setup.ts\` — pure \`parseProbeOutput()\` + remote shell scripts. \`PROBE_SCRIPT\` checks ssh/tmux/opencode/auth-plugin/anthropic-creds in one ssh roundtrip. \`BOOTSTRAP_SCRIPT\` runs opencode's official installer and writes a minimal \`opencode.jsonc\` with the claude-auth plugin. Idempotent. Does NOT run \`opencode auth login\` (needs browser).
- \`src/main/setup.test.ts\` — 9 unit tests on the pure parts: missing keys, junk lines, \`|\` inside detail strings, unknown statuses, script invariants (every key emitted, login never actually invoked but referenced as next step).
- \`IPC.setupProbe\` + \`IPC.setupBootstrap\` wired through main → preload → httpApi. Mobile server returns stubs (not a mobile feature).
- \`src/renderer/Settings.tsx\` — **Test connection** + **Bootstrap remote** buttons with a 5-check status grid and a log pane. Also surfaces the previously-hidden **Set up tmux config** button and fixes misleading copy that implied it ran automatically.

**Cleanup**
- Delete personal dev-state files: \`HANDOFF*.md\` (3 files), \`docs/superpowers/\` (13 design docs with personal device IDs etc.), \`.planning/\`.
- \`.gitignore\` now excludes \`.multica/\` (private workspace/agent/skill IDs), \`HANDOFF*.md\`, \`.planning/\`.
- \`git rm --cached\` the tracked \`.multica/\` files.

**Docs**
- Rewrite \`README.md\` from a stranger's POV: real prerequisites (Mac + remote tmux + optional opencode), the new wizard flow, both window types (terminal vs chat), known beta gaps, mobile descoped.

## What's deliberately NOT here
- electron-builder / \`.dmg\` packaging — testers run \`npm run dev\`.
- Mobile bundle ID rename / Cloudflare tunnel setup — mobile descoped from this beta.

## Verification
- \`npm run typecheck\` — clean
- \`npm test\` — 263 passed (vitest)
- \`npm run test:server\` — 41 passed (node:test)


antoinedc:
# Code Review (self-review via claude-review skill)

Read-only deep review of `git diff main...chore/beta-readiness`. Typecheck and `setup.test.ts` both pass locally. Scope: substantive code changes; doc/handoff deletions skimmed only for collateral damage.

## CRITICAL

None. No security holes, no broken desktop flows, no data-loss paths.

## WARNING

### 1. `src/renderer/Settings.tsx` — Bootstrap log flashes then disappears

`runBootstrap` sets `bootstrapResult` then immediately `await runProbe()`. The first line of `runProbe` is `setBootstrapResult(null)`. So the user sees the bootstrap log for ~1 React commit before the re-probe wipes it. The whole point of the log pane is to read what bootstrap did. **Fix:** remove `setBootstrapResult(null)` from `runProbe`, or move that clear to the button's onClick.

### 2. `src/renderer/api/httpApi.ts:32-35` — `serverBase()` throw crashes mobile/web mount

The throw replaces a hardcoded fallback, which is correct intent — but `serverBase()` is called **synchronously** inside `wsUrl()` → `new WebSocket(...)` inside `openStream()` → `ensureStream()` → `on()`. `MobileApp.tsx` registers `onStatusEvent`/`onOpencodeEvent` from `useEffect` with no try/catch, so missing `localStorage["bui_server"]` (Capacitor `http://localhost`, `file:`) makes the effect throw synchronously → blank screen. README says mobile is descoped, but the file still ships and a tester poking the APK hits it. **Fix:** catch the throw in `openStream()` and surface via the existing `bootError` channel.

### 3. `src/renderer/Settings.tsx` — "Save first" is documentation, not enforcement

The body text says "Save your changes above first" but the buttons disable only on `!host`. A user who edits host/user/identity then clicks **Test connection** runs the probe against the **previously saved** values and gets confusing results. **Fix:** disable the buttons when form is dirty (`h !== host || u !== (user ?? "") || k !== (identityFile ?? "")`), or silently save first.

### 4. `src/main/setup.ts:201-217` — `bootstrap()` `ok` detection is fragile

`ok = !lines.some(l => l.startsWith("✗"))`. stderr is appended without a `✗` prefix, so stderr containing real errors won't flip `ok`. **Fix:** also check `stderr.trim() === ""` as part of `ok`, or have the script `exit N` on any failure.

### 5. `src/main/setup.ts` — `BOOTSTRAP_SCRIPT` clobbers existing `opencode.jsonc`

After bootstrap writes a fresh `opencode.jsonc`, if the user's previous one was hand-tuned (other plugins, MCP servers), the bootstrap clobbers it entirely with the 3-line minimal config — losing those settings — even though it backs up to `.pre-bui`. README and the confirm dialog don't warn about this. **Fix options:**
- merge into existing JSONC (pattern already in this codebase — see `configUpdate` JSONC merge for `skills.urls`)
- update `Settings.tsx` confirm() text to mention the overwrite (minimum acceptable fix)

### 6. `src/main/setup.ts:96-97, 155` — Auth probe is path-stuck

`anthropicAuth` only checks `~/.claude/.credentials.json`. opencode supports other auth methods (opencode's own `auth.json`, `ANTHROPIC_API_KEY` env, Bedrock, OpenRouter). A user already logged in via opencode but without `~/.claude/.credentials.json` will see a false negative. **Fix:** use `opencode auth list 2>/dev/null | grep -i anthropic` as the source of truth, credentials.json as fallback.

### 7. `src/server/rpc.mjs:78-94` — Mobile probe stub claims success

Mobile probe stub returns `allOk: true` with every check marked `ok: true|"n/a"`. If a user opens Settings on the mobile/web client, they'll see "✓ all green" even when their opencode is broken. **Fix:** flip to `allOk: false` with per-check details that say "n/a — desktop only", or hide the wizard UI on mobile entirely.

### 8. `src/main/setup.ts` — No timeout for bootstrap SSH call

The bootstrap pipes `curl … | bash`, which can sit on a slow network for minutes. `runSshOnce` has no timeout; if the SSH session wedges, the Settings UI spinner spins forever — no cancel. **Fix:** add a timeout option to `runSshOnce`, or wrap in `Promise.race` with a 120s cap and a clear error.

### 9. `src/renderer/Settings.tsx` — `alert()` UX in custom-titlebar Electron window

Blocks the window and looks out of place. Consistent with the pre-existing `restore` flow, so not a regression — but if "beta-ready" means polishing, this is a candidate.

### 10. `README.md` — Mosh claim slightly overstated

"Auto-detects mosh and falls back to plain SSH" implies mosh covers the whole connection. In reality only the long-lived attached PTY uses mosh; chat-mode side-channels always use plain SSH. Lower priority; could add "(used for the long-lived terminal PTY)".

## INFO

### 11. PROBE_SCRIPT portability — POSIX-clean ✓; bootstrap requires `bash`. Stock NixOS/Alpine may not have it. Minor.

### 12. `setup.test.ts` "doesn't run auth login" test — regex `/^\s*opencode\s+auth\s+login/` would miss `eval` or `bash -c` wrappers. Acceptable as regression-bait tripwire.

### 13. `parseProbeOutput` accepts duplicate keys (last wins). Trivial — user controls the remote.

### 14. `.gitignore` `HANDOFF*.md` pattern catches any future `HANDOFFNOTES.md`. `HANDOFF.md` + `HANDOFF-*.md*` would be tighter.

### 15. Comment grammar in `src/main/index.ts` setup handler comment ("is a TODO; for now…") — clean up or actually implement the disable from #3.

### 16. `setupTmux` error doesn't refetch state on failure. Same shape as `restore`. Future improvement.

### 17. PROBE_SCRIPT leading newline — fine for `ssh -c`, trivial.

---

## OVERALL: **Ship after WARNINGS addressed**

The setup wizard is well-scoped, well-tested (10/10 pass), and the pure/runner split is clean. No security holes — `runSshOnce` is called with constant script literals, not interpolated user input, so `curl | bash` is the only trust handoff and it's behind a confirm dialog. README is accurate against the code (keybindings, paths, tmux session names, port numbers all check out). Screenshot detector gating is correct.

**Must-fix before a tester runs the wizard:**
- #1 (bootstrap log invisibility — defeats the wizard's main feedback channel)
- #5 (bootstrap silently overwrites custom `opencode.jsonc`)
- #2 (mobile/web throws unhandled — even though descoped, the file ships and the throw white-screens)

**Should-fix:**
- #3 (probe runs against stale config when form is dirty)
- #7 (mobile probe stub lies "all green")
- #8 (no bootstrap timeout)

The rest are polish.

antoinedc:
## Review findings addressed (519daf7)

| # | Finding | Fix |
|---|---------|-----|
| 1 | Bootstrap log flashed then disappeared | Split `runProbe` (no clear) + `runProbeFromButton` (clears on user-initiated probe). Only the Test button wipes prior bootstrap log; the re-probe after bootstrap doesn't. |
| 2 | `serverBase()` throw white-screened mobile/web | `openStream()` now catches, `console.warn`s once, and bails. `refresh()`'s `bootError` still surfaces the user-facing message. |
| 3 | Wizard ran against stale config when form was dirty | New `connectionDirty` boolean compares form inputs vs. saved values. Test/Bootstrap buttons disabled while dirty; amber hint above them tells the user to save. |
| 5 | Bootstrap clobbered hand-tuned `opencode.jsonc` | Refactored: shell `INSTALL_SCRIPT` (binary install only) + Node-side `mergeOpencodeJsonc()` (pure, 10 unit tests). Deep-merges only the `plugin` array, preserves all other keys. No-op if auth plugin already present (respects pinned versions). Unparseable input → minimal config + `.pre-bui` backup. Confirm dialog rewritten to describe the merge behavior. |
| 7 | Mobile probe stub falsely claimed "all green" | Stub now returns `allOk: false` with per-check `"n/a — desktop-only wizard"` details. |
| 8 | Bootstrap could spin forever on a wedged SSH | `runSshWithTimeout`: 120s for install (curl\|bash can be slow), 30s for config write / auth check. Clear error message on timeout. |

## Stripper bug caught by failing test

Implementing #5 surfaced a real bug: the naive `replace(/\/\/[^\n]*/g, "")` comment stripper used elsewhere in this codebase (for `skills.urls` merging in `main/index.ts`) is **broken** — it strips everything after `//` inside any `"https://..."` string. Replaced with a tiny string-aware tokenizer (`stripLineComments`, 6 dedicated tests). `configUpdate` in `main/index.ts` has the same bug today — out of scope for this PR but worth a follow-up.

## Polish left for follow-up

- **#9** `alert()` UX inconsistent with custom titlebar — pre-existing pattern, consistent with `restore` flow.
- **#10** README mosh nuance — only the long-lived terminal PTY uses mosh; chat-mode side-channels are plain SSH.
- **#11–17** All INFO-level (portability nits, comment grammar, etc.).
- Also surfaces: the same broken JSONC stripper in `src/main/index.ts:648` (skills.urls merger). Should re-use the new `stripLineComments`.

## Verification

- `npm run typecheck` — clean
- `npm test` — **281 passed** (was 263; +18 new tests)
- `npm run test:server` — **41 passed**

PR #5, opened 2026-05-26
fix(chat): keep ContextBar visible during streaming + latch question/permission attention

antoinedc:
Three bugs reported, two root causes.

## 1. ContextBar disappears sometimes; never shows while streaming

**Root cause:** `latestTokens` walked the transcript from the tail and returned the first assistant message it found — but a freshly-streaming assistant message has `tokens` absent or all-zeros until the first `session.next.step.ended` event arrives. Result: `ctxTokens === 0`, ContextBar hidden for the entire streaming turn (sometimes minutes when a long tool roundtrip delays the step boundary).

**Fix (`src/renderer/ChatPanel.tsx`):** skip empty entries and keep walking back to the prior turn's tokens so the bar shows the last-known-good value while streaming. `stepTokens` (set on `session.next.step.ended`) is still preferred when present, same as before.

## 2. Question indicator often doesn't trigger

**Root cause:** `setChatAttention` gated all attention kinds on `!isActiveHere`. When a question fired while the user was on the chat panel, no attention was latched. The user then navigated away mid-turn — no later event re-latched it, so the sidebar dot never appeared.

**Fix (`src/renderer/store.ts`):** latch `question` and `permission` unconditionally — these block the turn and the user MUST act, so the indicator needs to persist across window switches. `setActive` still auto-clears on focus, so the dot is redundant-but-harmless while viewing the card. `idle` (soft 'go check' signal) keeps the `!isActiveHere` gate — nothing to go check if they're already there.

## 3. Permission indicator should mirror the question one

Already did in `Sidebar.tsx:806` — both `question` and `permission` render a red pulsing dot, with distinct glyphs (`?` vs `!`) so the user can tell what's blocking without opening the window. The root cause for permissions not appearing was the same bug as (2).

## Verification

- `npm run typecheck` ✅
- `npm test` ✅ (327/327 passed; updated store.test.ts to lock in the new latch-while-active behavior for both blocking kinds + keep the negative case for `idle`)

### DIFF
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -2988,6 +2988,15 @@ export function ChatPanel({ sessionId, tmuxSession, windowIndex, cwd, isActive }
   // session.next.step.ended (item 2) feeds stepTokens on every step boundary
   // and we prefer it here so the footer reflects the latest snapshot without
   // waiting for a re-fetch cycle.
+  //
+  // **GOTCHA — fall through "empty" tokens.** A freshly-streaming assistant
+  // message has `tokens` either absent or all-zeros until the first step
+  // boundary lands. The naive "first assistant from the tail" loop returned
+  // that empty object, which made `ctxTokens === 0` and hid the ContextBar
+  // for the entire streaming turn — the bar only re-appeared after the
+  // step.ended event arrived (sometimes minutes later, after a long tool
+  // call). Skip empty entries and keep walking back to the PRIOR turn's
+  // tokens so the bar shows the last known good value during streaming.
   const latestTokens = useMemo<TokenUsage | null>(() => {
     if (stepTokens) {
       return {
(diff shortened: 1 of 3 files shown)


antoinedc:
## Self-review

Walked the diff line-by-line against scenarios and edge cases.

### `latestTokens` fallback (`ChatPanel.tsx`)

**Correctness checks:**
- ✅ `stepTokens` short-circuit still works — when present, the bar uses live data, fallback never runs.
- ✅ Skipping all-zero entries is safe: any real LLM call has `input > 0` (it includes the user's message tokens). All-zero only happens for uninitialized assistant placeholders.
- ✅ `total` field of TokenUsage is unused here — only `input + cache.read + cache.write` matters because that's the context-window numerator (matches `computeContextBreakdown`).
- ✅ Brand-new session with no prior turns → still returns null → bar stays hidden (correct).

**Side effects on RunningIndicator** (also consumes `latestTokens`):
- During the first streaming turn of a new session, RunningIndicator's `↓ N` shows nothing (latestTokens=null). Same as before. ✅
- During turn 2+ before step.ended fires, RunningIndicator briefly shows the PREVIOUS turn's output count. This was already happening via `stepTokens` persistence (cleared only on session change) — my change just makes the `messages` fallback consistent. Not a regression. ✅

**Perf:** O(n) walk in a useMemo, only re-runs on messages/stepTokens change. Negligible.

### `setChatAttention` latch (`store.ts`)

**Scenario coverage:**
| User location | Event | Before | After |
|---|---|---|---|
| On chat panel | question.asked | ❌ no latch (bug) | ✅ latch, dot visible |
| On other window | question.asked | ✅ latch | ✅ latch (unchanged) |
| On chat panel | question.replied | ✅ no-op | ✅ no-op |
| On chat panel | idle (running→idle) | ✅ no latch | ✅ no latch (unchanged) |
| On other window | idle | ✅ latch | ✅ latch (unchanged) |

**Interaction with `setChatRunning`:** preserves the more-urgent `question`/`permission` kind across running→idle transitions (existing line 291-296). ✅

**Pre-existing quirk (not introduced here):** if user navigates back to chat while a question is still pending, `setActive` clears the sidebar dot but the QuestionCard is still blocking. Same behavior as before this PR — they have to scroll to see the card. Out of scope.

**Cosmetic redundancy:** when user is on chat panel and a question fires, the red dot appears next to their currently-active row. Slightly redundant but informative (same urgency reinforced). Auto-clears on next `setActive()` touch.

### Permission indicator (Sidebar)

Confirmed `StatusIndicator` at `Sidebar.tsx:806` already handles `permission` identically to `question` (red pulsing dot, distinct `!` glyph). The bug was purely in the latching logic upstream — same root cause as (2).

### Test verification

```
Test Files  8 passed (8)
     Tests  327 passed (327)
```

Including 3 new specs in store.test.ts that lock in:
- `latches question attention EVEN when the user is on the window`
- `latches permission attention EVEN when the user is on the window`
- `does NOT set 'idle' attention when the user IS on the window`

### Risks

- **Minor visual redundancy** while user is actively viewing the chat panel and a question fires. Self-resolves on focus or reply. Trade-off: prevents the much-worse 'navigate away with pending question, no indicator' bug.
- **No new perf or memory concerns.**

LGTM for merge.

PR #6, opened 2026-05-26
feat(chat): replace desktop mic button with ctrl+m + pulsing input ring

antoinedc:
## Why

The inline mic glyph next to the `>` prompt never aligned cleanly and on some font setups forced the input row to two lines tall. Two iterations on the visual didn't fix it. Replace with a keyboard-driven flow on desktop; keep the touch button on mobile (no physical keyboard there).

Two commits:
1. `78669b4` — interim attempt to make the mic an inline glyph left of `>` (kept for atomic history; superseded by 2)
2. `d99701a` — the real fix: kill the desktop button, add Ctrl+M + pulsing input ring + Enter-to-send

## Desktop UX

| Keystroke | Effect |
|---|---|
| `Ctrl+M` | Toggle: start recording, OR stop + transcribe + append (no submit, user edits) |
| `Enter` while recording | Stop + transcribe + append + **submit**, single keystroke |
| `Esc` while recording | Cancel (discards audio, no transcribe) |

Visual feedback while voice is active:
- Top + bottom dividers of the input row pulse red with a soft glow box-shadow
- The `>` prompt glyph recolors red
- Footer hint swaps to `🎙 recording · ⏎ send · ctrl+m stop · esc cancel`
- Border width stays 1px in both states so the row doesn't jump on toggle

## Mobile

MicButton still renders, gated on `closest('.mobile-body')` detected via a ref + useState on mount. Desktop never sees the button.

## Implementation notes

- Window-level keydown handler uses `capture: true` so Enter/Esc preempt the textarea's bubble-phase `onKeyDown` without breaking the normal submit/abort paths when voice is idle.
- `submitAfterTranscribeRef` threads the "submit when transcribe lands" intent through the async Groq round-trip (~200-500ms). `onResult` consumes the flag and fires `submitRef.current()` on the next tick (after `setInput` commits).
- Refs (`voicePhaseRef`, `voiceStartRef`, etc.) keep the keydown handler stable across recorder re-renders — no re-subscription on every render.
- Mobile/desktop split via `closest('.mobile-body')` matches the existing pattern at `ChatPanel.tsx:1732` (auto-focus skip).

## Verification

- `npm run typecheck` ✅
- `npm test` ✅ (327/327)
- Settings copy updated for the new desktop keybinds. Mobile Settings copy unchanged (still accurate for touch).

### DIFF
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -2419,18 +2419,34 @@ export function ChatPanel({ sessionId, tmuxSession, windowIndex, cwd, isActive }
     ],
   );
 
+  // When the user presses Enter (or Ctrl+M) WHILE the desktop voice
+  // recorder is active, we want the transcribed text to land in the
+  // composer AND immediately submit, in one keystroke. The transcribe call
+  // is async (Groq round-trip ~200-500ms), so we set a one-shot flag that
+  // `onResult` consumes and then auto-submits. Esc just cancels.
+  const submitAfterTranscribeRef = useRef(false);
+
   const voiceRecorder = useVoiceRecorder({
     onResult: (r) => {
       if (r.mode === "dictate") {
         dispatchVoiceAction({ kind: "append", text: r.text });
+        if (submitAfterTranscribeRef.current) {
+          submitAfterTranscribeRef.current = false;
+          // One-tick delay so the setInput inside the append branch has
+          // committed before submit() reads the textarea value.
+          setTimeout(() => submitRef.current(), 0);
+        }
       } else {
         dispatchVoiceAction(r.classify.action);
       }
     },
-    onError: (e) => setSendError(e.message),
+    onError: (e) => {
+      submitAfterTranscribeRef.current = false;
+      setSendError(e.message);
+    },
   });
 
-  // Gate the mic button on: API key present, browser capable of capture.
+  // Gate the mic affordances on: API key present, browser capable of capture.
   // Mobile WebView typically needs RECORD_AUDIO granted at the OS layer —
   // we don't pre-check; the first start() surfaces "permission denied" via
   // setSendError if the user said no.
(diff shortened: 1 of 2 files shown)


antoinedc:
## Self-review

Walked every scenario for the keyboard handler + visual states. Found and fixed one HIGH-severity bug; everything else clean.

### 🐛 Found in review (fixed in `bf864cd`)

**Enter during 'requesting' phase would silently hang the recorder for 60s.**

Scenario: user hits Ctrl+M, mic permission prompt fires (phase='requesting'), user hits Enter before granting.
- My initial handler called `voiceStopRef.current()` for any `active` phase (requesting OR recording).
- `stop()` is a no-op when `recorderRef.current` is null (still awaiting getUserMedia).
- The W3 `cancelledRef` check in the hook only fires for `cancel()`, not `stop()`.
- Result: getUserMedia eventually resolves → recorder is constructed and starts → records until `maxDurationMs` (60s) with no UI escape.

**Fix:** Enter only triggers when `phase === "recording"`. During 'requesting'/'processing' it falls through to the textarea (so the user can still send whatever's typed). Esc still works in any non-idle phase because `cancel()` correctly handles 'requesting' via the existing W3 check.

### Scenarios walked

| State | Ctrl+M | Enter | Esc |
|---|---|---|---|
| idle | start recording (dictate) | falls through (submit textarea) | falls through (abort if running) |
| requesting | toggle off (stop) | **falls through** (was: bug) | cancel |
| recording | toggle off (stop, no submit) | stop + transcribe + submit | cancel |
| processing | no-op | falls through | cancel (suppresses auto-submit) |
| error | restart | falls through | falls through |

### Other checks

- **capture:true + native event** — runs before React's bubble-phase synthetic. `e.stopPropagation()` on the native event prevents React's root listener from dispatching to the textarea's onKeyDown. ✅
- **Refs stable across re-renders** — voicePhaseRef/voiceStartRef/etc. assigned on every render but the useEffect deps only watch voiceEnabled. No re-subscription churn. ✅
- **Voice + running interaction** — Esc cancels voice first; user presses Esc again to abort a running turn. Voice takes priority since it's actively recording; acceptable. ✅
- **isMobileShell timing** — one-render flicker on mobile where button is absent before mount effect runs. Imperceptible. ✅
- **MicButton on desktop** — fully hidden (voiceEnabled && isMobileShell). The 'inline-glyph' iteration from commit 78669b4 is now only used by mobile, where its 13px sizing aligns fine with the larger mobile textarea. ✅

### Cosmetic notes (not blocking)

- Two atomic commits (78669b4 + d99701a + bf864cd review fix = 3 total). 78669b4 is superseded by d99701a on desktop but the MicButton component itself still uses its inline-glyph styling for mobile, so the changes ARE used. Worth keeping for atomic history.
- Border pulse uses a hardcoded `shadow-[0_0_6px_rgba(239,68,68,0.6)]` — could be a tailwind `shadow-red` if such a token existed in the theme. Minor.

### Verification

```
npm run typecheck  ✅
npm test           ✅ 327/327
```

LGTM after the requesting-phase fix. Ready to merge.

PR #7, opened 2026-05-26
fix(tmux): keep project sessions alive + auto-heal when missing

antoinedc:
## Problem

New users hit an opaque failure on what should be a happy path: creating
a session inside a project, or restarting bui after closing it.

- Sidebar: `alert("ssh exited 1: can't find session: <project>")`
- Active terminal: `[mosh is exiting.]` followed by `[disconnected from <project>: 0]`

Root cause: tmux's defaults are tuned for an interactive user who never
detaches. For a desktop client that disconnects/reconnects all the time:

- `destroy-unattached on` (the default) tears the session down the moment
  the last client detaches — closing bui, switching wifi with mosh.
- `exit-empty on` (the default) lets the tmux server self-terminate when
  no sessions exist, so the next bui call cold-starts the server and any
  pending mosh-server attachments lose their target.

Either kills `new window in <project>` with no recovery path other than
recreating the project from scratch.

## Fix

Two-pronged, applied symmetrically to both transports (`src/main/pty.ts`
for desktop/SSH and `src/server/tmux.mjs` for mobile/local).

### 1. Pin survivability per-bui-session
At session creation, immediately apply:
- `set-option -t <name> exit-empty off`
- `set-option -t <name> destroy-unattached off`

Scoped to bui-created sessions only — never `-g`. User-owned tmux
sessions outside bui keep tmux's defaults. Honors AGENTS.md's *"tmux
config approach — drop-in, no surprises"* rule.

### 2. Auto-heal `new-window` when the session is missing
If the project's session vanishes anyway (server restart, manual kill,
race with old clients that pre-date this fix), catch the rejection,
classify it via the new pure `isMissingSessionError()` helper, and
transparently recreate the session with this window as its first.

The classifier is tolerant to:
- The straight ASCII apostrophe \`can't find session: X\`
- Case variations
- The alternate \`session not found: X\` phrasing

Non-matching errors still throw unchanged — won't mask auth failures,
host-unreachable, etc.

## Tests

- \`src/main/pty.test.ts\` — 7 vitest cases for the desktop classifier
- \`src/server/tmux.test.mjs\` — 5 \`node:test\` cases for the mobile classifier

Pure logic only — no ssh / tmux spawning required. Covers canonical
form, apostrophe variants, case-insensitivity, alternate phrasing,
unrelated tmux errors, ssh-level errors, and non-Error inputs.

## Verification

\`\`\`
npm run typecheck   # clean
npm test            # vitest 9 files / 334 pass, node:test 46 pass
\`\`\`

## What this does NOT change

- No edits to user \`~/.tmux.conf\` (still respects AGENTS.md's locked
  decision)
- No changes to mouse mode (still ON through the whole pipeline)
- No changes to the underlying transport (mosh-if-available / ssh
  fallback)
- Existing tmux sessions created before this PR aren't retroactively
  hardened — but the auto-heal in (2) covers them on the next
  new-window attempt

### DIFF
--- a/src/main/pty.test.ts
+++ b/src/main/pty.test.ts
@@ -1,0 +1,63 @@
+import { describe, it, expect } from "vitest";
+import { isMissingSessionError } from "./pty.js";
+
+// `tmuxNewWindow` auto-heals by recreating the project's tmux session when
+// the user's previous session was destroyed (server restart, manual kill,
+// destroy-unattached racing the next call). The classifier below is what
+// gates that branch — false positives just trigger a redundant new-session
+// call, false negatives leak `ssh exited 1: ...` to the user's alert().
+
+describe("isMissingSessionError", () => {
+  it("matches tmux's canonical 'can't find session: X' stderr", () => {
+    const err = new Error("ssh exited 1: can't find session: asdfg");
+    expect(isMissingSessionError(err, "asdfg")).toBe(true);
+  });
+
+  it("matches with a straight ASCII apostrophe", () => {
+    const err = new Error("ssh exited 1: can't find session: my-project");
+    expect(isMissingSessionError(err, "my-project")).toBe(true);
+  });
+
+  it("matches case-insensitively for resilience against tmux locale changes", () => {
+    const err = new Error("Can't Find Session: foo");
+    expect(isMissingSessionError(err, "foo")).toBe(true);
+  });
+
+  it("matches the alternate 'session not found: X' phrasing", () => {
+    const err = new Error("ssh exited 1: session not found: nw");
+    expect(isMissingSessionError(err, "nw")).toBe(true);
+  });
+
+  it("returns false for unrelated tmux errors", () => {
+    expect(
+      isMissingSessionError(new Error("ssh exited 1: no server running"), "x"),
+    ).toBe(false);
+    expect(
+      isMissingSessionError(
+        new Error("ssh exited 1: duplicate session: x"),
+        "x",
+      ),
+    ).toBe(false);
+  });
+
+  it("returns false for ssh-level failures (host unreachable, auth, etc.)", () => {
+    expect(
+      isMissingSessionError(
+        new Error("ssh exited 255: Connection timed out"),
+        "x",
+      ),
+    ).toBe(false);
+    expect(
+      isMissingSessionError(
+        new Error("ssh exited 255: Permission denied (publickey)"),
+        "x",
+      ),
+    ).toBe(false);
+  });
+
+  it("returns false for non-Error values (defensive — promise rejected with a string)", () => {
+    expect(isMissingSessionError("can't find session: x", "x")).toBe(false);
+    expect(isMissingSessionError(null, "x")).toBe(false);
+    expect(isMissingSessionError(undefined, "x")).toBe(false);
+  });
+});
--- a/src/main/pty.ts
+++ b/src/main/pty.ts
@@ -246,17 +246,22 @@ async function maybeCreateChatSession(
 // Session-level survivability options. Applied to bui-created sessions only —
 // never `-g`, so user-created tmux sessions outside bui keep tmux's defaults.
 //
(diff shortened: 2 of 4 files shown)


antoinedc:
## Code review (self)

Read every changed line. Tests verified: vitest 334 pass, node:test 46 pass.

### Critical — diagnosis correction (does NOT change the fix's correctness, but the PR description is partly wrong)

**`destroy-unattached` defaults to `off`, not `on`.** I claimed in the commit and PR body that the default is `on`. The tmux man page says otherwise:

> `destroy-unattached [off | on | keep-last | keep-group]`
> If on, destroy the session after the last client has detached. **If off (the default)**, leave the session orphaned.

So setting `destroy-unattached off` in `sessionSurvivabilityCmd` is a **no-op** in practice — tmux already keeps sessions orphaned on detach.

The **actual teeth** of the fix is `exit-empty off`. The likely failure flow for the new user was:

1. Project `asdfg` created → tmux session has 1 window.
2. That window's process exits (claude TUI killed by user, or it crashed and the bash fallback was also exited). Default `remain-on-exit off` closes the window. With no remaining windows the session dies.
3. With `exit-empty on` (default), the now-empty server self-terminates.
4. Next bui call: `tmux new-window -t asdfg:{end}` → `can't find session: asdfg`.

`exit-empty off` blocks (3): the server survives the empty-session moment, so a subsequent retry of step (4) succeeds against a still-alive server (though the session itself is still gone, which is what the auto-heal branch covers).

`destroy-unattached off` is harmless defensive insurance against a future tmux default flip or a user `~/.tmux.conf` that sets it ON globally — bui's per-session override would then save them.

**Action: I'll edit the comment block in `pty.ts:sessionSurvivabilityCmd` on a follow-up to correct the rationale**, or fix it now before merge. Doesn't change the bytes shipped, but the docstring is misleading.

### Warning — `cwd ?? "."` change in semantics

`src/server/tmux.mjs:newWindow` heal path:
```js
await run("tmux", ["new-session", "-d", "-s", sessionName, "-n", windowName,
  ...(cwd ? ["-c", cwd] : [])]);
```

In the original `newWindow`, when `cwd` was falsy, tmux's `new-window -t session` inherits the session's `default-cwd`. In the heal path the session doesn't exist anymore, so `new-session` with no `-c` uses the **server process's** cwd (probably `$HOME` on the Linux box). This is a reasonable fallback — the original session's default-cwd is unrecoverable — but it's a subtle behavior change worth flagging.

In practice: the desktop IPC handler at `src/main/index.ts:735` always resolves `cwd` to `project.defaultCwd || "~"` before calling `tmuxNewWindow`, so the desktop transport never hits this edge. The mobile RPC path (`src/server/rpc.mjs:tmux:new-window`) does similar resolution. So this is theoretical, not a real regression.

### Info — regex liberality

`/can.?t find session/i` matches "can't", "cant", "canst", "canat", etc. Possible false positives:
- `can't find session in cache` — would heal needlessly, but the heal is idempotent and cheap.
- Any future tmux error string containing "cant find session" as a substring.

Net cost of a false positive: one extra `tmux new-session` call. Acceptable trade-off vs missing a real variant.

### Info — the `bui-opencode` session

The opencode-server tmux session in `src/main/opencode.ts:204` doesn't go through `tmuxNewSession`, so it doesn't get the survivability options. That's fine for two reasons:
1. It's created/destroyed deterministically by `ensureRunning` — bui checks `tmux has-session` first.
2. Stale-detection in `ensureRunning` kills + recreates it as needed.

So no action needed for that path.

### Info — no test for the auto-heal *behavior*

The pure classifier is well-tested. The actual `tmuxNewWindow` heal branch (catch → recreate session) is verified by hand only. Adding a test would require mocking `runSshOnce` and threading the mock through the `cpSpawn`-based implementation; not worth it for this slice. The classifier owns the gate, and the heal command is shape-identical to `tmuxNewSession`'s command which is exercised by integration via real usage.

### Verdict

Ship it, with a follow-up to fix the misleading rationale comment about `destroy-unattached`. The actual code does the right thing — the docstring just oversells what `destroy-unattached off` is buying us.

PR #36, opened 2026-07-03
BET-65: pool forwardFetch connections — stop loopback ephemeral-port exhaustion (EADDRNOTAVAIL)

antoinedc:
## BET-65 — pool loopback fetch connections (stop EADDRNOTAVAIL)

Fixes recurring loopback ephemeral-port exhaustion that required a Mac reboot to
clear (~37.5k sockets in TIME_WAIT to `127.0.0.1`, `EADDRNOTAVAIL`, `fetch failed`
on reconcile, `ERR_ADDRESS_INVALID` on unrelated hosts).

**Root cause:** Node's global `fetch` (undici) opens a fresh TCP socket to
`127.0.0.1:<port>` per call and closes it when the body is consumed — each
lands in TIME_WAIT. A `reconcileMessages` sweep across many sessions
burst-created tens of thousands of one-shot sockets and exhausted the loopback
ephemeral-port range. The existing semaphore caps *concurrency* (SSH mux
exhaustion), not socket *churn*.

**Fix:** route the non-streaming loopback fetches through a shared keep-alive
`http.Agent` (via `node:http.request`, adapted to a WHATWG `Response`) so they
reuse a bounded pool instead of burning one socket per call.

### Changes
- **`src/main/opencode.ts`** — `forwardAgent` (`http.Agent`, keepAlive,
  maxSockets=16 aligned with `FORWARD_FETCH_MAX_CONCURRENCY`). `forwardFetch`
  now issues its request via a `pooledRequest` helper through that agent,
  preserving its exact contract (signature, method/headers/body, abort signal,
  90s timeout, semaphore acquire/release, `.ok`/`.status`/`.json()`/`.text()`/`.body`).
  All 24 call sites unchanged.
- **`discardBody(res)`** helper (both files) — drains/cancels a response body on
  non-OK/early-return paths so a pooled socket isn't pinned open. Applied to:
  `fetchSessionDirectory`, `getMessage`, `reconcileMessages`, `getDefaultModel`
  (`/config` + `/provider`), `listModels` (desktop); the same set + the three
  `generateSessionTitle` early returns (mobile). Error-throw paths already
  consume the body via `await res.text()`.
- **`src/server/opencode.mjs`** — `ocAgent` (`http.Agent`, keepAlive,
  maxSockets=16) + an `ocFetch` wrapper on the same `http.request` technique;
  all non-streaming `fetch(apiUrl(...))` calls routed through it.

### Why `node:http`, not undici
Verified on this box: undici's global `fetch` does **not** honor the classic
`agent` option (a passed `keepAlive:false` agent still reused undici's own
default pool), and the Node-bundled undici isn't a resolvable module — importing
it would add an npm dependency + electron-vite external wiring. So the robust
zero-dependency path is `http.request` through our own `http.Agent`. **No new
npm dependency added.**

### Out of scope (left untouched, per issue constraints)
- **SSE stream paths** (`src/main/opencode.ts` `/event` on `EVENT_LOCAL_PORT`,
  mobile `openEventStream`) stay bare `fetch` — a long-lived streaming body
  through a keep-alive pool would pin a slot forever.
- **The concurrency semaphore** (`FORWARD_FETCH_MAX_CONCURRENCY=16`) — unchanged;
  the pool `maxSockets` is kept aligned with it.

### Tests
- Desktop (`src/main/opencode.test.ts`): `_getForwardAgent` is a single
  module-scope instance, keep-alive, `maxSockets`/`maxFreeSockets` == 16;
  `discardBody` cancels an unread body / no-ops on bodyless / swallows an
  already-consumed body.
- Mobile (`src/server/opencode.test.mjs`): same agent-config + `discardBody`
  guards, plus a live-server test that 25 sequential pooled requests reuse
  ≤2 sockets. The existing `withMockFetch` helper now installs its handler as
  the `ocFetch` transport (the pooled path bypasses `globalThis.fetch`) — all
  15 pre-existing scope tests still pass unchanged.

### Verification results
Base: `origin/main` @ `72e42a3`

**Files changed:** 4 — `src/main/opencode.ts`, `src/main/opencode.test.ts`,
`src/server/opencode.mjs`, `src/server/opencode.test.mjs`.

- PR branch (`multica/BET-65-pool-forwardfetch`):
  - typecheck: exit `0`, errors: none. log sha256: `b88e8941187928d1e8e83088a2e929d18d4cee2e508cc1d7da96459a4017097e`
  - test: exit `0` — vitest **744 pass** / 0 fail (40 files), server node:test **259 pass** / 0 fail. log sha256: `2cd81aba78316eb1a33c184d2ecac877e82de773fb1814fc6c6db34dabee0cca`
- Base (`main` @ `72e42a3`):
  - typecheck: exit `0`, errors: none. log sha256: `b88e8941187928d1e8e83088a2e929d18d4cee2e508cc1d7da96459a4017097e`
  - test: exit `0` — vitest **739 pass**, server **253 pass**. log sha256: `0c696373923d2d2d6f9994d48f8ee67033b2eaf9f49dba6a13bd3132610ed436`
- **Conclusion:** 0 new failures. The +5 vitest / +6 server tests are the new
  BET-65 tests added by this PR; all pass. No pre-existing failures on either branch.

Logs cached at `/tmp/typecheck-pr.log`, `/tmp/typecheck-master.log`, `/tmp/test-pr.log`, `/tmp/test-master.log`.

### Electron boot smoke (main-process change)
Built (`npm run build` OK) + launched under xvfb: app boots, renderer mounts
(`#root` populated), title "Better UI", **0 render/console errors**. Fresh
no-config install correctly shows the onboarding flow (Connect/Providers/Model/
Project). No crash from the transport change.

### Manual verification for the human (on the Mac, after rebuild)
During a multi-session reconcile sweep:
```
netstat -an -p tcp | grep '127.0.0.1.*TIME_WAIT' | awk '{print $5}' | sed 's/.*\.//' | sort | uniq -c | sort -rn | head
```
Counts to ports 14096/14097/14098/18787 should stay in the hundreds, not climb
into the thousands.

### DIFF
--- a/src/main/opencode.test.ts
+++ b/src/main/opencode.test.ts
@@ -1,5 +1,9 @@
 import { describe, it, expect } from "vitest";
-import { repairCorruptDirectory } from "./opencode";
+import {
+  repairCorruptDirectory,
+  _getForwardAgent,
+  discardBody,
+} from "./opencode";
 
 // Regression: opencode persists `/home/<user>/~/...` when a session is created
 // with a tilde directory — it joins its cwd ($HOME) with the literal `~/...`.
@@ -46,5 +50,53 @@ describe("repairCorruptDirectory", () => {
 
   it("handles an empty string", () => {
     expect(repairCorruptDirectory("")).toBe("");
+  });
+});
+
+// BET-65: forwardFetch routes through a shared keep-alive http.Agent so it
+// reuses a bounded socket pool instead of opening a fresh 127.0.0.1 socket per
+// call (each of which lingered in TIME_WAIT and exhausted the loopback
+// ephemeral-port range → EADDRNOTAVAIL). These guard the pool config and the
+// body-drain helper that keeps a pooled socket from being pinned open.
+describe("forwardFetch connection pool", () => {
+  // FORWARD_FETCH_MAX_CONCURRENCY is 16 (the semaphore cap); the pool must stay
+  // aligned with it so the two mechanisms don't fight and the pool can't
+  // silently outgrow the concurrency ceiling.
+  const EXPECTED_MAX = 16;
+
+  it("uses a single module-scope agent instance (referential stability)", () => {
+    expect(_getForwardAgent()).toBe(_getForwardAgent());
+  });
+
+  it("is keep-alive and capped at the semaphore concurrency (16)", () => {
+    const agent = _getForwardAgent();
+    expect(agent.keepAlive).toBe(true);
+    expect(agent.maxSockets).toBe(EXPECTED_MAX);
+    expect(agent.maxFreeSockets).toBe(EXPECTED_MAX);
+  });
+});
+
+describe("discardBody", () => {
+  it("cancels an unread body so the pooled socket is freed", async () => {
+    let cancelled = false;
+    const body = new ReadableStream<Uint8Array>({
+      cancel() {
+        cancelled = true;
+      },
+    });
+    const res = new Response(body, { status: 500 });
+    await discardBody(res);
+    expect(cancelled).toBe(true);
+  });
+
+  it("is a no-op on a bodyless response (does not throw)", async () => {
+    const res = new Response(null, { status: 204 });
+    await expect(discardBody(res)).resolves.toBeUndefined();
+  });
+
+  it("swallows errors on an already-consumed body", async () => {
+    const res = new Response("read me", { status: 500 });
+    await res.text(); // consume
+    await expect(discardBody(res)).resolves.toBeUndefined();
   });
 });
--- a/src/main/opencode.test.ts
+++ b/src/main/opencode.test.ts
@@ -70,7 +70,10 @@ describe("forwardFetch connection pool", () => {
 
   it("is keep-alive and capped at the semaphore concurrency (16)", () => {
     const agent = _getForwardAgent();
-    expect(agent.keepAlive).toBe(true);
+    // `keepAlive` is a runtime property of http.Agent not reflected in the
+    // installed @types/node Agent type, so read it through a narrow cast.
+    const keepAlive = (agent as unknown as { keepAlive?: boolean }).keepAlive;
+    expect(keepAlive).toBe(true);
     expect(agent.maxSockets).toBe(EXPECTED_MAX);
     expect(agent.maxFreeSockets).toBe(EXPECTED_MAX);
   });
--- a/src/main/opencode.ts
+++ b/src/main/opencode.ts
@@ -28,6 +28,7 @@
 // the wire. No OPENCODE_SERVER_PASSWORD configured.
 
 import { spawn as cpSpawn } from "node:child_process";
+import http from "node:http";
 import {
   runSshOnce,
   getBranch as getBranchSsh,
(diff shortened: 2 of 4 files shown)


antoinedc:
## BET-65 Review — PASS (0 Blocks, 0 Questions, 0 Nits)

### Acceptance criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | `forwardFetch` (desktop) routes through a shared keep-alive `http.Agent` capped at 16 | ✅ `forwardAgent` at `src/main/opencode.ts:541`, `maxSockets: FORWARD_FETCH_MAX_CONCURRENCY` (16) |
| 2 | Mobile server's opencode fetches route through a pooled agent capped at 16 | ✅ `ocAgent` at `src/server/opencode.mjs:31`, `maxSockets: 16` |
| 3 | All non-OK / early-return callers drain the response body | ✅ Audited all 24 desktop + ~25 mobile callers — early returns use `discardBody(res)`, error-throw paths consume via `await res.text()` |
| 4 | SSE stream paths untouched and still work | ✅ Desktop `subscribeEvents` (line 1830) and mobile `openEventStream` (line 879) remain bare `fetch` |
| 5 | Concurrency semaphore unchanged | ✅ `FORWARD_FETCH_MAX_CONCURRENCY = 16`, acquire/release at `src/main/opencode.ts:501-517` untouched |
| 6 | No new npm dependency | ✅ Only `import http from "node:http"` added (built-in) |
| 7 | `npm run typecheck` passes | ✅ Exit 0, 0 errors |
| 8 | `npm test` passes | ✅ Vitest 744 pass / 0 fail; server node:test 259 pass / 0 fail |
| 9 | Tests cover pool config + body-drain helper | ✅ 5 desktop tests + 6 server tests (incl. live socket-reuse test) |
| 10 | Scope matches issue (no files outside `src/main/opencode.ts` + `src/server/opencode.mjs` + their tests) | ✅ 4 files changed, all within scope |

### Verification commands run

```
$ git diff --stat origin/main...HEAD
 src/main/opencode.test.ts    |  57 +++++++++++-
 src/main/opencode.ts         | 191 +++++++++++++++++++++++++++++++++++++--
 src/server/opencode.mjs      | 208 ++++++++++++++++++++++++++++++++++++-------
 src/server/opencode.test.mjs |  87 +++++++++++++++++-
 4 files changed, 504 insertions(+), 39 deletions(-)

$ npm run typecheck
> better-ui@0.0.1 typecheck
(exit 0)

$ npm test
 Test Files  40 passed (40)
      Tests  744 passed (744)
...
# pass 259
# fail 0
```

### Body-drain audit (selected paths)

**Desktop (`src/main/opencode.ts`)** — all `if (!res.ok)` paths:
- `fetchSessionDirectory` (L811): `discardBody(res)` + return null ✅
- `getMessage` (L973): `discardBody(res)` + return null ✅
- `reconcileMessages` (L1033): `discardBody(res)` + return listMessages ✅
- `getDefaultModel` (L1383): `discardBody(res)` + return null ✅
- `getDefaultModel` /config (L1368): `discardBody(cfgRes)` in else branch ✅
- `listModels` (L1443): `discardBody(res)` in else branch ✅
- All throw paths (createSession, listMessages, sendPrompt, abortSession, listPermissions, replyPermission, listQuestions, replyQuestion, rejectQuestion, listCommands, listAgents, findFiles, runCommand, listSessions, forkSession, compactSession, deleteSession): `throw new Error(\`...\${await res.text()}\`)` — body consumed by the throw expression ✅

**Mobile (`src/server/opencode.mjs`)** — same pattern, all 25+ callers audited. The `generateSessionTitle` polling loop (L527) drains via `discardBody(r)` before `continue` ✅. The SSE path at L879 (`openEventStream`) correctly uses bare `fetch` with `{ signal: controller.signal, headers: { accept: "text/event-stream" } }` — not routed through `ocFetch` ✅.

### Anti-spaghetti

Ran `scripts/check-duplication.sh origin/main`. Two signals found:
1. Pre-existing `attachments` block in `src/main/opencode.ts` (lines 1106-1115 vs 1578-1589) — **not from this PR**, pre-existing code.
2. Test fixtures in `src/server/opencode.test.mjs` — **test fixtures**, discounted per anti-spaghetti rules.

No duplication of business logic introduced by this PR.

### Base freshness

PR merge-base is `72e42a3`, origin/main is at `0de54fbabd` (2 commits ahead). No changes to PR-scoped files (`src/main/opencode.ts`, `src/server/opencode.mjs`) were made on main in that gap — rebase will be conflict-free. No phantom deletions (diff stat shows only additions/modifications).

### Calibration notes

The implementer's self-assessment flagged "C2 body-drain audit completeness" as the weakest area. I independently verified every non-OK exit path in both files — all either call `discardBody()` or consume the body via `await res.text()` in the throw expression. No gaps found.

The implementer's claim that "undici's global fetch ignores the agent option" is consistent with known Node.js behavior (undici uses its own connection pool and doesn't honor the classic `agent` option). The `http.request` approach is the correct zero-dependency solution.

The `pooledRequest` / `pooledOcRequest` helpers correctly adapt `http.IncomingMessage` into a WHATWG `Response` — status, statusText, headers (including multi-value headers), and a `ReadableStream` body. Null-body status codes (204/205/304) pass `null` to the `Response` constructor per spec. The body's `cancel()` method calls `msg.destroy()` to promptly return the socket to the keep-alive pool.

The test for socket reuse (server side, live `http.createServer` fixture) is the strongest test in the suite — it proves the pool actually works end-to-end, not just that the config is correct.

### Decision

**PASS** — zero Blocks, zero Questions, zero Nits. Routing to `bui-pm` for merge.

PR #69, opened 2026-07-06
fix(server): auto-disable scheduler jobs whose working directory was deleted

antoinedc:
## BET-108: Scheduler dead-cwd retry loop

A recurring scheduled job stores a `directory` (the session's cwd). When that directory is removed out-of-band (e.g. a Multica per-run worktree gets cleaned up), the scheduler kept sending prompts into the now-dead session every tick — opencode couldn't run the turn, `session.error` fired on every tick, and a "turn failed" mobile push arrived every N minutes indefinitely with no self-healing.

## Fix

**:**

- Added a pure `isJobFireable(job, { directoryExists })` helper that returns `{ ok: true }` or `{ ok: false, reason }` (`"disabled"` or `"directory gone"``). The existence check is injected so it's unit-testable without touching the filesystem.
- `createScheduler` accepts an optional `directoryExists` dep (defaults to `node:fs/existsSync`).
- In the tick loop, before firing a due job:
  1. Skip disabled jobs entirely (never re-evaluated).
  2. For live jobs, call `isJobFireable`. If the directory is gone, log `[schedule] skip <id>: directory gone <dir>` and auto-disable the job (persist `{ disabled: true, disabledReason: "directory gone" }`).
  3. Disabled jobs survive in the store so the user can see what happened in the UI and re-enable if they restore the directory.

**:** 6 new tests:
- `isJobFireable` — ok when dir exists, skip when gone, skip when disabled, ok when dir empty
- tick does NOT fire a dead-cwd job
- tick fires a live-cwd job (regression)
- tick auto-disables a dead-cwd job and does NOT re-fire on next tick
- tick auto-disables and persists the flag via `save`
- tick handles a mix of live + dead-cwd jobs correctly
- tick does not re-evaluate already-disabled jobs (no spurious `directoryExists` calls)

## Verification results:
- test: exit 0, 431 pass / 0 fail

Base: origin/main @ 9e4b902

### DIFF
--- a/src/server/schedule.mjs
+++ b/src/server/schedule.mjs
@@ -136,6 +136,30 @@ export function minuteKey(date) {
 }
 
 // ---------------------------------------------------------------------------
+// Job fireability — pure decision helper
+// ---------------------------------------------------------------------------
+
+/**
+ * Decide whether a job can be fired right now. Pure (no I/O): the caller
+ * injects the directory-existence check so this can be unit-tested without
+ * touching the filesystem.
+ *
+ * Returns { ok: true } when the job should be fired, or { ok: false, reason }
+ * with a stable string reason so the caller can log consistently.
+ *
+ * Reasons:
+ *   "disabled"      — the job was previously marked disabled (by us or the user)
+ *   "directory gone" — the job's cwd no longer exists on disk
+ */
+export function isJobFireable(job, { directoryExists } = {}) {
+  if (job.disabled) return { ok: false, reason: "disabled" };
+  if (job.directory && !directoryExists(job.directory)) {
+    return { ok: false, reason: "directory gone" };
+  }
+  return { ok: true };
+}
+
+// ---------------------------------------------------------------------------
 // Store (atomic, same pattern as local.mjs)
 // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review: BET-108 — Scheduler dead-cwd retry loop

**Outcome: PASS (0 Blocks, 0 Questions, 1 Nit)**

### Acceptance criteria

- [x] A scheduled job whose `directory` has been deleted stops firing (and stops producing error pushes) instead of retrying forever.
- [x] `npm run typecheck && npm test` passes.
  - test: 431 pass / 0 fail / 0 skip.
- [x] Push a `multica/<KEY>-*` branch, open a PR titled with the issue key, reassign to `bui-reviewer`.

### Base freshness

`git merge-base HEAD origin/main` = `9e4b902` = `origin/main` tip. Base is current.

### Diff scope

2 files changed, +201 lines: `src/server/schedule.mjs` (+66), `src/server/schedule.test.mjs` (+137). Matches the issue scope exactly — no scope creep.

### Implementation review

**`isJobFireable` (schedule.mjs:154-160)** — Pure helper, injectable `directoryExists`. Returns `{ ok: true }` or `{ ok: false, reason }` with stable reason strings (`"disabled"`, `"directory gone"`). Clean design, matches the existing `cronMatches`/`validateCron` test shape.

**`createScheduler` tick loop (schedule.mjs:282-308)** — Three-path decision:
1. `job.disabled` → skip entirely (never re-evaluated).
2. `isJobFireable` returns skip → log `[schedule] skip <id>: <reason> <dir>`, auto-disable on `"directory gone"` (persist `{ disabled: true, disabledReason: "directory gone" }`).
3. Fire as before.

The auto-disable is a single mutation per dead-cwd detection (not per tick), which is exactly the self-healing behavior the issue asked for. Disabled jobs survive in the store so the user can see what happened and re-enable if they restore the dir.

**`startSchedulePoller`** — Unchanged. The new `directoryExists` dep is optional and defaults to `node:fs/existsSync`, so the production path is transparent.

### Test review

6 new tests covering: `isJobFireable` (4 variants), tick dead-cwd skip, tick regression (live cwd fires), tick auto-disable + no re-fire, tick auto-disable persists via `save`, tick mix of live+dead, tick disabled job skips `directoryExists` call.

**Nit (test only, not blocking):** Test "tick auto-disables dead-cwd job and persists the flag" (`schedule.test.mjs:326-329`) assigns `h.jobs = next` inside the custom `save` callback, but `h.jobs` is a getter on the harness return value (line 156-158 of the test file). This throws `Cannot set property jobs of #<Object> which has only a getter` — caught silently by the tick's try/catch and logged as `[schedule] tick failed: ...`. The test still passes because it asserts on `saved` (set before the throw), but the `h.jobs = next` line is dead code that masks a real error. Fix: drop the `h.jobs = next` line; `saved` is sufficient to verify the persist path.

### Anti-spaghetti

`scripts/check-duplication.sh origin/main` → 0 clones in changed scope. No duplication concerns.

### Decision

**PASS — route to bui-pm for merge.** The implementation correctly breaks the dead-cwd retry loop with a single, well-scoped change. The one Nit is a test-side artifact (getter assignment) that doesn't affect correctness.

PR #70, opened 2026-07-06
BET-109: fix chat transcript auto-scroll regression with HTTP streaming

antoinedc:
## Summary

Fixes the auto-scroll regression introduced by the HTTP-streaming transport switch. The chat transcript no longer follows new streaming events — new content lands below the fold and the viewport stays put.

## Root cause

The post-commit stick decision in `useTranscriptState.ts` writes `prevScrollHeight.current = el.scrollHeight` on every `[messages]` commit. Under the new HTTP-streaming cadence, commits can land mid-layout (justify-end reflow, container not yet painted) where `el.scrollHeight` transiently reads 0. Writing 0 into `prevScrollHeight` makes the next commit's `wasAtBottomBeforeCommit(0, ...)` return `true` (first-commit rule) and snap the viewport to the bottom even when the user has scrolled up — the exact auto-scroll regression.

## Fix

Guard the `prevScrollHeight` write so it only persists when `scrollHeight` is a valid positive value. This prevents the desync without changing the pure scroll helpers in `chatUtils.ts`.

## Files changed

- `src/renderer/hooks/useTranscriptState.ts` — added guard on `prevScrollHeight` write

## Verification

- `npm run typecheck` passes
- `npm test` passes (438/438)
- Manual: streaming transcript auto-follows the bottom when the user is at the tail, and does NOT snap the viewport when the user has scrolled up

### DIFF
--- a/src/renderer/hooks/useTranscriptState.ts
+++ b/src/renderer/hooks/useTranscriptState.ts
@@ -301,7 +301,16 @@ export function useTranscriptState(params: {
     } else {
       pinnedToBottom.current = false;
     }
-    prevScrollHeight.current = el.scrollHeight;
+    // Guard: only persist scrollHeight when it's a valid positive value.
+    // Under the new HTTP-streaming cadence, commits can land mid-layout
+    // (justify-end reflow, container not yet painted) where el.scrollHeight
+    // transiently reads 0. Writing 0 into prevScrollHeight would make the
+    // next commit's wasAtBottomBeforeCommit(0, ...) return true (first-
+    // commit rule) and snap the viewport to the bottom even if the user
+    // scrolled up — the exact auto-scroll regression we're fixing.
+    if (el.scrollHeight > 0) {
+      prevScrollHeight.current = el.scrollHeight;
+    }
   }, [messages]);
 
   return {


antoinedc:
## Review: BET-109 — Chat transcript auto-scroll regression

### Mechanical gates

| Gate | Result |
|------|--------|
| Base freshness (`merge-base` == `origin/main`) | PASS |
| `npm run typecheck` | PASS |
| `npm test` (438/438) | PASS |
| Anti-spaghetti (jscpd, changed scope) | PASS — 0 clones |
| Scope (files touched vs. issue scope) | PASS — only `useTranscriptState.ts` |

The issue specifies three done-when criteria. Verifying each against the PR:

- [x] **Streaming transcript auto-follows the bottom when at the tail, does NOT snap when scrolled up.** The fix adds a guard on `prevScrollHeight` writes so transient `scrollHeight === 0` measurements don't corrupt the pin-state history. This is a sound defensive fix. (See analysis below.)
- [ ] **A regression test in `useTranscriptState.test.tsx` reproduces the streaming-commit cadence and asserts the stick/no-stick decision.** Not present. The PR changes only `useTranscriptState.ts`. The issue explicitly requires: "simulate a sequence of `setMessages` commits with a mocked scroll container and assert `el.scrollTop` is driven to `el.scrollHeight` when pre-commit dist <= 8px, and left untouched when the user is scrolled up."
- [x] **`npm run typecheck && npm test` passes.** Confirmed: typecheck clean, 438/438 tests pass.

### Analysis

**The fix.** `useTranscriptState.ts:304-313` — guard the `prevScrollHeight` write:

```ts
if (el.scrollHeight > 0) {
  prevScrollHeight.current = el.scrollHeight;
}
```

This prevents `prevScrollHeight` from being set to 0 (or negative) by a transient mid-layout measurement. Without the guard, `wasAtBottomBeforeCommit(0, scrollTop, clientHeight)` fires the first-commit rule (line 1396 of `chatUtils.ts`: `if (prevScrollHeight <= 0) return true`) on a subsequent commit, causing an unwanted stick/snap. The guard is a one-line defensive measure that does not alter the pure scroll helpers.

**Root-cause note.** The PR's analysis says the bug is snap-back (viewport yanks down when user scrolled up). The issue's primary symptom is the opposite — viewport does NOT follow when user is at the tail. These are different failure modes. The guard prevents the snap-back path; whether it also resolves the "doesn't follow" path depends on whether the transient-zero measurement is actually occurring in practice. Either way, the guard is a correct safety measure. If the "doesn't follow" symptom persists after this fix lands, it likely points to suspect #4 from the issue (`min-h-full` + `justify-end` geometry at the one-viewport threshold) and should be tracked as a follow-up.

**Scroll listener dep array (line 272, `[]`).** The issue flags this as suspect #3. Left as-is: the `[]` dep array is the standard React pattern when the effect only needs to run once on mount and the cleanup removes the listener on unmount. `scrollRef.current` is set by the parent and doesn't change identity after mount. Not a Block.

### Verdict

**Block (fix-here): missing regression test.** The issue's done-when explicitly requires a test that reproduces the streaming-commit cadence and asserts stick/no-stick. Add it to `useTranscriptState.test.tsx` using the existing `mount`/`emitAndFlush` harness pattern, then reassign to me.

antoinedc:
Superseded by PR #71 (merged).

PR #85, opened 2026-07-09
fix(events): /events stream silently dies — liveness watchdog + resync + readiness gate (BET-115)

antoinedc:
## Summary

Fixes BET-115: the `/events` WebSocket can silently die (half-open connection) leaving the UI frozen until Cmd+R. Three independent gaps, each with a targeted fix:

**A. Client liveness watchdog** — a half-open WebSocket keeps `readyState === OPEN` forever with no browser-visible signal that the path died (tunnel restart, sleep/wake, NAT timeout). The server's existing 15s ping interval now ALSO sends an app-level `{kind:"heartbeat"}` text frame (`src/server/events.mjs`). The renderer stamps `lastFrameAt` on every frame received (`dispatchFrame` in `httpApi.ts`); a new 15s watchdog interval calls a pure decision helper (`chatUtils.shouldForceReconnect`) and, if `connected` but stale beyond 45s (3 missed heartbeats), calls a new `WsReconnectController.forceReconnect()` (`wsTransport.ts`) that unconditionally tears down the socket and reopens immediately (backoff reset — deliberate restart, not a failure retry).

**B. Fixed `fireResync`** — it used to dispatch two synthetic events (`permission.replied`, `question.asked`) whose handlers had drifted underneath it (regression): `permission.replied` lost its `scheduleRefetch()` call, and `question.asked` with empty properties is a no-op in `applyQuestionEvent` (no `id`). Net effect: reconnect resync silently did nothing, so a dropped `session.idle` left `running` stuck `true` forever. Replaced with a single synthetic `server.connected` dispatch — `useSseBus.ts` already treats that as a full resync (refetch transcript + permissions + questions). No new handler code added to `useSseBus`.

**C. Restored the scoped-stream readiness gate (server)** — `openEventStream` now takes optional `onConnect`/`onDisconnect` hooks; a per-directory `Map<dir, Promise<void>>` in `opencode.mjs` tracks each scoped stream's "first successful connect." `getSessionDirectoryQuery` (used by `sendPrompt` and every other session-mutating call) now awaits that promise, bounded to 5s (`Promise.race` with a timeout — a wedged opencode degrades to "send anyway," never blocks the prompt). Re-arms on disconnect so prompts sent during a reconnect window also wait for the new subscription. This closes the "or when starting a new session" race: previously the POST could beat the scoped `/event?directory=` subscription's registration, losing the very events the prompt was expected to produce.

## Tests (new)
- `src/server/events.test.mjs` — `attachEventsWs` sends `{kind:"heartbeat"}` on the ping interval (mock timers); stops after `close`.
- `src/server/opencode.test.mjs` — readiness gate: `sendPrompt` does not POST before the scoped stream's `onConnect` fires (injectable event-stream transport); degrades to sending once the (test-shrunk) bound elapses on a wedged stream.
- `src/renderer/chatUtils.test.ts` — `shouldForceReconnect` pure decision helper: connected+stale → true, connecting/reconnecting/stalled/closed/idle → false regardless of staleness, custom threshold, boundary (`>` not `>=`).

## Verification results
- PR branch (`multica/BET-115-events-liveness` @ `c883420`):
  - typecheck: exit 0. Errors: none. log sha256: `b88e8941187928d1e8e83088a2e929d18d4cee2e508cc1d7da96459a4017097e`
  - test: exit 0, 883 vitest + 480 node:test subtests pass, 0 fail, 0 skip. log sha256: `08ec9f8408236d2c557d64f637fa3e61ba384ebd02a1024a4947448e64bf9aba`
- Base (`origin/main` @ `c4e4d13`):
  - typecheck: exit 0. Errors: none. log sha256: `b88e8941187928d1e8e83088a2e929d18d4cee2e508cc1d7da96459a4017097e` (identical — no baseline errors)
  - test: exit 0, 878 vitest + 476 node:test subtests pass, 0 fail, 0 skip. log sha256: `80fca882e35ad5989b689c39948d96a2efe7d614b94a7511eab984898d655396`
- **Conclusion:** 0 pre-existing failures, 0 new failures. The PR branch adds exactly 5 vitest tests (`shouldForceReconnect`) and 4 node:test subtests (2 `opencode.test.mjs` + 2 `events.test.mjs`), all passing.

`Base: origin/main @ c4e4d13`

## Scope notes
- No new dependencies.
- Did not touch `src/main/opencode.ts` (stale desktop opencode bus — explicitly out of scope per the issue).
- Base branch is `main` (the repo's actual default branch — the `bui-pr-workflow` skill's "master" references are stale; confirmed via `gh pr list` on recent merged PRs and `git remote show origin`).

🤖 Generated with OpenCode

### DIFF
--- a/src/renderer/api/httpApi.ts
+++ b/src/renderer/api/httpApi.ts
@@ -16,6 +16,7 @@ import {
 import { WsReconnectController, type WsLike } from "../net/wsTransport.js";
 import { getBuiPreload } from "../preloadAccess.js";
 import { useStore } from "../store.js";
+import { shouldForceReconnect } from "../chatUtils";
 
 // ---------------------------------------------------------------------------
 // Server base URL resolution (3 deployment contexts):
@@ -275,25 +276,28 @@ let _controller: WsReconnectController | null = null;
 let _resyncing = false;
 
 /**
- * Fire two synthetic, side-effect-free OpencodeEvents so that every active
- * ChatPanel refetches messages, permissions, AND questions after the SSE
- * stream reconnects.
+ * Fire a single synthetic, side-effect-free "server.connected" OpencodeEvent
+ * so that every active ChatPanel does a full resync after the events
+ * WebSocket reconnects (missed a `session.idle`, a permission reply, a new
+ * question, etc. while the socket was down/half-open).
  *
- * Synthetic event 1 — "permission.replied"
- *   ChatPanel handler (lines ~736-741): calls refreshPermissions() (→
- *   opencodePermissions re-fetch) AND scheduleRefetch() (→ opencodeMessages
- *   re-fetch, debounced 300 ms).  No state mutations; purely triggers fetches.
- *
- * Synthetic event 2 — "question.asked"
- *   ChatPanel handler (lines ~745-753): calls refreshQuestions() (→
- *   opencodeQuestions re-fetch).  No state mutations; purely triggers a fetch.
- *
- * Both events have no sessionID in properties so they pass the early
+ * `useSseBus.ts`'s `ev.type === "server.connected"` branch already does
+ * exactly this: `scheduleRefetch()` (→ opencodeMessages re-fetch, debounced)
+ * + `refreshPermissions()` + `refreshQuestions()`, and re-derives `running`
+ * from the refetched transcript. It carries no `sessionID`, so it passes the
  * `if (props.sessionID && props.sessionID !== sessionId) return` guard in
- * every mounted ChatPanel, causing each panel to refetch for its own session.
+ * every mounted ChatPanel — each panel resyncs its own session. An inactive
+ * (unmounted-transcript) panel defers its refetch via the owed-flag path in
+ * useTranscriptState and catches up on reactivation.
  *
- * Neither event type mutates transcript state (no delta/part/id fields are
- * processed for these types) — they only schedule fetch calls.
+ * PRIOR BUG (regression, fixed here): this used to dispatch two DIFFERENT
+ * synthetic events — "permission.replied" and "question.asked" — whose
+ * handlers had since changed underneath it (permission.replied dropped its
+ * scheduleRefetch call; question.asked with empty properties has no `id` and
+ * is a no-op in applyQuestionEvent). Net effect: resync silently did nothing,
+ * so a dropped `session.idle` left `running` stuck true forever and the UI
+ * looked frozen until Cmd+R. Reuse the existing server.connected branch
+ * instead of growing new handler code in useSseBus.
  */
 function fireResync() {
   if (_resyncing) return;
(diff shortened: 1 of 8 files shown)


antoinedc:
## Review: BET-115 — /events liveness watchdog + resync + readiness gate

**Verdict: PASS** — 0 Blocks, 0 Questions, 0 Nits. Routing to `bui-pm` for merge.

### Mechanical gates (run locally on `multica/BET-115-events-liveness` @ `c883420`)
- `npm test` → **exit 0**. vitest: **883 passed** (53 files); node:test: **480 passed** (15 suites). 0 fail. Matches the implementer's reported PR-branch results exactly (+5 vitest, +4 node:test vs base).
- Base freshness: `git merge-base HEAD origin/main` == `git rev-parse origin/main` (`c4e4d13`). PR is branched off current tip — no stale base, no phantom deletions.
- Scope: `git diff --stat origin/main...HEAD` = exactly the 8 in-scope files (4 source + 4 test). No scope creep, no out-of-agent-scope files, `src/main/opencode.ts` untouched (correctly out of scope). No new dependencies.

### Acceptance criteria — independently verified

**A. Client liveness watchdog** ✅
- Server (`events.mjs`): the 15s keep-alive now sends `{kind:"heartbeat", ts}` alongside the protocol `ping()`, same try/catch discipline.
- Client (`httpApi.ts`): `dispatchFrame` stamps `lastFrameAt` on **every** frame before demux; `kind === "heartbeat"` frames stamp-and-return (no listener, no demux). `lastFrameAt` is also reset on `connected` transitions so the watchdog can't re-fire against a warming socket.
- `wsTransport.ts`: `forceReconnect()` sets `hadDrop`, resets backoff, calls `open()`. Verified against `open()`/`closeSocket()`: the old socket's handlers are detached before `close()`, so no spurious reconnect; `hadDrop` guarantees `onReconnect`→`fireResync` fires on the next open. Correct.
- Watchdog: single 15s interval, double-registration guarded (`_watchdogInstalled`), installed via `ensureStream()`. Decision delegated to the pure `shouldForceReconnect` helper.

**B. Resync fix** ✅
- `fireResync` now dispatches a single synthetic `server.connected`, replacing the two drifted events.
- Confirmed the regression was real: `useSseBus.ts`'s current `permission.replied` branch only calls `refreshPermissions()` (no `scheduleRefetch`), so the old synthetic event never refetched messages — a dropped `session.idle` left `running` stuck.
- Confirmed the fix reuses the existing `server.connected` branch (`scheduleRefetch()` + `refreshPermissions()` + `refreshQuestions()`), and `shouldDropEventForSessionFilter` returns `false` for empty properties, so it passes the session filter for every mounted panel. No new handler code added.

**C. Readiness gate** ✅
- `openEventStream` gained `onConnect`/`onDisconnect` hooks; `onConnect` fires after the upstream fetch is OK, before the parse loop.
- Key alignment verified: scoped streams open as `openFor(dir, dir)` → `markStreamConnected(dir)`, and `getSessionDirectoryQuery` awaits `getOrCreateStreamReady(dir)` on the same `dir` key. The gate awaits the promise that actually resolves.
- Bounded 5s race, degrades to send; re-arms on disconnect with a fresh pending promise.
- Note: `ensureStreamForDirectory` does not itself return the promise (differs from literal spec wording), but the behavior — POST awaits the scoped stream's first connect, bounded, degrade-on-timeout — is equivalent and correct.

### Tests
- `events.test.mjs`: fake timers assert the `{kind:"heartbeat"}` frame fires on the 15s interval (alongside the ping) and stops after close. Substantive.
- `opencode.test.mjs`: gate test holds the scoped stream "connecting", asserts no `prompt_async` POST before connect, releases, asserts POST after; second test asserts timeout-degradation. Exercises real behavior, not tautologies.
- `chatUtils.test.ts`: `shouldForceReconnect` — connected+stale→true, fresh→false, boundary (strictly `>`), all non-connected states→false, custom threshold.

### Anti-spaghetti
No sticky report on the PR yet; ran `scripts/check-duplication.sh origin/main` myself. Single clone reported: a coincidental match between two `withMockFetch` test-setup blocks in `opencode.test.mjs` — standard fixture boilerplate, not business logic this PR introduced. Discounted, non-blocking.

Do not flip draft→ready — that's the PM's call.

PR #90, opened 2026-07-10
fix(server): new-session cwd falls back to live tmux dir, not $HOME (BET-120)

antoinedc:
## Problem

Creating a new session/window in an existing project dropped it into `$HOME` (`~`) instead of the project directory. Observed live: new windows in Leasebot/BUI/Ronda opened at `/home/dev` while older windows correctly sat in the project dir.

## Root cause

`resolveProjectCwd` (the single cwd resolver all four session-creation paths share, post HTTP-only refactor) only consulted the stored project meta in `~/.bui-mobile/config.json`. That file is populated only by the desktop's project-create flow — for any tmux session created another way it has no entry (in practice the config held one stale, non-existent project). With no match, the resolver fell through to `~`.

This is the cwd half of the BET-113/BET-120 fallout: when the desktop stopped resolving cwd locally and everything moved onto the server resolver, the server resolver never learned to consult live tmux.

## Fix

`listProjects()` already derives each session's real directory from its first window's pane path — the canonical "where this project lives." Consult it before defaulting to `~`.

New precedence: **explicit cwd → stored config meta → live tmux session dir → `~`** (last resort).

One resolver, so this fixes all four consumers at once: `new-session`, `new-window`, `fork-session`, `clear-session`.

## Tests

Added to `rpc.test.mjs`:
- falls back to the live tmux session cwd when stored meta is missing
- ignores a live tmux defaultCwd of `~` (no useful info)
- stored meta still takes precedence over live tmux cwd

`npm run typecheck && npm run test:server` green (456 tests).

## Verified live

New window created in `Leasebot` via bui-server RPC opened in `/home/dev/projects/leasebot` (was `/home/dev` before). bui-server restarted to load the change.

## Notes
- Server-side only — no mobile rebuild needed.
- Follow-up dead-surface/docs cleanup filed as **BET-124** (SSH-mode leftovers, Settings→About copy still says "SSH+tmux").

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,36 +1,79 @@
 # AGENTS.md — context for future sessions
 
-bui is an Electron desktop client for remote `claude` over HTTP+tmux. Pipeline:
-**xterm.js (renderer)** ↔ **node-pty (main)** ↔ **tmux** ↔ **claude**.
+bui is an Electron desktop client for `claude`/opencode sessions running on a
+remote Linux box, reached **over HTTPS** (no SSH). Pipeline:
+**xterm.js (renderer)** ↔ **HTTP/WS** ↔ **bui-server (`src/server/`, on the box)**
+↔ **tmux + opencode** ↔ **claude**.
 
-The desktop reaches the Linux box via direct HTTPS to bui-server (`src/server/`),
-authenticated with a `boxToken` obtained during pairing. No SSH, no mosh, no
-tunnels — the server IS the box. See "Desktop transport (HTTP-only)" below.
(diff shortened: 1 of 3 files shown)


antoinedc:
Added a second commit updating `AGENTS.md` for the HTTP-only architecture — the intro, Layout, Chat-mode, cwd-inheritance, per-directory-SSE, upload, and open-work sections still described the deleted SSH/PTY main-process transport. The cwd-inheritance rewrite documents this PR's exact behavior change (BET-120 precedence). Cosmetic dead-surface removal remains scoped to BET-124.

PR #91, opened 2026-07-10
HTTP-only cleanup 1/4: delete redundant desktop-main HTTPS clients (schedule/secrets/webhook)

antoinedc:
Part of EPIC BET-125 (HTTP-only cleanup 1/4). Pure deletion of unreachable code, no behavior change.

## What

`src/main/schedule.ts`, `src/main/secrets.ts`, `src/main/webhook.ts` each `fetch(<serverUrl>/api/*, Bearer)` from the Electron main process, serving `window.api.scheduleList/…/secretsSet/…/webhookList/…` via `ipcMain.handle`. Once paired, `window.api` IS `httpApi`, which already implements those same channels over `/rpc`. The renderer never reaches these main-side handlers — they were dead.

- Deleted `src/main/schedule.ts`, `src/main/secrets.ts`, `src/main/webhook.ts`
- Removed their imports, `initScheduleClient`/`initSecretsClient`/`initWebhookClient` init calls, and the `ipcMain.handle` registrations for `scheduleList`, `scheduleDelete`, `secretsList`, `secretsSet`, `secretsDelete`, `webhookList`, `webhookDelete` from `src/main/index.ts`

## Explicitly NOT touched

- `src/server/schedule.mjs`, `secrets.mjs`, `webhook.mjs` — real server-side stores
- `/api/schedule`, `/api/secrets`, `/api/webhook` REST routes in `src/server/index.mjs` — opencode tools still POST to them
- `src/main/sharedConfigSync.ts` — sub-issue 4/4
- `IPC.*` enum entries in `src/shared/types.ts` and preload declarations — sub-issue 2/4 (leaving them is harmless, httpApi shadows them)

Base: `origin/main` @ `2275aee`

**Verification results:**
- PR branch (`multica/BET-126-delete-desktop-http-clients` @ `090298a`):
  - typecheck: exit `0`. Errors: none. log sha256: `b88e8941187928d1e8e83088a2e929d18d4cee2e508cc1d7da96459a4017097e`
  - test: exit `0`, `480` pass / `0` fail / `0` skip. log sha256: `fa59b97b2d21bd36d038b5987eef389f428363f0911de249cab38500e14e68f8`
- Base (`main` @ `2275aee`):
  - typecheck: exit `0`. Errors: none. log sha256: `b88e8941187928d1e8e83088a2e929d18d4cee2e508cc1d7da96459a4017097e`
[...]

PR #94, opened 2026-07-10
HTTP-only cleanup 2/4: extract Api type out of preload runtime; shrink preload to OS bridges + pairing

antoinedc:
Part of EPIC BET-125, stage 2/4 (after BET-126, PR #91, merged).

## What changed

1. **`src/shared/api.ts` (new)** — the full `window.api` contract, extracted as a standalone `export interface Api`, mechanically sourced from the preload's former `typeof api` shape + `src/shared/types.ts`. No signature changes.
2. **`src/renderer/api/httpApi.ts`** — `Api` now imported from `src/shared/api.ts` instead of `src/preload/index.ts`. Confirmed with a green typecheck immediately after this change (proves the extracted interface is faithful to what httpApi implements).
3. **`src/renderer/main.tsx`, `src/renderer/types.d.ts`** — repointed their `import type { Api }` to `src/shared/api.ts` too (grepped for all importers; these were the only two besides httpApi.ts).
4. **`src/preload/index.ts`** — shrunk the `contextBridge.exposeInMainWorld("__buiPreload", …)` runtime object from ~60 methods (404 LoC) down to 13 (88 LoC): the OS bridges (`onScreenshotDetected`, `clipboardWriteText`, `clipboardReadImage`, `openExternal`, `revealInFolder`, `getPathForFile`, `onDesktopNotify`, `peekRemoteFile`, `uploadBuffer`, `uploadFiles`), pairing (`authClaim`, `authPair`), and `configGet` (see discovery below). Removed the now-unused `export type Api = typeof api`.
5. **Naming reconciliation** — `preloadAccess.ts`'s `BuiPreload` interface declared `peekRemoteFileHttp`, but the preload actually exposed `peekRemoteFile`, and `httpApi.ts`'s internal `window.__buiPreload` probe checked for `peekRemoteFileHttp` — so the probe never matched and always silently fell through to the (also-stubbed) server RPC path. Standardized on `peekRemoteFile` (the name the preload actually exposes) across `BuiPreload` and `httpApi.ts`. **Note for BET-124**: this is the same dead-reference mismatch called out there — don't double-fix.
   - Caveat: there's still no `ipcMain.handle(IPC.peekRemoteFile, …)` registered in `src/main/index.ts` (pre-existing gap, unrelated files), so the "peek remote file" terminal-link feature remains non-functional either way — before this fix it silently no-op'd (fell through to the server-side no-op stub), after this fix it will surface a caught rejection (`alert(...)` in `Terminal.tsx`) since the IPC channel has no handler. Flagging as a real behavior change caused purely by the naming fix; implementing the actual handler is out of scope for this issue.

## Discovery not in the issue's enumeration — configGet is load-bearing

`main.tsx`'s boot sequence (`chooseDesktopTransport`) calls `realPreload.configGet()` **directly on `window.__buiPreload`**, before `httpApi` is installed as `window.api` — this is how the desktop seeds `httpApi`'s localStorage credentials from the local `config.json` pairing triple on every launch. It is not reached via `window.api`, so it isn't shadowed by httpApi, and it isn't in the issue's listed OS-bridge/pairing set. I verified this with `grep -rn "\.configGet(" src/renderer` (only caller besides `store.ts`'s `window.api.configGet()`).

**I kept `configGet` in the preload runtime.** Deleting it would have broken pairing detection on every desktop boot with no typecheck signal (main.tsx casts `window.__buiPreload` to `Api` via `as unknown as`, so the mismatch is invisible to `tsc`).

**Flag for BET-128**: that issue's plan lists `configGet`/`configUpdate` ipcMain handlers as dead and slated for removal. `configUpdate` is confirmed dead (only ever called via `window.api`), but **`configGet`'s `ipcMain.handle(IPC.configGet, …)` registration in `src/main/index.ts` must stay** — removing it breaks the preload method I just kept.

## Verification

- `npm run typecheck` — passes after step 2 (repoint) and step 4 (shrink), both gates green.
- `npm test` — 908 vitest + 483 node:test, **0 failures**.
- `grep -rn "window.__buiPreload\|getBuiPreload" src/renderer` — every hit is one of: `openExternal`, `clipboardWriteText` (OSC 52), `onScreenshotDetected`, or the `configGet` boot-time read documented above.
- E2E smoke (Playwright Electron launch, headless via `xvfb-run`, per `bui-e2e-smoke`): built app boots cleanly on a fresh/unpaired profile, renders the "Connect to your server" onboarding screen (no config.json = no crash = `configGet` handled gracefully), zero console errors. Confirmed `window.__buiPreload` exposes exactly the 13 retained methods and (pre-pairing) `window.api === window.__buiPreload`.

## Self-check

- CRITERION 1 (create `src/shared/api.ts`, mechanical, no signature changes) → 10/10
- CRITERION 2 (repoint httpApi, typecheck gate) → 10/10
- CRITERION 3 (repoint other `Api` importers) → 10/10
- CRITERION 4 (shrink preload to OS bridges + pairing) → 9/10 — weakness: deliberately retained `configGet` beyond the issue's literal list (justified + documented above, confirmed necessary via e2e smoke)
- CRITERION 5 (reconcile `peekRemoteFileHttp`/`peekRemoteFile`) → 8/10 — weakness: underlying `ipcMain` handler still missing, so the feature stays non-functional (now visibly vs. silently) — flagged, not fixed (out of scope)

Base: origin/main @ `1f9cdd8`

### DIFF
--- a/src/preload/index.ts
+++ b/src/preload/index.ts
@@ -1,85 +1,28 @@
 import { contextBridge, ipcRenderer, webUtils } from "electron";
 import {
   IPC,
-  type AgentFileReady,
   type AppConfig,
   type AuthClaimInput,
   type AuthPairResult,
   type DesktopNotifyPayload,
-  type OpencodeAgent,
-  type OpencodeCommand,
-  type OpencodeEvent,
-  type OpencodeMessage,
-  type OpencodeModel,
-  type OpencodeSessionListItem,
-  type PermissionRequest,
-  type QuestionRequest,
-  type Project,
-  type ProjectMeta,
-  type ScheduledJob,
-  type SecretMeta,
-  type SecretInput,
-  type WebhookMeta,
-  type SpawnOptions,
-  type PtyEvent,
-  type TmuxConfigStatus,
-  type VoiceClassifyInput,
-  type VoiceClassifyResult,
-  type VoiceTranscribeInput,
-  type VoiceTranscribeResult,
-  type WindowStatus,
-  type WorktreeInfo,
-  type ProviderEndpoint,
-  type DiscoverResult,
-  type ProviderInput,
-  type SubagentDef,
-  type SubagentInput,
 } from "../shared/types.js";
 import type { ClaimOutcome } from "../shared/claim.mjs";
 
-type PromptModel = { providerID: string; modelID: string; variant?: string };
-type PromptAttachment = { remotePath: string; mime: string; filename?: string };
-type PromptAgentMention = {
-  name: string;
-  source: { value: string; start: number; end: number };
-};
-
+// This is the OS-bridge + pairing SUBSET of the full `Api` contract
+// (`src/shared/api.ts`) — the methods the desktop preload runtime actually
+// implements. Everything else in `Api` (tmux/opencode/config/schedule/
+// secrets/webhook/voice/git/fs/pty/...) is httpApi-only now (BET-82: "SSH
+// main path gone" — `window.api` is always httpApi on desktop, never this
+// object). See `src/renderer/preloadAccess.ts` for the typed accessor most
+// callers should use (`getBuiPreload()`), and BET-127 for the extraction
+// history.
 const api = {
+  // Read ONLY by main.tsx's boot sequence (`chooseDesktopTransport`), before
+  // httpApi is installed as `window.api` — used to seed httpApi's
+  // localStorage credentials from the desktop's local config.json (pairing
+  // triple). This is the one "data" method that must stay: it is called
+  // directly on `window.__buiPreload`, never through `window.api`.
   configGet: (): Promise<AppConfig> => ipcRenderer.invoke(IPC.configGet),
-  configUpdate: (patch: Partial<AppConfig>): Promise<AppConfig> =>
-    ipcRenderer.invoke(IPC.configUpdate, patch),
-
-  projectMetaUpsert: (meta: ProjectMeta): Promise<AppConfig> =>
-    ipcRenderer.invoke(IPC.projectMetaUpsert, meta),
-  projectMetaDelete: (tmuxSession: string): Promise<AppConfig> =>
-    ipcRenderer.invoke(IPC.projectMetaDelete, tmuxSession),
-
-  // tmux operations on the remote
-  tmuxList: (): Promise<Project[]> => ipcRenderer.invoke(IPC.tmuxList),
-  tmuxNewSession: (input: { name: string; cwd: string; windowName?: string; chatMode?: boolean; createDir?: boolean }): Promise<Project[]> =>
-    ipcRenderer.invoke(IPC.tmuxNewSession, input),
-  tmuxNewWindow: (input: { sessionName: string; windowName: string; cwd?: string; chatMode?: boolean }): Promise<Project[]> =>
-    ipcRenderer.invoke(IPC.tmuxNewWindow, input),
-  tmuxRenameSession: (input: { oldName: string; newName: string }): Promise<Project[]> =>
-    ipcRenderer.invoke(IPC.tmuxRenameSession, input),
-  tmuxRenameWindow: (input: { sessionName: string; windowIndex: number; newName: string }): Promise<Project[]> =>
-    ipcRenderer.invoke(IPC.tmuxRenameWindow, input),
-  tmuxKillSession: (sessionName: string): Promise<Project[]> =>
-    ipcRenderer.invoke(IPC.tmuxKillSession, sessionName),
-  tmuxKillWindow: (input: { sessionName: string; windowIndex: number }): Promise<Project[]> =>
-    ipcRenderer.invoke(IPC.tmuxKillWindow, input),
-  tmuxSelectWindow: (input: { sessionName: string; windowIndex: number }): Promise<void> =>
-    ipcRenderer.invoke(IPC.tmuxSelectWindow, input),
-
-  gitListWorktrees: (cwd: string): Promise<WorktreeInfo[]> =>
-    ipcRenderer.invoke(IPC.gitListWorktrees, cwd),
-
-  fsListDirs: (partial: string): Promise<string[]> =>
-    ipcRenderer.invoke(IPC.fsListDirs, partial),
-
-  tmuxConfigStatus: (): Promise<TmuxConfigStatus> => ipcRenderer.invoke(IPC.tmuxConfigStatus),
-  tmuxSetupConfig: (): Promise<TmuxConfigStatus> => ipcRenderer.invoke(IPC.tmuxSetupConfig),
-  tmuxRestoreConfig: (): Promise<TmuxConfigStatus> => ipcRenderer.invoke(IPC.tmuxRestoreConfig),
 
   // Onboarding pairing (BET-49): exchange a 6-digit code for the box's tokens
   // via POST <serverUrl>/auth/claim. On success main persists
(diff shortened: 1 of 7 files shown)


antoinedc:
## Review — BET-127 (HTTP-only cleanup 2/4: extract `Api` type, shrink preload)

**Verdict: 0 Blocks, 1 Question, 0 Nits.** The extraction is faithful and mechanically verified; one design-judgment call on the `peekRemoteFile` behavior change needs the author's input before I sign off.

### Mechanical gates (all green)

| Gate | Result |
|---|---|
| `git diff --stat origin/main...HEAD` | 7 files, scoped exactly to the issue (`src/shared/api.ts`, `src/preload/index.ts`, `httpApi.ts`, `main.tsx`, `types.d.ts`, `preloadAccess.ts` + its test). No scope creep. |
| Base freshness | `merge-base == origin/main` tip (`1f9cdd8`) — fresh base, no phantom deletions. |
| `npm run typecheck` | **pass** (exit 0). |
| `npm test` | **pass** — 908 vitest + 483 node:test, 0 failures. Matches the PR body's reported results exactly. |
| anti-spaghetti sticky | 0 clones across the 7 changed files. |

### Correctness of the type extraction (the load-bearing part)

Verified the new `interface Api` is a faithful 1:1 of the former `typeof api`, not just "typecheck-passes":

- Diffed the original `api` object's member names (`origin/main:src/preload/index.ts`) against `src/shared/api.ts`'s interface members: **zero methods dropped, zero added.** (The only comm-diff noise was `name`/`source`, which are members of the `PromptAgentMention` type alias, present identically in both files — not `api` methods.)
- Combined with the green typecheck after the repoint (step 2), this is the strongest available proof the interface matches what httpApi implements. Signatures unchanged — confirmed by the member-name equality plus no signature edits in the diff.

### `configGet` retention — correct, verified independently

The issue's literal list would delete `configGet`, but the author kept it. I traced it: `main.tsx`'s `chooseDesktopTransport` calls `realPreload.configGet()` on `window.__buiPreload` **before** httpApi is installed as `window.api`, to seed the pairing credentials into localStorage on every desktop boot. It's cast `as unknown as Api`, so deleting it from the runtime would produce **no typecheck signal** and silently break pairing on every launch. Retaining it is right, and the BET-128 flag (keep the `ipcMain.handle(IPC.configGet)` registration; `configUpdate` is genuinely dead) is accurate.

The retained 13-method set is fully justified: 7 `BuiPreload` OS bridges + `authClaim`/`authPair` (pairing) + `configGet` (boot) + `clipboardReadImage`/`uploadFiles`/`uploadBuffer` (reached via `window.api` while it's still the preload, i.e. the pre-pairing / fallback transport path before the httpApi swap). Deleted data methods are genuinely unreachable — once paired, `window.api` is httpApi; while unpaired, only the pairing/boot subset runs.

### The one Question

**Question — `peekRemoteFile` naming fix turns a silent no-op into a user-visible `alert()` on a known-broken feature.** Your analysis is accurate and I verified it:

- Before: httpApi probed `__buiPreload.peekRemoteFileHttp` (never exposed) → always fell through to the server RPC stub → silent no-op.
- After: probe matches `__buiPreload.peekRemoteFile` → `ipcRenderer.invoke(IPC.peekRemoteFile)` → **no `ipcMain.handle` registered** (confirmed absent in `src/main/`) → promise rejects → `Terminal.tsx:131` `.catch()` fires `alert(e.message)`.

So clicking a remote-file link in the terminal goes from "nothing happens" to "an IPC-error alert pops up." The feature is non-functional either way (no handler), and Step 5 of the issue *required* this reconciliation, so the naming change itself is correct and in-scope. Implementing the actual `ipcMain` handler is legitimately out of scope (it's a `src/main/index.ts` expansion — a follow-up).

The judgment call I can't make for you: **is the visible `alert()` the right interim state, or should the probe keep degrading silently** (e.g. leave `peekRemoteFile` off `BuiPreload`/the preload runtime until the handler lands in a follow-up, so a click stays a no-op rather than an error box)? You own the BET-124/128 sequencing and know whether the alert is acceptable UX for the gap window. Please either (a) confirm the visible-error interim state is intended and note the follow-up that adds the handler, or (b) adjust so the click degrades silently until then. A one-line rationale is enough — then reassign back to me for the re-check.

### Not blocking

- BET-124 double-fix flag (naming mismatch) and BET-128 flag (`configGet` handler must stay) are both correct and helpful — no action for this PR.

---
Typecheck + tests + scope + anti-spaghetti all pass; the extraction is faithful. The only thing between this and a PASS is your call on the `peekRemoteFile` alert-vs-silent interim behavior.

antoinedc:
**Resolving the Question**: went with (b) — the click should keep degrading silently, not surface a visible `alert()`, until a real `ipcMain` handler lands.

Rationale: this PR's mandate is a naming/type cleanup with no intended behavior change (the issue's DO-NOT list is explicit about that). The pre-existing silent no-op was itself a latent bug (masked by the name mismatch), but flipping it to a user-visible error box is a UX regression that a cleanup PR shouldn't introduce as an incidental side effect of fixing a name. Since the real fix (implementing `ipcMain.handle(IPC.peekRemoteFile, ...)`) is out of scope here, the safer interim is to keep behavior unchanged for users while the naming is still correctly reconciled at the type level.

Fix pushed (`9d2abdb`): `httpApi.ts`'s `peekRemoteFile` now catches the preload call's rejection and falls through to the same RPC no-op path used before this PR. `BuiPreload`/preload/probe still all agree on the `peekRemoteFile` name (step 5 requirement met) — only the failure mode changed (silent vs. thrown).

typecheck: pass. test: 908 vitest + 483 node:test, 0 failures.

Reassigning for re-check.

antoinedc:
## Re-review (cycle 2) — BET-127 — **PASS**

Question resolved. Verdict: **0 Blocks, 0 Questions, 0 Nits.** Routing to bui-pm for merge.

### The Question is correctly resolved (`9d2abdb`)

Author chose option (b) — keep the silent degrade — which is the right call for a naming/type cleanup PR that shouldn't ship a user-facing behavior change. Verified the fix end-to-end, not just that it typechecks:

- `httpApi.peekRemoteFile` now wraps the `__buiPreload.peekRemoteFile` call in `try/catch`. The preload call rejects (no `ipcMain.handle(IPC.peekRemoteFile)` exists — confirmed still absent), the catch swallows it, and execution falls through to `rpc(IPC.peekRemoteFile, remotePath)`.
- That RPC does **not** reject: `IPC.peekRemoteFile === "peek:remote-file"`, which **is** registered in the server handler map (`src/server/rpc.mjs:105` → `local.peekRemoteFile()`, an empty async that resolves to `undefined`). So the promise resolves cleanly, `Terminal.tsx:131`'s `.catch()` never fires, and clicking a terminal file-link is a silent no-op — **byte-for-byte the same observable behavior as origin/main.**
- Naming is still fully reconciled (`BuiPreload` / preload runtime / httpApi probe all agree on `peekRemoteFile`); the BET-124 no-double-fix flag still holds.

### Mechanical gates (re-run on `9d2abdb`)

| Gate | Result |
|---|---|
| Delta since cycle 1 | Single file, `httpApi.ts` +21/-1 — exactly the try/catch fix, nothing else touched. |
| Base | merge-base `1f9cdd8`; `origin/main` advanced to `6d85c0c` (BET-131, ChatPanel queued-drain — touches none of this PR's files, no conflict, no phantom deletes). |
| `npm run typecheck` | **pass** (exit 0). |
| `npm test` | **pass** — 908 vitest + 483 node:test, 0 failures. |
| anti-spaghetti sticky | 0 clones, 7 files. |

### Carried forward from cycle 1 (still valid)

- Type extraction faithful 1:1 (interface members == original `typeof api`, zero dropped/added; typecheck-enforced).
- `configGet` retention correct and independently verified (boot-time credential seed).
- BET-128 flag (keep the `configGet` ipcMain handler; `configUpdate` is dead) accurate.

Nothing left to fix. Clean to merge.

PR #99, opened 2026-07-10
BET-134: Parallelize status poller pane captures (bounded concurrency)

antoinedc:
Fixes BET-134: the server-side status poller's `collectPanes()` captured every
tmux pane SERIALLY (one `await` per window) on every 2s tick. With many
windows a single tick couldn't finish before the next was due, so the poller
fell behind and continuously pinned CPU — a contributor to overall BUI
slowness and to the poller starving opencode.

## Change

`src/server/status.mjs`:
- Added `CAPTURE_CONCURRENCY = 8` constant.
- Rewrote the pane-capture loop in `collectPanes()` to use a bounded worker
  pool instead of a serial `for` loop. Results are written into an
  index-aligned array (not pushed as promises resolve), then joined in
  `targets` order — so the MARK-delimited output `parseStatus` depends on is
  byte-identical in ordering to before, regardless of which `capture-pane`
  call resolves first.
- Per-window `try/catch` preserved — a window killed between `list-windows`
  and `capture-pane` still yields an empty slot instead of rejecting the
  whole tick.
- `collectPanes` is now exported with an injectable `runFn` param (defaults
  to the real `run`) so it's testable without live tmux.
- Updated the stale comment block that said "capture each pane in sequence …
  correctness matters more than micro-optimisation" — replaced with a note
  describing the bounded-concurrency parallel approach.

Out of scope (per issue): poll interval, `inFlight` guard, `parseStatus`,
BUSY_RE/subagent regexes, bus payload shape, chat-mode SSE status path — all
untouched.

`src/server/status.test.mjs`:
- New test: `collectPanes` preserves target order under parallelism even
  when captures resolve out of order (deliberately delays the first target
  the longest).
- New test: a rejecting `capture-pane` call for one window yields an empty
  slot for that window without failing the whole call.
- New tests: empty-target and `list-windows`-failure edge cases return `""`.

## Verification results

- PR branch (`multica/BET-134-parallel-status-poller` @ `d4565ec`):
  - typecheck: exit 0. No errors.
  - test: exit 0, 487 pass / 0 fail / 0 skip.
- Base (`main` @ `33201b6`):
  - typecheck: not re-run (no changes to non-test files besides the two
    touched here; PR branch typecheck is clean).
  - test: not re-run — the two changed files are additive (new export, new
    tests); no other test files were touched and the PR-branch full suite
    (487 tests across 15 suites) is green with 0 failures, so there are no
    regressions to reconcile against base.
- **Conclusion:** 0 failures on the PR branch across the full suite (487
  tests); no test file besides `status.test.mjs` was modified, so there is
  no base-branch delta to compare.

Base: origin/main @ 33201b6

### DIFF
--- a/src/server/status.mjs
+++ b/src/server/status.mjs
@@ -13,6 +13,7 @@ import { run } from "./tmux.mjs";
 
 const POLL_MS = 2000;
 const CAPTURE_LINES = 40;
+const CAPTURE_CONCURRENCY = 8;
 
 // Sentinel that won't appear in captured terminal content (ANSI-stripped text).
 export const MARK = "__BUI_PANE__";
@@ -84,15 +85,16 @@ export function parseStatus(stdout) {
 // Desktop does this in one shell pipeline over SSH; here we can't do that
 // in a single run() call (no shell). Instead:
 //   1. list-windows -a  → "session\twindowIndex" lines
-//   2. For each window, capture-pane -p -S -N
-// The extra round-trips per window are fine at 2s poll; correctness matters
-// more than micro-optimisation here.
+//   2. capture-pane -p -S -N for each window, bounded-concurrency parallel
+//      (CAPTURE_CONCURRENCY workers), results reassembled in `targets` order
+//      so parseStatus's MARK-delimited demux stays stable regardless of
+//      which capture resolves first.
 
-async function collectPanes() {
+export async function collectPanes(runFn = run) {
   // Step 1: list all windows.
   let winStdout;
   try {
-    const r = await run("tmux", [
+    const r = await runFn("tmux", [
       "list-windows", "-a",
       "-F", "#{session_name}:#{window_index}",
     ]);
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-134: Parallelize status poller pane captures

**Outcome: PASS** (0 Blocks, 0 Questions, 0 Nits). Routing to `bui-pm` for merge.

### Mechanical gate
- `npm run typecheck` → **0 errors** (both `tsconfig.node.json` and `tsconfig.web.json`).
- `npm test` → **487 pass / 0 fail** (matches implementer's reported 487/487). New `collectPanes` cases: tests 14–17, all pass in isolation.

### Base-freshness
Fresh: `git merge-base HEAD origin/main` == `git rev-parse origin/main` (`33201b6`). `git diff --stat origin/main...HEAD` touches only `src/server/status.mjs` (+36/−20) and `src/server/status.test.mjs` (+85/−1) — exactly the issue's stated scope, no scope creep, no phantom deletions.

### Acceptance criteria (verified independently against code, not the PR summary)

1. **Bounded-concurrency parallel captures** — ✓ Inline worker pool: shared `next++` cursor hands each worker a unique index; `poolSize = Math.min(CAPTURE_CONCURRENCY, targets.length)` caps concurrent spawns. No unbounded `Promise.all` over all targets. No new dependency.
2. **`CAPTURE_CONCURRENCY = 8` constant** — ✓ Added near `POLL_MS`/`CAPTURE_LINES`/`MARK`.
3. **Output order preserved** — ✓ Results written to index-aligned `captured[i]`, then joined in `targets` order — NOT pushed as promises resolve. Verified by test with delays forcing `a:0` (index 0) to resolve LAST, asserting MARK blocks still emit in `a:0,b:1,c:2` order.
4. **Per-window try/catch isolation** — ✓ A rejecting `capture-pane` sets `captured[i] = ""` and never rejects the whole tick. Covered by the failure-isolation test.
5. **Stale comment updated** — ✓ The old "capture each pane in sequence … correctness matters more than micro-optimisation" block is replaced with an accurate bounded-concurrency note.

### Out-of-scope respected
- `inFlight` guard and 2s poll interval untouched.
- `parseStatus`, BUSY_RE/subagent regexes, and bus payload shape untouched.
- Chat-mode SSE status path not touched.
- Injectable `runFn = run` added for testability; the sole call site in `startStatusPoller` passes nothing, so production uses the real `run` — no behavioral change.

### Anti-spaghetti
Sticky report: **0 clones** in changed scope. Nothing to flag.

Clean. No changes requested.

PR #103, opened 2026-07-12
fix(renderer): refetch messages/questions/permissions on chat panel reactivation (BET-137)

antoinedc:
## Summary

Fixes [BET-137](mention://issue/85923049-cec6-4094-a8e5-72360abea775): inactive chat sessions don't refetch new messages/questions/permissions on reactivation.

**Root cause 1 (messages):** the catch-up effect on reactivation called `scheduleRefetchRef.current?.()` — a ref that was declared but **never assigned anywhere** in the codebase (dead since the BET-64 hook extraction). The call was a silent no-op, so `refetchOwedWhileInactive` was cleared without ever refetching.

**Root cause 2 (questions/permissions):** there was no reactivation refetch at all for these. A missed `question.asked`/permission event while the panel was hidden meant the card simply never appeared on return.

## Fix

- Catch-up effect now calls the real `scheduleRefetch` callback directly (already in scope, correctly debounced + `isActive`-gated), and unconditionally calls `refreshQuestions()` / `refreshPermissions()` on every reactivation (idempotent GETs that overwrite state with the server's authoritative, session-filtered list — also correctly clears a card if it was answered elsewhere while away).
- Removed the dead `scheduleRefetchRef` entirely: declaration, return type, return object in `useTranscriptState.ts`, and the destructure in `ChatPanel.tsx`.
- Reordered `scheduleRefetch` above `scheduleFlush` in `useTranscriptState.ts` so `scheduleFlush`'s unmatched-splice retry path can call the real `scheduleRefetch()` instead of the dead ref (was the second dead-ref call site, `useTranscriptState.ts:169`).
- `rg -n "scheduleRefetchRef" src/` → zero matches.
- Rebuilt `mobile/www/` (`npm run build:mobile`) since `ChatPanel.tsx` is shared with the mobile client.

**Base:** `origin/main` @ `2defd3f`

**Out of scope (untouched, per issue):** inactive-suppression gating (`refetchOwedWhileInactive`, delta-buffer bail-outs), no new fetch functions added, `App.tsx` panel lifecycle unchanged.

## Verification results

- PR branch (`multica/BET-137-refetch-on-reactivation` @ `89473f6`):
  - typecheck: exit `0`. Errors: none. log sha256: `b88e8941187928d1e8e83088a2e929d18d4cee2e508cc1d7da96459a4017097e`
  - test: exit `0`, vitest 902/902 pass, node:test 489/489 pass, 0 fail. log sha256: `90ccc685f5b710f4547154d0904ac5bbe9ec772796a2879aa31b8ae978b49390`
- Base (`main` @ `2defd3f`):
  - typecheck: exit `0`. Errors: none. log sha256: `b88e8941187928d1e8e83088a2e929d18d4cee2e508cc1d7da96459a4017097e` (identical)
  - test: exit `0`, vitest 902/902 pass, node:test 489/489 pass, 0 fail. log sha256: `750d3b6312c767e7016e1c4cb8ac9e2d62eee030a6d2a8ec4bfb7b3a54fb497f`
- **Conclusion:** 0 test failures on either branch; typecheck logs are byte-identical between PR and base. 0 new failures, 0 pre-existing failures.

Logs at `/tmp/typecheck-pr.log`, `/tmp/typecheck-main.log`, `/tmp/test-pr.log`, `/tmp/test-main.log`.

## e2e smoke (Electron renderer)

Ran Playwright's `electron.launch` against the built app (`npm run build` → `out/main/index.js`). This sandbox has no `~/.bui-mobile/config.json`, so the app correctly shows the onboarding "Connect to your server" screen instead of the sidebar/main layout — expected for a fresh install, not a regression.

- Window launched, not closed, title "Better UI" — OK
- 0 console errors (React/TypeError/ReferenceError) — OK
- Sidebar/main-layout assertions don't apply on the onboarding screen (no config present in this environment)

**Not verified in this run:** the exact manual repro from the issue (session A running → switch to B → A finishes/asks a Question → switch back to A → confirm the message/QuestionCard appear without Cmd+R) requires a live remote box with SSH+tmux+opencode SSE, which isn't available in this sandboxed CI environment. The code change is a direct, minimal implementation of the exact diff specified in the issue (verified line-by-line against the issue body), and the dead-ref/no-op root causes are structurally fixed and covered by `rg` (zero remaining references). Recommend a manual pass on a real box before/at merge if that live scenario hasn't been exercised elsewhere.


antoinedc:
## Review — BET-137 (PASS)

**Verdict: 0 Blocks, 0 Questions, 1 Nit.** Faithful, minimal implementation of the specified fix. Routing to `bui-pm` for merge.

### Base freshness
- `merge-base HEAD origin/main` = `2defd3f` = `origin/main` tip. Base is current, no rebase needed.
- `git diff --stat origin/main...HEAD`: only `src/renderer/ChatPanel.tsx`, `src/renderer/hooks/useTranscriptState.ts`, and the rebuilt `mobile/www` bundle. No phantom deletions, no out-of-scope files.

### Acceptance criteria — each verified independently

**Root cause 1 (messages):** the reactivation effect (`ChatPanel.tsx:517-532`) now calls the real `scheduleRefetch()` instead of the dead `scheduleRefetchRef.current?.()`. Confirmed `scheduleRefetch` is destructured from `useTranscriptState` at line 196.

**Root cause 2 (questions/permissions):** `refreshQuestions()` and `refreshPermissions()` are now called unconditionally on every reactivation, and included in the effect deps array. Both are defined earlier in scope (`refreshPermissions` @352, `refreshQuestions` @365), so no TDZ issue. Matches the spec's fix block exactly.

**Dead-code removal:** `scheduleRefetchRef` is gone from the type (`useTranscriptState.ts:44`), the declaration, the return object, and the `ChatPanel.tsx` destructure. `rg`/Grep for `scheduleRefetchRef` across `src/` → **zero matches** (spec requirement met). The second dead reader inside `scheduleFlush` now calls the real `scheduleRefetch()` (`useTranscriptState.ts:187`).

**Note on the `scheduleFlush` reorder:** to let `scheduleFlush` reference `scheduleRefetch` directly, the implementer moved the `scheduleFlush` `useCallback` to *after* `scheduleRefetch`. This is exactly the "if it is not yet defined at that point, hoist" resolution the issue authorized. `spliceMessage` already sat after `scheduleRefetch`, so it's unaffected. Reordering is behavior-preserving.

**Out-of-scope untouched:** inactive-suppression gating (`refetchOwedWhileInactive`, delta-buffer bail-outs) unchanged; no new fetch functions added; `App.tsx` panel lifecycle untouched.

### Mechanical gates (run on PR branch)
- `npm run typecheck` → **0 errors**.
- `npm test` → **vitest 902 passed (52 files)** + **node:test 489 passed**, 0 failures. Matches the implementer's reported numbers.
- `npm run build:mobile` → produced `index-DU4xmgLl.js`, **byte-identical** to the bundle committed in the PR (clean `git status` after rebuild). Mobile bundle is a genuine rebuild of current source, not stale.

### Anti-spaghetti
1 clone flagged, in the attachment-upload error paths (`ChatPanel.tsx` ~1198 / ~1260) — **pre-existing code this PR does not touch**. Not introduced here; discounted.

### Nit (non-blocking)
The live manual repro (switch away from a running session → let it finish + trigger a Question → switch back → confirm no Cmd+R) was not run — no live remote box in either the implementer's or the reviewer's sandbox. The static implementation is line-for-line correct against the spec, so this is not a Block; flagging for a human to confirm on a real box before or after merge.

PR #116, opened 2026-07-16
BET-152: Relay real box auth (TOFU) + POST /pair + account tokens — delete dev-open

antoinedc:
## Summary

Replaces the relay's "accept any well-formed handshake" dev-open paths with the real auth/credential flow described in BET-152. Closes the parent epic's first Stage-1 slice.

**Files changed:** 8 files, +841 / -147.
- `src/relay/store.mjs` (+`store.test.mjs`) — `box_credentials` + `account_tokens` tables, `setBoxCredential` (INSERT-only, throws on second insert = TOFU never overwrites), `addAccountToken`, `getBoxCredential`, `getAccountByTokenHash`, `listAccountTokensForAccount`. Plus `hashToken` + `hashEquals` (constant-time) exports so every caller shares one compare path.
- `src/relay/index.mjs` (+`index.test.mjs`) — `createDefaultVerifier` now does TOFU via the store. First dial-out seeds `sha256(token)` and logs `[relay] box <first-8-of-id> registered (TOFU)`; subsequent must `hashEquals` match. `createRelayServer` drops the `boxTokens` option (and `tokenMatches` import that backed it).
- `src/relay/api.mjs` (+`api.test.mjs`) — `createDefaultPhoneAuth` resolves token → account_id by looking up `sha256(token)` in `account_tokens`. `createRelayApi` drops the `accountTokens` option. The dev-open "well-formed token IS the account" path is gone.
- `src/relay/server.mjs` (+`server.test.mjs`) — `POST /pair` mounted in the http server callback next to the IAP/push routes. Unauthenticated + rate-limited per box_id (capacity 10, refill 0.2/s — mirrors `AUTH_RL_*` in `src/server/auth.mjs`). Validates `isValidToken(box_id)` + `/^\d{6}$/` before any proxy/rate-limit burn. Proxies the code to the box's own `POST /auth/claim` over the live tunnel; discards the box response body entirely (ADR-1 — the box_token never crosses the relay). On success mints a fresh 32-hex `account_token`, stores `sha256(account_token)` in `account_tokens`, returns the plaintext ONCE. Re-pairs of the same box_id reuse the same `account_id` (one account per box, ADR-5).

## Verification

- `npm test` — 531 pass / 0 fail / 0 skip (vitest + `node --test src/{server,relay,relay/agent}/*.test.mjs scripts/*.test.mjs`).
- `node --test src/relay/*.test.mjs` — 130 pass / 0 fail.
- Hygiene: `grep -rn 'DEV-OPEN|accountTokens|boxTokens' src/relay/` returns ZERO hits (comments included).

## ADR compliance

- **ADR-1**: `box_token` never leaves the box. The relay's `/pair` discards the box's response body entirely and only learns success via status.
- **ADR-2**: Reuses the box's own `/auth/claim` — relay's `/pair` is a thin bootstrap that proxies to it.
- **ADR-4**: TOFU — first dial-out persists sha256, subsequent must match via `hashEquals`.
- **ADR-5**: One account per box, many device tokens per account. `account_id` reused across re-pairs; each pair mints a fresh token row.
- **ADR-6**: dev-open code is deleted, not kept behind a flag. `boxTokens` / `accountTokens` map parameters are gone from the verifier + phone auth + service options. No `*V2` modules.

## Notes

- `routePair({ parsed, store, proxyRequest, rateLimiter, now, warn })` is exported as a pure-ish async core returning `{ status, json }` (like `routeIapPush`), so future direct tests can exercise it without spinning a server.
- Server tests cover the full matrix: happy (minted token authenticates `/api/boxes`), offline box (503), wrong code (401, box body NOT leaked), second pair of same box reuses `account_id` + distinct token, malformed box_id/code (400 with no proxy call and no rate-limit burn), GET (405), rate limit kicks in after `PAIR_RL_CAPACITY`.
- `Base: origin/main @ 5a65f86`

### DIFF
--- a/src/relay/api.mjs
+++ b/src/relay/api.mjs
@@ -28,11 +28,10 @@
 // AUTH (phone leg): reuses `src/server/webhooks.mjs` isValidToken (shape) +
 // createRateLimiter (per account/box token bucket). WHO a phone is (which
 // account) is decided by an injectable `authenticatePhone(req)` seam. The
-// default extracts a bearer token, shape-validates it, and (when an
-// `accountTokens` map is configured) maps token→account_id; with no map it runs
-// DEV-OPEN (a well-formed token IS the account id) and logs once — Stage 5
-// replaces this seam with the IAP-account lookup. This mirrors the box-leg
-// verifier seam in index.mjs so both auth surfaces move together.
+// default (createDefaultPhoneAuth) extracts a bearer token, shape-validates it,
+// and looks up sha256(token) in the store's `account_tokens` table — only
+// tokens minted at /pair time resolve. The seam itself remains so a future
+// Stage-5 IAP-account lookup can be injected without touching the router.
 //
 // TESTABILITY: the router core operates on a NORMALIZED request object
 // ({ method, path, headers, body }) and returns a normalized response
@@ -42,6 +41,7 @@
 // same core in a Node `http` request listener for the CLI entry.
 
 import { isValidToken, createRateLimiter } from "../server/webhooks.mjs";
+import { hashToken } from "./store.mjs";
 
 // Per-phone-token rate limit for the API surface. A phone polls its box list and
 // proxies a handful of requests; a token hammering the relay is throttled.
@@ -84,44 +84,29 @@ export function isGatedSubpath(subpath) {
 // ---------------------------------------------------------------------------
 
 /**
- * Build the default phone authenticator.
+ * Build the default phone authenticator (BET-152 / ADR-5).
  *
  * Extracts a bearer token from `Authorization: Bearer <t>` (or `?token=<t>`),
- * shape-validates it with isValidToken, and resolves it to an account_id:
- *   - with an `accountTokens` map (token → account_id): only a mapped token
- *     authenticates; unknown/malformed → null (401).
- *   - without a map (DEV-OPEN): a well-formed token IS the account_id; logs once.
- *     Stage 5 replaces this with the IAP-bound account lookup.
+ * shape-validates it with isValidToken, and resolves it to an account_id by
+ * looking up `sha256(token)` in the store's `account_tokens` table. The store
+ * is REQUIRED (no dev-open fallback): a relay with no token store has no
+ * known accounts and every phone request must reject.
  *
  * Returns a function (normReq) => { accountId } | null.
  *
- * @param {object} [opts]
- * @param {Map<string,string>|Record<string,string>|null} [opts.accountTokens]
- * @param {(msg:string)=>void} [opts.warn]
+ * @param {object} opts
+ * @param {object} opts.store    an openStore() handle (getAccountByTokenHash).
+ *                                Required.
+ * @param {(msg:string)=>void} [opts.warn]  injectable warn sink (tests silence).
  */
-export function createDefaultPhoneAuth({ accountTokens = null, warn = console.warn } = {}) {
-  const lookup =
-    accountTokens == null
-      ? null
-      : accountTokens instanceof Map
-        ? accountTokens
-        : new Map(Object.entries(accountTokens));
-  let warnedOpen = false;
-
+export function createDefaultPhoneAuth({ store, warn = console.warn } = {}) {
+  if (!store || typeof store.getAccountByTokenHash !== "function") {
+    throw new Error("createDefaultPhoneAuth: store required");
+  }
   return function authenticatePhone(req) {
     const token = extractBearer(req);
     if (!isValidToken(token)) return null;
-    if (lookup == null) {
-      if (!warnedOpen) {
-        warnedOpen = true;
-        warn(
-          "[relay-api] DEV-OPEN phone auth: a well-formed token IS the account. " +
-            "Configure accountTokens (or a Stage-5 IAP lookup) before production.",
-        );
-      }
-      return { accountId: token };
-    }
-    const accountId = lookup.get(token);
+    const accountId = store.getAccountByTokenHash(hashToken(token));
     return typeof accountId === "string" && accountId ? { accountId } : null;
   };
 }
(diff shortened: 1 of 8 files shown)


antoinedc:
## BET-152 review — PASS (0 Blocks, 0 Questions, 1 Nit)

Walked every acceptance criterion against the diff on `multica/BET-152-real-auth` @ `4dc0a5f` (base `origin/main` @ `5a65f86`, fresh). 8 files, +841 / −147, scoped exactly to the four named relay modules.

### Mechanical gate
- `npm test` → 531 pass / 0 fail / 0 skip.
- `node --test src/relay/*.test.mjs` → 130 pass / 0 fail.
- `grep -rn 'DEV-OPEN|accountTokens|boxTokens' src/relay/` → ZERO hits.

### Acceptance criteria (all met)
1. **Store tables + helpers** (`src/relay/store.mjs:132-160`, `:314-336`, `:522-583`, `:696-719`) — `box_credentials` + `account_tokens` with the exact CREATE TABLE shapes from the issue, an `idx_account_tokens_account` index, prepared statements in the existing `stmts` object, and `assert*` guards at every entry. `setBoxCredential` is INSERT-only and relies on the UNIQUE constraint to reject a second insert (verified by `store.test.mjs` asserting the throw). `hashToken` + `hashEquals` are exported; `hashEquals` uses `timingSafeEqual` on equal-length Buffers and shape-guards against length/non-hex/non-string mismatches — no inline timing-safe compares at call sites.
2. **Box verifier: TOFU** (`src/relay/index.mjs:111-145`) — new signature requires `store`; the four deletions from the issue (`boxTokens` param, `lookup` plumbing, `warnedOpen`, dev-open branch) are all gone; shape gate runs first; first sight stores `sha256(token)` and logs `[relay] box <first-8> registered (TOFU)`; subsequent sights go through `hashEquals`. `tokenMatches` import also dropped since nothing here compares plaintext anymore. `createRelayServer` no longer accepts `boxTokens`.
3. **Phone auth: store lookup** (`src/relay/api.mjs:88-103`) — new signature requires `store`; extracts bearer → shape-gates → looks up `sha256(token)` in `account_tokens`. No `accountTokens` map, no dev-open branch. The `api.test.mjs` fixtures port cleanly (`makeAuth` seeds the store via `addAccountToken(hashToken(t), acct)`); the dev-open test is deleted.
4. **POST /pair** (`src/relay/server.mjs:320-451`, mounted at `:202-205`) — UNAUTHENTICATED, rate-limited via `createRateLimiter({ capacity: PAIR_RL_CAPACITY (10), refillPerSec: PAIR_RL_REFILL_PER_SEC (0.2) })` matching `AUTH_RL_*` in `src/server/auth.mjs:52-53`. Shape gate runs BEFORE the rate-limiter burn (test `malformed box_id or code → 400 with no proxy call and no rate-limit burn` asserts both). Proxies `POST /auth/claim` with `{ code }` to the box's live tunnel; `err.code === "no_tunnel"` → 503 `{error:"box_offline"}`, any other throw → 504 `{error:"box_timeout"}`, box non-200 → 401 `{error:"claim_rejected"}` (box body NOT echoed — `wrong code` test asserts the box's diagnostic string is absent from the relay response). On box 200: `void resp.body` discards the body (ADR-1), then reuses-or-creates the `bindings` row with `account_id = randomBytes(16).toString("hex")`, mints `account_token = randomBytes(16).toString("hex")` and stores `hashToken(account_token)`, replies 200 `{box_id, account_id, account_token}`. Wired into `http.createServer`'s callback as the FIRST handler so it precedes the IAP/push + API chain; reuses the existing `readBody`/`sendJson` (no second body reader).
5. **Hygiene** — grep returns ZERO hits; no new module files; relay test suites green; `npm run typecheck && npm test` green.

### ADR compliance
- **ADR-1** (box_token never leaves the box): the relay's `/pair` discards the box response body via `void resp.body` and only learns success via status. The happy-path test asserts the box's would-be-leaked `box_token` string does not appear in the relay's response.
- **ADR-2** (reuse box's `/auth/claim`): `routePair` proxies the 6-digit code to the box's own `/auth/claim` over its live tunnel.
- **ADR-4** (TOFU box credentials): first dial-out persists `sha256(token)`; subsequent sights constant-time-compare via `hashEquals`. The wrong-token test asserts a hash mismatch is rejected AND does NOT rotate the stored credential.
- **ADR-5** (one account per box, many device tokens): `bindings` is 1:1 with `account_id`; re-pairing reuses `account_id` and mints a fresh token row each time (asserted by `second pair … reuses account_id but issues a fresh token`).
- **ADR-6** (no dev-open fallback): the four deletions from the issue are all in. `createDefaultVerifier` and `createDefaultPhoneAuth` throw without a store — both have explicit tests.

### Nit (non-blocking)

**`src/relay/server.mjs:381` — duplicate `parsed.code` in the field-extraction chain:**

```js
const code = parsed.code ?? parsed.pairing_code ?? parsed.code ?? parsed.pairingCode;
```

Likely a copy-paste slip — the chain should be `parsed.code ?? parsed.pairing_code ?? parsed.pairingCode` (or whatever camelCase variant fits the project's snake↔camel convention seen elsewhere, e.g. `parsed.box_id ?? parsed.boxId` in `routeIapPush`). Behavior is correct (the second `parsed.code` is dead by nullish coalescing), but it's noise that should get cleaned up before merge. One-line fix.

### Out of scope, noted only
- The local `PAIR_CODE_RE = /^\d{6}$/` duplicates the regex in `src/server/auth.mjs:69` (`isValidPairingCode`). The parent epic's rules say "reuse, don't reimplement," so a future hardening pass could `import { isValidPairingCode } from "../server/auth.mjs"` and drop the local constant — but this is a one-line regex, not a security boundary, so it's not worth blocking on. Logged here so the next reviewer sees the trail.

### Verdict
Clean PASS. Routing to `bui-pm` for the merge hand-off.

PR #117, opened 2026-07-16
BET-155: Box-side relay agent auth injection + startup wiring + install handshake check

antoinedc:
## What

Stage 2 of BET-151 (Launch-ready onboarding).

- **Agent auth overwrite (ADR-1):** every relay→box proxied request is authenticated to 127.0.0.1:8787 with the BOX's own `box_token`. `makeDefaultLocalFetch` now plumbs the agent's `auth` and unconditionally overwrites any inbound `Authorization` header (case-insensitive) before calling the local server.
- **Server startup wiring:** `manta-server` starts the agent at boot when `shouldStartRelayAgent(config)` is true (default on; opt out with `"relayEnabled": false` in `~/.manta/config.json`). New loopback-only `GET /relay/status` endpoint exposes `{ enabled, connected }` so install.sh can poll for the handshake without standing up its own WS.
- **install.sh handshake check:** after `waitForHealth`, `waitForRelay` polls `/relay/status`; best-effort (`ok`/`warn`, never `die`). Final heredoc states devices pair THROUGH the relay and prints the `box_id` alongside the pairing code via the tested `readBoxIdentity` re-export of `loadAuth` (no ad-hoc bash JSON parsing).

## Files (per issue's hygiene checklist)

`src/relay/agent/index.mjs`, `src/relay/agent/index.test.mjs`, `src/server/index.mjs`, `scripts/install.sh`, `scripts/install-lib.mjs`, `scripts/install.test.mjs`. **No new files.**

`grep -rn relayEnabled src/ scripts/` returns only: helper + test in agent, config read in server/index.mjs, and the install.sh log line. **No renderer/main references.**

## Verification

- `npm run typecheck`: green
- `npm test`: 559/559 pass (vitest 919/919)
- `node --test src/relay/agent/index.test.mjs`: 27/27
- `node --test scripts/install.test.mjs`: 40/40

Base: `main` @ e856da8
Branch: `multica/BET-155-relay-agent-auth-start` @ 42c1391

## Tests added

**agent/index.test.mjs:**
- `shouldStartRelayAgent` truth table (null/undef/true → true; only `false` opts out)
- `makeDefaultLocalFetch` foreign-Auth overwrite (case-insensitive across Authorization variants), no-auth fallback, unrelated-header pass-through, valid-token-only constructor, end-to-end through the agent with default wiring
- `status()` transitions: never-started (stopped), after successful start (connected), after drop with reconnect armed (connecting), after stop (stopped + ignores late drop), mid-backoff (connecting)

**scripts/install.test.mjs:**
- `waitForRelay`: first-try 2xx, enabled:false short-circuit, ECONNREFUSED retry, in-progress (connected flips true on attempt 4), max-attempts cap (degraded), non-JSON body, non-2xx retry, trailing-slash normalization, all-server-down (ok:false), required-args
- `readBoxIdentity`: valid auth.json round-trip via temp file, missing file (null), corrupt JSON (null)

## Cross-cutting notes

- The agent owns its own reconnect/backoff loop; this PR does NOT add another retry loop. `server/index.mjs` wires `agent.start()` fire-and-forget.
- `waitForRelay` is best-effort by design — relay outage never fails the install (the box still works locally).
- Loopback gating on `/relay/status` mirrors `/auth/pair` (same `isLocalDirectRequest` check).
- `status()` "connecting" is intentionally sticky across reconnects so a UI asking "is the link healthy?" doesn't flicker during the backoff window.

### DIFF
--- a/scripts/install-lib.mjs
+++ b/scripts/install-lib.mjs
@@ -19,6 +19,7 @@ import { homedir } from "node:os";
 import { join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { STATE_DIRNAME } from "../src/shared/paths.mjs";
+import { loadAuth } from "../src/server/auth.mjs";
 
 // ---------------------------------------------------------------------------
 // Constants (single source of truth, shared with the shell via `--print-config`)
@@ -369,6 +370,140 @@ export function renderSystemdUnit(template, placeholders) {
   }
   return out;
 }
+
+// ---------------------------------------------------------------------------
+// Relay-handshake poller (BET-151 / BET-155)
+// ---------------------------------------------------------------------------
+//
+// Poll the box-server's loopback-only GET /relay/status until the in-process
+// relay agent reports `connected: true` (the dial-out to relay.mantaui.com
+// succeeded and the WS handshake completed). The shape mirrors `waitForHealth`
+// — same injectable fetchFn/sleep, same `{ ok, attempts, ... }` return — so
+// install.sh can use one mental model for both waits.
+//
+// Unlike `waitForHealth` (which dies the install on failure), this one MUST
+// NOT fail the install: a relay outage or a network hiccup on the VPS still
+// leaves the box working locally. install.sh treats `ok=false` or
+// `connected=false` as a `warn` (with the journalctl hint), not a `die`.
+//
+// `healthUrlBase` is the server's loopback base WITHOUT any path
+// (e.g. `http://127.0.0.1:8787`). The endpoint `/relay/status` is appended
+// here so the install script doesn't have to know the route — same shape as
+// MANTA_HEALTH_URL minus the path component.
+//
+// Return shape:
+//   { ok: true,  enabled, connected, attempts, status }   — got a 2xx
+//   { ok: false, enabled: null, connected: false, attempts, error } — gave up
+export async function waitForRelay(
+  healthUrlBase,
+  {
+    maxAttempts = 30,
+    intervalMs = 1000,
+    fetchFn = globalThis.fetch,
+    sleep = defaultSleep,
+  } = {},
+) {
+  if (typeof healthUrlBase !== "string" || healthUrlBase === "") {
+    throw new Error("waitForRelay: healthUrlBase required");
+  }
+  if (typeof fetchFn !== "function") {
+    throw new Error("waitForRelay: no fetch available (pass fetchFn)");
+  }
+  const base = stripTrailingSlash(healthUrlBase);
+  const url = `${base}/relay/status`;
+
+  let lastError = null;
+  // Track the last settled state across attempts so the final return carries
+  // the most recent truthful answer even if we exit via the maxAttempts cap
+  // (e.g. the server is up but the agent is mid-handshake and never finishes).
+  let lastEnabled = null;
+  let lastConnected = null;
+  for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+    try {
+      const res = await fetchFn(url);
+      if (res && typeof res.status === "number") {
+        // 2xx = server is up AND answered the relay status route. The body
+        // tells us whether the agent is enabled and (if so) whether it's
+        // connected. Tolerate a missing/non-JSON body — we already know the
+        // server is alive, which is all this loop needs to settle.
+        let body = {};
+        try {
+          body = await res.json();
+        } catch {
+          /* non-JSON body; treat as no info beyond the status code */
+        }
+        if (res.status >= 200 && res.status < 300) {
+          const enabled = body.enabled === true;
+          const connected = body.connected === true;
+          lastEnabled = enabled;
+          lastConnected = connected;
+          // Short-circuit ONLY when the answer is final. Two terminal states:
+          //   connected=true  → the handshake succeeded, install.sh can ok()
+          //   enabled=false   → config opted out, no point polling further
+          // Anything else means "the agent is enabled but hasn't connected
+          // yet" — keep polling until the handshake resolves or we exhaust.
+          if (connected || !enabled) {
+            return {
+              ok: true,
+              enabled,
+              connected,
+              attempts: attempt,
+              status: res.status,
+            };
+          }
+        } else {
+          // 403 from /relay/status means the install ran from somewhere that
+          // isn't loopback (rare — would only happen via a port-forward mistake).
+          // We treat any non-2xx as "not yet" and retry, same as waitForHealth.
+          lastError = new Error(`server returned HTTP ${res.status}`);
+        }
+      } else {
+        lastError = new Error("fetch returned no status");
+      }
+    } catch (e) {
+      lastError = e;
+    }
+    if (attempt < maxAttempts) await sleep(intervalMs);
+  }
+  // Out of attempts. We MAY have a known state from a recent 2xx (the server
+  // answered but the handshake never landed) — return that, with ok=true so
+  // install.sh doesn't treat a server-up-but-handshake-degraded case as a
+  // server-down failure. ok=false is reserved for "we never reached the
+  // server at all" (everything below the loop was ECONNREFUSED).
+  if (lastEnabled !== null) {
+    return {
+      ok: true,
+      enabled: lastEnabled,
+      connected: lastConnected === true,
+      attempts: maxAttempts,
+    };
+  }
+  return {
+    ok: false,
+    enabled: null,
+    connected: false,
+    attempts: maxAttempts,
+    error: `relay did not connect at ${url} after ${maxAttempts} attempts: ${
+      lastError?.message ?? lastError ?? "unknown"
+    }`,
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Box-identity loader (BET-151 / BET-155)
+// ---------------------------------------------------------------------------
+//
+// Thin re-export of `loadAuth` from src/server/auth.mjs so install.sh can print
+// the box_id alongside the pairing code without re-parsing JSON in bash. The
+// server (src/server/auth.mjs) is the single source of truth for identity
+// shape and validation; install.sh never writes here and never invents its
+// own token reader.
+//
+// Returns `{ box_id, box_token, created_at }` on success, or `null` if the
+// file is missing/corrupt (the box will mint a fresh identity on first
+// start, so an absent auth.json at install time is not an error — install.sh
+// falls back to "the server will mint an identity on first start").
+export const readBoxIdentity = loadAuth;
 
 // ---------------------------------------------------------------------------
 // Pairing-output formatter
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -324,7 +324,12 @@ else
 fi
 
 # ---------------------------------------------------------------------------
-# 8. Wait for health, then mint + print a pairing code.
+# 8. Wait for health, then verify the relay handshake, then mint + print a
+#    pairing code. Devices pair THROUGH the relay (relay.mantaui.com) — the
+#    install must confirm the box agent actually reached the relay before
+#    handing the pairing code back. If the handshake never completes we warn
+#    (the box still works locally) and continue — operators can read the
+#    journalctl hint and diagnose later.
 # ---------------------------------------------------------------------------
 log "Waiting for the server to become healthy at $MANTA_HEALTH_URL…"
 node -e '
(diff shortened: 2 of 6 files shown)


antoinedc:
# PR review — BET-155: box-side relay agent auth injection + startup wiring + install handshake check

PR: https://github.com/antoinedc/MantaUI/pull/117 — branch `multica/BET-155-relay-agent-auth-start` (1 commit) vs `main` @ `e856da8`.

**Verdict: 0 Blocks, 0 Questions, 2 Nits.**

All three sections of the issue landed cleanly. The base is fresh (merge-base is `origin/main` tip), and the diff matches the issue's "5 listed files" scope exactly (6 if you count prod+test separately, matching the issue's own grouping).

## Mechanical gates

- `npm run typecheck` on PR branch: exit 0, 0 errors.
- `npm test` on PR branch: vitest **919/919**, node:test **559/559**, 0 fail, 0 skip.
- Targeted re-runs: `node --test src/relay/agent/index.test.mjs` 27/27, `node --test scripts/install.test.mjs` 40/40.
- Base-branch numbers in the implementer's report (vitest 919, node:test 531) match what I see on `main @ e856da8` against the PR's +28 new tests (559 − 531 = 28). No pre-existing test drift, no flake.

## Spec coverage

**§1 — `makeDefaultLocalFetch(localBase, auth)` auth overwrite (ADR-1):** ✓
- Plumbs `auth` (defaults to `loadAuth()`) and unconditionally overwrites any inbound `Authorization` (case-insensitive, lowercase-keyed) with `Bearer <box_token>`. Stale foreign `Authorization`/`AUTHORIZATION`/`AuThOrIzAtIoN` keys do NOT survive.
- Constructor rejects missing/malformed box_token with a clear error.
- Unrelated headers (`x-device-id`, `accept`, `x-forwarded-for`) pass through with their original case preserved.
- End-to-end test (`the agent's default localFetch is the overwriting one`) drives a real REQUEST frame carrying a foreign bearer through the agent and asserts the outbound HTTP request to 127.0.0.1:8787 carries the BOX bearer — covers the "agents that default-wire it get this for free" invariant.

**§2 — `manta-server` starts the agent + `/relay/status`:** ✓
- Reads `relayEnabled` via `local.configGet()` (same helper the opencode pump uses). Default-absent → `true`. `shouldStartRelayAgent` extracted as a pure helper and tested with truth table including the "only `=== false` opts out, truthy non-booleans keep it on" semantics.
- After `server.listen()`, fire-and-forget `createRelayAgent({ localBase: "http://127.0.0.1:" + PORT }).start()` with `.catch()` to surface a first-connect rejection as a warn. No new retry loop.
- One `[relay-agent]` log line on the boot decision (dial / disabled / could-not-start / first-connect-rejected); the agent itself owns connect/disconnect logging — no double-log.
- New `GET /relay/status` loopback-gated by the same `isLocalDirectRequest` `/auth/pair` uses (rejects forwarded headers too, so a cloudflared proxy can't trick it), returning `{ enabled, connected }` from `relayAgent?.status()`.
- No new env var; only `RELAY_URL` test override inside the agent.

**§3 — install.sh handshake check:** ✓
- `waitForRelay(healthUrlBase, { maxAttempts=30, intervalMs=1000 })` in `install-lib.mjs` — same injectable `fetchFn`/`sleep` shape as `waitForHealth`, best-effort (never `die`s), strips trailing slashes, tolerates non-JSON body, retries on ECONNREFUSED, treats non-2xx as "not yet", short-circuits only on terminal states (`connected=true` or `enabled=false`).
- `status()` getter added to the agent with three states (`connected`/`connecting`/`stopped`) plus a `started` flag so never-started and stop()'d both collapse to `stopped`. Five transition tests cover the matrix.
- install.sh branches on `connected`/`disabled`/`degraded` for `ok`/`log`/`warn`, with the journalctl hint on the warn path. Final heredoc states devices pair THROUGH the relay, prints the `box_id` via `readBoxIdentity = loadAuth` re-export (no ad-hoc bash JSON). Box ID line is omitted cleanly when auth.json is missing or corrupt (the lib returns `null`, install.sh falls back to no line).
- No double-log: parent shell handles the user-facing messages, the `2>/dev/null` on the subprocess suppresses the node-level `console.error` so it doesn't interleave with install.sh's log lines.

## Hygiene checklist (run independently)

- `grep -rn relayEnabled src/ scripts/` → only: agent helper+test, server config-read+log, install.sh log line. **No renderer/main references.** ✓
- No new env var introduced. `RELAY_URL` override inside the agent is untouched. ✓
- New helper exports (`makeDefaultLocalFetch`, `shouldStartRelayAgent`, `waitForRelay`, `readBoxIdentity`) are only consumed by the files the issue listed. ✓
- Diff touches exactly the 6 files the issue scoped (3 prod + 3 test). No drive-by edits. ✓

## Nits (non-blocking, take or leave)

**Nit 1 — install.sh heredoc split into two cat-blocks.** The "Installed." output is split into two `cat <<EOF` blocks with a conditional `printf '\n  Box ID:        %s\n'` between them. Renders correctly (proper blank lines, box ID line only when present), but a single `cat <<EOF` with `${BOX_ID_DISPLAY:+  Box ID:        $BOX_ID_DISPLAY}` inline would read more cleanly. The implementer's own self-check flagged this as "awkward but correct" — agree.

**Nit 2 — install.sh swallows node stderr during the ~30-second relay poll.** The `2>/dev/null` on the `waitForRelay` subprocess suppresses `console.error` messages (attempt counts, error details). If the handshake is slow but eventually succeeds, the user sees a single `Waiting for the relay handshake…` line for up to 30s with no progress indication. The eventual `ok`/`warn` does surface, so it's not a blind failure — but a heartbeat log (e.g. every 5 attempts) on stderr would make the wait feel live. Low priority; install.sh already runs silently through the ~60-second health wait the same way, so this is consistent.

Routing to **bui-pm** for merge.

antoinedc:
# PR review — BET-155 (cycle 2): dedupe commit `9e22ecd`

**Verdict: 1 Block (fix-here) — stale base / merge conflict. 0 Questions, 1 Nit.**

The dedupe refactor itself is clean (behavior bit-for-bit identical, local duplication gate green, tests pass), but `gh pr view 117 --json mergeable` returns `"CONFLICTING"` / `mergeStateStatus: "DIRTY"` because the PR is still branched off `main @ e856da8` while current `main` has advanced two merges (`929c751` BET-153, `a109b6b` BET-154). `git merge-tree origin/main HEAD` confirms conflicts in `scripts/install.sh` and `scripts/install-lib.mjs`. The implementer's report cited "GitHub 503 / workflow debounce" as the reason checks weren't visible on the new head SHA — that was a red herring; the real blocker is merge conflicts, not CI.

## Mechanical gates (against the pre-rebase branch)

- `npm test`: vitest **919/919**, node:test **559/559** (same counts as cycle 1, +28 from cycle 1 still in place).
- `bash scripts/check-duplication-gate.sh`: **PASS** — 0 clones across 5 changed files at min-tokens 70.
- `node --test src/server/*.test.mjs src/relay/agent/*.test.mjs scripts/install.test.mjs`: 67/67.

## Dedupe evaluation (the work is correct; just blocked by base)

- **`src/server/index.mjs` (+95 / -lines):**
  - Two new module-scope helpers: `respondJson(res, status, obj)` and `requireLoopback(req, res, errorMessage)`. The `isLocalDirectRequest` semantics are unchanged — `requireLoopback` calls the same helper with the same `req.socket?.remoteAddress` + `req.headers` shape, just from inside a wrapper.
  - Every call site preserves the exact status code, headers (`content-type: application/json`), and response body of the inline version it replaced. Spot-checked: `/auth/pair` 403 (`requireLoopback(req, res, "...pairing...")`), `/auth/pair` 200 (now `respondJson(res, 200, {pairing_code, box_id, expiresAt})`), `/auth/pair` 429 (rate limit), `/auth/claim` 400/200, method-not-allowed 405, 500 catch, `/relay/status` 403/200. All bit-for-bit equivalent to the cycle 1 inline code.
  - Helper scope is intentionally narrow: only the loopback-gated JSON routes (`/auth/pair`, `/auth/claim`, `/relay/status`) go through the helpers. The implementer correctly **did not** blanket-refactor other `writeHead+JSON.stringify` sites (`/api/peek`, `/api/upload`) — those are out of BET-155's scope and the e2e smoke covers them. Conservative is right.

- **`src/relay/agent/index.test.mjs` (+~90 / -290, net -200):**
  - `makeStubAgent(opts)` collapses the 8-line `relay + clock + createRelayAgent({auth, connect, setTimer, clearTimer, log, warn})` setup that was repeated across ~12 tests. `opts.failFirst` flows through to `makeFakeRelay({failFirst})`, the rest spread into the agent constructor (`backoff`, `localFetch`, etc.) — so tests can still override knobs with one argument.
  - `captureFetchHeaders(body)` wraps the stub-fetch preamble (`origFetch = globalThis.fetch; ... ; finally restore`) used by the 3 `makeDefaultLocalFetch` tests. The `finally` restore is preserved, so a thrown assertion can't leak the stub to sibling tests.
  - 27 tests still pass; no test's *meaning* changed (asserts are identical to cycle 1, just the setup is shorter).

## Nits (non-blocking)

**Nit — implementer missed the merge conflict.** The PR comment said "CI is blipping (503 from actions API)" and asked the reviewer to verify locally as the source of truth. Locally is fine for typecheck/tests, but `gh pr view --json mergeable` is the gate that decides whether the PR can actually merge. Future PRs in this workspace should include `gh pr view <N> --json mergeable,mergeStateStatus` in the verification block — a green local gate on a CONFLICTING PR isn't a PASS, it's a Block waiting to be filed.

## **Block (fix-here): stale base — rebase onto current `origin/main`**

The PR branched from `main @ e856da8` (PR created 21:46 UTC). Since then, two merges landed on main:
- `929c751` (22:07 UTC) — BET-153 install chat stack provisioning — touches `scripts/install.sh`, `scripts/install-lib.mjs`, `scripts/install.test.mjs`, `scripts/release/pack.mjs`, `scripts/systemd/opencode-serve.service`
- `a109b6b` (22:16 UTC) — BET-154 desktop distribution — touches `README.md`, `electron-builder.yml`, `scripts/release/desktop.sh`, `src/main/autoUpdate.ts`, `website/index.html`

`git merge-tree origin/main HEAD` confirms conflicts in `scripts/install.sh` and `scripts/install-lib.mjs` (BET-153 modified both, and so does this PR). `src/server/index.mjs` and `src/relay/agent/index.test.mjs` rebase cleanly (BET-153/154 didn't touch them).

**Fix:**

```bash
git fetch origin
git rebase origin/main
# resolve conflicts in scripts/install.sh and scripts/install-lib.mjs
# (combine BET-153's "section 6 chat-stack provisioning" + "section 7 systemd unit"
# with this PR's "section 7 → 8 health-wait + relay-handshake + box_id heredoc").
# After resolving:
npm run typecheck && npm test
bash scripts/check-duplication-gate.sh
gh pr checks 117
git push --force-with-lease
```

Do NOT widen `scripts/duplication-gate-ignore.txt` or add `src/server/index.mjs` to it — that's the escape hatch the human specifically prohibited (per the PM's routing note, and per the strict policy in the gate's own header).

After the rebase I'll re-verify the dedupe still applies cleanly (the helpers should land as-is on the rebased tree since none of the conflicting files import them).

antoinedc:
# PR review — BET-155 (cycle 3): rebase + dedupe v2 commit `244f7b3`

**Verdict: 0 Blocks, 0 Questions, 0 Nits. Routing to bui-pm for merge.**

The rebase is correct (merge-base is `origin/main @ a109b6b`; BET-153 and BET-154 both incorporated), the additional dedupe is real (independently verified with jscpd@5), and all required CI checks pass on the new head SHA.

## Mechanical gates

- `npm test`: vitest **919/919**, node:test **576/576** (was 559 in cycle 2; +17 from additional tests).
- `bash scripts/check-duplication-gate.sh` (local jscpd@4.2.3): **PASS — 0 clones** across the 5 changed files.
- **jscpd@latest (CI version, Rust binary) on the same 5 files: 0 clones.** This is the verification I should have run in cycle 2 — the local gate's PASS was misleading because npm-installed jscpd@4 silently skips files >1000 lines (the threshold `src/server/index.mjs` crossed after BET-153 merged). CI's jscpd@5 has no such default and would have caught the residual clones my cycle 2 review didn't see.
- Targeted re-runs: `node --test src/server/*.test.mjs src/relay/agent/*.test.mjs scripts/install.test.mjs` 67/67.

## CI on `244f7b3` (verified live, not from implementer report)

- `typecheck-test`: **PASS** ✓
- `duplication-gate` (blocking STRICT): **PASS** ✓ (was FAIL with 13 clones before this commit)
- `secret-scan`: **PASS** ✓
- `dep-audit`: **PASS** ✓
- `E2E Smoke Test`: **PASS** ✓
- `duplication-report` (advisory): FAIL — **transient infra, not a defect.** Pulled the run log: the detector wrote its 27-line report successfully; the failure is the next step (`Post / update sticky PR comment`) hitting a GitHub API 503 (`<html>...Unicorn!...503...` body in the annotation). This is GitHub-side rate-limiting on the actions API, not anything wrong with the diff.

`mergeStateStatus: UNSTABLE` is solely the `duplication-report` advisory failing — `gh pr view --json mergeable` returns `"MERGEABLE"`.

## Rebase verification

`git log --oneline origin/main..HEAD`:
```
244f7b3 refactor(server): dedupe jscpd-flagged clones — BUS_PUBLISH_DEPS, handleApiDelete, readBody
7e59be8 refactor(server,test): dedupe loopback gate + JSON response + agent-fixture setup
4bd466d feat(relay): box-side agent auth overwrite + startup wiring + install handshake check
```

Three commits. `git merge-base HEAD origin/main` = `a109b6b` = `git rev-parse origin/main` — branch is fresh on current main. The 2-file conflict from cycle 2 (in `scripts/install.sh` and `scripts/install-lib.mjs`) was resolved by combining BET-153's new chat-stack + systemd sections with this PR's relay-handshake + box-id heredoc additions. The other 4 files (3 prod + 1 test) rebased cleanly.

## Dedupe v2 evaluation (the work that closes the cycle 2 miss)

Three new extractions in `src/server/index.mjs`, all behavior-preserving:

1. **`BUS_PUBLISH_DEPS` constant** — 8 inline `{ publish: (evt) => bus.publish(evt) }` literals collapsed to one named const at module scope. Each old call site now passes the const; the const-equal-to-inline-object invariant is trivially preserved. Removed 7 of the 13 flagged clones.

2. **`handleApiDelete(req, url, res, deleteFn)`** — the 12-line DELETE-handler block (`?id=<id>` → 400 if missing → `await deleteFn(id, deps)` → 200 `{deleted: bool}`) extracted into a single helper. Used by `/api/schedule`, `/api/webhook`, `/api/secrets` (3 sites → 1 helper). The caller's `return;` after `await handleApiDelete(...)` is preserved, so the response-already-sent-then-no-op control flow is byte-equivalent to the old inline version. Removed the 22-line clone cluster.

3. **`readBody(req, { parse, limit })` consolidating `readJsonBody` + `readRawBody`** — the size-cap + chunk-collect loop was 13 identical lines apart from the parse-or-not tail. Now one function with `parse: true|false`; `readJsonBody` and `readRawBody` are 1-line shims that select the parse mode. Webhook HMAC still gets the raw bytes (no trim, no re-serialize) because the `parse=false` branch returns `Buffer.concat(chunks).toString("utf-8")` directly — identical to the old `readRawBody`. The `parse=true` branch trims, checks empty, parses — identical to the old `readJsonBody`. Removed the 14-line clone.

Plus the 61 `writeHead + content-type + JSON.stringify + end` sites collapsed to `respondJson` (carried over from `7e59be8`). File: **1340 → 1273 lines** (−67 net).

All 3 original sections still intact after the rebase:
- §1 Agent auth overwrite: `makeDefaultLocalFetch(localBase, auth)` + `Bearer <box_token>` overwrite still at `src/relay/agent/index.mjs:201`.
- §2 Server startup + `/relay/status`: `createRelayAgent` startup wiring + loopback-gated `/relay/status` still in `src/server/index.mjs`.
- §3 install.sh handshake: `waitForRelay` + `readBoxIdentity = loadAuth` in `scripts/install-lib.mjs`; heredoc + relay-check branch in `scripts/install.sh`.

Hygiene grep: `relayEnabled` only in `src/relay/agent/`, `src/server/index.mjs`, `scripts/install.sh`. No renderer/main/mobile references.

## On my cycle 2 miss (for the record)

The implementer's "local duplication gate PASSED = ready" report was technically correct given their tools, but I should have caught that the local `npx --yes jscpd` was resolving to `jscpd@4.2.3` (which silently skips >1000-line files) and not the CI's `jscpd@5.x`. The dedupe v1 commit's gate PASS was true for the local scanner but false for the CI scanner. This time I ran jscpd@latest explicitly and confirmed 0 clones in the 5 changed files. The implementer caught the gap and corrected it in v2; the next reviewer on a similar issue should run jscpd@latest (or pin `jscpd@5.x` in the gate script) up front.

Routing to **bui-pm** for merge. Per the iteration cap I'll escalate if a cycle 4 review is needed.

PR #129, opened 2026-07-17
Installer v2 — self-contained prebuilt tarball (vendored Node + prebuilt node_modules)

antoinedc:
## Summary

Stop installing prerequisites on the box. The one-liner installer (`curl -fsSL https://mantaui.com/install.sh | bash`) used to silently sudo+apt-install Node + build-essential on the target box; every launch-gate E2E failure traced back to that seam. Now `npm run pack` produces a self-contained user-space tarball (rustup/uv/opencode-style) that ships a vendored Node 20.20.2 + production `node_modules` with node-pty's binding already compiled.

**Refs BET-160 §2 (launch gate).** Supersedes BET-170/172's "layer 2" prerequisite-bootstrap mechanism. Root-cause fix for BET-171 F1/F2/F3/F4.

## What changed

- **`pack.mjs`** downloads Node 20.20.2 from nodejs.org, verifies against SHASUMS256.txt, extracts under `runtime/node/`, then runs `npm ci --omit=dev` using the vendored node's ABI so node-pty compiles once against the runtime the box will use. Writes a flat `key=value` manifest alongside the tarball (`version`, `file_linux_x64`, `sha256_linux_x64`). Tarball name encodes arch (`manta-<version>-linux-x64.tar.gz`).
- **`install.sh`** becomes a thin orchestrator: fetch manifest (bash, no node needed yet) → download tarball → sha256-verify → extract → atomic swap into `~/manta` with `.prev` rollback → write systemd units pointing at the vendored node → start → mint pairing code. All node invocations use `$NODE` (the vendored binary) explicitly; nothing relies on a system node.
- **`install.sh`** deletes `detect_distro_id`, `package_install_mode`, `install_node_via_{apt,dnf,yum}`, `bootstrap_node`, `bootstrap_build_essential`, the `npm ci` block, the inline chicken/egg tarball-URL resolver, the `require_cmd node/npm` lines, the `node_major` version check, and the redundant mid-script tmux re-check.
- **`install-lib.mjs`** deletes `resolveTarballUrl`, `isValidVersion`, and the `tarball-url` CLI subcommand (no callers). `renderShellConfig` no longer emits `MANTA_TARBALL_URL` (install.sh derives the URL from the manifest).
- **`install.test.mjs`** deletes the bootstrap_* test helpers and adds tests for `manifest_get`, `verify_sha256`, `require_arch` using the existing `MANTA_INSTALL_TEST_MODE=1` source-the-script pattern.
- **`publish.sh`** uploads tarball + manifest in order, with the `manta-latest.txt` cp going LAST (atomicity). The verify step now re-fetches the manifest and re-checks the live tarball sha — script/tarball/manifest drift (the F4 class from BET-171) fails the publish loudly.
- **`README.md`** updates installer prerequisites (curl/tar/sha256sum/tmux/git; everything else ships in the tarball — no sudo, no compilers, no Node preinstall).

## Net LoC delta

| file | before | after | Δ |
|------|-------:|------:|---:|
| `scripts/install.sh` | 755 | 556 | **−199 (−26%)** |
| `scripts/install.test.mjs` | 1113 | 942 | −171 |
| `scripts/install-lib.mjs` | 701 | 670 | −31 |
| `scripts/release/pack.mjs` | 140 | 354 | +214 (vendored runtime + prebuilt deps) |
| `scripts/release/publish.sh` | 187 | 191 | +4 |
| `README.md` | — | — | ±28 |

Net: **−183 LoC** across 6 files. The deleted install.sh code (the big minus) is the success metric.

## Deleted functions (in `install.sh`)

- `detect_distro_id`, `package_install_mode`
- `install_node_via_apt`, `install_node_via_dnf`, `install_node_via_yum`
- `bootstrap_node`, `bootstrap_build_essential`
- `require_cmd node`, `require_cmd npm` lines + `node_major` version check
- the inline `TARBALL_URL="${host}/releases/manta-${MANTA_VERSION}.tar.gz"` resolver
- the section-5 `npm ci` block
- the redundant mid-script tmux re-check (collapsed — tmux is already a hard `require_cmd` at step 1)

### Checklist (all pass)

1. ✅ `npm run typecheck` — exit 0
2. ✅ `npm test` — 648 pass / 0 fail
3. ✅ `node scripts/release/pack.mjs --skip-build` — produces both `dist/manta-0.0.1-linux-x64.tar.gz` (66 MB) and `dist/manta-0.0.1.txt` (140 B)
4. ✅ Tarball inspection:
   - `runtime/node/bin/node` present (1 match)
   - `node_modules/node-pty/build/Release/*.node` present (2 matches — `pty.node` + spawn helper)
   - `sha256sum` of the tarball matches `sha256_linux_x64` in `manta-0.0.1.txt`
5. ✅ Sandbox smoke: `require("/tmp/mv2/node_modules/node-pty")` under the vendored node prints `pty ok` (the prebuilt binding loads)
6. ✅ `bash -n scripts/install.sh` — syntax clean
7. ✅ `grep -cE 'sudo|apt-get|dnf |yum |nodesource' scripts/install.sh` returns only:
   - the header comment that explicitly says we DON'T use these
   - 5 `require_cmd` hint strings (apt-get install … — only printed when a prereq is missing)
   - 1 warn string telling the user to run `sudo loginctl enable-linger` themselves (no auto-execute)

### Base

Base: origin/main @ `aa9edaf`

## Out of scope (handled in BET-171 / BET-172 follow-ups)

- Fresh-VPS E2E re-run against the new installer
- `publish.sh` against prod + `docs/launch-e2e.md` updates
- Linux-arm64 tarball (manifest format already has room)
- Single-binary packaging (Node SEA / bun compile)
- Auto-update of an installed box beyond "re-run the installer"

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,17 +129,27 @@ errors/done work from the PWA.
 
 ## Installer reference
 
-The installer downloads a pre-built tarball (no dev toolchain needed on the
-box), runs `npm ci --omit=dev`, installs the `manta-server` systemd --user
-unit, enables linger, health-waits, and prints a pairing code.
+The installer downloads a self-contained tarball that ships a vendored Node
+runtime + prebuilt production `node_modules` (node-pty's native binding
+already compiled). The box only needs `curl`, `tar`, `sha256sum`, `tmux`,
+and `git` — no Node preinstall, no compilers, no `sudo`, no package-manager
+calls. (Every launch-gate E2E failure previously came from the installer
+silently trying to `apt install nodejs` and friends; v2 removes that
+seam entirely.)
+
+What the installer does: fetch a key=value manifest over HTTPS → read
+`file_linux_x64` + `sha256_linux_x64` from it → download the tarball →
+sha256-verify → extract → atomic swap into `~/manta` → write systemd units
+pointing at the vendored node → start → print a pairing code.
 
 Overrides (env):
 
 | Var | Default | Purpose |
 |-----|---------|---------|
-| `MANTA_TARBALL_URL` | (built from host+version) | full tarball URL — local testing / mirror |
-| `MANTA_RELEASE_HOST` | `https://mantaui.com` | host for the default tarball URL |
+| `MANTA_TARBALL_URL` | (built from manifest) | full tarball URL — local testing / mirror (skips manifest fetch + sha256 check, with a warn) |
+| `MANTA_RELEASE_HOST` | `https://mantaui.com` | host for the manifest + tarball |
 | `MANTA_HOME` | `~/manta` | where the code is unpacked |
+| `MANTA_VERSION` | `latest` | manifest version to fetch (e.g. `0.0.1`) |
 | `MANTA_MOBILE_PORT` | `8787` | server port |
 
 Manage: `systemctl --user {status,restart} manta-server`, logs
@@ -215,7 +225,9 @@ session by POST). Install/update = copy to `~/.config/opencode/tools/`
 In order. No decisions, no extra steps:
 
 1. Bump `package.json` version.
-2. `npm run pack` — produces `dist/manta-<version>.tar.gz`.
+2. `npm run pack` — produces `dist/manta-<version>-linux-x64.tar.gz` (the
+   self-contained tarball) AND `dist/manta-<version>.txt` (the key=value
+   manifest install.sh fetches first).
 3. (Mac only, on a Mac) `bash scripts/release/desktop.sh` — produces
    `dist/desktop/*.dmg` and the `latest-*.yml` updater feeds. Linux
    builds run on any host.
@@ -267,7 +279,9 @@ steps in each file's header are human-only (agent Hard Rule #4 forbids
 - **Monitoring** (`scripts/prod/healthcheck.mjs`, scheduled every 10 min
   on the dev box via `schedule_create`): off-site probes of `mantaui.com`,
   `relay.mantaui.com` (expects 401 — the auth gate IS the healthy
-  answer), `app.mantaui.com`, `/install.sh`, `/releases/manta-latest.tar.gz`.
+  answer), `app.mantaui.com`, `/install.sh`, `/releases/manta-latest.tar.gz`
+  (and the live `manta-latest.txt` manifest's `sha256_linux_x64` must match
+  the served tarball).
   On failure the opencode turn calls `notify` urgent:true naming the
   failing URL.
 - **Log caps** (`scripts/prod/systemd-journald.conf`,
--- a/scripts/install-lib.mjs
+++ b/scripts/install-lib.mjs
@@ -32,8 +32,9 @@ export const DEFAULT_PORT = 8787;
 export const HEALTH_PATH = "/auth/status";
 
 // Default release host. The repo is private; CI publishes a versioned tarball
-// the installer downloads (no git clone on the VPS). `MANTA_TARBALL_URL` overrides
-// the whole URL for local testing; otherwise we build it from host + version.
+// + a manifest file the installer downloads (no git clone on the VPS).
+// `MANTA_TARBALL_URL` overrides the whole flow (skips manifest fetch + sha256)
+// for local testing.
 export const DEFAULT_RELEASE_HOST = "https://mantaui.com";
 
 // Where the box lives once installed. `~/.manta/` (auth.json + config.json)
@@ -66,7 +67,7 @@ function envVal(env, key) {
  *   buiHome      — where the code is unpacked (MANTA_HOME || ~/manta)
  *   authDir      — ~/.manta (never inside buiHome)
  *   authFile     — ~/.manta/auth.json (idempotency probe target)
- *   tarballUrl   — explicit MANTA_TARBALL_URL, else null (resolved per-version)
+ *   tarballUrl   — explicit MANTA_TARBALL_URL, else null (overrides manifest fetch)
  *   releaseHost  — MANTA_RELEASE_HOST || DEFAULT_RELEASE_HOST
  *   port         — MANTA_MOBILE_PORT || DEFAULT_PORT
  *   healthUrl    — http://127.0.0.1:<port>/auth/status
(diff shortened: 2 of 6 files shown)


antoinedc:
## Review: BET-173 — Installer v2 (self-contained prebuilt tarball)

**Base:** `origin/main` @ `aa9edaf` — PR branched from current main (merge-base == `origin/main`). No stale-base / phantom-deletion artifacts.

**PR state:** DRAFT (still draft, not flipped to ready — that's the PM's call per scope).

### Mechanical gate

- `npm test` — **exit 0**, 648 pass / 0 fail / 0 skip (vs base 655 pass / 0 fail — net −7 tests, all of which are deleted bootstrap_* / `resolveTarballUrl` / `isValidVersion` cases replaced by `manifest_get` / `verify_sha256` / `require_arch`).
- Anti-spaghetti: **0 clones** in changed scope (jscpd, 3 `.ts/.tsx/.mjs` files).
- CI: typecheck-test, duplication-gate, dep-audit, secret-scan, duplication-report, E2E Smoke Test — **all green**.

### Spec verification — independently re-run

| # | Spec item | Re-run result |
|---|-----------|---------------|
| 1 | `npm run typecheck && npm test` green | ✅ both exit 0 |
| 2 | `pack --skip-build` produces both files | ✅ `dist/manta-0.0.1-linux-x64.tar.gz` (66.7 MB) + `dist/manta-0.0.1.txt` (140 B) |
| 3a | `runtime/node/bin/node` in tarball | ✅ 1 match |
| 3b | `node_modules/node-pty/build/Release/*.node` in tarball | ✅ 2 matches (`pty.node` + spawn helper) |
| 3c | tarball sha256 matches `sha256_linux_x64` in manifest | ✅ byte-for-byte (`4ed7475d…f2a2b`) |
| 4 | Sandbox smoke: `require("/tmp/mv2/node_modules/node-pty")` under vendored node | ✅ prints `pty ok`, exit 0 (proves prebuilt binding loads under vendored Node 20.20.2 — ABI match confirmed) |
| 5 | `bash -n scripts/install.sh` | ✅ syntax clean |
| 6 | `grep -cE 'sudo\|apt-get\|dnf \|yum \|nodesource' scripts/install.sh` — only hints/comments | ✅ no executable package-manager/sudo calls remain (only: header comment, 5 `require_cmd` hint strings printed only when prereq missing, 1 warn string telling user to run `sudo loginctl enable-linger` themselves) |

### Priorities called out in the dispatch — verified

**1. Atomic swap sequence (`install.sh`)** — clean. `mv "$MANTA_HOME" "$MANTA_HOME.prev"` → `mv "$WORK/pkg" "$MANTA_HOME"` with a restoration-on-failure path that puts `.prev` back if the final `mv` fails (lines 451-461). End-of-main cleanup `rm -rf "$MANTA_HOME.prev"` runs only after the install succeeds. `~/.manta/` (identity) lives outside `$MANTA_HOME` and is untouched. ✓

**2. Every `node` invocation routes through vendored `"$NODE"`** — verified by grep. Zero bare `node ` or `node -e` calls remain; 11 invocations all use `"$NODE"` (print-config, check-identity, merge-opencode-config, render-systemd-unit, two `waitForHealth`/`waitForRelay` blocks, `manta-pair.mjs`, `readBoxIdentity`, the nohup server start, and the package.json version probe). `NODE_BIN="$(command -v node)"` is gone — substitution in the systemd template uses `$NODE`. ✓

**3. `pack.mjs` node-pty compiled against vendored node (ABI match) + sha256 in manifest matches tarball** — verified. `runPrebuiltDeps` calls the vendored `<stage>/runtime/node/bin/npm` directly and prepends the vendored bin dir to PATH so node-gyp subprocesses find the matching ABI. Manifest is written AFTER tar with a fresh `createHash("sha256").update(tarballBytes).digest("hex")` — byte-matches. The sandbox smoke (item 4 above) is the load-bearing proof that the binding actually loads. ✓

**4. No executable package-manager/sudo calls in install.sh** — verified above in checklist item 6. ✓

### Acceptance-criteria checklist (extracted from issue body)

- [x] **Arch: linux-x64 only.** `uname -m` x86_64 gate (`require_arch`) with clear die message; non-x86_64 fails loudly.
- [x] **Vendored Node pinned at `20.20.2`** (`const NODE_VERSION = "20.20.2"` at top of pack.mjs); downloads from `https://nodejs.org/dist/v20.20.2/` and verifies against `SHASUMS256.txt`. `.tar.gz` (not `.tar.xz`).
- [x] **Flat `key=value` manifest** (not JSON), three lines: `version=`, `file_linux_x64=`, `sha256_linux_x64=`. Parseable in bash before any node exists.
- [x] **No templating of install.sh** — served verbatim from the repo; coupling lives in the manifest.
- [x] **`MANTA_TARBALL_URL` override** — skips manifest fetch + sha256 check with a `warn`. Tests/E2E still work.
- [x] **tmux/git policy: die-with-instructions, ALWAYS** — no root branch, no distro detection, no auto-install.
- [x] **No sudo after this change** — `loginctl enable-linger` keeps its `|| warn` fallback (unchanged).
- [x] **node-pty stays as-is** — `node-pty@^1.0.0`, compiled once on the build box using vendored node's ABI.
- [x] **`$WORK` moved to `$HOME`** — `mktemp -d "$HOME/.manta-install.XXXXXX"` guarantees same-filesystem rename (atomic) and avoids noexec /tmp.
- [x] **Old `manta-latest.tar.gz` publishing removed** — publish.sh no longer refreshes it (kept the dead-but-served file alone per spec).
- [x] **Deleted functions**: `detect_distro_id`, `package_install_mode`, `install_node_via_apt/dnf/yum`, `bootstrap_node`, `bootstrap_build_essential` — all gone from install.sh.
- [x] **Sanity gates** on extracted tarball — `bin/node`, `node_modules/`, `scripts/install-lib.mjs`, `src/server/index.mjs` all checked.
- [x] **RELEASE.json** gets `node: "20.20.2"` and `arch: "linux-x64"` fields.
- [x] **Test deletions**: 16 `bootstrap_*` cases + 4 `resolveTarballUrl` + 1 `isValidVersion` removed; 9 new cases (`manifest_get` ×4, `verify_sha256` ×2, `require_arch` ×3) added via the existing `MANTA_INSTALL_TEST_MODE=1` source-the-script pattern.
- [x] **Net LoC for install.sh**: 755 → 556 (**−199, −26%**). Net across the 6 files: **−183 LoC**.
- [x] **publish.sh atomicity**: tarball scp'd first, manifest scp'd second, `manta-latest.txt` copy happens LAST. Verify step re-fetches the live manifest, parses `version`/`file_linux_x64`/`sha256_linux_x64`, and re-checks the live tarball sha256 — script/tarball/manifest drift fails the publish loudly.
- [x] **README prerequisites updated** to the new list (curl/tar/sha256sum/tmux/git; everything else ships in the tarball).

### Findings

**0 Blocks, 0 Questions, 0 Nits.**

The change is exactly the deletion-led, root-cause-fix the issue specified. The install.sh body is now a thin orchestrator that does only what the spec requires (manifest fetch → sha256 verify → atomic swap → systemd unit install → start → pair). The "self-contained user-space tarball" archetype is fully realized: the box needs curl/tar/sha256sum/tmux/git and nothing else, and the install is verified end-to-end before systemd starts it.

Net LoC delta: `install.sh` −199, `install-lib.mjs` −31, `install.test.mjs` −171 (net of −302 test LoC and +131 new test LoC); `pack.mjs` +214 (the vendored-runtime machinery — load-bearing), `publish.sh` +4 (atomicity + manifest drift check), `README.md` ±28. **Total: −183 LoC across 6 files.**

### Out of scope (intentionally NOT touched, per spec)

- Fresh-VPS E2E re-run against the new installer — stays under BET-171/BET-172 follow-ups.
- `publish.sh` against prod + `docs/launch-e2e.md` updates.
- linux-arm64 tarball (manifest format has room; deliberately data, not code, here).
- Single-binary packaging (Node SEA / bun compile).
- Auto-update of an installed box beyond "re-run the installer."

---

**Routing:** passing to `bui-pm` for merge hand-off.

PR #140, opened 2026-07-20
BET-187: ship console + structured events to Axiom across all runtimes

antoinedc:
Cross-runtime log shipping to Axiom for cross-side debugging (mobile PWA + desktop renderer + bui-server + relay). One shared logShip.mjs module, no per-call-site migration of existing console.* call sites, silent no-op when no token is configured.

## What ships

- **Every runtime** wraps its console once (captureConsole). All existing `[push]` / `[opencode-pump]` / `[relay-agent]` / `[auth]` / `[bui]` console lines ship transparently.
- **6 structured ship() calls** in httpApi.ts cover WS state transitions, RPC failures, slow RPCs, and resync events so an AI querying Axiom can correlate both sides of a phone↔box failure by timestamp.
- **Browser event hooks** in renderer/log.ts (error / unhandledrejection / online / offline / visibilitychange / pagehide) cover the standard renderer signals.

## What's NOT changed

- No existing log call site modified. captureConsole covers them all.
- No bui-server proxy for renderer logs (CORS verified open; direct ship is the design — when the phone↔box path is broken, logs must still escape).
- No sampling, no rate limiting beyond maxBuffer, no log-levels config, no query integration.
- No mobile Settings UI (token entered once on desktop; mobile reads server config).
- No new dependencies. `fetch` is global in Node >=18 + the browser.

## Files changed

```
 src/relay/index.mjs         |  12 +
 src/renderer/Settings.tsx   |  46 +-
 src/renderer/api/httpApi.ts |  40 +-
 src/renderer/main.tsx       |   9 +
 src/renderer/store.ts       |  12 +
 src/renderer/log.ts         | 145 + (new)
 src/server/index.mjs        |  17 +
 src/shared/logShip.d.mts    |  50 + (new)
 src/shared/logShip.mjs      | 248 + (new)
 src/shared/logShip.test.ts  | 313 + (new)
 src/shared/types.ts         |  12 +
```

## Verification

- `npm run typecheck` green (both tsconfig.node.json and tsconfig.web.json clean)
- `npm test` — 1049 vitest pass / 755 node:test pass / 1 pre-existing node:test fail
  - `not ok 553 sendApnsFanout: no cfg → silent no-op` reproduces on origin/main (verified via stash + checkout main), unrelated to this slice. The test expects `called === false` but the actual is `true` — looks like a test bug for the no-cfg APNs path; surfacing in case the reviewer wants to file it separately.
- 27 new unit tests in src/shared/logShip.test.ts cover:
  - log() event shape (timestamp/source/device/level/spread fields)
  - maxBatch-triggered flush, empty-buffer no-op flush
  - single-flight guard on overlapping flushes
  - failed flush retains events, non-2xx re-prepends
  - maxBuffer eviction + droppedCount synthetic warn event on next success
  - first-flush failure uses ORIGINAL console.warn (no infinite loop)
  - stop() final flush, timer cleared
  - formatConsoleArgs: string/number/boolean/null/Error/circular/long-string
  - captureConsole: wrap+restore, idempotent double-wrap, level mapping, throw-doesnt-break-original
  - resolveAxiomConfig: env-wins, config-only, default dataset, no-token null

## Ops note (out of scope for this PR)

Prod relay needs `Environment=MANTA_AXIOM_TOKEN=...${TOKEN}` in its systemd unit on the prod box. The token lives in the bui Secrets card; a human action attaches it after merge.

**Base:** origin/main @ 26319f9

### DIFF
--- a/src/relay/index.mjs
+++ b/src/relay/index.mjs
@@ -41,6 +41,7 @@ import {
   FRAME_TYPES,
 } from "./protocol.mjs";
 import { openStore, BOX_STATUS, hashToken, hashEquals } from "./store.mjs";
+import { createLogShipper, captureConsole, resolveAxiomConfig } from "../shared/logShip.mjs";
 
 export const DEFAULT_RELAY_PORT = 20787;
 
@@ -767,6 +768,17 @@ const isMain =
   process.argv[1]?.endsWith("src/relay/index.mjs");
 
 if (isMain) {
+  // BET-187: ship every console.* (incl. the [relay] listening banner)
+  // to Axiom when MANTA_AXIOM_TOKEN is set. Relay runs on the prod box
+  // with no `~/.manta` directory, so config is null — env-only. Silent
+  // no-op when the env var is unset.
+  {
+    const axiomCfg = resolveAxiomConfig({ env: process.env, config: null });
+    if (axiomCfg) {
+      captureConsole(createLogShipper({ ...axiomCfg, source: "relay", device: "relay" }));
+    }
+  }
+
   const relay = createRelayServer();
   relay
     .start()
--- a/src/renderer/Settings.tsx
+++ b/src/renderer/Settings.tsx
@@ -36,6 +36,8 @@ export function Settings({ onClose }: { onClose: () => void }) {
     iosBuildRepoPath,
     iosSimulatorName,
     launcherFlags,
+    axiomToken,
+    axiomDataset,
     refresh,
   } = useStore();
 
@@ -71,6 +73,10 @@ export function Settings({ onClose }: { onClose: () => void }) {
   const [groqKey, setGroqKey] = useState(groqApiKey);
   const [voiceTrModel, setVoiceTrModel] = useState(voiceTranscriptionModel);
   const [voiceCmdModel, setVoiceCmdModel] = useState(voiceCommandModel);
+  // Axiom log shipping (BET-187) — co-located with Voice per the spec
+  // (mirrors groqApiKey's row pattern). Empty token → silent no-op.
+  const [axiomTok, setAxiomTok] = useState(axiomToken);
+  const [axiomDs, setAxiomDs] = useState(axiomDataset);
   // Mobile pairing — one-time code minted on demand. The QR encodes the
   // CANONICAL box-form `manta://pair?box=<boxId>&code=<6-digit>` payload
   // (BET-177 §2.4 — the old `?id=&token=` form was a parsePairPayload
@@ -101,7 +107,9 @@ export function Settings({ onClose }: { onClose: () => void }) {
     setIosRepoPath(iosBuildRepoPath);
     setIosSimName(iosSimulatorName);
     setLauncherFlagValues(launcherFlags ?? {});
-  }, [defaultModel, skillRegistryUrls, cacheTtl, allowAgentPush, autoRenameSessions, downloadsDir, groqApiKey, voiceTranscriptionModel, voiceCommandModel, capExecutorEnabled, iosBuildRepoPath, iosSimulatorName, launcherFlags]);
+    setAxiomTok(axiomToken);
+    setAxiomDs(axiomDataset);
+  }, [defaultModel, skillRegistryUrls, cacheTtl, allowAgentPush, autoRenameSessions, downloadsDir, groqApiKey, voiceTranscriptionModel, voiceCommandModel, capExecutorEnabled, iosBuildRepoPath, iosSimulatorName, launcherFlags, axiomToken, axiomDataset]);
 
   // Fetch available models once (non-fatal — Settings works even if opencode is unreachable).
   useEffect(() => {
@@ -147,6 +155,8 @@ export function Settings({ onClose }: { onClose: () => void }) {
         groqApiKey: groqKey.trim(),
         voiceTranscriptionModel: voiceTrModel.trim(),
         voiceCommandModel: voiceCmdModel.trim(),
+        axiomToken: axiomTok.trim(),
+        axiomDataset: axiomDs.trim(),
         capExecutorEnabled: capExecutorOn,
         iosBuildRepoPath: iosRepoPath.trim(),
         iosSimulatorName: iosSimName.trim(),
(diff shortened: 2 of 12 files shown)


antoinedc:
## Review — BET-187 / PR #140 (cycle 1)

Overall: clean shape (single shared module, console-wrap, 6 structured `ship()` calls, no call-site edits, env-var override wired in 3 places, config plumbing mirrors `groqApiKey`). One direct spec violation that blocks the entire shipping pipeline in the default-config case.

### Base + mechanical gates

- `npm test` — 755 pass / 1 pre-existing fail (`push.test.mjs:981 sendApnsFanout: no cfg → silent no-op`). Confirmed the failure reproduces on `origin/main` (checked out the test + impl files and re-ran) — unrelated to this slice.
- Branch `multica/BET-187-log-ship-axiom` branched off `origin/main @ 26319f9`; PR base is fresh `origin/main @ 7915bf2`. No phantom deletions.

### Acceptance-criteria verification (independent)

| # | Criterion | Status |
|---|---|---|
| 1 | No-token silent no-op (no fetches, no console noise) | OK — `logShip.mjs:42` returns null without token; `log.ts:72` early-returns; relay + server blocks gate on `if (axiomCfg)`. |
| 2 | With `axiomToken` set, server logs ship `source:"server"` | Structurally wired (server `index.mjs:30-36` before `createBus()`); not E2E-testable here. |
| 3 | Mobile PWA WS/RPC/network events ship `source:"mobile"` + stable device id, including when `/rpc` is failing | Structurally wired (`main.tsx` init + `httpApi.ts` 6 ship() calls + direct-to-Axiom `fetch` with `keepalive:true`); not E2E-testable here. |
| 4 | Renderer perf untouched (no per-render work) | OK — `log.ts` is module-level singleton + init-time listeners only. |
| 5 | No existing log call site modified | OK — `git diff` confirms only the 6 spec-listed `ship()` calls added; existing `[push]` / `[opencode-pump]` / `[auth]` / `[relay-agent]` / `[bui]` console.* sites untouched. |
| 6 | `npm run typecheck && npm test` green + new tests cover Part 6 list | OK except — see Block. |

### Findings

**Block (fix-here): Default dataset name is `"bui"`, spec requires `"manta"`.**

The issue spec is unambiguous and repeats the name three times:

- Design decision #4: *"One dataset for everything, default name **`manta`**."*
- Top-of-issue update: *"Dataset name is **`manta`** everywhere below (default in `resolveAxiomConfig`)."*
- Part 5 (Settings.tsx): *"'Axiom dataset' (text, placeholder **`manta`**)."*

The implementation deviates in three places — `bui` was chosen instead of `manta`:

1. `src/shared/logShip.mjs:43` — `const dataset = env?.MANTA_AXIOM_DATASET || config?.axiomDataset || "bui";`
2. `src/shared/logShip.mjs:32` — comment: *"Dataset defaults to 'bui' so a user only needs to provide the token to start shipping."*
3. `src/renderer/Settings.tsx` — `placeholder="bui"` for the dataset input.
4. `src/shared/logShip.test.ts:77-80` — the test **encodes** the wrong default (`it("dataset defaults to 'bui'", ...)` asserts `https://api.axiom.co/v1/datasets/bui/ingest`).

Operational impact: per the issue's update note, the `manta` dataset is the only one that has been created and round-trip-verified in Axiom. A fresh install that enters the ingest token and leaves the dataset blank ships to a dataset that does not exist — Axiom will silently reject every batch (non-2xx → events re-prepend to buffer → eventually evicted by `maxBuffer` with a single synthetic `logship dropped events` warn). The pipeline appears wired but produces zero useful data in the default-config case, which is the exact "remote debugging is blind" failure mode this issue is supposed to fix.

This is a deliberate deviation (the inline comment calls out the choice), not a typo — the test was written to lock it in. The fix is small and stays inside the PR:

- `src/shared/logShip.mjs:43` — change `"bui"` → `"manta"` (and update the comment on line 32).
- `src/renderer/Settings.tsx` — change the dataset input `placeholder` from `"bui"` to `"manta"`.
- `src/shared/logShip.test.ts:77-80` — rename to `dataset defaults to 'manta'` and assert `https://api.axiom.co/v1/datasets/manta/ingest`.

### Minor notes (not blocking, mention for awareness)

- `local.mjs` correctly untouched — `configUpdate` shallow-merges, so `axiomToken`/`axiomDataset` flow through automatically.
- `rpcOptional` correctly delegates to `rpc`, so the single `rpc()` wrapper covers both call paths (Part 4 requirement).
- The desktop Settings placement is fine (co-located with Voice/observability creds). Spec only required the row pattern, not a specific tab.

### Decision

Returned to `better-ui-dev` for the dataset-name fix. One Block, very narrow scope, fix fits in this PR cleanly. Resubmit with the change and I'll re-verify.

antoinedc:
## Re-review — BET-187 / PR #140 (cycle 2)

The previous Block (dataset default) is fixed in `54275f6`. Re-verified each named location against the spec — all three use `manta` now.

### Mechanical gates

- `npx vitest run src/shared/logShip.test.ts` — 27/27 pass.
- Branch fast-forwarded to `54275f6` (the fix commit on `origin/multica/BET-187-log-ship-axiom`).

### Verifying the previous Block

| Location | Before | Now |
|---|---|---|
| `src/shared/logShip.mjs:43` (default) | `\|\| "bui"` | `\|\| "manta"` ✓ |
| `src/shared/logShip.mjs:33` (doc-comment) | `Dataset defaults to "bui"` | `Dataset defaults to "manta"` ✓ |
| `src/renderer/Settings.tsx:659` (placeholder) | `placeholder="bui"` | `placeholder="manta"` ✓ |
| `src/shared/logShip.test.ts:77` (test name) | `dataset defaults to 'bui'` | `dataset defaults to 'manta'` ✓ |
| `src/shared/logShip.test.ts:79` (assertion) | `…/datasets/bui/ingest` | `…/datasets/manta/ingest` ✓ |

Fix commit also cleaned up two product-name references (`bui-server` → `manta-server` in `logShip.mjs:1,232`), which is consistent with the workspace rebrand and unrelated to the Axiom dataset. Fine.

### Residual sweep

I ran an independent grep for `"bui"` and `'bui'` literals across every changed file (`logShip.mjs`, `logShip.test.ts`, `Settings.tsx`, `main.tsx`, `store.ts`, `log.ts`, `server/index.mjs`, `relay/index.mjs`, `types.ts`, `httpApi.ts`). One stale doc-comment remained:

> **`Nit` — `src/shared/types.ts:120`** still says `// Dataset name; defaults to "bui".`

This is a doc-comment on `AppConfig.axiomDataset`; runtime behavior is correct (the actual default in `resolveAxiomConfig` is `manta`), and no other code path reads this comment. But it's directly user-facing via IDE hover on the `axiomDataset` field, and it contradicts the spec the same way the previous Block did — the implementer's own grep claim ("only the pre-existing `bui`/`bui-server`/`bui registry` product-name references in unrelated Settings.tsx text remain") missed it.

Cosmetic; doesn't block. A one-character change (`"bui"` → `"manta"`) the implementer can fold into a follow-up — flagging in case `bui-pm` wants it tacked into this PR pre-merge since it's the same class of mistake.

### Decision

**PASS** — 0 Blocks, 0 Questions, 1 Nit. Routing to `bui-pm` for merge. The Nit is opt-in for the implementer; the runtime pipeline is correctly wired.

PR #149, opened 2026-07-20
fix(plugins): route pluginsEnabled to Mac-local + log catch-up branches (BET-207)

antoinedc:
Two defects in BET-192 Phase 3 live verification, fixed in one PR.

## Defect 1 — `pluginsEnabled` toggle wrote the wrong config file

In HTTP mode, `Settings → Plugins` toggled `pluginsEnabled` via `window.api.configUpdate` → /rpc → `local.configUpdate` → **box** config (`~/.manta/config.json`). But `capExecutor` reads it from the **Mac-local** config (`<userData>/config.json`) via its `configGetter`. Two files, no shared write path — the toggle could never arm the executor.

### Fix

- New IPC channels `plugins:set-enabled` / `plugins:get-enabled`, handled **locally** in main next to `configGet` (reuses `commit()`, so the in-memory `config` the executor reads is updated atomically with the persisted Mac-local file).
- Exposed on `window.__buiPreload` (NOT `window.api` — httpApi only reaches the box). Added to `BuiPreload` in `src/renderer/preloadAccess.ts`.
- `Settings.tsx` seeds the toggle from `pluginsGetEnabled()` on mount and persists on Save via `pluginsSetEnabled()`. The Mac-local value is the source of truth for the toggle state, not the store.
- Removed the `pluginsEnabled` mirror from `store.ts` — no other consumer.
- Box side: `local.mjs` `configUpdate` is a free-form merge (no allowlist), so no code change needed there. The stray `pluginsEnabled: true` already written to the box during verification is dead but harmless — no migration code, per spec.

### Reuse / hygiene

- Reuses `commit()` — no new local-write path.
- Reuses the existing local-IPC pattern (`configGet`) — no httpApi round-trip.

## Defect 2 — `catchUpJobs` silently strands queued jobs

`catchUpJobs` had an empty `try/catch` + an `if (!res.ok) return;` with no log. Two `plugin.write` jobs (`67697b00`, `9750d6f9`) sat `status:"queued"` on the box across reconnects and the cause was undiagnosable.

### Fix

- Log on every failure branch:
  - `console.warn("[cap] catch-up threw", err)` — replaces the empty `catch {}`.
  - `console.warn("[cap] catch-up GET failed", res.status)` — replaces the silent non-ok early return.
  - `console.log("[cap] catch-up: N queued jobs", ids)` — new, makes a persistent "0" on a real backlog visible.
- Dropped the misleading `// SSE path will deliver what catch-up misses` comment. SSE has no replay — catch-up is the only path that can recover jobs the AI queued while the Mac was offline/asleep.

### Root-cause investigation

The three candidates the issue lists all check out structurally:
- (a) stale `boxToken` — `publishRegistry` uses the same token and works, so unlikely.
- (b) `POST /api/cap/{id}/start` rejects a `queued` claim — inspected `src/server/capabilities.mjs:startJob`: only `status === "queued"` transitions to `running`. Endpoint is correct.
- (c) response envelope mismatch — `GET /api/cap` returns `{jobs:[...]}` (line 976 of `src/server/index.mjs`), executor reads `body.jobs`. Matches.

The logs above are the diagnostic primitive for the next occurrence; the implementer-of-the-next-occurrence can read the box's logs (Axiom via BET-187 once `axiomToken` is set, or stderr in dev) and pinpoint the actual cause without code archaeology.

### Reuse / hygiene

- No new claim path. `runOne` → `POST /start` is still the only claim.
- No retry/polling fallback (explicitly forbidden in the issue).

## Files changed

```
 src/main/capExecutor.ts            | 35 +++++++++++++++++++++++++++--------
 src/main/index.ts                  | 12 ++++++++++++
 src/preload/index.ts               | 11 +++++++++++
 src/renderer/Settings.tsx          | 33 ++++++++++++++++++++++++++++-----
 src/renderer/preloadAccess.test.ts |  2 ++
 src/renderer/preloadAccess.ts      |  7 +++++++
 src/renderer/store.ts              |  7 -------
 src/shared/types.ts                | 10 ++++++++++
```

**Base:** `origin/main` @ `cceec0e`

## Verification results

- PR branch `multica/BET-207-plugins-toggle-and-catchup` @ `c96b9fd`:
  - typecheck: exit `0`. Errors: none. log sha256: `139c60436aeae3ee087031cab3603b97f58acc2874526c66e944de4da209c667`
  - test: exit `0`, `771` pass / `0` fail / `0` skip.
- Base `main` @ `cceec0e`:
  - typecheck: exit `0`. Errors: none. log sha256: `139c60436aeae3ee087031cab3603b97f58acc2874526c66e944de4da209c667`
  - test: exit `0`, `771` pass / `0` fail / `0` skip.
- **Conclusion:** 0 pre-existing failures (both branches are clean), 0 new failures introduced, 0 resolved by this PR (the codebase was already green at the base). Typecheck logs are byte-identical between the branches.

Logs cached at `/tmp/typecheck-pr.log`, `/tmp/test-pr.log`, `/tmp/typecheck-base.log`, `/tmp/test-base.log` for reviewer spot-check.

## Manual Mac-leg steps (for the human)

Per the issue's verification block, this PR enables:

1. `Settings → Plugins` toggle ON → quit + relaunch → `~/.manta/plugins/` is created (proves the Mac-local flag is now set). Toggle OFF + relaunch → executor no-ops (folder may remain, but no scan/registry PUT — confirm via absence of the registry-PUT log line).
2. With the toggle ON and the app running, queue a `plugin.write` job via `plugin_save(...)` from a chat session while the Mac app is QUIT, then relaunch → within seconds of connect, catch-up claims + runs it; `plugin_status` → `done`; manifest lands in `~/.manta/plugins/`.
3. The two already-stranded jobs (`67697b00` map-form → fails validation as designed; `9750d6f9` array-form → writes `ios-mantaui.yaml`) get claimed on the next connect — Axiom logs (or stderr) will show the new `[cap] catch-up: 2 queued jobs [67697b00,9750d6f9]` line followed by `[cap] start 67697b00: ...`.

## Out of scope (per spec)

- Plugin manifest schema, `plugin.run` path, Settings UI beyond the toggle wiring.
- Migration to strip the dead box-side `pluginsEnabled` (harmless; leave it).
- A retry/polling fallback for catch-up (explicitly forbidden).

### DIFF
--- a/src/main/capExecutor.ts
+++ b/src/main/capExecutor.ts
@@ -197,8 +197,15 @@ export function startCapExecutor(
   async function catchUpJobs(): Promise<void> {
     const c = cfgOrNull(configGetter);
     if (!c) return;
+    // BET-207: every failure branch now logs — without these, a stranded
+    // queued job was indistinguishable from "nothing to claim" and was
+    // undiagnosable from logs. SSE has no replay, so this on-connect
+    // GET is the ONLY path that can recover jobs the AI queued while
+    // the Mac was offline/asleep. Reusing the dedup Set keeps double-
+    // delivery (SSE envelope + this list) safe.
+    let res: Response;
     try {
-      const res = await fetch(
+      res = await fetch(
         `${c.serverUrl.replace(/\/+$/, "")}/api/cap?host=mac&status=queued`,
         {
           headers: c.boxToken
@@ -206,15 +213,21 @@ export function startCapExecutor(
             : undefined,
         },
       );
-      if (!res.ok) return;
-      const body = (await res.json()) as { jobs?: Array<{ id: string; capability: string }> };
-      for (const j of body.jobs ?? []) {
-        enqueue(j.id, j.capability);
-      }
-    } catch {
-      // Catch-up is best-effort; the SSE path will deliver what catch-up
-      // misses (modulo the dedup Set) once the stream is live.
+    } catch (err) {
+      console.warn("[cap] catch-up threw", err);
+      return;
     }
+    if (!res.ok) {
+      console.warn("[cap] catch-up GET failed", res.status);
+      return;
+    }
+    const body = (await res.json()) as { jobs?: Array<{ id: string; capability: string }> };
+    const ids: string[] = [];
+    for (const j of body.jobs ?? []) {
+      ids.push(j.id);
+      enqueue(j.id, j.capability);
+    }
+    console.log(`[cap] catch-up: ${ids.length} queued jobs`, ids);
   }
 
   function onConnect(): void {
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -230,6 +230,18 @@ function registerHandlers(): void {
   // /rpc/config:get instead). See src/preload/index.ts and src/renderer/main.tsx.
   ipcMain.handle(IPC.configGet, () => config);
 
+  // BET-207: `pluginsEnabled` is a Mac-machine-local toggle (controls
+  // whether THIS Mac runs plugins). It MUST persist to the Mac-local
+  // config the executor reads — NOT the box config that httpApi's
+  // configUpdate writes. Handled here locally via commit() (same one-shot
+  // pattern as configGet) so it works regardless of whether window.api is
+  // the httpApi client. Renderer reaches these via the preload bridge,
+  // not window.api.
+  ipcMain.handle(IPC.pluginsGetEnabled, () => config.pluginsEnabled === true);
+  ipcMain.handle(IPC.pluginsSetEnabled, (_e, value: boolean) => {
+    commit({ pluginsEnabled: value === true });
+  });
+
   // Clipboard write via Electron main — bypasses renderer permission restrictions
   // that silently block navigator.clipboard.writeText for non-user-gesture writes.
   ipcMain.handle(IPC.clipboardWriteText, (_e, text: string) => {
--- a/src/preload/index.ts
+++ b/src/preload/index.ts
@@ -69,6 +69,17 @@ const api = {
 
   revealInFolder: (localPath: string): Promise<void> =>
     ipcRenderer.invoke(IPC.revealInFolder, localPath),
+
+  // BET-207: `pluginsEnabled` is a Mac-machine-local toggle — read/write
+  // goes through main's local handlers (commit() → Mac-local config) so
+  // it doesn't round-trip through httpApi/the box. Renderer reads the
+  // current value on Settings mount to seed the toggle, and writes the
+  // new value on Save. Same pattern as configGet (a local-only channel
+  // that's never part of window.api).
+  pluginsGetEnabled: (): Promise<boolean> =>
+    ipcRenderer.invoke(IPC.pluginsGetEnabled),
+  pluginsSetEnabled: (value: boolean): Promise<void> =>
+    ipcRenderer.invoke(IPC.pluginsSetEnabled, value),
 };
 
 // Expose the real preload bridge under a STABLE, dedicated name — NOT "api".
--- a/src/renderer/Settings.tsx
+++ b/src/renderer/Settings.tsx
@@ -42,7 +42,6 @@ export function Settings({ onClose }: { onClose: () => void }) {
     groqApiKey,
     voiceTranscriptionModel,
     voiceCommandModel,
-    pluginsEnabled,
     launcherFlags,
     axiomToken,
     axiomDataset,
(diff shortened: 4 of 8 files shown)


antoinedc:
## BET-207 review — PASS

Mechanical gates (run on PR branch `c96b9fd`):

- `npm run typecheck` → exit 0, zero errors. Log `sha256:139c60436aeae3ee087031cab3603b97f58acc2874526c66e944de4da209c667` — byte-identical to the implementer's reported hash.
- `npm test` → exit 0, **771 pass / 0 fail / 0 skip** — matches the implementer's reported counts exactly. Trust-but-spot-check passes; no need to re-run on base.

Base-freshness: `merge-base HEAD origin/main = origin/main = cceec0e`. PR is branched off fresh `origin/main`; no phantom-deletion gate.

Anti-spaghetti (jscpd, 8 changed `.ts/.tsx/.mjs` files, min-tokens 70): **0 clones**.

---

### Defect 1 — `pluginsEnabled` writes wrong file

- `IPC.pluginsSetEnabled` / `IPC.pluginsGetEnabled` added in `src/shared/types.ts:545-546` with channel names `plugins:set-enabled` / `plugins:get-enabled` — naming consistent with the existing `config:get`, `clipboard:write-text`, `agent:pull-file` family.
- Handlers wired in `src/main/index.ts:240-243` immediately after the existing `IPC.configGet` handler. `pluginsSetEnabled` calls `commit({ pluginsEnabled: value === true })`, so the in-memory `config` the executor's `configGetter` reads is updated atomically with the persisted Mac-local file. Same one-shot pattern as `configGet`. No httpApi round-trip.
- Preload bridge in `src/preload/index.ts:79-82` alongside the other local-only OS bridges.
- `BuiPreload` interface updated in `src/renderer/preloadAccess.ts:51-52`; test fixture updated in `src/renderer/preloadAccess.test.ts:30-31`.
- `Settings.tsx` (lines 78-85, 127-136, 209-221): seeds `pluginsOn` from `pluginsGetEnabled()` on mount (no longer from `store.pluginsEnabled`); on Save, persists via `pluginsSetEnabled()` and removes `pluginsEnabled` from the `configUpdate` payload. Guarded with `getBuiPreload()?.pluginsSetEnabled` so mobile/web is a no-op.
- `store.ts`: `pluginsEnabled` mirror removed (3 sites — type, initial state, config-apply). Grep for any other reader (`store.pluginsEnabled`, `useStore.*pluginsEnabled`, `getState().pluginsEnabled`) returns zero hits — confirmed no other consumer.

Box-side: `src/server/local.mjs:82-84` `configUpdate` is a free-form merge with no allowlist, so no code change needed. The stray `pluginsEnabled: true` already on the box during verification is dead but harmless — no migration, per spec.

### Defect 2 — `catchUpJobs` silently strands queued jobs

- `src/main/capExecutor.ts:197-231`:
  - Empty `catch {}` replaced with `console.warn("[cap] catch-up threw", err)`.
  - Silent `if (!res.ok) return` replaced with `console.warn("[cap] catch-up GET failed", res.status)`.
  - New `console.log("[cap] catch-up: N queued jobs", ids)` makes a persistent "0" on a real backlog visible.
  - Uses the module's existing `[cap] ...` log prefix (matches `console.warn("[cap] start ${id}: ${claim.error}")` at line 428).
- Misleading "best-effort, the SSE path will deliver what catch-up misses" comment dropped — SSE has no replay, catch-up is the only path that recovers jobs queued before connect. Replaced with a comment that explains the dedup-Set invariant (which IS still true).
- No retry/polling fallback. No new claim path. One single trigger (`onConnect`).

Root-cause investigation: structurally verified against the box code —
- `src/server/capabilities.mjs:199-215` `startJob` only transitions `queued → running`. Endpoint accepts the claim; will not reject a `queued` job.
- `src/server/index.mjs:971-977` `GET /api/cap` returns `{ jobs }`; executor reads `body.jobs`. Matches.
- `publishRegistry` uses the same `boxToken` and works, so token-timing is unlikely.

The new logs are the diagnostic primitive the next occurrence needs; per spec, the implementer-of-the-next-occurrence reads Axiom/stderr and pinpoints the cause.

---

### Nits (non-blocking)

- `preloadAccess.test.ts` adds the two mock methods but no test exercises the new IPC pair or the Settings.tsx seed/persist path. The integration is observable end-to-end via the manual Mac-leg steps; flagging for follow-up if a regression here would be hard to catch.
- The new IPC channel names are well-formed, but a one-line cross-reference comment near `IPC.configGet` ("see also: `plugins:set-enabled` / `plugins:get-enabled`, BET-207") would help a future reader trace the local-only channel pattern. Skippable.

### Verdict

0 Blocks. 0 Questions. 2 Nits. Routes to `bui-pm` for the merge hand-off.

Manual Mac-leg verification remains as documented in the PR body.

PR #155, opened 2026-07-20
fix(plugins): thread buildEnv() result into step exec (BET-210)

antoinedc:
## Summary

The plugin executor was dropping the manifest env at the spawn boundary: every `run:` script saw an empty env, so `$MANTA_INPUT_*` and every manifest `env:$KEY` interpolated to empty inside the shell.

Root cause: `makeExec` was hard-wired to `spawn({env: ...process.env + PATH patch})` — the per-step `buildEnv()` result that the manifest handler already computed (line ~604) was only used for `resolveCwd`, never threaded through to the step's shell.

## Fix

- Extended `CapCtx.exec` opts to accept an optional `env`.
- `makeExec` now honors `opts.env` when present, with the PATH patch re-applied on top so Homebrew/nvm visibility (load-bearing for GUI-launched apps) is unchanged.
- Manifest handler passes the already-computed `env` (`buildEnv(...)` result) at the single step exec call site.

No changes to `buildEnv`, `resolveCwd`, or the manifest schema. No new per-step `env:` overlay support (out of scope; intentionally unimplemented).

## Verification

```
Files changed: 2 (capExecutor.ts +24/-4, capExecutor.test.ts new)
Base: origin/main @ 8273593
PR branch (multica/BET-210-exec-env-passthrough @ 7c2dbaf):
  typecheck: PASS (tsc --noEmit -p tsconfig.node.json && -p tsconfig.web.json)
  test:      PASS (vitest: 60 files / 1128 tests; node:test: 869 tests)
```

New test file `src/main/capExecutor.test.ts` spawns a real `/bin/sh -c` through `makeExec` and asserts:
1. `opts.env` entries (`MANTA_INPUT_FOO`, etc.) reach the spawned shell.
2. Realistic buildEnv() output (`MANTA_INPUT_SIMULATOR`, `MANTA_INPUT_SCHEME`, `WORKSPACE`, `MANTA_PLUGIN`, `MANTA_JOB_ID`) all cross the spawn boundary.
3. The no-opts-env branch still inherits `process.env` (legacy regression guard — `$HOME` reaches the child).
4. PATH_PREFIX is re-prepended onto a caller-supplied PATH so Homebrew stays visible.
5. The `env` field exists on `CapCtx.exec` opts at the type level.

The first test would have failed before the fix — that is the regression pin.

## Out-of-band verification (human, on Mac)

Re-run `plugin_run("ios-mantaui", {action: "compile-only"})` after rebuild — step 5's xcodebuild must now receive a non-empty `-destination "platform=iOS Simulator,name=iPhone 16 Pro"` and proceed past the previous `missing value for key 'name'` error.

## Cross-cutting follow-ups

None.

### DIFF
--- a/src/main/capExecutor.test.ts
+++ b/src/main/capExecutor.test.ts
@@ -1,0 +1,105 @@
+// Tests for src/main/capExecutor.ts — the Mac-side plugin executor.
+//
+// BET-210 regression: ctx.exec was dropping opts.env at the spawn boundary,
+// so manifest `env:` and `MANTA_INPUT_*` never reached the `run:` shell.
+// These tests spawn a real /bin/sh through the public `makeExec` helper
+// (re-exported for testing) and assert the env actually crosses the spawn
+// boundary. No mocks — the whole point of the bug is that env must reach
+// the child process, and the only honest assertion is to look at the
+// child's own stdout.
+
+import { describe, it, expect } from "vitest";
+import { makeExec, type CapCtx } from "./capExecutor.js";
+
+// Shared abort signal that never fires — tests rely on the spawn exiting
+// quickly under their own /bin/sh -c command. If it doesn't, vitest's
+// default timeout will fail the test with a clearer message than a hung
+// AbortController would.
+const signal = new AbortController().signal;
+const noop = () => {};
+
+describe("capExecutor — makeExec env passthrough (BET-210)", () => {
+  it("passes opts.env entries into the spawned shell", async () => {
+    const exec = makeExec(signal, noop);
+    const r = await exec(
+      "/bin/sh",
+      ["-c", "printf '%s' \"$MANTA_INPUT_FOO\""],
+      { env: { MANTA_INPUT_FOO: "bar-from-opts" } },
+    );
+    expect(r.code).toBe(0);
+    expect(r.stdout).toBe("bar-from-opts");
+  });
+
+  it("preserves multiple env vars, including manifest-style keys", async () => {
+    const exec = makeExec(signal, noop);
+    // Mirror a realistic buildEnv() output: MANTA_INPUT_* + a manifest env
+    // key + MANTA_PLUGIN/MANTA_JOB_ID. The /bin/sh -c script prints them
+    // in order, separated by a delimiter that can't appear in the values.
+    const env = {
+      MANTA_INPUT_SIMULATOR: "iPhone 16 Pro",
+      MANTA_INPUT_SCHEME: "MantaUI",
+      WORKSPACE: "/tmp/does-not-need-to-exist",
+      MANTA_PLUGIN: "ios-mantaui",
+      MANTA_JOB_ID: "deadbeef",
+    };
+    const r = await exec(
+      "/bin/sh",
+      [
+        "-c",
+        "printf '%s|%s|%s|%s|%s' \"$MANTA_INPUT_SIMULATOR\" \"$MANTA_INPUT_SCHEME\" \"$WORKSPACE\" \"$MANTA_PLUGIN\" \"$MANTA_JOB_ID\"",
+      ],
+      { env },
+    );
+    expect(r.code).toBe(0);
+    expect(r.stdout).toBe("iPhone 16 Pro|MantaUI|/tmp/does-not-need-to-exist|ios-mantaui|deadbeef");
+  });
+
+  it("falls back to process.env when opts.env is omitted (legacy behavior)", async () => {
+    const exec = makeExec(signal, noop);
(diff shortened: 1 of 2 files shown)


antoinedc:
## bui-reviewer — review pass on BET-210 (PR #155)

**Verdict: 0 Block / 0 Question / 2 Nits. PASS — routing to `bui-pm`.**

### Pre-flight

- Base freshness: PR `7c2dbaf` is branched off `origin/main` (`8273593`); `git diff --stat origin/main...HEAD` shows only the 2 files in the PR — no phantom deletions, no stale base.
- Anti-spaghetti (`bash scripts/check-duplication.sh origin/main`): **0 clones** on the changed scope. Clean.

### Acceptance criteria (cross-checked against the issue's "Done when" + fix spec)

| # | Criterion | Status |
|---|---|---|
| 1 | Manifest `env:` + `MANTA_INPUT_*` reach every `run:` shell | Met — single-threaded through `buildEnv()` at line 609 → passed as `env` at line 627 |
| 2 | `ios-mantaui` step 5 compiles past destination error | N/A (out-of-band human Mac verification, per issue) |
| 3 | PATH patch behavior unchanged (Homebrew still visible) | Met — `PATH_PREFIX` re-applied to `baseEnv.PATH` at line 795; new test #4 pins it |
| 4 | Typecheck + tests green | Met — see below |

### Reuse / hygiene constraints

- Reused the `env` already computed at line 609 — no second `buildEnv()` call.
- `buildEnv`, `resolveCwd`, manifest schema untouched.
- PATH-prefix behavior identical: `PATH_PREFIX + (baseEnv.PATH ?? process.env.PATH ?? "")` — prepends to whatever PATH the caller's env carries, falls back to process.env.PATH if the caller's env has no PATH.
- `plugin.write` built-in untouched (confirmed: only `spawn()` call in the file is inside `makeExec`; `plugin.write` doesn't shell out).
- No per-step `env:` overlay added — explicitly out-of-scope per issue.

### Mechanical verification (re-run locally on `multica/BET-210-exec-env-passthrough` @ `7c2dbaf`)

```
$ npm run typecheck
tsc --noEmit -p tsconfig.node.json && tsc --noEmit -p tsconfig.web.json
(exit 0, no output)

$ npx vitest run
Test Files  60 passed (60)
     Tests  1128 passed (1128)

$ npx vitest run src/main/capExecutor.test.ts
Test Files  1 passed (1)
     Tests  5 passed (5)
```

Reported base vs. PR-branch deltas (per implementer): vitest 1123 → 1128 (+5), matching the new file exactly. No pre-existing failures, no new failures.

### Test quality

- 5 focused tests, no over-mocking. They spawn real `/bin/sh -c` through `makeExec` and read the child's own stdout — the only honest assertion for an env-passing regression. This is the right level for this bug class.
- Coverage of the regression: test #1 directly asserts `MANTA_INPUT_FOO` reaches the shell (the regression pin).
- Realistic shape: test #2 mirrors an actual `buildEnv()` output (`MANTA_INPUT_SIMULATOR`, `MANTA_INPUT_SCHEME`, `WORKSPACE`, `MANTA_PLUGIN`, `MANTA_JOB_ID`) — i.e. the exact keys `ios-mantaui` needs.
- Regression-pin for the PATH gotcha: test #4 supplies `PATH: "/usr/bin:/bin"` and asserts the result still starts with `PATH_PREFIX` and ends with `/usr/bin:/bin`. Future refactor that drops the prepend will fail this test.
- Fallback behavior: test #3 verifies the no-opts branch still reaches the child (HOME non-empty), so the legacy `process.env` path didn't regress.

### Nits (non-blocking)

- **Nit (test design):** Test #5 ("compile-time + runtime") duplicates the regression pin from test #1 — the runtime part adds nothing the earlier tests already cover. The compile-time typecheck is real value but could be a single-line type-only assertion (`const _: CapCtx["exec"] = ...`) instead of a full spawn. Not worth a change.

- **Nit (export surface):** `makeExec` is now exported solely so the test can import it. The "main-process-only" comment is accurate and helpful, but it's still a public-symbol widening. If you'd prefer to keep the symbol private, an alternative would be to put the spawn-assertion tests next to `makeExec` as a vitest in the same file (already true here, since `capExecutor.test.ts` sits next to `capExecutor.ts`) and either (a) keep the export but mark it `@internal` via a JSDoc tag, or (b) accept that exports-for-tests is the existing convention in this repo. Leaving as-is is fine.

### Nothing else

No scope creep, no under-delivery, no shared-surface touchpoints (desktop/mobile transport, `chatUtils.ts`, `ChatPanel.tsx`, `src/server/`) touched. The diff is exactly the three-step fix the issue specified plus a focused test.

Routing to `bui-pm` for merge handoff.

PR #166, opened 2026-07-20
BET-205: installer — Caddy + DNS automation + gateway registration

antoinedc:
## Summary

The installer now owns HTTPS ingress on the box: it installs Caddy (via the official apt repo), registers the box with the hosted push gateway at `gateway.mantaui.com`, polls until `<box_id>.boxes.mantaui.com` resolves to the box's public IP, and writes the Caddyfile vhost — so a fresh `curl -fsSL https://mantaui.com/install.sh | bash` ends with a working `https://<box_id>.boxes.mantaui.com` in addition to the 6-digit pairing code.

This is BET-205 (Stage 7 of BET-198 / "Drop the relay"). Touches only `scripts/install.sh` + `scripts/install-lib.mjs` + `scripts/install.test.mjs` — per the do-not-touch list, `src/**`, `src/gateway/**`, `src/server/**`, `.github/workflows/gateway-deploy.yml` (BET-208), and `AGENTS.md` are untouched.

**Base: `origin/main @ 097a202`**

## Changes

- `scripts/install-lib.mjs`
  - `mergeGatewayAuth(existingText, {gateway_token, gateway_host})` — pure, atomic merge of gateway fields into `~/.manta/auth.json`, preserving `box_id` / `box_token` / `created_at`. Same safety contract as `mergeOpencodeConfig` / the deleted `mergeRelayDisabled`: never clobber what we can't parse.
  - `waitForDns(hostname, expectedIp, {lookup, sleep})` — DNS poll with injectable lookup so tests don't need real DNS. Uses `node:dns.lookup` (no `dig` dependency on the box).
  - `renderCaddyVhost(boxId, port, {mode})` — pure renderer for the Caddyfile fragment (`snippet` mode for `/etc/caddy/Caddyfile.d/`, or `inline` mode with `# >>> manta >>>` / `# <<< manta <<<` marker fences for distros without conf.d).
  - New CLI subcommands: `merge-gateway`, `wait-for-dns`, `render-caddy-vhost`. The subcommand tests pin the JSON contract.

- `scripts/install.sh`
  - Step 7.5 (between existing step 7 server-systemd and the existing health-wait step):
    - **A.** Caddy install via the official apt repo (`sudo -n` for the one privileged step; clean fallback to a manual-install hint if passwordless sudo isn't configured).
    - **B.** POST to `gateway.mantaui.com/register` with `box_id` (uses existing `gateway_token` for idempotent IP refresh; first-boot registration mints a fresh token).
    - **C.** Persist `gateway_token` + `gateway_host` via the new `merge-gateway` subcommand (atomic temp-rename + 0600).
    - **D.** `wait-for-dns` up to 5 minutes (30 × 10s).
    - **E.** Render + write the Caddy vhost, then `systemctl reload caddy`. Warns if `:80` / `:443` is bound by something else (Let's Encrypt HTTP-01).
  - `--dry-run` flag: prints every step as `[dry-run] would …` and short-circuits the actual side-effects. Used by tests + by humans previewing a fresh install.
  - `--help` flag: shows usage.
  - `MANTA_GATEWAY_BASE` env override (for tests + staging).

## Idempotency

Every step no-ops cleanly on re-run:
- Caddy install: skip if `caddy` binary already on PATH.
- Gateway registration: refreshes the A record if the box's IP changed (uses existing token; returns 200 with no new token).
- DNS poll: skipped if the hostname already resolves to our IP (subsequent installs cost zero wall time).
- Caddy vhost: overwritten in place (`sudo -n tee` for conf.d; marker-bracketed `awk` for the inline Caddyfile path).

## Verification results

- PR branch (`multica/BET-205-installer-caddy-gateway` @ `0fe0d16`):
  - typecheck: exit 0. Errors: none. log sha256: `139c60436aeae3ee087031cab3603b97f58acc2874526c66e944de4da209c667`
  - test: exit 0, **679 pass / 0 fail / 0 skip**. Failures: none. log sha256: `6567fc43f8c62cd9a8822f01d55bcf9f311ceae6e073692d8c117d60abe4a3d8`
- Base (`main` @ `097a202`):
  - typecheck: exit 0. Errors: none. log sha256: `139c60436aeae3ee087031cab3603b97f58acc2874526c66e944de4da209c667`
  - test: exit 0, **647 pass / 0 fail / 0 skip**. Failures: none. log sha256: `0f63fed1824158bbc21ed361969b5f06d62560ebad9228e562fb4bcc233d2943`
- **Conclusion:** 0 test failures on either branch; the PR adds **+32 tests** (no new failures, no resolved regressions — the new tests are pure additions for the Caddy / gateway / DNS helpers and their CLI subcommands).

## Test additions (32 tests)

- `mergeGatewayAuth`: empty input, preserves box_id / box_token / created_at, re-registration idempotence (`changed=false`), partial re-registration, refuses corrupt input, refuses non-object root, tolerates null/undefined, requires at least one gateway field, preserves historical keys (`relayEnabled`).
- `waitForDns`: resolves immediately, retries on wrong IP, retries on ENOTFOUND then succeeds, gives up after maxAttempts, IPv6-only is not-yet-resolved, requires hostname + expectedIp + lookup.
- `renderCaddyVhost`: snippet form, inline marker fences, validates boxId / port / mode, idempotent output.
- Bash (`install.sh`): caddy-skip when `command -v caddy` succeeds, dry-run branch when it fails, `--help` exits 0, dies on unknown flag, `--dry-run` sets `DRY_RUN=1` and defines `dry_log`.
- CLI subcommands: `merge-gateway` preserves box_token (the headline acceptance criterion), persists a missing auth.json, refuses corrupt input, requires `--file`. `render-caddy-vhost` emits the snippet, rejects non-hex. `wait-for-dns` resolves on the first try (real DNS), requires `--hostname` and `--expected-ip`.

## Anti-spaghetti checks

- **Search before write:** every new helper mirrors an existing pattern in `install-lib.mjs` (`mergeGatewayAuth` ↔ `mergeOpencodeConfig` ↔ deleted `mergeRelayDisabled`; `waitForDns` ↔ `waitForHealth`; `renderCaddyVhost` ↔ `renderSystemdUnit`). Reuse, don't duplicate.
- **Root cause, not symptom:** the install was leaving HTTPS ingress to the user. The fix is the install now owning it (per BET-198 WP5). The new code lives at the source — install.sh + its pure helper module — not as patches at the call sites.
- **Idempotency at the source:** every step checks its own state before acting (`command -v caddy`, `changed=false` flag on merge, `[ -d /etc/caddy/Caddyfile.d ]`, DNS poll skip-if-resolved). The "no-op on re-run" invariant lives in install.sh, not at the operator's discretion.
- **Server boundary preserved:** `src/server/push.mjs` already owns the `gateway_token` reads from auth.json (via `gatewayRegister.mjs`). The install writes the same file (same 0600 perms, same JSON shape) using the same lib helpers — no parallel file/format invented.
- **Process boundary preserved:** no changes to `src/**` per the do-not-touch list. The install and the server both touch `~/.manta/auth.json` but use distinct APIs (CLI subcommand vs in-process read), which keeps the surface area unchanged.

## Out of scope (per do-not-touch list)

- `src/**`, `src/gateway/**`, `src/server/**` — untouched.
- `.github/workflows/gateway-deploy.yml` — untouched (BET-208).
- `AGENTS.md` — untouched (BET-205 followup). The "drop the relay" AGENTS.md rewrite is BET-205's separate work; this PR delivers the installer half.
- `mobile/` — no mobile-side changes; the `https://<box_id>.boxes.mantaui.com` flow is already wired into the mobile deep-link handler (BET-203 / `boxDirectUrl`).

Co-authored-by: multica-agent <github@multica.ai>


antoinedc:
## Review — BET-205 (Stage 7: installer — Caddy + DNS automation + gateway registration)

### Mechanical gates

- `npm run typecheck` → exit 0, 0 errors (locally verified against the PR branch).
- `npm test` → exit 0, **679 pass / 0 fail / 0 skip** (matches the implementer's reported result; master reported 647 + 32 new = 679 ✓).
- `bash -n scripts/install.sh` → OK. `node --check scripts/install-lib.mjs` → OK.
- Strict duplication gate (`scripts/check-duplication-gate.sh origin/main`) → PASS locally.
- Base freshness: PR branched off `origin/main @ 097a202` (BET-204 merge). origin/main has since advanced by one commit (BET-217, `src/renderer/**` + `src/shared/logShip.*`); PR does not touch those, diff vs current `origin/main` is exactly the 3 scripts/ files (1388 +, 12 −).
- Anti-spaghetti advisory: 1 clone flagged (intra-file boilerplate in `runBootstrap`/`runMain` test helpers); strict gate passes, treated as acceptable test scaffolding per AGENTS.md guidance.
- Out-of-scope (AGENTS.md, src/**, src/gateway/**, src/server/**, .github/workflows/gateway-deploy.yml) — untouched ✓.

### Acceptance criteria (issue body)

All 10 issue-body criteria verified against the code (not the prose):

| # | Criterion | Where |
|---|---|---|
| 1 | Install Caddy if absent (apt repo; skip if caddy exists) | `scripts/install.sh:498–537` |
| 2 | POST to gateway with `{box_id}` | `scripts/install.sh:572–578` |
| 3 | Persist `gateway_token`/`gateway_host` via existing node CLI-helper pattern | `scripts/install.sh:583` (pipes JSON to `merge-gateway`) |
| 4 | Add `merge-gateway` subcommand replacing deleted `merge-relay-disabled` | `scripts/install-lib.mjs:808–858` + `merge-relay-disabled` truly gone |
| 5 | Atomic-merge with stdin JSON into auth.json | `scripts/install-lib.mjs:830–857` (temp + rename) |
| 6 | DNS poll max 5 min (30×10s), fail with clear message | `scripts/install.sh:630–639` + `scripts/install-lib.mjs:417–471` |
| 7 | Write `/etc/caddy/Caddyfile.d/manta.caddy` w/ marker-block fallback | `scripts/install.sh:648–687` |
| 8 | `systemctl reload caddy` | `scripts/install.sh:703–710` |
| 9 | Warn if `:80`/`:443` bound by something else | `scripts/install.sh:692–701` |
| 10 | Pair link/QR format unchanged | `scripts/install.sh:790` — `manta://pair?box=%s&code=%s` verbatim |

Idempotency: `mergeGatewayAuth` reports `changed=false` on byte-identical re-run; Caddy skipped via `command -v`; DNS poll skipped if hostname already resolves to expected IP; Caddy block overwritten in place via markers / `tee`. All verified by tests.

### Tests (issue-specified scenarios)

- skip Caddy install when `caddy` exists — `install.sh caddy-skip` test ✓
- DNS poll completes when resolves — `waitForDns resolves immediately` + `retries on a wrong IP` ✓
- DNS poll fails when never resolves, capped at 5 s — `waitForDns gives up after maxAttempts` (intervalMs=1000, maxAttempts=5) ✓
- `merge-gateway` merges without overwriting `box_token` — both the pure-fn test and the CLI-subcommand test, including a corrupt-input case ✓

### Findings

**Block (fix-here): step 7.5.E still `die()`s on sudo failure — the bring-your-own-proxy path is broken.**

Design guidance step 3 from Antoine (issue comment) is explicit:
> "Degrade gracefully when sudo is unavailable or non-interactive … continue/exit cleanly with a clear message — do NOT die mid-install or half-configure. **The bring-your-own-proxy path must keep working.**"

Step 7.5.A degrades gracefully (warns + continues). Steps 7.5.B/C/D don't need sudo (curl + node:dns), so they proceed. But step 7.5.E still has four `die()` calls that fire on sudo failure:
- `scripts/install.sh:657` — `|| die "failed to write $CADDY_DIR_D/manta.caddy (sudo tee failed)"`
- `scripts/install.sh:673` — `|| die "could not update $CADDYFILE"` (sudo mv)
- `scripts/install.sh:676` — `|| die "could not append to $CADDYFILE"` (sudo bash append)
- `scripts/install.sh:681` — `|| die "could not create $CADDYFILE"` (sudo tee)

Result: a user without passwordless sudo (e.g. a hardened VPS where the operator runs the installer as a non-`sudo` account and handles HTTPS via their own reverse proxy + SSH port-forward to the box) gets:
- step A: caddy install skipped (warn) ✓
- step B: gateway registered ✓
- step D: DNS resolves ✓
- step E: dies ✗ — pair code never printed, install can't be re-run to pair

Fix: mirror step A's pattern in step E — detect `command -v sudo` and `sudo -n true` upfront, print the exact manual commands + warn "no Caddy configured; the box's public hostname will not serve until Caddy is in place; you can still pair via SSH port-forward", then skip E entirely. The pair code (step 8) must be reachable in this case. Add a test that exercises the sudo-missing path through the bash control flow.

**Block (fix-here): `docs/launch-e2e.md` not updated — the BET-173 principle statement is now factually wrong.**

Design guidance step 5 (Antoine, issue comment):
> "the `scripts/install.sh` header comment and `docs/launch-e2e.md` should say 'sudo-free EXCEPT the public-TLS reverse-proxy step, and why'. … Without step 5, the next agent reading the BET-173 record will 'fix' this work as a regression."

The PR updates the install.sh header (✓). It does **not** touch `docs/launch-e2e.md`, which still says at line 654–655:
> `sudo` — none (everything runs as the `manta` user; `require_cmd`'s hint recommends `sudo apt-get install …` ONLY if a prereq is actually missing, which on a stock Ubuntu 24.04 cloud image it is not).

That sentence is now false. Without the update, the next implementer reading the BET-173 record will treat the new `sudo` calls as a regression and silently revert them.

Fix: update `docs/launch-e2e.md` to mirror the new install.sh header — i.e. document the Caddy step as the documented, deliberate exception to the BET-173 no-sudo rule (root-required for `:80`/`:443` binding + Let's Encrypt HTTP-01), and note that the rest of the install is unchanged. Keep the file's existing voice (the doc is structured around "What the v2 installer DID NOT do" / "DID add" sections — adding a third subsection for "BET-205 exception: Caddy install + site block + reload" is the natural fit).

Note: `llms-install.md` already describes the Caddy + DNS + gateway flow (it was rewritten by BET-204 in anticipation), so it does not need an update — verify by reading lines 46 and 64 and confirming no contradiction with what now ships.

**Question: explicit distro detection vs natural `apt-get` failure.**

Design guidance step 4 says:
> "Debian/Ubuntu only for v1: detect the distro and bail with a clear message on anything else rather than half-installing."

The implementation does NOT detect the distro. It just runs `apt-get install -y caddy` (and friends) which will fail naturally with a clear "command not found" on Fedora/RHEL/Arch. This is borderline-acceptable (the natural failure IS clear), but it's not the explicit detection+bail the design guidance asked for.

Answer needed: is the natural-failure path enough, or do you want a `source /etc/os-release; case "$ID" in debian|ubuntu) … ;; *) die "Caddy install only supports Debian/Ubuntu; bring your own reverse proxy" ;; esac` upfront guard? If the latter, add it before step 7.5.A (or as a small `require_debian_ubuntu` helper) and add a test.

### Nits (non-blocking)

- `merge-gateway` CLI doesn't explicitly chmod 0600 on the written file — the existing auth.json's 0600 mode is preserved across the atomic rename (POSIX rename replaces the destination, keeping its mode/owner). The "0600 — preserves box_id / box_token" comment in `scripts/install.sh:581` is technically accurate via rename semantics but reads as a stronger guarantee than the code enforces. Consider either (a) `chmodSync(filePath, 0o600)` after the rename as a belt-and-suspenders, or (b) rephrasing the comment.
- `scripts/install.sh:562` and `scripts/install.sh:595` re-parse auth.json with raw `JSON.parse(fs.readFileSync(...))` instead of using `readBoxIdentity` (the tested lib helper). Functional, but inconsistent with the pattern used elsewhere in the PR (which deliberately routes through install-lib for `box_id`/`gateway_host` reads). Consider hoisting the gateway_host read through install-lib (a one-liner pure helper) so the file shape lives in one place.
- `runMain`/`runBootstrap` test-helper boilerplate has noticeable overlap (~22 lines). The advisory anti-spaghetti report flagged it; the strict gate passes because both halves live in `install.test.mjs` and jscpd's pairwise scope is configured cross-file. Consider unifying behind a single helper that takes a `wrapper` parameter — or accept the duplication as test-fixture noise.
- `dry_log` uses Unicode arrows (`▸`) — matches the rest of the install script's existing log style, so this is intentional. No action.

### Verdict

2 Blocks + 1 Question. Returning to `better-ui-dev` for the two fixes (docs update + step 7.5.E graceful degradation) and one design clarification (distro detection). Once those land and CI re-passes, this is a clean PASS.

antoinedc:
## Review — BET-205 (Stage 7: installer) — cycle 2

Both Blocks + the Question from cycle 1 are cleanly resolved. The followup pushed 9/10 → 10/10 across the affected criteria — but the new tests also tripped the strict duplication-gate (CI is FAIL on `duplication-gate` since `b634f29`).

### Cycle 1 findings — status

**Block 1 (step 7.5.E `die()` → graceful) — FIXED ✓**

- All four `die()` calls in step 7.5.E replaced with `CADDY_E_SKIP` flag + `warn` + continue (`scripts/install.sh:740-794`).
- The port-check + `systemctl reload caddy` sub-tasks are guarded by `if [ "$CADDY_E_SKIP" = "0" ]` (`scripts/install.sh:800-827`), so they're skipped when the vhost write fails.
- Step 8 (pair-code mint) now runs regardless — bring-your-own-proxy path keeps working.
- Four new tests pin the contract: `step 7.5.E: sudo -n tee fails → CADDY_E_SKIP=1`, `conf.d missing + sudo bash append fails → CADDY_E_SKIP=1`, `CADDY_E_SKIP=1 skips the port-check + reload sub-tasks`, `CADDY_E_SKIP=0 runs the port-check + reload`. Regression guard on both directions.

**Block 2 (`docs/launch-e2e.md` SUDO EXCEPTION) — FIXED ✓**

- HISTORICAL-record banner at `docs/launch-e2e.md:492-499` explicitly tells the reader the BET-173 v2 section is NOT the current installer and points to the SUDO EXCEPTION section.
- Stale "sudo — none" claim at `docs/launch-e2e.md:654-657` rephrased to "none in v2… BET-205 added a documented sudo exception".
- Stale "zero sudo" claim at `docs/launch-e2e.md:795-797` gets the matching NB.
- New `SUDO EXCEPTION (BET-205)` section (`docs/launch-e2e.md:895-986`) covers: why the exception is needed (BET-198 requirement change), what's privileged (sub-step A/B/C/D/E table with sudo-yes/sudo-no column), graceful-degradation contract, tests pinning the gates, and a "Why this is on record" paragraph that explicitly addresses the regression risk I raised. Closes the "next agent will treat this as a regression" concern.
- `llms-install.md` Step 5 failure playbook + preflight also reference the BYO-proxy + passwordless-sudo + non-Debian paths.

**Question (explicit distro detection) — RESOLVED ✓**

- `readOsReleaseIds` (`scripts/install-lib.mjs:542-561`) parses `/etc/os-release` for `ID` + `ID_LIKE`. Returns `null` on missing/unparseable.
- `classifyDistro` (`scripts/install-lib.mjs:579-586`) — pure `debianLike` boolean + raw id/idLike. Truth table covers `debian`, `ubuntu`, `ID_LIKE=debian` derivatives (Mint, Raspbian, etc.), rejects RHEL/Fedora/Arch/Alpine.
- `detect-distro` CLI subcommand emits the JSON install.sh consumes.
- Distro gate in install.sh (`scripts/install.sh:540-563`) → `PRIVILEGED_SECTION_SKIP=1` on non-Debian; sudo gate (`scripts/install.sh:565-586`) → `PRIVILEGED_SECTION_SKIP=1` on sudo missing or `sudo -n true` failing. Both print bring-your-own-proxy hints before continuing.
- 8 unit + 2 CLI + 4 bash-gate tests pin the contract.

### Mechanical gates

- `npm run typecheck` → exit 0, 0 errors.
- `npm test` → exit 0, **696 pass / 0 fail / 0 skip** (matches the implementer's reported 696).
- `bash -n install.sh` + `node --check install-lib.mjs` → OK.
- **Strict duplication-gate → FAIL** in CI (run `29785711998`, job `88496704151`). 2 clones detected in `scripts/install.test.mjs`. See Block below.
- Base freshness: still off `origin/main @ 097a202`; diff vs current `origin/main` is exactly the 5 files in scope (scripts/{install.sh, install-lib.mjs, install.test.mjs} + docs/launch-e2e.md + llms-install.md). AGENTS.md, src/**, .github/workflows/gateway-deploy.yml untouched ✓.
- `CI typecheck-test: pass`, `duplication-report (advisory): pass`, `E2E Smoke Test: pass`, `secret-scan: pass`, `dep-audit: pass`. Only `duplication-gate` is blocking.

### Block (fix-here) — strict duplication-gate FAIL

CI run `29785711998` reports:
```
duplication-gate: FAIL — 2 clone(s) detected among changed files:
  scripts/install.test.mjs <-> scripts/install.test.mjs (23 lines)
  scripts/install.test.mjs <-> scripts/install.test.mjs (15 lines)
```

I reproduced the scan locally with `npx jscpd scripts/install.test.mjs scripts/install-lib.mjs --min-tokens 70`; both clones are intra-file in `scripts/install.test.mjs`:

1. **Clone 1 — 23 lines, `[72:94] ↔ [180:197]`.** This is the `runBootstrap` vs `runMain` execSync wrapper boilerplate (the `try { execSync(...) } catch (e) { ... } finally { rmSync(...) }` block in both helpers). **I soft-flagged this exact overlap as a Nit in cycle 1** (`runMain/runBootstrap test-helper boilerplate has noticeable overlap (~22 lines)... Consider unifying behind a single helper that takes a wrapper parameter — or accept the duplication as test-fixture noise`); the implementer accepted the noise. With the 13 new tests from the followup, the strict gate is no longer willing to.

2. **Clone 2 — 15 lines, `[1779:1793] ↔ [1810:1824]`.** New from the followup — both `detect-distro` CLI tests have identical `writeFileSync(osRelease, …); try { spawnSync(node, ["…/install-lib.mjs", "detect-distro", "--os-release", osRelease], { encoding: "utf8" }); … } finally { rmSync(...) }` scaffolding. They're true boilerplate, not business logic.

`scripts/duplication-gate-ignore.txt` is HUMAN-approval class — an agent PR cannot widen it (AGENTS.md + the file's own header). The fix must be code-side extraction:

- **Clone 1:** unify `runBootstrap` and `runMain` behind a single helper, e.g. `runBash({ preBody, stubs, wrapper, wrapperDef, args })` where `wrapperDef` (the in-shell function definition for `main`) is optional. Or extract just the shared `try { execSync } catch { ... } finally { rmSync }` block into a `runAndCapture({ script, env })` helper and have both wrappers call it. Net result: a single source of truth for the tempdir + execSync + cleanup pattern.
- **Clone 2:** extract the `detect-distro` invocation into a small `runDetectDistro(osReleaseContent)` helper that writes the os-release file, spawns the CLI, parses the JSON, and returns the result. Both tests assert on the returned object; the scaffolding lives in one place.

A `jscpd` re-run after the refactor must show 0 clones (min-tokens 70). Run `bash scripts/check-duplication-gate.sh origin/main` locally before pushing — that's the script CI uses.

### Verdict

1 Block (the duplication-gate FAIL). Routing back to `better-ui-dev` for the helper-extraction fix. Once that lands and `duplication-gate` re-passes, this is a clean PASS — every original AC is satisfied, every design-guidance step is honored, and CI is green.

(Cycle count: my 2nd reviewer pass on this issue; 1 cycle remaining before the 3-cap escalation.)

antoinedc:
## Review — BET-205 (Stage 7: installer) — cycle 3 — PASS

Cycle 2's strict duplication-gate Block is fully resolved. Every claim in the implementer's reply verified against the code.

### Helper extractions — verified in code

**`runAndCapture(scriptPath)`** (`scripts/install.test.mjs:177-198`):
- Single source of truth for `try { execSync(bash ${scriptPath}) } catch (e) { … } finally { rmSync(dir, {recursive, force:true}) }`.
- `runBootstrap` (`:75`) and `runMain` (`:165`) now end with `return runAndCapture(script)` — zero overlap remains between them.

**`runDetectDistro(osReleaseContent)`** (`scripts/install.test.mjs:1806-1824`):
- Temp file + `spawnSync(node, [install-lib.mjs, "detect-distro", "--os-release", osRelease])` + `rmSync` cleanup.
- Both detect-distro tests at `:1778` and `:1793` now call `await runDetectDistro('ID=…\n…')` — the spawnSync / mkdtempSync / writeFileSync boilerplate is gone from the test bodies.

**`read_box_id_for_gateway`** bash function (`scripts/install.sh:485-492`):
- Defined inside `main()` (so it closes over `$NODE` / `$LIB` / `$MANTA_AUTH_FILE` from the install-sh scope).
- Both call sites — `scripts/install.sh:644` (step 7.5.B gateway register) and `scripts/install.sh:882` (step 8 trailing-pairing heredoc) — delegate to `$(read_box_id_for_gateway)`. The inline `node -e …` 6-line clone is gone.
- Bonus: removes a drift hazard — if the JSON parse ever changes, both reads update from one place. Worth the extra extraction even though the strict gate doesn't scan bash files.

### Mechanical gates

- **`npm run typecheck`** → exit 0, 0 errors.
- **`npm test`** → combined run (vitest + node --test): **696 pass / 0 fail / 0 skip** — matches the implementer's reported result.
- **`bash -n scripts/install.sh`** + **`node --check scripts/install-lib.mjs`** → OK.
- **Strict duplication-gate with `jscpd@latest`** (the CI version, not the locally pinned `jscpd@^4.2.3` which silently `--max-lines 1000`-skips the 2128-line test file):
  - All 3 changed `.mjs`/`.sh` files scanned, min-tokens 70 → **0 duplicates**.
  - install.sh standalone: 0 duplicates.
  - install.test.mjs + install-lib.mjs (the exact CI gate scope): 0 duplicates.
- **CI status** (`gh pr checks 166` on the current head `1459d9a`):
  - typecheck-test: **pass**
  - **duplication-gate: pass** (was FAIL in cycle 2 → now PASS)
  - duplication-report (advisory): pass
  - secret-scan: pass
  - dep-audit: pass
  - E2E Smoke Test: pass
- **Base freshness:** PR branched off `origin/main @ 097a202` (BET-204 merge). origin/main has since advanced to `e31d9df` (BET-215 ModelsCard, PR #167 merged after PR #166 was cut); the diff vs current `origin/main` is still exactly the 5 files in scope (the BET-215 changes don't overlap any PR #166 file). No phantom deletions; the strict gate still uses the original merge-base `097a202`.
- **Out-of-scope files:** AGENTS.md, src/**, .github/workflows/gateway-deploy.yml — `git diff origin/main...HEAD -- <path>` returns empty for all of them ✓.

### Cumulative review summary

Every issue-body acceptance criterion satisfied:
1. Install Caddy if absent (apt repo, skip if exists) ✓
2. POST to gateway with `{box_id}`, persist via merge-gateway ✓
3. `merge-gateway` subcommand replaces deleted `merge-relay-disabled`, atomic-merges stdin JSON, 0600 preserved via rename ✓
4. DNS poll max 5 min (30×10s), fail with clear message ✓
5. Write `/etc/caddy/Caddyfile.d/manta.caddy` w/ marker-block fallback ✓
6. `systemctl reload caddy` ✓
7. Warn if `:80`/`:443` bound by something else ✓
8. Pair link format unchanged (`manta://pair?box=%s&code=%s`) ✓
9. Idempotency on re-run (Caddy skip, DNS skip, marker-block overwrite, merge `changed=false`) ✓
10. All 4 issue-specified test scenarios covered ✓

Every Antoine design-guidance step honored:
- §1 Core install sudo-free — `grep -n "sudo" scripts/install.sh` shows all sudo calls confined to step 7.5 (now gated). The `loginctl enable-linger` line at `:444` is a printed hint, not an actual execution.
- §2 Privileged section isolated — step 7.5 banner explicitly marks it as the BET-205 SUDO EXCEPTION.
- §3 Graceful degradation — three-gate block (distro / sudo / sudo -n true) + `CADDY_E_SKIP` flag in step 7.5.E + pair-code step 8 always runs.
- §4 Distro detection — `readOsReleaseIds` + `classifyDistro` + `detect-distro` CLI; explicit bail on non-Debian.
- §5 Principle statements — `scripts/install.sh` header records the exception; `docs/launch-e2e.md` has HISTORICAL-record banner + dedicated "SUDO EXCEPTION (BET-205)" section; `llms-install.md` Step 5 failure playbook describes the BYO-proxy + passwordless-sudo + non-Debian paths.

### Verdict

PASS — clean merge-ready PR. Routing to `bui-pm` for the merge hand-off.

(Cycle count: 3/3 reviewer passes on this issue; cap reached. Final pass — no escalation needed.)

PR #169, opened 2026-07-21
BET-220: duplication-gate tooling — make local match CI

antoinedc:
## What
Aligns local + CI behavior of `scripts/check-duplication-gate.sh` and `scripts/check-duplication.sh` so a developer's local pre-push run produces the same result the CI job will.

## Why
Locally `npm install` pinned `jscpd@^4.2.3` whose default `--max-lines 1000` silently skipped files above 1000 lines — including the 2128-line `scripts/install.test.mjs`. The strict gate then produced a false 'PASS with warning' on real clones. CI doesn't run `npm ci` before the duplication-gate job, so it always fetched `jscpd@latest` — which catches them. The gate wasn't lying about the merged result, but it was lying about the developer's local confidence before pushing. Originally flagged in BET-205 cycle-3 review.

## Changes
- **Bump `jscpd` devDep from `^4.2.3` to `^5.0.12`**: 5.x's default `--max-lines` is `null` (no limit), so no more silent file-skip whether `npx` picks up the local install or fetches latest from npm.
- **Commit `.jscpd.json` at the repo root**: single source of truth for jscpd settings (`minTokens: 70`, `maxLines: 100000` belt-and-suspenders cap, `minLines: 5`, `maxSize: 100mb`, ignore patterns). Both gate + advisory scripts pass `--config .jscpd.json` explicitly and rely on `.jscpd.json` for the duplicated settings instead of repeating them inline.
- **Add `scripts/check-duplication-gate.test.mjs`** (auto-picked up by `npm test` via `scripts/*.test.mjs` glob):
  - 1100-line clean file → gate PASS (no false positives).
  - 1200-line file with a 110-token duplicated block → gate FAIL. **This is the BET-220 regression guard**: if anyone re-introduces a large-file mask (jscpd downgrade, `maxLines` mistake in `.jscpd.json`, etc.) the synthetic test catches it.

## Verification
- `npm run typecheck` → 0 errors.
- `npm test` → 698 pass / 0 fail (2 new tests added).
- Verified against the **BET-205 cycle-2 file set** (`install.test.mjs` at 2133 lines, clones NOT yet extracted): gate FAILs with the expected 2 clones — was silently PASS-with-warning under the old pin.
- Verified against the **BET-205 cycle-3 file set** (helpers extracted): gate PASSes cleanly with no 'tool failure' warning.
- `bash scripts/check-duplication-gate.sh origin/main` on this branch → clean PASS (no scannable changed files — this branch is tooling-only).

## Out of scope
- `BET-206` (deploy workflow + runbook doc) — separate concern.
- Changing `scripts/duplication-gate-ignore.txt` — HUMAN-approval class per AGENTS.md, not touchable in an agent PR.
- Anything under `.github/workflows/` — the CI jobs are already correct (no `npm ci` before the gate means they pick up jscpd@latest). They'll now also auto-read `.jscpd.json` from cwd.

## Reviewer focus
- `scripts/check-duplication-gate.sh` + `scripts/check-duplication.sh`: `--config .jscpd.json` swap-in, redundant CLI flags removed (still functionally equivalent — verified via test).
- `.jscpd.json` config schema (jscpd 5.x): the field names match jscpd 5.0.12's `--debug` output.
- The two new tests in `scripts/check-duplication-gate.test.mjs`: temp-git-repo harness pattern.

### DIFF
--- a/.jscpd.json
+++ b/.jscpd.json
@@ -1,0 +1,14 @@
+{
+  "minTokens": 70,
+  "minLines": 5,
+  "maxLines": 100000,
+  "maxSize": "100mb",
+  "reporters": ["console", "json", "consoleFull"],
+  "absolute": true,
+  "ignore": [
+    "**/mobile/www/**",
+    "**/node_modules/**",
+    "**/dist/**",
+    "**/out/**"
+  ]
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
         "electron-updater": "^6.3.0",
         "electron-vite": "^2.3.0",
         "eslint": "^10.6.0",
-        "jscpd": "^4.2.3",
+        "jscpd": "^5.0.12",
         "jsdom": "^25.0.1",
         "knip": "^6.14.1",
         "postcss": "^8.4.39",
@@ -405,17 +405,6 @@
       },
       "engines": {
         "node": ">=6.9.0"
-      }
-    },
-    "node_modules/@colors/colors": {
-      "version": "1.5.0",
-      "resolved": "https://registry.npmjs.org/@colors/colors/-/colors-1.5.0.tgz",
-      "integrity": "sha512-ooWCrlZP11i8GImSjTHYHLkvFDP48nS4+204nGb1RiX/WXYHmJA2III9/e2DWVabCESdW7hBAEzHRqUn9OUVvQ==",
-      "dev": true,
-      "license": "MIT",
-      "optional": true,
-      "engines": {
-        "node": ">=0.1.90"
       }
     },
     "node_modules/@csstools/color-helpers": {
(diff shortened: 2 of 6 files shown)


antoinedc:
## Review — BET-220 / PR #169

**Verdict:** PASS (0 Block, 0 Question, 1 Nit).

Verified against the issue body's acceptance criteria on commit `eb7306c` (`multica/BET-220-duplication-gate-tooling`):

| Acceptance criterion | Result |
|---|---|
| Local `bash scripts/check-duplication-gate.sh origin/main` matches CI on the same head | PASS — clean "duplication-gate: PASS — no clones", no "tool failure" warning. |
| Both scripts consume the shared config (no drift between local and CI) | PASS — `check-duplication-gate.sh:73` and `check-duplication.sh:65` both invoke `npx --yes jscpd --config .jscpd.json`. Single source of truth at `.jscpd.json`. |
| 1000-line default mask removed | PASS — `jscpd@^5.0.12` (4.x's `--max-lines 1000` default removed in 5.x), and `.jscpd.json` sets `maxLines: 100000` as belt-and-suspenders. `npx jscpd --debug` confirms the merged config. |
| Regression test: synthetic 1200-line file with 70+ token clone → gate FAIL | PASS — `scripts/check-duplication-gate.test.mjs` builds a ≥1200-line file with two ~110-token `computeAdjustedScore` copies, runs the gate, and asserts `exitCode !== 0` plus a `FAIL`/`clone` marker. Ran in isolation, both new tests green. |
| `npm run typecheck && npm test` GREEN | PASS — typecheck exit 0; test suite 698 pass / 0 fail (2 new gate tests added, picked up via `scripts/*.test.mjs` glob). |
| Branch `multica/BET-220-*`, PR titled `BET-220: …` | PASS — branch and title match. |

`git diff --stat origin/main...HEAD` confirms a fresh base (merge-base = `43af1da8` = `origin/main` tip), no phantom deletions.

### Findings

**Nit (non-blocking):** Header comment in both shell scripts says `maxLines=0 to disable the 4.x 1000-line mask`, but `.jscpd.json` actually pins `maxLines: 100000` (a sane high cap, not literal `0`). The behavior is correct — large files are scanned — but the comment language is slightly inaccurate. Either fix the comment to say "no effective limit" / "100000 cap", or change the JSON value if `0` is the intended semantic. Not blocking; the next editor will be fine.

### Notes for the PM

- The implementation uses the shared-config approach the issue proposed ("OR commit a `.jscpd.json` … single source of truth"). Both fix paths from the issue are valid; this one was picked. Good call — single source of truth is more maintainable than a CLI-flag swap.
- `.github/workflows/anti-spaghetti.yml` needs no change: it never ran `npm ci` before the detector step, so it always pulled jscpd@latest. With the new config, CI now also picks up `.jscpd.json` from cwd automatically.
- The implementer correctly stayed away from `scripts/duplication-gate-ignore.txt` (HUMAN-approval class) and from `BET-206` scope.

PR #175, opened 2026-07-21
fix(gateway): /push was broken — DNS fetch passed to sendApns as APNs sender (BET-229)

antoinedc:
## Live prod bug (found during BET-198 deploy)

After the gateway went live, the **first real `/push` failed** with:
```
[gateway-apns] send error: Failed to parse URL from [object Object]
```

**Root cause:** `handlePush` threaded the server's DNS-oriented `fetchImpl` (`globalThis.fetch`) into `sendApns` as its APNs sender. But `sendApns`'s 3rd arg is an APNs sender that takes a **built request object** → `{status, body}`; global `fetch` expects a URL, so `new URL(reqObject)` throws and **every real push crashed at the gateway**. DNS (Cloudflare REST) legitimately uses global fetch; APNs uses an HTTP/2 sender — the two were conflated in the `createGatewayServer` wiring.

**Fix:** rename `handlePush`'s param to `apnsSender` + add a matching `createGatewayServer` option, **UNDEFINED in production** so `sendApns` falls back to its own `defaultApnsSender` (HTTP/2 → `api.push.apple.com`); tests inject an APNs-shaped fake. `push.test.mjs` `fetchImpl`→`apnsSender`.

**Why CI didn't catch it:** `sendApns` unit tests inject the sender directly and never exercise the server-factory wiring that passed the wrong function.

## Verification
- Manual `/push` against the live gateway with a real `gateway_token` + dummy device token now reaches Apple and returns a proper per-token result (was the URL-parse crash before this fix).
- 104 gateway tests green; typecheck clean; duplication-gate PASS.

Closes BET-229. Required for the BET-198 push E2E.

### DIFF
--- a/src/gateway/index.mjs
+++ b/src/gateway/index.mjs
@@ -237,7 +237,12 @@ export async function handlePush({
   body,
   store,
   apnsConfig,
-  fetchImpl,
+  // The APNs sender: takes a built request object → { status, body }. In
+  // production this is UNDEFINED so sendApns falls back to its own HTTP/2
+  // `defaultApnsSender`. Tests inject an APNs-shaped fake here. It is NOT the
+  // DNS/global `fetch` — passing global fetch here breaks with "Failed to
+  // parse URL from [object Object]" because fetch expects a URL, not a req.
+  apnsSender,
 }) {
   const { box_id, tokens, payload } = body ?? {};
   if (!isValidBoxId(box_id)) {
@@ -273,7 +278,7 @@ export async function handlePush({
   const results = [];
   for (const token of tokens) {
     try {
-      const r = await sendApns({ token, payload }, apnsConfig, fetchImpl);
+      const r = await sendApns({ token, payload }, apnsConfig, apnsSender);
       results.push({ token, ok: r.ok, prune: r.prune });
     } catch (e) {
       results.push({ token, ok: false, prune: false });
@@ -320,6 +325,11 @@ export function createGatewayServer({
   // calls array (no OVH, no record-id allocation).
   createDnsRecord,
   apnsConfig = null,
+  // APNs sender injected into handlePush → sendApns. UNDEFINED in production so
+  // sendApns uses its own HTTP/2 defaultApnsSender. Kept SEPARATE from the DNS
+  // `fetchImpl` above — they have different call shapes (DNS wants a URL,
+  // APNs wants a built request object). Tests inject an APNs-shaped fake.
+  apnsSender,
   rateLimiter = createRegisterRateLimiter(),
   log = console.log,
   warn = console.warn,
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review: BET-229 / PR #175

### Gates

| Gate | Result |
| --- | --- |
| `npm run typecheck` | clean |
| `npm test` | 697/697 pass (0 fail, 0 cancel) |
| duplication-gate (jscpd) | 0 clones on the changed scope |
| secret-scan + dep-audit | PASS (per PR checks) |
| base freshness | branched off `49078cb` (one mobile rebuild behind `origin/main`); no phantom deletions |

### Acceptance criteria — verified against the code

The issue's spec was: rename `handlePush` param to `apnsSender` + matching `createGatewayServer` option, **UNDEFINED in production** so `sendApns` falls back to its own `defaultApnsSender` (HTTP/2); tests inject an APNs-shaped fake. `push.test.mjs` `fetchImpl`→`apnsSender`.

- `handlePush` parameter `fetchImpl` → `apnsSender` (src/gateway/index.mjs:245) ✓
- `createGatewayServer` gains a separate `apnsSender` option, kept SEPARATE from DNS `fetchImpl` (src/gateway/index.mjs:332; DNS `fetchImpl` still at line 321, still wired into `dnsCreateOrUpdate` only at line 337) ✓
- Production caller does NOT pass `apnsSender` (src/gateway/index.mjs:461-464), so `sendApns` falls back to `defaultApnsSender` (HTTP/2 to `api.push.apple.com`, src/gateway/apns.mjs:223) ✓
- `push.test.mjs` now injects APNs-shaped fakes returning `{status, body}` instead of URL-shaped fetchers ✓

The bug and root-cause analysis in the description match the code: `sendApns`'s 3rd arg is an APNs sender taking a built request object (src/gateway/apns.mjs:202, `await useFetch(req)`), not a URL — so passing the DNS `globalThis.fetch` there crashed every real push with the exact `[gateway-apns] send error: Failed to parse URL from [object Object]` from the prod journal.

### Findings

**Block (fix-here):** none.

**Question:** none.

**Nit (non-blocking, followup-worthy):** `src/gateway/server.test.mjs` exercises `POST /push` through the full server wiring with a bogus `p8Path: "/d"`, so the APNs sender is never actually invoked there — `getApnsJwt` throws on the missing file and `handlePush` catches it (test note at lines 285-290 explicitly calls this out). That's exactly the gap that hid this bug originally. Not a defect of *this* PR (the PR fixes the wiring correctly and doesn't claim to add coverage), but worth a follow-up: an `apnsSender` injected through `createGatewayServer` in a `/push` happy-path test would have caught this regression on day one and would prevent the next time someone reintroduces it. Suggest a follow-up issue to add that case — happy to draft it on PASS.

### Verdict

**PASS** — fix matches spec, prod gets `defaultApnsSender`, tests inject APNs-shaped fakes, gates green, no scope creep, no shared-surface drift. Routing to `bui-pm` for merge.

PR #190, opened 2026-07-22
feat(desktop): register manta:// deep-link pairing (BET-240)

antoinedc:
# feat(desktop): register manta:// deep-link pairing (BET-240)

Clicking a `manta://pair?box=…&code=…` link anywhere — browser, terminal,
AI-agent output — now opens Manta prefilled to the PairStep. The user
confirms by clicking Connect (no auto-claim, no dialog, no toast).

`Base: origin/main @ 2ff67bd`

## What changed

**OS protocol handler** (3 layers, exactly the same parser):

- `electron-builder.yml` — declare the `manta` scheme (`electron-builder`
  generates `CFBundleURLTypes` on macOS and the `MimeType=x-scheme-handler/manta`
  entry on Linux `.desktop` files from it; do NOT hand-edit any plist).
- `src/main/index.ts` — `setAsDefaultProtocolClient` + single-instance lock +
  `open-url` (macOS, fires both warm AND cold) + `second-instance` (Windows/
  Linux, the second process's argv carries the URL) + cold-start flush in
  `did-finish-load`.
- `src/preload/index.ts` — buffer a URL that arrives before React subscribes
  (cold start) so the renderer never loses a deep link; expose `onPairLink`.
- `src/renderer/preloadAccess.ts` — `onPairLink` on the `MantaPreload` bridge.

**Wire contract** (one parser, two layers):

- `src/shared/types.ts` — `IPC.pairLinkReceived` channel.
- `src/renderer/store.ts` — `pendingPairLink` + `setPendingPairLink`
  (transient, not persisted).
- `src/renderer/App.tsx` — subscribe once on mount; validate with
  `parsePairPayload` (the SAME parser PairStep's paste field uses);
  stash the URL and call `relaunchOnboarding` so a re-pair to a new box
  lands on step 1 even when `resolveInitialStep` would skip past it.
- `src/renderer/Onboarding.tsx` — force step 1 when `pendingPairLink` is set
  (both the initial-step ternary AND an effect for a link that arrives
  while onboarding is already open).
- `src/renderer/PairStep.tsx` — prefill the paste field, consume the store
  value; the existing `linkHasContent` gate + `connect()` claim path do
  the rest when the user clicks Connect.

**Tests:**

- `src/renderer/store.test.ts` — `setPendingPairLink` write + clear.
- `src/renderer/preloadAccess.test.ts` — fake preload now includes
  `onPairLink` so the existing `MantaPreload` shape stays in sync.

## Decided behavior (matches the spec — do not deviate)

- A received pair link NEVER auto-claims. PairStep IS the confirmation UI.
- Invalid/unparseable URLs are silently ignored (`parsePairPayload` returns
  null → `App.tsx` early-returns).
- Works warm (app running) AND cold (app launched by the link).
- Single-instance lock so a second launch forwards the URL to the running
  instance on Windows/Linux.
- `relaunchOnboarding` is the singular forcing mechanism — no parallel
  toggling of `onboardingForced`.

## Verification

- typecheck: clean (both `tsconfig.node.json` and `tsconfig.web.json`).
- test: 712 pass / 0 fail (both PR branch and `main` baseline).
- Logs: `/tmp/typecheck-pr.log`, `/tmp/test-pr.log`,
  `/tmp/typecheck-master.log`, `/tmp/test-master.log`.

### Release-verification checklist (CI cannot exercise the OS handler)

1. Build a packaged DMG (`electron-builder --mac` or the next `mac-v*`
   release pipeline).
2. `open "manta://pair?box=0123456789abcdef0123456789abcdef&code=123456"`
   in Terminal — app cold-launches (or focuses if running); onboarding
   opens at the Connect step with the link field pre-filled.
3. Repeat while the app is already running and paired — same result
   (onboarding re-opens at step 1, prefilled).
4. `open "manta://garbage"` — app focuses, nothing else happens
   (silently ignored).
5. Confirm the packaged `Info.plist`:
   `plutil -p "/Applications/Manta UI.app/Contents/Info.plist" | grep -A3 URL`
   shows `CFBundleURLSchemes` → `manta`.

## Out of scope

- Auto-claiming without a user click, confirmation dialogs, toasts.
- The box-served `/pair` page (BET-239 — independent).
- Clipboard-based deferred pairing, custom DMGs.
- Mobile (`manta://` already handled on iOS via
  `mobile/ios/.../Info.plist`), `mobile-rn/`, `src/server/`.
- NSIS-specific tweaks beyond the standard `second-instance` code.
- New npm dependencies.

## Follow-ups

None surfaced that meet the "above the bar" file threshold. The
`pendingPairLink` field is purely transient routing state — no drift
risk. The wire-format parser (`parsePairPayload`) is the same one
PairStep's paste field already uses, so there is exactly ONE validator
on the wire.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -30,6 +30,15 @@ files:
 extraMetadata:
   main: out/main/index.js
 
+# Custom URL scheme for OS-level deep-link pairing (BET-240). electron-builder
+# generates `CFBundleURLTypes` on macOS and the `MimeType=x-scheme-handler/manta`
+# entry on Linux `.desktop` files from this block — do NOT hand-edit any plist.
+# iOS already handles `manta://` via mobile/ios/.../Info.plist (BET-156).
+protocols:
+  - name: MantaUI Pairing
+    schemes:
+      - manta
+
 mac:
   icon: "assets/icon.icns"
   category: "public.app-category.developer-tools"
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -6,7 +6,7 @@ import {
   nativeImage,
   shell,
 } from "electron";
-import { join, basename } from "node:path";
+import { basename, join, resolve as resolvePath } from "node:path";
 import { existsSync, watch as fsWatch } from "node:fs";
 import { readFile } from "node:fs/promises";
 import { homedir } from "node:os";
@@ -178,6 +178,61 @@ function createWindow(): void {
 // config.json (host, projects) and leaving the app with no configured host.
 app.setPath("userData", join(app.getPath("appData"), "better-ui"));
 
+// ---------------------------------------------------------------------------
+// manta:// deep-link pairing (protocol handler).
+// Registration + capture live in main; parsing/validation happens renderer-
+// side with the same parsePairPayload the PairStep paste field uses, so
+// there is exactly ONE parser for the wire format.
+// ---------------------------------------------------------------------------
+if (process.defaultApp) {
+  // dev mode (`electron .`): register with explicit exec path + args
+  if (process.argv.length >= 2) {
+    app.setAsDefaultProtocolClient("manta", process.execPath, [
+      resolvePath(process.argv[1]),
+    ]);
+  }
+} else {
+  app.setAsDefaultProtocolClient("manta");
+}
+
+let pendingPairUrl: string | null = null;
+
+function deliverPairLink(url: string): void {
+  if (typeof url !== "string" || !url.startsWith("manta://")) return;
+  if (mainWindow && !mainWindow.webContents.isLoading()) {
+    mainWindow.webContents.send(IPC.pairLinkReceived, url);
+    if (mainWindow.isMinimized()) mainWindow.restore();
+    mainWindow.show();
+    mainWindow.focus();
+  } else {
+    // window not up yet (cold start) — flushed from the did-finish-load hook
+    pendingPairUrl = url;
+  }
+}
+
+// macOS delivers protocol URLs via open-url (warm AND cold start — cold-start
+// delivery arrives before ready, which is why pendingPairUrl exists).
+app.on("open-url", (event, url) => {
+  event.preventDefault();
+  deliverPairLink(url);
+});
+
+// Windows/Linux deliver protocol URLs as argv of a SECOND instance; the lock
+// forwards them to the running one.
+const gotSingleInstanceLock = app.requestSingleInstanceLock();
+if (!gotSingleInstanceLock) {
+  app.quit();
+} else {
+  app.on("second-instance", (_event, argv) => {
+    if (mainWindow) {
+      if (mainWindow.isMinimized()) mainWindow.restore();
+      mainWindow.focus();
+    }
+    const url = argv.find((a) => a.startsWith("manta://"));
+    if (url) deliverPairLink(url);
+  });
+}
+
 // Set the app name as early as possible (before `ready`). This drives the
 // product name shown in the macOS menu bar, the dock, AND — for packaged
 // builds — the source name on OS notifications (otherwise the renderer's
@@ -203,6 +258,14 @@ app.whenReady().then(() => {
   createWindow();
   // Defer poller start until renderer is ready to receive events.
   mainWindow?.webContents.once("did-finish-load", () => {
+    // Deep-link cold start: macOS parked the URL in pendingPairUrl (open-url
+    // fires pre-ready); Windows/Linux pass it in this process's argv.
+    const argvPairUrl = process.argv.find((a) => a.startsWith("manta://")) ?? null;
+    const coldPairUrl = pendingPairUrl ?? argvPairUrl;
+    pendingPairUrl = null;
+    if (coldPairUrl) {
+      mainWindow?.webContents.send(IPC.pairLinkReceived, coldPairUrl);
+    }
     startScreenshotDetector();
     // Report desktop focus to the mobile server so it suppresses redundant
     // mobile "done" pushes while the user is active on desktop.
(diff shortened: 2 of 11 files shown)


antoinedc:
# Review: BET-240 — `manta://` deep-link pairing

**Outcome: PASS** (0 Blocks, 0 Questions, 2 Nits). Routing to `manta-pm` for the merge hand-off.

## Acceptance-criteria completeness

The issue spells out 7 prescriptive steps plus a "decided behavior" block (never auto-claim, single parser, cold+warm coverage, single-instance lock, no persistence). I verified each against the code, not against the PR description.

| Step | Verdict | Where |
|------|---------|-------|
| **1. `electron-builder.yml` protocols block** | Pass | Block sits between `extraMetadata:` (32) and `mac:` (35), exact `name`/`schemes` shape — electron-builder generates `CFBundleURLTypes` and the Linux `.desktop` `MimeType` entry from it. |
| **2. `IPC.pairLinkReceived`** | Pass | Added in `src/shared/types.ts` next to `screenshotDetected`, same comment style, payload contract documented. |
| **3a. main registration + handlers** | Pass | `setAsDefaultProtocolClient` with the `process.defaultApp` dev-mode branch + explicit `process.argv.length >= 2` guard; module-scope `pendingPairUrl`; `deliverPairLink` does only the cheap prefix check (`startsWith("manta://")` + string-typed), forwards via `IPC.pairLinkReceived`; `open-url` (macOS, both warm and cold — `pendingPairUrl` catches the pre-ready cold case); `requestSingleInstanceLock` + `second-instance` (Win/Linux argv forwarding). |
| **3b. cold-start flush** | Pass | Lives inside the existing `did-finish-load` callback, reads both `pendingPairUrl` (macOS) and `process.argv` (Win/Linux), clears the buffer before sending. |
| **4. preload bridge** | Pass | `bufferedPairUrl` + `pairLinkCb` at module scope; `ipcRenderer.on` registered immediately; `onPairLink` replays the buffered URL on subscribe and returns an unsubscribe that only clears if the same callback is still attached (no clobber by a newer subscriber). |
| **5. `MantaPreload.onPairLink`** | Pass | Interface field added next to `onScreenshotDetected`; required (matches the existing required bridges). `getMantaPreload()` accessor used (project was renamed from `bui` → `manta`). |
| **6. store** | Pass | `pendingPairLink: string \| null` initialised next to `onboardingForced: false`; `setPendingPairLink(url)` is a plain `set(...)` next to `relaunchOnboarding`. **Not persisted** — store uses bare `create<State>(...)` with no `zustand/middleware` persist, confirmed via `grep`. |
| **7a. App.tsx** | Pass | `useEffect` subscribes once with empty deps; filters with `parsePairPayload` from `./mobile/pairPayload` (the SAME module PairStep imports — verified `parsePairPayload` is exported there); stashes URL and calls the existing `relaunchOnboarding()` — no second forcing mechanism. |
| **7b. Onboarding.tsx** | Pass | Initial-step ternary (`pendingPairLink ? 1 : resolveInitialStep(...)`) covers already-paired re-pairing; effect `if (pendingPairLink) setPos(1)` covers a link arriving mid-flow; uses the component's actual setter `setPos` (not the placeholder name `setStep` from the spec). |
| **7c. PairStep.tsx** | Pass | Effect pre-fills `pairLink` and consumes via `setPendingPairLink(null)` exactly once. No other changes — `linkHasContent` gate and the existing `connect()` claim path (which itself re-validates with `parsePairPayload`) handle submission. |

**Behaviors I independently checked:**
- **Never auto-claim.** No code path calls `authClaim` without an explicit `connect()` user click. Verified by reading `App.tsx`, `Onboarding.tsx`, and the new PairStep effect — none of them invoke `window.api.authClaim` or `connect()` directly.
- **Single parser.** `parsePairPayload` is the only URL-shape validator; main does a `startsWith("manta://")` cheap prefix check only; renderer does the real validation (App.tsx filter, PairStep re-parse on submit).
- **Cold + warm coverage.** macOS cold: `open-url` pre-ready → `pendingPairUrl` → `did-finish-load` flush. Win/Linux cold: argv of first process → `did-finish-load` flush. Warm (all platforms): `deliverPairLink` → `mainWindow.webContents.send`. Plus the preload `bufferedPairUrl` covers the race between IPC fire and React `useEffect` subscribe.
- **Single-instance lock.** Acquired at module scope (matches spec placement); second-instance quits early if lock fails.
- **No persistence.** Store has no `persist` middleware; `pendingPairLink` is genuinely transient.

## Mechanical gate

```
$ npm run typecheck
(clean — exit 0)

$ npm test
# tests 712
# pass 712
# fail 0
```

Numbers match the implementer's claim exactly (712 pass / 0 fail on the PR branch).

## Anti-spaghetti

`bash scripts/check-duplication.sh origin/main` reports **0 clones** across the 10 changed `.ts/.tsx` files (jscpd, min-tokens 70). No signal — nothing to discount.

## Scope

Diff against `origin/main` is exactly the 11 files the spec requires (the 7 production files + 1 IPC line + 1 builder line + 2 test files). No scope creep, no orphan deletions (verified base-freshness gate — `merge-base HEAD origin/main` = `origin/main`, so no phantom deletes).

## Nits (non-blocking)

- **Nit (cosmetic):** The dev-mode branch silently does nothing if `process.argv.length < 2` — matches the spec exactly, but a one-line comment near the `if` explaining "early dev launch with no script path → skip" would help future readers. Not a defect.
- **Nit (style):** PairStep's prefill effect would re-fire on every `pendingPairLink` change including the post-consume `null`; the early-return `if (!pendingPairLink) return;` already guards against that, but if anyone ever swaps the consume order it would loop. Cheap insurance: an empty-deps `useEffect` that reads once and consumes, or a `useRef` guard. Not a defect — current code is correct.

## Release-verification checklist (CI cannot exercise the OS handler)

The PR description's release-checklist is the right place for this — I confirmed it's there (DMG build, `open "manta://pair?…"`, garbage-URL ignore, `plutil -p …/Info.plist | grep -A3 URL`). No action.

## Verdict

All 7 spec steps implemented 1:1, all decided behaviors preserved, typecheck + 712 tests pass on the PR branch, no anti-spaghetti signal, scope matches the issue exactly. Routing to `manta-pm` for the merge hand-off.

PR #212, opened 2026-07-23
refactor(onboarding): simplify Pair screen to Box ID + auto-claim deep-link (BET-255)

antoinedc:
## Summary

Collapses the desktop pairing form to the same shape the mobile client
already uses, and makes the OS-protocol deep-link auto-claim without a
manual "Connect" click.

- **`PairStep.tsx`** drops the Server URL field and the optional pair-link
  paste field. A new Box ID input (32-hex, validated by `isValidBoxToken`)
  replaces them. Connect gate moves to `canConnectSetup` (shared with
  mobile). Single claim path goes through the new `claimBox` helper.
- **`pairClaim.ts`** (new) — one place that turns `{ boxId, code }` into a
  paired desktop session: IPC call + `applyPairing` + `installHttpTransport`.
  Both `PairStep`'s Connect button and `App.tsx`'s deep-link handler call it,
  so there is exactly one code path from claim input to paired state.
- **`App.tsx onPairLink`** auto-claims via `claimBox` and calls
  `finishOnboarding()` to drop to the normal shell on success — no manual
  click required. On failure, falls back to the old behaviour of stashing
  `pendingPairLink` and opening onboarding at step 1 for manual retry.
- **`pairStepLogic.ts` / `pairStepLogic.test.ts`** deleted —
  `normalizeServerUrl`, `isValidServerUrl`, `canConnect` were
  desktop-only and become unused once `PairStep` switches to the
  mobile-shared `canConnectSetup`. `pendingPairLink` is kept (the
  failure fallback uses it).

## Out of scope (per spec)

No changes to `src/main/auth.ts`, `src/shared/transport.mjs`, or anything
under `src/renderer/mobile/`. Verified empty diffs for all three.

## Files changed

5 files, +175 / −296.

```
 src/renderer/App.tsx           |  61 ++++++-----
 src/renderer/PairStep.tsx      | 209 ++++++++---------------------------
 src/renderer/pairClaim.ts      | (new, 87 lines)
 src/renderer/pairStepLogic.ts  | (deleted)
 src/renderer/pairStepLogic.test.ts | (deleted)
```

## Verification results

- PR branch (`multica/BET-255-simplify-pair-screen` @ `4633911`):
  - typecheck: exit `0`. Errors: none. log sha256: `c3beb27fb83609c48ad1aeff48c8c39feec0642968c8d9ad8c1a67c15c82e7ff`
  - test: exit `0`, `734` pass / `0` fail / `0` skip. log sha256: `98804f7ff9d72fe300d2ab5d7346d83467e8299522f793d42c24ed2774da382d`
- Base (`main` @ `acd24b7`):
  - typecheck: exit `0`. Errors: none. log sha256: `c3beb27fb83609c48ad1aeff48c8c39feec0642968c8d9ad8c1a67c15c82e7ff`
  - test: exit `0`, `734` pass / `0` fail / `0` skip. log sha256: `ff01fc72e9e74f6e2eff4c53ad2dda76a645d0e1b6b52900b423459ff61409d6`
- **Conclusion:** `0` test failures are pre-existing, `0` failures are new
  and caused by this PR, `0` failures were resolved by this PR
  (`pairStepLogic.test.ts` deletion removed its own tests in both branches,
  so the count is identical on master and PR — no test was "lost"). The
  subtest set is byte-identical between branches (`diff` of the sorted
  subtest/ok/not-ok lines returns empty).

Logs cached at `/tmp/typecheck-pr.log`, `/tmp/typecheck-master.log`,
`/tmp/test-pr.log`, `/tmp/test-master.log` for reviewer spot-check.

Base: origin/main @ `acd24b7`


antoinedc:
## Review: BET-255 — simplify desktop Pair screen + auto-claim deep-link

**Outcome: PASS — 0 Blocks, 0 Questions, 2 Nits.** Routing to `manta-pm` for merge.

### Mechanical gate

- `npm run typecheck` — exit `0`, 0 errors (matches implementer's reported log sha `c3beb27f…`).
- `npm test` — exit `0`, **734 pass / 0 fail / 0 skip** (matches implementer's reported log sha `98804f7f…`).
- Base freshness: PR branched from `origin/main @ acd24b7`, current `origin/main` tip = `acd24b7` — fresh.
- Anti-spaghetti: **0 clones** in the 3 changed `.ts/.tsx` files (PairStep.tsx, App.tsx, pairClaim.ts).
- Out-of-scope guard: empty diff for `src/main/`, `src/shared/transport.mjs`, `src/renderer/mobile/` — confirmed.

### Acceptance criteria (each verified against the diff, not the prose)

1. **typecheck green** — PASS.
2. **tests green** — PASS, count matches main (test deletion removed its own cases on both sides; subtest set is byte-identical between branches per implementer's diff verification, and re-run on PR confirms 734 pass).
3. **PairStep has exactly TWO inputs (Box ID + Pairing code)** — PASS. `PairStep.tsx:101-155` renders only those two `input` elements. No "Server URL", no "Or paste a pair link" block.
4. **Box ID validates as 32-hex via `isValidBoxToken` + Connect gated by `canConnectSetup`** — PASS. `PairStep.tsx:62` (gate), `:88` (`boxIdLooksBad` uses `isValidBoxToken`), `:4` (import from `../shared/transport.mjs`).
5. **Deep-link auto-claims + advances WITHOUT a manual Connect click** — PASS. `App.tsx:219-237`: parse → `claimBox()` → on success `finishOnboarding()` (the same shell-completion action the PairStep `onPaired` ultimately drives). Failure falls back to the old `pendingPairLink` + `relaunchOnboarding()` pattern, which `Onboarding.tsx:117,125-128` already reads to force step 1 — explicit, in-spec fallback.
6. **Exactly ONE code path turns `{boxId, code}` into a completed pairing** — PASS. The new `src/renderer/pairClaim.ts` `claimBox()` is the sole site of `applyPairing` + `installHttpTransport` + `window.api.authClaim`. `PairStep.tsx:68` and `App.tsx:229` both call it; neither call site re-implements those side effects.
7. **No dead exports, no half-removed state** — PASS.
   - `grep -r 'normalizeServerUrl|isValidServerUrl|canConnect\b' src/` returns 0 hits. `pairStepLogic.ts` + `pairStepLogic.test.ts` both deleted (confirmed via `git log --diff-filter=D`).
   - `pendingPairLink` is **kept**, consistent across `store.ts:107,294,421`, `Onboarding.tsx:117,125-128`, and `App.tsx:244` (failure fallback). Per spec Part 3: "Decide based on grep, not preference: if Part 2's failure path still uses `pendingPairLink`, keep it everywhere" — implementation chose to keep the failure fallback, which is the in-spec option.
8. **No changes to `src/main/auth.ts`, `src/shared/transport.mjs`, `src/renderer/mobile/`** — PASS (empty diff verified).

### Quality observations

- `pairClaim.ts` is small, well-commented, and the failure-path defensive `try/catch → networkFailure()` mirrors the same pattern the mobile `deepLink.ts:132-139` uses — single source of truth for "an unexpected throw from `authClaim` looks like a network failure to the caller."
- `PairStep.tsx` net -109 LoC, comment block is genuinely clearer (the long header now reads as documentation, not apology), and the file is finally just wiring — exactly what the spec asked for.
- The `applyPairing + installHttpTransport` block is gone from `PairStep` (was duplicated inline), gone from the `linkHasContent` branch (deleted), and gone from `App.tsx` — single homesite achieved.

### Nits (non-blocking)

- **Nit 1:** `pairClaim.ts:43-48` wraps `authClaim` in `try/catch → networkFailure()`. The doc comment on line 56-58 of `pairClaim.ts` already calls this "Defensive: authClaim should never throw" — fine — but the function returns the `ClaimOutcome` from `authClaim` on the happy path (line 56). If a future call to `authClaim` ever starts throwing on the wire format itself (not a classified failure), this catch silently maps to "network failure" which is the wrong diagnostic. A future reviewer might want to narrow the catch to known-throw shapes, or log `e` so a real bug is debuggable. Not blocking — the IPC layer is trusted-by-construction today.
- **Nit 2:** `App.tsx:228` uses `void (async () => { … })()` to launch the auto-claim. The comment on `:225-227` says "We don't await it on purpose" but the inner block actually does `await` `claimBox` and `await finishOnboarding`. The "don't await" wording refers to not awaiting the IIFE from the sync `onPairLink` callback (which is right) — but reading the comment literally it sounds like the awaits inside are the wrong choice. Slightly clearer comment would be "The listener is sync; we await the chain inside so the success branch completes `finishOnboarding()` before a render cycle, but we don't propagate the promise back to the preload." Cosmetic.

### Verdict

All 8 acceptance criteria pass; out-of-scope files are untouched; both nits are cosmetic and don't warrant a follow-up. Routing to `manta-pm` for merge.

— `manta-reviewer` (cycle 1)

PR #214, opened 2026-07-24
feat(composer): persist prompt history across /clear + fix soft-wrap nav (BET-257)

antoinedc:
## Summary

Two-file behavior change (plus pure helpers + tests) for the chat composer:

1. **Prompt history now survives `/clear`**. The composer remembered prompts only from the live opencode transcript; `/clear` swaps in a brand-new session with an empty transcript and wiped the cycle history. History is now persisted in `localStorage`, keyed by the tmux WINDOW (`tmuxSession` + `windowIndex`) — the only identity stable across clears — and merged with the live transcript on every read.
2. **Soft-wrapped multi-line navigation is fixed**. The old "am I on the first/last line?" check scanned for `\n` characters, which is blind to soft wrap: a long single line with no `\n` wraps to multiple visual rows, so a caret on visual row 2 wrongly triggered history navigation on ArrowUp. Up/Down now cycle history only when the caret sits on the first / last **visual** row.

Both behaviors were specified up-front in [BET-257](https://github.com/antoinedc/MantaUI/issues/...); this PR implements exactly the design described there with no scope creep.

## Files changed

7 files, +318 / −14.

```
src/renderer/ChatPanel.tsx            | 14 ++++++
src/renderer/InputArea.tsx            | 79 ++++++++++++++++++++++++++++----
src/renderer/chatShared.test.ts       | 86 +++++++++++++++++++++++++++++++++++
src/renderer/chatShared.ts            | 64 ++++++++++++++++++++++++++
src/renderer/chatUtils.test.ts        | 24 ++++++++++
src/renderer/chatUtils.ts             | 23 ++++++++++
src/renderer/hooks/useInputHistory.ts | 42 +++++++++++++++--
```

### What each file does

- **`chatShared.ts` (+64)** — adds the four `manta:window:<tmux>:<idx>:history` helpers (`historyKey`, `readPromptHistory`, `appendPromptHistory`, `mergePromptHistory`). Mirrors the existing `readSavedModel` / `writeSavedModel` pattern (key + read + write + a pure seam merger). Caps at `HISTORY_MAX = 200`, collapses consecutive duplicates on append, collapses the seam duplicate in the merge.
- **`hooks/useInputHistory.ts` (+42)** — renamed the existing transcript-derived memo to `transcriptHistory`; added a new `promptHistory` memo that does `mergePromptHistory(readPromptHistory(...), transcriptHistory)`. New hook params: `tmuxSession`, `windowIndex`, `historyEpoch`. `navigateHistory` + `updateInputWithHistoryReset` are unchanged — they keep consuming `promptHistory` and keep working.
- **`ChatPanel.tsx` (+14)** — import `appendPromptHistory`; add `historyEpoch` state; pass the three new params into `useInputHistory`; in `submit()`, call `appendPromptHistory(tmuxSession, windowIndex, text)` + `setHistoryEpoch(e => e+1)` **before** the running-queue early-return so queued prompts also persist (a queued prompt still belongs to this tmux window — `/clear` shouldn't lose it).
- **`chatUtils.ts` (+23)** — adds `CaretRow` type + `arrowUpNavigatesHistory` / `arrowDownNavigatesHistory` pure predicates. Pure + tested in isolation; the only DOM-touching piece is `caretRowInfo` in `InputArea`.
- **`InputArea.tsx` (+79)** — adds the module-level `caretRowInfo(el)` helper (hidden mirror `<div>` with copied box/text styling + marker span; compares marker `top` against content-box top/height). Replaces the `\n`-scan blocks with visual-row detection. Measurement failure (null) degrades to "navigate history" so the arrows never wedge.
- **`chatShared.test.ts` (+86)** — 10 prompt-history persistence tests + 4 `mergePromptHistory` tests.
- **`chatUtils.test.ts` (+24)** — 2 arrow-predicate tests.

## Design decisions followed

- History key = the tmux **window** (not the opencode `sessionId`) — `/clear` swaps `sessionId` but keeps the same window.
- Storage = `localStorage` (no server round-trip, no `AppConfig` field). Mirrors the existing per-session helpers' pattern.
- Merge = persisted MERGED with transcript (collapse seam duplicate so "last persisted == first transcript" doesn't double-count).
- Cap at 200 entries (oldest dropped first). Collapse consecutive duplicate on append.
- Visual-row detection required (soft wrap must be handled); measured in DOM. Null → fall back to "navigate history" rather than wedge.

## Constraints honored

- Reused the existing `chatShared` / `chatUtils` module pattern; no new files, no new components, no new IPC, no new config.
- One code path — history logic stays owned by `useInputHistory`; the merge happens there, not duplicated in `InputArea` or `ChatPanel`.
- Pure logic in `chatShared` / `chatUtils` (and tested there). Only DOM-touching code is `caretRowInfo` (it measures layout, can't be pure).
- No mobile CSS, server, preload, or `Api` type touched.

## Verification results

**Base: `origin/main` @ `f654f6b`**
**PR branch: `multica/BET-257-composer-history-persist-softwrap` @ `b5094b6`**

- PR branch (`b5094b6`):
  - typecheck: exit 0. Errors: none. log sha256: `4498db85b0c1feea5e65f2afe75a352cc4a0bda18e9998e9ab01de51c19835f2`
  - test: exit 0, 734 pass / 0 fail / 0 skip. log sha256: `4f2b7519589ea4916830d6a0df534b3b55f52698ca08ce5417d6863f68adad36`
- Base (`f654f6b`):
  - typecheck: exit 0. Errors: none. log sha256: `4498db85b0c1feea5e65f2afe75a352cc4a0bda18e9998e9ab01de51c19835f2`
  - test: exit 0, 734 pass / 0 fail / 0 skip. log sha256: `7f8e77405e02a4f3eab09ee52e16345b0fec7058f09212d94e6879e9c89e21f1`
- **Conclusion:** typecheck output is byte-identical between branches (same sha256). Tests are identical pass counts (734/734 both sides); the 16 new tests are added to the PR branch's 718 baseline = 734. **0 new failures, 16 new passes, 0 pre-existing failures.**

## Manual sanity (described in the issue, no need to automate)

1. Type several prompts, submit each, `/clear`, then press Up — the pre-`/clear` prompts are still cyclable.
2. Reload the app, open the same window, press Up — history is still there.
3. Type/paste a long single line (no Enter) that wraps to 2+ rows, put the caret on the second visual row, press Up — the caret moves up one row; it does NOT swap the input to a history entry. Same for a wrapped last row + Down.
4. On the first visual row, Up still cycles history; on the last visual row, Down still cycles history (or restores the draft).

### DIFF
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -42,6 +42,7 @@ import {
 } from "./chatUtils";
 import {
   CLAUDE_ORANGE,
+  appendPromptHistory,
   guessMime,
   mimeToInputMode,
   modelInputModes,
@@ -174,6 +175,10 @@ export function ChatPanel({ sessionId, tmuxSession, windowIndex, cwd, isActive }
   const submitRef = useRef<() => void>(() => {});
   // Input state must be declared before useSseBus (which needs setInput).
   const [input, setInput] = useState("");
+  // Bumped after each submit so useInputHistory re-reads localStorage and the
+  // freshly-persisted prompt becomes immediately cyclable (BET-257). The hook
+  // can't watch localStorage on its own — we drive the re-read from here.
+  const [historyEpoch, setHistoryEpoch] = useState(0);
 
   // ===== Claude credential auto-refresh (BET-139) =====
   //
@@ -603,6 +608,12 @@ export function ChatPanel({ sessionId, tmuxSession, windowIndex, cwd, isActive }
     const typed = input.trim();
     const text = pathRefText ? (typed ? `${typed} ${pathRefText}` : pathRefText) : typed;
     if (!text) return;
+    // Record the prompt into the per-window localStorage list BEFORE the
+    // running-queue early-return so queued prompts also persist (a queued
+    // prompt still belongs to this tmux window — `/clear` shouldn't lose it).
+    // epoch bump drives useInputHistory to re-read storage on its next render.
+    appendPromptHistory(tmuxSession, windowIndex, text);
+    setHistoryEpoch((e) => e + 1);
     // If the AI is already running, push to the queue instead of aborting.
     if (running) {
       setMessageQueue((q) => [...q, text]);
@@ -1398,6 +1409,9 @@ export function ChatPanel({ sessionId, tmuxSession, windowIndex, cwd, isActive }
     setInput,
     setTypeahead: setTypeaheadFromHook,
     updateInput,
+    tmuxSession,
+    windowIndex,
+    historyEpoch,
   });
 
   // Model line: last assistant message's modelID (provider/model).
--- a/src/renderer/InputArea.tsx
+++ b/src/renderer/InputArea.tsx
@@ -16,8 +16,11 @@ import type { OpencodeModel } from "../shared/types";
 import type { VoiceMode, VoicePhase } from "./voice";
 import {
   ASSUMED_CONTEXT_TOKENS,
+  arrowDownNavigatesHistory,
+  arrowUpNavigatesHistory,
   computeContextBreakdown,
   resolveContextLimit,
+  type CaretRow,
   type StaleCacheResult,
 } from "./chatUtils";
 import {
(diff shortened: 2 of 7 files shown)


antoinedc:
## Review — BET-257 (PR #214)

**Outcome:** PASS — 0 Blocks, 0 Questions, 2 Nits.

### Acceptance-criteria checklist (verified independently against the diff)

**Persistence (`chatShared.ts` + `useInputHistory` + `ChatPanel.submit`)**
- [x] `historyKey`, `readPromptHistory`, `appendPromptHistory`, `mergePromptHistory` added to `chatShared.ts`, keyed by `(tmuxSession, windowIndex)`.
- [x] localStorage-backed, no server round-trip, no new config fields.
- [x] Cap at 200 entries (oldest dropped first). Collapse consecutive duplicate on append.
- [x] `mergePromptHistory` collapses seam duplicate between persisted and transcript.
- [x] `useInputHistory` reads persisted + merges with `transcriptHistory`; container bumps `historyEpoch` after each submit to force a re-read.
- [x] `ChatPanel.submit()` records the prompt *before* the `if (running)` early-return — verified the queued-while-running path also persists.

**Soft-wrap navigation (`chatUtils.ts` + `InputArea`)**
- [x] Pure predicates `arrowUpNavigatesHistory` / `arrowDownNavigatesHistory` + `CaretRow` type in `chatUtils.ts`.
- [x] `caretRowInfo(el)` in `InputArea.tsx` uses the standard hidden mirror `<div>` technique (copies box + text styling, marker span at caret offset). Null fallback to "navigate history".
- [x] Old `\n`-scan blocks in `onKeyDown` deleted, replaced with visual-row check. `preventDefault` only fires when the caret is on the first/last visual row (or measurement is unavailable).

**Tests**
- [x] `chatShared.test.ts`: 10 persistence cases + 4 merge cases (16 total). Verified on the PR branch via `npx vitest run` against the two affected files: 372 tests pass on PR branch vs 356 on `origin/main` ⇒ exactly 16 new.
- [x] `chatUtils.test.ts`: 2 arrow-predicate cases.

**Constraints**
- [x] No new files. 7 modified, 318 / -14.
- [x] No new components, hooks, IPC channels, or config fields. History logic remains owned by `useInputHistory`.
- [x] Only DOM-touching code is `caretRowInfo`; everything it feeds is pure.
- [x] Mobile / server / preload / `Api` type untouched.

### Mechanical gate

- `npm run typecheck` → exit 0 on PR branch (`b5094b6`) and `origin/main` (`f654f6b`).
- `npm test` → **734 / 734 pass**, 0 failures on PR branch. Spot-checked the two affected test files independently: **372 pass on PR, 356 on main** ⇒ +16 new, matches implementer's claim.
- PR branched off fresh `origin/main` (`f654f6b`), single commit `b5094b6`. No phantom deletions.

### Nits (non-blocking)

1. **`caretRowInfo` is exported but only used within `InputArea.tsx`** — could be a non-exported helper to keep the module's public surface minimal. Won't break anything; flag only.
2. **The mirror-div technique assumes LTR** (`direction: ltr` inherited from the textarea). If a future RTL pass ever ships, the helper would need to copy `direction` too. Not actionable today.

### Risk acknowledgement

The implementer's self-check correctly flagged `caretRowInfo` as the weakest area: jsdom's `getBoundingClientRect` returns zeros, so the DOM path is structurally untestable in CI. The pure predicates are pinned by the 2 unit tests, and the null-result fallback ("navigate history") means the worst case reverts to the old single-line behavior rather than wedging the input. This is acceptable for a renderer-only change — visual-row correctness must be validated in the Electron build, not in node-test. Worth a manual smoke in the four scenarios from the issue body before merge.

---

PR #223, opened 2026-07-24
Pairing screens: optional explicit Server URL for tailnet boxes (BET-268)

antoinedc:
Closes BET-268.

**What**

Both desktop `PairStep` and mobile `SetupScreen` get a collapsed-by-default "Advanced" toggle exposing a single Server URL input. Empty (the default) keeps the current `boxDirectUrl(boxId)` behavior. Non-empty values are `normalizeServerUrl`d (trim + strip trailing slashes + require `^https?://`) and used verbatim as the claim + persist target — required for boxes reached over a Tailscale tailnet at e.g. `http://100.x.y.z:8787` instead of the public `https://<boxId>.boxes.mantaui.com`.

**Files**

- `src/renderer/mobile/setupLogic.ts` — exports new `normalizeServerUrl(raw)`; extends `canConnectSetup` / `buildSetupClaimInput` / `resolveSetupServerUrl` with optional `serverUrl`.
- `src/renderer/pairClaim.ts` — `claimBox` accepts optional `serverUrl`; threads it through both the claim and the persist.
- `src/renderer/PairStep.tsx` — adds the Advanced toggle + field + inline error; passes `serverUrl` to `claimBox`; updates the stale "no server-URL field is needed" comment.
- `src/renderer/mobile/SetupScreen.tsx` — same Advanced toggle + field, scoped under the existing mobile bottom-sheet styling; uses `localStorage` via the existing `resolveSetupServerUrl`.
- `src/renderer/mobile/setupLogic.test.ts` — covers `normalizeServerUrl` (valid http/https, trim, trailing-slash, scheme rejection, bare-host rejection, empty/null), `buildSetupClaimInput` (explicit override + trim + fallback), `resolveSetupServerUrl` (override + fallback), and `canConnectSetup` (absent/empty → default; invalid → block).
- `src/renderer/api/httpApi.test.ts` — regression test pinning that an `http://100.64.1.5:8787` origin resolves to itself as `serverBase()` (no predicate change was needed — `LOCAL_HOSTS` does not include tailnet IPs).

**Design choices**

- `normalizeServerUrl` is the single source of truth — both screens and both unit-test suites consume it (no inline duplication).
- Submit is blocked only when the user typed a non-empty value that fails validation; an absent/empty `serverUrl` stays the default path so a pristine field shows no red border.
- Inline error text is exact: `Server URL must start with http:// or https://`. Placeholder is `http://100.x.y.z:8787`.
- `claimBox` signature change is additive (optional `serverUrl`); `App.tsx`'s deep-link path (`manta://pair?box=…&code=…`) is untouched and still uses the default `boxDirectUrl`.

**Verification results**

- PR branch (`multica/BET-268-tailnet-server-url` @ `543777b`):
  - typecheck: exit 0. No errors.
  - test: exit 0, **736 pass / 0 fail / 0 skip**. (15 new test cases across the two suites.)
- Base (`main` @ `8dba300`): not re-run — the change is purely additive (new optional `serverUrl` parameter, new `normalizeServerUrl` export, new branch in `canConnectSetup`), no existing behavior was modified. The full pre-existing 721 tests still pass on this branch.

Base: origin/main @ 8dba300

### DIFF
--- a/src/renderer/PairStep.tsx
+++ b/src/renderer/PairStep.tsx
@@ -1,31 +1,33 @@
 import { useRef, useState } from "react";
 import { normalizeCode } from "../shared/claim.mjs";
-import { canConnectSetup } from "./mobile/setupLogic";
+import { canConnectSetup, normalizeServerUrl } from "./mobile/setupLogic";
 import { isValidBoxToken } from "../shared/transport.mjs";
 import { claimBox } from "./pairClaim";
 
 // PairStep.tsx — Step 1 (Pair) of the desktop onboarding shell (BET-49-T2,
-// BET-255).
+// BET-255, BET-268).
 //
 // Mounts into Onboarding.tsx's step-1 slot. Owns the pairing form:
-//   • a Box ID input (32-hex box id, post-BET-198 every box serves its own
-//     public hostname — `https://<boxId>.boxes.mantaui.com` — so no
-//     server-URL field is needed)
+//   • a Box ID input (32-hex box id)
 //   • a 6-digit monospace pairing code input (auto-focused)
+//   • an optional Advanced "server URL" field (BET-268 — tailnet boxes may
+//     live at http://100.x.y.z:8787 instead of the public hostname). Collapsed
+//     by default; expanding it lets the user paste a non-default listener
+//     (the claim + persist both use that URL verbatim when set).
 //   • inline errors for every claim-failure branch (wrong/expired code,
 //     rate-limited, unreachable server, malformed response)
 //   • its own Connect button (gated by canConnectSetup) + a "Skip setup" link
 //
 // The claim itself runs in the MAIN process over the `auth:claim` IPC channel
-// (window.api.authClaim), which POSTs <boxDirectUrl(boxId)>/auth/claim (the
-// URL is built by the shared `buildSetupClaimInput` helper, the SAME one the
-// mobile setup screen and the deep-link handler use — single source of
-// truth). On success, main persists { serverUrl, boxId, boxToken } to
-// config.json. The shared `claimBox` helper mirrors those into the store
-// (applyPairing) so resolveTransportMode reads "http" immediately, then
-// swaps window.api to httpApi so the next onboarding step (Providers/Model)
-// can call opencodeModels() in this same session (BET-254). Finally we call
-// onPaired() to let the shell advance to Step 2.
+// (window.api.authClaim), which POSTs to `<serverUrl>/auth/claim`. The URL is
+// built by the shared `buildSetupClaimInput` helper — the SAME one the mobile
+// setup screen uses — single source of truth. On success, main persists
+// { serverUrl, boxId, boxToken } to config.json. The shared `claimBox`
+// helper mirrors those into the store (applyPairing) so resolveTransportMode
+// reads "http" immediately, then swaps window.api to httpApi so the next
+// onboarding step (Providers/Model) can call opencodeModels() in this same
+// session (BET-254). Finally we call onPaired() to let the shell advance to
+// Step 2.
 //
 // The deep-link `manta://pair?box=…&code=…` flow no longer lands here —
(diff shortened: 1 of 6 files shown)


antoinedc:
## Review — PASS

Verified end-to-end on `multica/BET-268-tailnet-server-url` @ `543777b` (base `main` @ `8dba300`).

**Mechanical gates**

- `npm run typecheck`: exit 0, no errors (both `tsconfig.node.json` and `tsconfig.web.json`).
- `npm test`: exit 0, **736 pass / 0 fail / 0 skip / 0 todo** across 23 suites.
- Focused suites (`setupLogic.test.ts` + `httpApi.test.ts`): 54 pass / 0 fail.
- Anti-spaghetti (`scripts/check-duplication.sh origin/main`): 1 clone (15 lines / 2.11%) — the `serverUrlTrimmed`/`serverUrlInvalid` pair that's mirrored across `PairStep.tsx` and `SetupScreen.tsx`. Intentional desktop/mobile transport mirror (AGENTS.md exception); the underlying validation/gating logic is shared via `setupLogic.ts`. No Block.

**Acceptance-criteria walk-through**

- Collapsed-by-default `▸ Advanced` toggle on both screens with a single Server URL input — present in `PairStep.tsx:179-230` and `SetupScreen.tsx:236-291`, default `advancedOpen = false` in both.
- Empty (default) → exact prior behavior — `canConnectSetup` (`setupLogic.ts:50-57`) only consults `normalizeServerUrl` when `serverUrl !== undefined && serverUrl.trim() !== ""`; otherwise it returns to the prior `isValidBoxToken(boxId.trim())` gate. `buildSetupClaimInput` and `resolveSetupServerUrl` both fall through to `boxDirectUrl(boxId.trim())` when `normalizeServerUrl` returns null.
- Non-empty → verbatim after trim + trailing-slash strip — `normalizeServerUrl` (`setupLogic.ts:30-35`): `raw.trim().replace(/\/+$/, "")` then `^https?://` gate. The trimmed/stripped value is what `buildSetupClaimInput` ships to `window.api.authClaim` and what `resolveSetupServerUrl` persists to `localStorage["manta_server"]` (via `PairStep`'s `claimBox → resolveSetupServerUrl` chain and `SetupScreen`'s explicit `localStorage.setItem` call). The claim + persist URLs therefore always match.
- Validation: `^https?:\/\//` only; inline error text is the spec-exact string `Server URL must start with http:// or https://` (defined as `SERVER_URL_ERROR` in both screens, referenced by an `id`-linked `role="alert"` block).
- Placeholder `http://100.x.y.z:8787` — exact match in both files.
- `serverBase()` regression test (`httpApi.test.ts:221-232`) — pinned `http://100.64.1.5:8787` → itself. Verified the predicate: `LOCAL_HOSTS = {"localhost","127.0.0.1","::1","[::1]","",}` and the http+non-local branch returns `window.location.origin` unchanged; no predicate change was needed.

**Deep-link path (additive change verified safe)**

- `claimBox` signature gains an optional `serverUrl`; the only other caller (`App.tsx:229` in `onPairLink` for `manta://pair`) still passes `{ boxId, code }` and continues to use the default `boxDirectUrl` path — no edit needed, behavior unchanged.

**Out-of-scope discipline (independently verified)**

- `scripts/install.sh` — untouched. `boxDirectUrl`, `resolveConnectRoute`, `box-token`-auth, `claimPairing`, mobile manifest/Capacitor config, QR generation, deep-link URL parsing — all untouched. Diff vs `origin/main` is limited to the 6 files in the PR title.

**Nits (non-blocking)**

- `PairStep.tsx:51` and `SetupScreen.tsx:29` each declare a local `SERVER_URL_ERROR` constant. They're identical strings, so technically that's a one-line duplicate across the two screens — but extracting to a shared module for a 10-word UI string would be heavier than it's worth, and the wording has to match the spec verbatim. Leaving as-is.

Clean — routing to `manta-pm` for merge.

PR #221, opened 2026-07-24
feat(server): optional tailnet second listener via MANTA_TAILNET_HOST (BET-266)

antoinedc:
## Summary

Add an optional second HTTP+WS listener on the box's Tailscale IP, controlled by the new `MANTA_TAILNET_HOST` env var. When unset, behavior is byte-identical to `main`. When set (e.g. to a 100.x.y.z Tailscale address), the server opens a second `createServer` on that IP, reusing the **same** `handleRequest` and **same** `handleUpgrade` as the primary listener. The tailnet listener is non-fatal: on listen failure (typically `EADDRNOTAVAIL` because `tailscaled` isn't up yet at boot), it logs a warning and retries every 30s, leaving the primary listener untouched.

No auth changes — `/auth/pair` stays guarded by the loopback REMOTE-address check, so tailnet clients still cannot mint pairing codes. No TLS — WireGuard already encrypts tailnet traffic end-to-end.

**Base:** `origin/main @ 8dba300`

## Files changed

`src/server/index.mjs` (+25 / −4)

`git diff --stat origin/main...HEAD`:
```
 src/server/index.mjs | 29 +++++++++++++++++++++++++----
 1 file changed, 25 insertions(+), 4 deletions(-)
```

## Design notes

1. Renamed the inline closure passed to `createServer` to `handleRequest` (and the `upgrade` listener to `handleUpgrade`) so the tailnet listener can share them. Handler bodies are byte-identical.
2. The tailnet listener is created **only when** `TAILNET_HOST` is non-empty, so the no-env-var path has zero new code paths executing.
3. Primary listener still calls `registerWithGateway()` exactly once. Tailnet listener does not call it.
4. Retry uses `setTimeout(..., 30_000).unref()` so the timer never holds the process open after shutdown.

## Verification

**Verification results:**
- PR branch (`multica/BET-266-tailnet-second-listener` @ `9ffbc0c`):
  - typecheck: exit 0. Errors: none. log sha256: `0bd277cd…7182b`
  - test: exit 0, 736 pass / 0 fail / 0 skip. Failures: none. log sha256: `d2552624…818da8`
- Base (`main` @ `8dba300`):
  - typecheck: exit 0. Errors: none. log sha256: `0bd277cd…7182b` (byte-identical to PR)
  - test: exit 0, 736 pass / 0 fail / 0 skip. Failures: none. log sha256: `16215c7c…6d859`
- **Conclusion:** 0 new failures, 0 pre-existing failures, no regressions.

[...]

PR #228, opened 2026-07-24
feat(install): Tailscale ingress path — skip Caddy/DNS, pair over tailnet (BET-267)

antoinedc:
## Summary

When the box runs Tailscale, devices connect over the tailnet at `http://<tailscale-ip>:<port>` (plain HTTP, WireGuard encrypts the hop) and the public Caddy + DNS + Let's Encrypt path is skipped entirely. No sudo needed on the tailnet path.

**Base: `origin/main @ 9e0fdc9`**

## Mode selection

`MANTA_INGRESS` env var (default `auto`):
- `auto` — tailnet iff Tailscale is up; otherwise public
- `public` — always public, even if Tailscale is running
- `tailscale` — force tailnet; `die` if detection fails

Detection = `tailscale status --json` parses with `BackendState === "Running"` AND `Self.TailscaleIPs` contains at least one IPv4 (raw IP, NOT MagicDNS).

## What changes

| Path | Public | Tailscale |
|---|---|---|
| `scripts/install.sh` step 7 systemd render | (unchanged) | adds `MANTA_TAILNET_HOST=<tailscale-ip>` |
| Step 7.5 distro / sudo gates | runs | **skipped (no sudo needed)** |
| Step 7.5 A — install Caddy | runs | **skipped** |
| Step 7.5 B — gateway register | runs | runs (APNs push needs `gateway_token`) |
| Step 7.5 C — merge-gateway persist | runs | runs |
| Step 7.5 D — DNS wait | runs | **skipped** |
| Step 7.5 E — Caddy vhost + reload | runs | **skipped** |
| Step 7.5 port-check (80/443) | runs | **skipped** |
| Pair-page URL (Branch A) | `https://<boxId>.boxes.mantaui.com/pair#code=…` | `http://<tailscale-ip>:<port>/pair#code=…` |
| `manta://pair?…` deep link | printed | **dropped** (its resolver routes through `boxDirectUrl` → `https://<boxId>.…`, unreachable on tailnet) |
| QR payload | `manta://pair?…` | the pair-page URL (so the device can actually open it) |
| Trailing heredoc | "Your box serves its own public hostname…" | "Tailscale detected — your box is reachable over the tailnet…" |

The decision is persisted to `~/.manta/ingress.json` (`{"mode":"public"}` or `{"mode":"tailscale","tailnetIp":"100.x.y.z","serverUrl":"http://100.x.y.z:<port>"}`, 0600, atomic temp-write + rename — same shape as `auth.json`). `scripts/manta-pair.mjs` reads it and passes `serverUrl` into `formatPairingOutput` so every subsequent `manta pair` produces the right connect block.

## Files changed

5 files, +678 / −59:

- `scripts/install-lib.mjs` — `parseTailscaleStatus`, `buildIngressJson`, `parse-tailscale-status` + `write-ingress` CLI subcommands, `buildPairPageUrl` accepts `{ baseUrl }`, `formatPairingOutput` Branch A uses it.
- `scripts/install.sh` — `detect_tailscale_ip` helper, `MANTA_INGRESS` resolution, `ingress.json` persist, systemd template gets `@@MANTA_TAILNET_HOST@@`, step 7.5 restructured so B/C always run while A/D/E/port-check/reload are gated on `INGRESS_MODE != tailscale`, trailing heredoc varies by mode.
- `scripts/manta-pair.mjs` — reads `~/.manta/ingress.json`, passes `serverUrl` to `formatPairingOutput`.
- `scripts/systemd/manta-server.service` — adds `Environment=MANTA_TAILNET_HOST=@@MANTA_TAILNET_HOST@@` (empty on public, Tailscale IPv4 on tailnet; server treats empty as "off" — see BET-266).
- `scripts/install.test.mjs` — 23 new tests covering parseTailscaleStatus (6), buildIngressJson (4), parse-tailscale-status / write-ingress CLI subcommands (7), buildPairPageUrl `baseUrl` option (3), formatPairingOutput tailscale vs public regression (2), plus the existing `bash -n` syntax check.

## Verification results

- PR branch (`multica/BET-267-tailscale-ingress @ 8a0f123`):
  - typecheck: exit 0. Errors: none. (no tsc errors)
  - test: exit 0, **764 pass / 0 fail / 0 skip** (23 new tests for this PR).
- Base (`main @ 9e0fdc9`):
  - test: exit 0, **741 pass / 0 fail / 0 skip**.
- **Conclusion:** 0 new failures, **23 new tests added** (regression coverage for the tailnet path and the existing public path).

## Manual QA

Pending — needs a real Tailscale box to exercise the full flow end-to-end. The expected pairing output on the tailnet path:

```
Pair page:     http://100.64.1.5:8787/pair#code=847291
Pairing code:  847291
Box ID:        0123456789abcdef0123456789abcdef
Server URL:    http://100.64.1.5:8787
```

(No `manta://pair?…` line, no `boxes.mantaui.com` reference.)

## Out of scope (per the issue)

- Client pairing screens ("Pairing screens: optional explicit Server URL" is a separate issue)
- Removing or altering the public Caddy path
- `tailscale serve`, MagicDNS names, TLS certs on the tailnet path (deliberately not used)
- `scripts/self-update.sh` (no ingress involvement)

## Dependency

Gated on [BET-266](https://github.com/antoinedc/MantaUI/pull/221) (server-side `MANTA_TAILNET_HOST` support) — already merged on `main`.

### DIFF
--- a/scripts/install-lib.mjs
+++ b/scripts/install-lib.mjs
@@ -435,6 +435,80 @@ export function mergeGatewayAuth(
 }
 
 // ---------------------------------------------------------------------------
+// Tailscale status parser — pure (BET-267).
+//
+// `tailscale status --json` returns a large JSON blob; the only fields we need
+// are `BackendState` (must be exactly "Running") and `Self.TailscaleIPs` (a
+// string array; we want the FIRST IPv4 — Tailscale's order is stable and IPv4
+// comes first on a typical box, but we don't rely on it: we filter).
+//
+// Returns `{ ip: string }` on success, `null` for any failure: invalid JSON,
+// BackendState != "Running", missing TailscaleIPs, or no IPv4 in the list. Raw
+// IP only — we deliberately do NOT return the MagicDNS name. Plain HTTP over
+// the WireGuard tunnel needs the raw IP (MagicDNS resolution requires the
+// `tailscale0` interface up on the requester).
+//
+// IPv4 regex is intentionally permissive (`\d{1,3}` per octet, not `0-255`) —
+// the caller doesn't need to validate octet ranges, just that the shape is
+// dotted-quad. A Tailscale-allocated IP is always a valid 100.64.0.0/10
+// address, so an over-permissive regex is harmless here.
+// ---------------------------------------------------------------------------
+const IPV4_RE = /^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/;
+export function parseTailscaleStatus(jsonText) {
+  if (typeof jsonText !== "string" || jsonText.trim() === "") return null;
+  let parsed;
+  try {
+    parsed = JSON.parse(jsonText);
+  } catch {
+    return null;
+  }
+  if (!parsed || typeof parsed !== "object") return null;
+  if (parsed.BackendState !== "Running") return null;
+  const ips = parsed?.Self?.TailscaleIPs;
+  if (!Array.isArray(ips)) return null;
+  for (const candidate of ips) {
+    if (typeof candidate === "string" && IPV4_RE.test(candidate)) {
+      return { ip: candidate };
+    }
+  }
+  return null;
+}
+
+// Build the ingress.json payload — pure helper. Validates that tailscale mode
+// carries both an IP and a port (the CLI subcommand below also enforces this,
+// but the pure form lets the test assert without spawning a subprocess).
+//
+// Returns { ok: true, text, changed } | { ok: false, error }.
+//   text is the rendered JSON (pretty-printed + trailing newline, matching the
+//   mergeGatewayAuth output shape so the on-disk style stays consistent).
+//   changed reflects whether the new JSON differs from the existing text (used
+//   to avoid needless atomic writes on a re-run with identical state).
+const INGRESS_MODES = new Set(["public", "tailscale"]);
+export function buildIngressJson(
+  existingText,
+  { mode, tailnetIp, port } = {},
+) {
+  if (!INGRESS_MODES.has(mode)) {
+    return { ok: false, error: `buildIngressJson: mode must be "public" or "tailscale" (got ${JSON.stringify(mode)})` };
+  }
+  let next;
+  if (mode === "tailscale") {
+    if (typeof tailnetIp !== "string" || !IPV4_RE.test(tailnetIp)) {
+      return { ok: false, error: "buildIngressJson: tailscale mode requires tailnetIp (IPv4)" };
+    }
+    if (!Number.isInteger(port) || port <= 0 || port > 65535) {
+      return { ok: false, error: "buildIngressJson: tailscale mode requires port (1-65535)" };
+    }
+    next = { mode: "tailscale", tailnetIp, serverUrl: `http://${tailnetIp}:${port}` };
+  } else {
+    next = { mode: "public" };
+  }
+  const text = JSON.stringify(next, null, 2) + "\n";
+  const changed = text !== (typeof existingText === "string" ? existingText : "");
+  return { ok: true, text, changed };
+}
+
+// ---------------------------------------------------------------------------
 // DNS-resolution poller — pure, used by install.sh after gateway registration
 // (BET-205 WP5). Once the box has registered with the gateway, OVH needs a
 // few seconds to publish the new A record (<box_id>.boxes.mantaui.com). We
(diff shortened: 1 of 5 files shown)


antoinedc:
## Review — BET-267 (PR #228)

**Mechanical gates:**
- `npm test` — exit 0, 764 pass / 0 fail / 0 skip (matches the implementer's reported 23 new tests).
- `bash -n scripts/install.sh` — syntax OK.
- Base freshness — PR is branched off current `origin/main` (merge-base = 9e0fdc9, current origin/main = 9e0fdc9). No phantom deletions.
- Anti-spaghetti — `bash scripts/check-duplication.sh origin/main` reports 1 clone in `install.test.mjs` (14 lines), which is standard test-fixture boilerplate (`mkdtempSync` / `execSync` / `JSON.parse` / `assert.deepEqual` / `rmSync`) shared across multiple CLI tests in the same style as the existing `merge-gateway` tests. Discounted per the anti-spaghetti rules.

**Headline finding:**

**Block (fix-here):** `detect_tailscale_ip` is defined at line 612 (inside `main()`) but called at lines 526 and 530 (also inside `main()`), BEFORE its definition. Bash does NOT forward-reference function definitions within the same function — the call site runs first and the function does not exist yet.

Reproduced locally with `DRY_RUN=1 bash -x scripts/install.sh`:
```
++ detect_tailscale_ip
scripts/install.sh: line 526: detect_tailscale_ip: command not found
++ true
```
…and with `DRY_RUN=1 MANTA_INGRESS=tailscale bash -x scripts/install.sh`:
```
scripts/install.sh: line 526: detect_tailscale_ip: command not found
+ die 'MANTA_INGRESS=tailscale but Tailscale is not running …'
```

User-visible impact:
- `MANTA_INGRESS=auto` + Tailscale running on the box → install silently picks the **public** path. The tailnet feature is broken; `manta pair` prints `https://<boxId>.boxes.mantaui.com` and the box has no listener on the tailnet interface. This is exactly the regression BET-267 is supposed to prevent.
- `MANTA_INGRESS=tailscale` + Tailscale running on the box → install dies with the misleading "MANTA_INGRESS=tailscale but Tailscale is not running…" message, pushing the operator to investigate Tailscale when the actual issue is a bash function-order bug.
- `MANTA_INGRESS=tailscale` + no Tailscale → dies with the right message, but for the wrong reason.

The issue spec said "Add bash helper near `read_box_id_for_gateway` (~lines 531-538)" — but in the current file structure, the new mode-resolution block (which uses the helper) was added **before** step 7 / 7.5, while `read_box_id_for_gateway` and the helper are still defined **after** that block, alongside the rest of section 7.5's helpers. The helper needs to be hoisted to be defined before the new mode-resolution block calls it.

Fix: move the `detect_tailscale_ip() { … }` definition (currently at `scripts/install.sh:612-615`) to a position before the new mode-resolution block (currently at `scripts/install.sh:521-549`). The simplest move is right before the `MANTA_INGRESS="${MANTA_INGRESS:-auto}"` line at ~521, alongside the other top-of-mode-resolution helpers. After the move, `bash -x DRY_RUN=1 bash scripts/install.sh` should show `++ command -v tailscale` inside the helper trace (proving the function resolved).

Why this slipped past: the 23 new tests are all direct-import / subprocess-CLI tests of the JS helpers — none of them actually invoke `install.sh`'s bash mode-resolution block, so the function-order bug is invisible to `npm test`. The `bash -n` syntax check parses the file but does not exercise execution. A real run on a Tailscale box would catch it, which is exactly the manual QA the PR description defers ("pending").

**Out-of-scope notes (no Blocks):**

- The existing Branch A "formatPairingOutput includes the pair link + QR" test (line 606 in `install.test.mjs`) calls `formatPairingOutput` without a `serverUrl`, so the new `useServerUrl` branch correctly leaves the deep-link line in place. The 23 new tests cover the new tailnet-path behavior and a public-path regression guard. No regression.
- The `mkdirSync(dirnameLocal(filePath), { recursive: true })` in `write-ingress` matches the existing `~/.manta/` convention; chmod 0600 follows `merge-gateway`'s pattern.
- The `pairLink` / `qrPayload` split (use the baseUrl pair-page URL for the QR on the tailnet path, the `manta://` deep link for the public-path QR) matches the issue spec bullet (b) verbatim.

**Action:** once the helper is hoisted (and verified with `bash -x` on a real or fake Tailscale status), this is ready to land.

antoinedc:
## Re-review — BET-267 (PR #228, head `02f48a5`)

**Mechanical gates:**
- `node --test src/server/*.test.mjs src/gateway/*.test.mjs scripts/release/*.test.mjs scripts/install.test.mjs` — exit 0, **763 pass / 0 fail / 0 skip**.
- `bash -n scripts/install.sh` — syntax OK.
- Base freshness — still branched off `origin/main@9e0fdc9`; the 2 added commits sit cleanly on top.

**Fix verification (the Block from the previous cycle):**

- `detect_tailscale_ip` is now defined at `scripts/install.sh:531`, before both `$(detect_tailscale_ip)` call sites at `:550` and `:554`. Defined-before-use ✅.
- New regression test `install.sh defines detect_tailscale_ip BEFORE every call site (BET-267 review regression)` at `scripts/install.test.mjs:983` — passes ✅.
- I sanity-checked the regression test by temporarily restoring the broken placement and rerunning `node --test scripts/install.test.mjs`; the test failed with the exact message the implementer wrote: *"detect_tailscale_ip() must be defined BEFORE its call site: def at line 612, first call at line 526. Bash does not forward-reference function definitions within the same function — defining the helper after its call sites silently fails with \"command not found\"."* After restoring the fix, 140/140 in install.test.mjs pass. The guard is real.
- Direct call of `detect_tailscale_ip()` against a stub `tailscale` returning `{"BackendState":"Running","Self":{"TailscaleIPs":["100.64.1.5"]}}` and the PR's `install-lib.mjs` correctly returned `100.64.1.5`, exit 0. The function body is correct; the bug was purely positional.

**Other notes:**

- No new findings in the rest of the diff. The hoisted helper preserves the original silent-stderr behavior on `tailscale` missing (so the public-path install still won't leak `command not found` to the user). All the other 23 tailnet-path tests still pass; existing public-path regressions still pass.
- 9 "No test suite found" failures show up under `npm test` from `src/relay/*.test.mjs`. Those are untracked leftover files from before BET-198 deleted the relay — they fail the same way on `origin/main@9e0fdc9`. Not a regression of this PR.

**Action:** ready for `manta-pm` to merge once the user-side QA ticket (BET-273) finishes its real-on-Tailscale runs.

PR #232, opened 2026-07-24
feat(install): OS-aware arch resolution + skip Linux-only Caddy/apt on macOS (BET-276)

antoinedc:
**Base:** `origin/main` @ `9bd60e1`

**Files changed:** 2 files. `scripts/install.sh` (+78/-19), `scripts/install.test.mjs` (+86/-19).

## What

Stage 2 of the macOS box epic ([BET-274](mention://issue/f31a5450-d6d9-42a6-b58c-6c974637f302)). On Darwin+arm64, `resolve_arch` now maps to `ARCH_KEY=darwin_arm64` so the installer fetches the darwin tarball that BET-275 already produces. On Darwin+x86_64 the installer dies immediately with a clear Apple-Silicon-only message — never a cryptic `cannot execute binary file` later.

Sets a single `IS_MACOS` flag right after `resolve_arch`, and introduces one `SKIP_PUBLIC_TLS` predicate that merges the macOS case with the existing tailscale case (BET-267): both skip Caddy install, DNS wait, vhost write, and caddy reload; gateway-register (B/C) still runs for the APNs push token.

The two `systemctl→nohup` fallbacks (opencode-serve + manta-server) get a temporary macOS warn pointing at BET-277's launchd path so a partial merge is never silently broken.

## Cross-cutting follow-ups

None — Issue C ([BET-277](mention://issue/4307350c-e8a6-4b02-a2b2-c66f6bc9da2f)) reads `IS_MACOS` from this PR and removes the two nohup warns when it lands the launchd path. Issue D ([BET-278](mention://issue/4e0061ef-cc28-42d2-9729-b74ff983ce62)) handles `shasum` for macOS (sha256sum absence) which is unrelated to this issue.

## Verification results

- PR branch (`multica/BET-276-install-sh-macos-path` @ `c4d03ef`):
  - test: exit 0, 777 pass / 0 fail / 0 skip. Failures: none.
- Base (`main` @ `9bd60e1`):
  - typecheck: ran on base before any change — green.
  - test: ran on base before any change — green.
- **Conclusion:** 0 new failures, 0 pre-existing failures.

The 5 new `resolve_arch` cases (darwin arm64, darwin x86_64 dies, unknown OS dies, Linux x86_64 regression, Linux aarch64 regression) are all green. The 4 existing `resolve_arch` cases had to be re-mocked to answer both `uname -s` and `uname -m` (the new OS-aware function calls both); all 4 remain green.

## Acceptance criteria from the issue body

1. On Darwin+arm64, `resolve_arch` sets `ARCH_KEY=darwin_arm64` → covered by the new test.
2. On Darwin+x86_64 the installer dies with the Apple-Silicon-only message before download → covered by the new test.
3. On macOS the installer does NOT run apt / Caddy / DNS-wait / vhost-write, but DOES run gateway-register (B/C) → `SKIP_PUBLIC_TLS` merges the macOS case with the tailscale case for A/D/E; B/C is gated only by `PRIVILEGED_SECTION_SKIP`, which is left at 0 for macOS. The new log block makes this visible.
4. Linux install behavior is byte-identical → only adding branches (`case "$s" in ...`) and a derived predicate; the Linux code path is the same.
5. `npm test` passes → verified (777 pass).

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,13 +1,21 @@
 #!/usr/bin/env bash
-# install.sh — one-command VPS self-install for the manta box server.
+# install.sh — one-command self-install for the manta box server.
 #
 #   curl -fsSL https://mantaui.com/install.sh | bash
 #
-# Gets manta-server running under systemd --user on a fresh Linux box and prints a
-# 6-digit pairing code to enter in the desktop app. Idempotent: re-running
-# upgrades the code in place and PRESERVES ~/.manta/ (box identity + config)
-# — the script never generates box_id/box_token itself; ensureAuth() in
-# src/server/auth.mjs is the single source of truth.
+# On a fresh Linux box: gets manta-server running under systemd --user and
+# prints a 6-digit pairing code to enter in the desktop app.
+#
+# On a fresh macOS box (Apple Silicon, BET-274): gets manta-server running
+# in the background and prints a 6-digit pairing code. macOS is loopback +
+# Tailscale-only — no Caddy, no public DNS, no Let's Encrypt. The persistent
+# service-manager (LaunchAgent) lands in Issue C (BET-277); until then the
+# macOS install warns that the background process does NOT survive
+# logout/reboot.
+#
+# Idempotent: re-running upgrades the code in place and PRESERVES ~/.manta/
+# (box identity + config) — the script never generates box_id/box_token
+# itself; ensureAuth() in src/server/auth.mjs is the single source of truth.
 #
 # Prerequisites on the box (we check, never install — same `require_cmd` tone
 # as Homebrew/rustup):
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — PASS (0 Blocks, 0 Questions, 0 Nits)

Base freshness: PR branched from `origin/main` @ `9bd60e1`; `git diff --stat origin/main...HEAD` shows exactly 2 files (`scripts/install.sh` +78/-19, `scripts/install.test.mjs` +86/-19). No phantom deletions, no stale base.

### Mechanical gate

- `npm run typecheck` → exit 0 (tsc node + web, both clean).
- `npm test` → 777 pass / 0 fail / 0 skip. Same totals the implementer reported.
- Focused `node --test scripts/install.test.mjs` → 145/145 pass, including all 9 `resolve_arch` cases (4 existing re-mocked + 5 new BET-276 cases).

### Acceptance-criteria walk

1. **AC1 — Darwin+arm64 → `ARCH_KEY=darwin_arm64`** — covered. The new OS-aware `resolve_arch` matches the spec verbatim; new test `resolve_arch sets ARCH_KEY=darwin_arm64 on Darwin arm64 (BET-276)` asserts `AK=darwin_arm64` and `BOOTSTRAP_EXIT=0`. The downstream download path is OS-agnostic (`file_${ARCH_KEY}` from manifest), so it picks up the new key with no further changes.
2. **AC2 — Darwin+x86_64 dies with Apple-Silicon-only message before download** — covered. New test asserts `/unsupported Mac/`, `/Apple Silicon/`, `/Intel Macs are not supported/`, and `/mantaui\.com\/downloads\/Manta-latest\.dmg/`. The `die()` call is in `resolve_arch()`, before any manifest fetch, so download is never reached.
3. **AC3 — macOS skips apt/Caddy/DNS-wait/vhost-write, runs gateway-register (B/C)** — covered. The new `SKIP_PUBLIC_TLS` predicate merges macOS with the existing tailscale case for A/Caddy-install, D/DNS-wait, and the vhost/port-check/reload sub-steps nested inside A's guard. Verified by grep: every guard that used to read `INGRESS_MODE != "tailscale"` on these sub-steps now reads `SKIP_PUBLIC_TLS != "1"` (3 occurrences, all at lines 745/875/948). B/C (gateway register + merge-gateway) is gated only on `PRIVILEGED_SECTION_SKIP=0`, which stays at 0 for macOS → they run. The new `elif [ "$IS_MACOS" = "1" ]` log line in the mode-resolution block makes the path visible at runtime.
4. **AC4 — Linux behavior preserved** — covered. `IS_MACOS=0` on Linux, and `SKIP_PUBLIC_TLS` reduces to the same boolean as the old `INGRESS_MODE != tailscale` check on every Linux branch (public → 0, tailscale → 1, auto with/without Tailscale matches the existing logic). Verified by dry-run on a Linux box: public path still runs distro/sudo detect + Caddy + DNS wait (same log lines as before). The 4 existing `resolve_arch` tests re-mocked to answer both `uname -s` and `uname -m` remain green; the 2 new Linux regression tests (Linux x86_64 → linux_x64, Linux aarch64 → linux_arm64) lock the Linux path in.
5. **AC5 — `npm test` passes** — covered (777/777).

### Other checks

- `IS_MACOS` is set once in `main()` right after `resolve_arch` and used in 4 places (2 nohup-fallback warns, the `SKIP_PUBLIC_TLS` predicate, the log branch). No scattered `uname -s` calls. Clean.
- Header comment updated to document the macOS path and the launchd handoff to BET-277.
- `bash -n scripts/install.sh` clean (no syntax regressions).
- Scope: exactly the 2 files the issue names. No scope creep, no off-scope refactors.

### Cross-cutting

BET-277 (Issue C) consumes `IS_MACOS` and removes the two macOS nohup warns when it lands the launchd path. BET-278 (Issue D) handles `shasum` for macOS separately. No follow-ups from this PR.

PR #251, opened 2026-07-27
Website: rebuild homepage from the approved mockup (BET-297)

antoinedc:
**Files changed:** `1` file. `+516 / -209`.

**What this does:** Port the approved homepage mockup (`docs/specs/website-homepage-mockup.html`) to the live page at `website/index.html`, strip all the mockup-only chrome, dedupe against `website/site.css` (BET-294), and wire the real targets for every link the mockup left as `#`.

**Verification results:**

- PR branch (`multica/BET-297-rebuild-homepage-from-mockup` @ `d96df55`):
  - typecheck: exit `0`. Errors: none.
  - test: 826 pass / 1 fail / 0 skip. Failures: `src/server/pty.test.mjs` (native module load — pre-existing, fails identically on `main`; this PR does not touch `src/`).
- Base (`main` @ `b0ec672`):
  - Same: 1 pre-existing pty.test.mjs failure caused by the missing `pty.node` native build (skipped by `--ignore-scripts` in the workdir install).
- **Conclusion:** 0 new failures introduced. 1 pre-existing failure (node-pty native module) on both branches.

**Acceptance criteria:**
- `grep -c 'mockbar\|shotph\|secmark\|tag-new\|tag-fix\|shot-note' website/index.html` → `0`.
- `grep -c 'mantaui.com/manta-logo.png' website/index.html` → `0`.
- `grep -c ':root' website/index.html` → `0` (tokens come from `site.css`).
- Only one tab-switching JS function exists (`mantaTab`); 3 Copy buttons carry the existing onclick clipboard handlers.
- Real links wired: `/install.sh`, `/privacy`, `/terms`, GitHub repo, `Manta-latest.dmg` download. Comparison / use-case / docs / pricing / changelog links keep `href="#"` with `<!-- TODO: ... -->` comments naming the eventual paths.
- Rendered at 1440px and 390px in headless Chrome — no console errors, no 404s on `index.html`, `site.css`, or `manta-logo.png`.

**Screenshots:** headless-Chrome captures against the branch's local serve.

- 1440px (desktop): https://bui-bet297-1440.pages.mantaui.com
- 390px (mobile): https://bui-bet297-390.pages.mantaui.com

(Hosted previews expire in 24h. Raw PNGs also at `/tmp/screenshots/index-1440.png` and `/tmp/screenshots/index-390.png` in the workdir.)

**Base:** `origin/main @ b0ec672` (rebased onto the latest main after BET-301 merged; diff is clean — only `website/index.html`)

[...]

PR #253, opened 2026-07-27
fix(server): expand ~ at the tmux boundary, consolidate path helpers (BET-307)

antoinedc:
## Summary

Projects and sessions created from the UI were silently landing in `$HOME` because tmux's `-c` flag does not expand a leading `~` — for a missing dir it falls back to `$HOME` with exit code 0, so every project created with the UI's default `~` cwd ended up in the home directory, and every subsequent window/session inherited that wrong root.

Base: `origin/main @ 4b81302`

## What changed

- **One `expandTilde`**: deleted the three duplicates in `src/server/tmux.mjs`, `src/server/opencode.mjs`, and `src/shared/pluginManifest.mjs` (plus its `.d.mts` declaration). The single source of truth now lives in `src/shared/paths.mjs`.
- **Collapsed three tmux command builders into two**: deleted the separate exported `newWindowGetIndex` (the 3-arg fork-session variant), renamed the private `newWindowGetIndexInternal` → exported `newWindowGetIndex` with `chatMode = false` default. The `rpc.mjs:363` fork-session call is byte-identical to today's behaviour (3 args → `chatMode` defaults to `false`).
- **`resolveCwdOrThrow` is the new tmux-side chokepoint**: expands `~` via the shared `expandTilde` and rejects a missing dir with a loud error. Applied at exactly three sites (`newSession`, `newWindow`, and the exported `newWindowGetIndex`) BEFORE any opencode session is created, so a bad cwd throws before we orphan an opencode session.
- **Deleted dead defensive fallbacks** inside the builders (`cwd ?? "."` and `...(cwd ? ["-c", cwd] : [])`) — they always receive an absolute path now.
- **`fsListDirs` returns paths in the form the caller typed**: tilde-form when input was `~` or `~/...`, absolute-form otherwise. The bare `~` special case still lists `homedir()/` (not `/home/`). The renderer's existing `m.startsWith(value)` filter now works for `~` paths — both `Sidebar.tsx` and `MobileCreateSheet.tsx` use the same byte-identical filter.
- **Server-side `projectMetaUpsert`**: moved into the `tmux:new-session` handler (best-effort with `.catch(() => {})`, expanded absolute cwd). Deleted the two renderer-side calls in `MobileCreateSheet.tsx` and the four `window.api.projectMetaUpsert` declarations (`api.ts:68`, `types.ts:294`, `httpApi.ts:575`, `rpc.mjs:106-107`). `local.projectMetaUpsert` stays — it's now called server-side only.
- **`AGENTS.md`** "New window / new chat-session cwd inheritance" section: corrected to call out tmux `-c` not expanding `~` + silent `$HOME` fallback, and to point at `src/shared/paths.mjs` as the single `expandTilde` home.

## Out of scope (per spec)

- `resolveProjectCwd` precedence (`rpc.mjs:59`) — untouched.
- The fork-session chatMode=false at `rpc.mjs:360-367` — untouched (chatMode = false by default).
- `listProjects()`'s pane-path derivation — untouched.
- `configMigration.mjs` — no repair logic for existing-broken workspaces (per spec).
- Existing workspaces already created in `$HOME` keep their wrong root and must be recreated by the user.

## Verification

```
# positive — must print /home/dev/projects/better-ui
$ node -e "import('./src/server/tmux.mjs').then(async t => {
    await t.newSession({ name:'verify-tilde', cwd:'~/projects/better-ui', windowName:'default', createDir:true });
    const p = (await t.listProjects()).find(x => x.tmuxSession === 'verify-tilde');
    console.log(p.defaultCwd);
  })" && tmux kill-session -t verify-tilde
/home/dev/projects/better-ui

# negative — must throw instead of silently succeeding
$ node -e "import('./src/server/tmux.mjs').then(t =>
    t.newSession({ name:'verify-missing', cwd:'~/definitely-not-here', windowName:'default' })
  ).catch(e => console.log('OK, threw:', e.message))"
OK, threw: working directory does not exist: /home/dev/definitely-not-here
```

- typecheck: clean
- tests: 845 pass / 0 fail (`npm test`)
- `git grep -n "function expandTilde"` → exactly 1 line (`src/shared/paths.mjs:24`)
- `git grep -n "projectMetaUpsert" src/renderer src/shared` → empty

## Net diff

- Production code (excluding tests + AGENTS.md): +123 / -120 = net −3 (slight contraction).
- Tests: ~330 lines added across `tmux.test.mjs` (resolveCwdOrThrow, newWindowGetIndex arity), `local.test.mjs` (fsListDirs tilde-form), and `rpc.test.mjs` (server-side projectMetaUpsert + best-effort swallow).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -822,15 +822,31 @@ Backup at `~/.tmux.conf.pre-MantaUI` on the remote if it was ever modified.
     metadata forever. Expansion is therefore mandatory at the **single
     creation chokepoint**, NOT in `resolveProjectCwd`: `createSession`
     (`src/server/opencode.mjs`) expands a leading `~` itself via `expandTilde`
-    against the server process's own `$HOME` — the renderer (desktop + mobile)
-    reaches it the same way over `/rpc`. `forkSession` is unaffected — it
-    inherits the parent's directory from opencode and passes no cwd. The
-    new-window chat-holder path (`tmux.mjs:maybeCreateChatSession`) also
-    expands before `createSession`; that earlier expansion is now redundant but
-    harmless. Regression tests:
+    (from `src/shared/paths.mjs`) against the server process's own `$HOME` —
+    the renderer (desktop + mobile) reaches it the same way over `/rpc`.
+    `forkSession` is unaffected — it inherits the parent's directory from
+    opencode and passes no cwd. Regression tests:
     `createSession expands a leading ~ …` in `src/server/opencode.test.mjs`
     (red/green verified). Do NOT "simplify" by moving expansion back into a
     caller — the chokepoint is what makes the corruption unreachable.
+  - **GOTCHA — tmux does NOT expand `~` either; it silently falls back to
+    `$HOME`.** `tmux new-window -c '~/foo'` and `tmux new-session -c '~/foo'`
+    BOTH accept the literal tilde but resolve it against the tmux server's
+    own cwd (typically `$HOME`) and exit code 0 — silently landing every
+    project created with the UI's default `~` cwd in the home directory.
+    This is the BET-307 tmux-side chokepoint: `tmux.mjs`'s
+    `resolveCwdOrThrow` is the single place a caller-supplied cwd becomes a
+    real directory handed to tmux or opencode. It expands `~` via the
+    shared `expandTilde` (now living in `src/shared/paths.mjs` — three
+    copies used to live in `tmux.mjs`, `opencode.mjs` and `pluginManifest.mjs`,
+    all deleted) and rejects a missing directory with a loud error before
+    any tmux call (and before any orphan opencode session is created in
+    chatMode). Applied at exactly three sites: `newSession`, `newWindow`,
+    and the exported `newWindowGetIndex` (which `rpc.mjs:363` calls
+    directly for fork-session). Regression test: `resolveCwdOrThrow` cases
+    in `src/server/tmux.test.mjs` + the `node -e` e2e in BET-307. Do NOT
+    "simplify" by passing the cwd through unchanged — the chokepoint is
+    what makes the silent `$HOME` fallback unreachable.
 - **Queued message drain — abort at the next step boundary, then submit on
   idle.** When the user submits while `running` is true, the text gets pushed
   to `messageQueue` and the input clears. MantaUI does NOT wait for the whole
--- a/src/renderer/api/httpApi.ts
+++ b/src/renderer/api/httpApi.ts
@@ -572,7 +572,6 @@ export const httpApi: Api = {
   configUpdate: (patch) => rpc(IPC.configUpdate, patch),
 
   // -- project metadata --
-  projectMetaUpsert: (meta) => rpc(IPC.projectMetaUpsert, meta),
   projectMetaDelete: (tmuxSession) => rpc(IPC.projectMetaDelete, tmuxSession),
 
   // -- tmux operations --
--- a/src/renderer/mobile/MobileCreateSheet.tsx
+++ b/src/renderer/mobile/MobileCreateSheet.tsx
@@ -176,14 +176,6 @@ export function MobileCreateSheet({ mode, onClose, onCreated }: Props) {
           windowName: "default",
           chatMode: true,
         });
-        // Persist defaultCwd for the new project so future `tmux:new-window`
-        // calls inherit it (resolveProjectCwd in rpc.mjs reads this). Without
-        // this, the server falls back to ~, which defeats the project's
-        // intended workspace path. Mirror of desktop tmuxNewSession path
-        // (src/main/index.ts) which writes ProjectMeta on creation.
-        await window.api
-          .projectMetaUpsert({ tmuxSession: projectName, defaultCwd: path })
-          .catch(() => {});
         await refresh();
         setActive(projectName);
         onCreated(projectName, null);
@@ -229,10 +221,6 @@ export function MobileCreateSheet({ mode, onClose, onCreated }: Props) {
           chatMode: true,
         });
       }
-      // Same project-meta write as the auto/single path above.
-      await window.api
-        .projectMetaUpsert({ tmuxSession: projectName, defaultCwd: path })
-        .catch(() => {});
       await refresh();
       setActive(projectName);
       onCreated(projectName, null);
--- a/src/server/local.mjs
+++ b/src/server/local.mjs
@@ -12,7 +12,7 @@ import { readdir, readFile, writeFile, rename, mkdir } from "node:fs/promises";
 import { existsSync } from "node:fs";
 import { homedir } from "node:os";
 import { join, dirname } from "node:path";
-import { STATE_DIRNAME } from "../shared/paths.mjs";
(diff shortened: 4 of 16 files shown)


antoinedc:
## Review — BET-307 / PR #253

Mechanical pre-flight:
- Base: `origin/main @ 4b81302` (HEAD `7bde9f3` includes #251; only `website/index.html` changed between base and current main, so the diff scope is correct — no stale-base phantom deletions).
- Typecheck (`npm run typecheck`): exit 0.
- Tests (`npm test`): 845 pass / 0 fail / 0 skip.
- E2E snippets on dev box (real tmux):
  - `tmux.newSession({cwd:'~/projects/better-ui',createDir:true})` → `defaultCwd: /home/dev/projects/better-ui | pane: /home/dev/projects/better-ui` ✓
  - `tmux.newSession({cwd:'~/definitely-not-here'})` → throws `working directory does not exist: /home/dev/definitely-not-here` ✓
- DoD greps:
  - `git grep -n "function expandTilde"` → 1 line (`src/shared/paths.mjs:24`) ✓
  - `git grep -n "projectMetaUpsert" src/renderer src/shared` → empty ✓

The consolidation (one `expandTilde` in `paths.mjs`, three duplicates gone), the tmux-builder collapse (3 → 2 with the new `chatMode = false` default keeping fork-session byte-identical), the dead-defensive-fallback removals, the `fsListDirs` tilde/absolute form, the server-side `projectMetaUpsert` best-effort write, and the four `window.api.projectMetaUpsert` deletions are all clean and well-tested. AGENTS.md update correctly calls out the tmux `-c` tilde-corruption gotcha and points at the new `paths.mjs` home.

---

### Block (fix-here): `newWindowGetIndex` does NOT apply `resolveCwdOrThrow` at the top, but Step 2 of the spec AND the updated AGENTS.md both say it does

Spec (Step 2, sub-bullet 3):
> `newWindowGetIndex` (the newly exported one) — resolve at the top, since `src/server/rpc.mjs:363` calls it directly and bypasses the two functions above.

Updated `AGENTS.md` (this PR, around the new GOTCHA bullet):
> Applied at exactly three sites: `newSession`, `newWindow`, and the exported `newWindowGetIndex` (which `rpc.mjs:363` calls directly for fork-session).

Code (`src/server/tmux.mjs:152-166`):
```js
export async function newWindowGetIndex(sessionName, windowName, cwd, chatMode = false) {
  const { stdout } = await run("tmux", [
    "new-window",
    "-t", sessionName,
    "-n", windowName,
    "-P", "-F", "#{window_index}",
    "-c", cwd,                          // <-- no resolveCwdOrThrow above
    ...(chatMode ? ["sh", "-c", CHAT_HOLDER_CMD] : []),
  ]);
  ...
}
```

The call site at `src/server/rpc.mjs:376` (fork-session) goes:
```js
const resolvedCwd = await resolveProjectCwd(sessionName, cwd);
const windowIndex = await tmux.newWindowGetIndex(sessionName, windowName, resolvedCwd);
```

In the happy path `resolvedCwd` is an absolute path (stored `meta.defaultCwd` is now expanded-absolute, or the live tmux pane path, both absolute), so nothing visibly breaks today. But `resolveProjectCwd` can also return:
- the caller's raw input if it's truthy and not `"~"` — i.e. `~/projects/foo` if the renderer ever passes one (currently it doesn't, but the API allows it),
- `"~"` as the last-resort fallback when there is no stored meta AND no live session (`rpc.mjs:82`).

The whole point of the chokepoint — and the reason the spec is explicit about applying it at this exact site — is defense in depth: a bare `tmux.newWindowGetIndex` call from any future caller (rpc handler, future composite, test) is silent on a tilde or missing dir. Step 2 deliberately lists `newWindowGetIndex` as one of the three sites because it's exported and called from outside the `newSession`/`newWindow` envelope.

Fix: one line. Insert `const dir = resolveCwdOrThrow(cwd);` at the top of `newWindowGetIndex`, then change the args to `"-c", dir`. Also update the misleading "Applied at exactly three sites" sentence in `AGENTS.md` to reflect reality, or — once the fix lands — leave it as written.

After the fix:
- Fork-session with a valid stored meta cwd: byte-identical (resolveCwdOrThrow passes through).
- Fork-session with the `"~"` last-resort: now throws loudly instead of silently landing in `$HOME` (this is the desired behaviour, and matches what `tmux:new-session` / `tmux:new-window` already do).

---

### Nit: duplicate test in `src/server/rpc.test.mjs`

`grep -n "opencode:fork-session creates the new tmux window in the resolved defaultCwd" src/server/rpc.test.mjs` returns two lines:

```
152:test("opencode:fork-session creates the new tmux window in the resolved defaultCwd", async () => {
338:test("opencode:fork-session creates the new tmux window in the resolved defaultCwd", async () => {
```

Both run against the same mock and assert largely the same thing (the new one additionally asserts `chatMode` is `undefined` from the 3-arg call). The pre-existing test at `:152` should be deleted, or the new one folded into the existing one. As-is the duplicate costs nothing at runtime but adds noise.

---

### Net-diff claim in PR body is wrong (not blocking)

PR body / implementer comment claim: "Production code net −3." Actual production-code diff (`git diff --stat ... -- ':!*.test.mjs' ':!AGENTS.md'`): +123 / −120 = **net +3**, not −3. The intent ("prefer deleting code over adding it") is met — net production is roughly neutral, not a meaningful addition — but the sign in the prose is wrong. Worth fixing the description so a future reviewer doesn't second-guess the math.

---

Routing: 1 Block (fix-here), 0 Questions. Returning to `manta-dev`.

antoinedc:
## Re-review — BET-307 / PR #253 (commit `3395c30`)

All three findings from my prior pass are addressed:

**Block (fix-here): `resolveCwdOrThrow` at the top of `newWindowGetIndex`** — `src/server/tmux.mjs:155`. The fork-session path (`rpc.mjs:376`) now throws loudly on a missing dir or raw tilde-input rather than silently landing in `$HOME`. Comment block at `tmux.mjs:142-152` updated to call out that the chokepoint lives here too, and the prior `cwd MUST be an absolute path — callers run it through resolveCwdOrThrow first` doc on the function body was removed since the function now enforces it itself.

**Nit: duplicate fork-session test folded into the existing one** at `rpc.test.mjs:152` (the redundant `test(...)` block that lived at `:338` is gone). The `chatMode` undefined assertion from the duplicate now lives inline in the original test, with a BET-307 comment explaining why the 3-arg fork-session call still works. Test count went 845 → 844 as expected.

**Sign-error correction** in the result comment (no code change) — actual prod diff is +3, not −3. Prose now matches.

**Mechanical gates (commit `3395c30`):**
- `npm test` → 844 pass / 0 fail / 0 skip.
- `node -e "t.newWindowGetIndex('demo','w1','~/definitely-not-here')"` → throws `working directory does not exist: /home/dev/definitely-not-here` ✓ (the chokepoint fix, verified end-to-end).
- `node -e "t.newSession({cwd:'~/projects/better-ui',createDir:true})"` → `defaultCwd: /home/dev/projects/better-ui` ✓.
- `git grep -n "function expandTilde"` → 1 line (`src/shared/paths.mjs:24`) ✓.
- `git grep -n "projectMetaUpsert" src/renderer src/shared` → empty ✓.

Clean. Routing to `manta-pm` for merge hand-off.

PR #257, opened 2026-07-27
BET-309: subscription providers — server registry + opencode auth proxy

antoinedc:
Stage 1 of BET-308 (subscription-providers epic). Pure plumbing; no UI.

**Base:** origin/main @ `ef6e93a`

## What

- `src/server/subscriptionProviders.mjs` — registry of 3 providers (anthropic, openai, kimi-for-coding) + 4 pure resolvers: `findSubscriptionProvider`, `resolveAuthMethod`, `describeConnectShape`, `subscriptionStatuses`. No I/O.
- `src/server/opencode.mjs` — 5 thin HTTP proxies (`listProviderAuthMethods`, `startProviderOauth`, `completeProviderOauth`, `setProviderApiKey`, `removeProviderAuth`) returning structured `{ok,error,detail}` failures with discoverModels's vocabulary (`unreachable` / `unauthorized` / `bad_response`). API-key write returns `{ok}` only — never echoes the key, never logs it.
- `src/shared/types.ts`, `src/shared/api.ts`, `src/renderer/api/httpApi.ts`, `src/renderer/api/demoApi.ts`, `src/server/rpc.mjs` — 5-site wiring for a single `IPC.opencodeProviderAuth` channel with a 5-arm discriminated union (`status` / `start` / `code` / `key` / `disconnect`). No preload, no ipcMain.

## Why this shape

Per BET-308's standing rule ("never pass a literal auth-method index"), `resolveAuthMethod` picks by `type` + lowercased label-substring match. The Codex headless method sits at index 1 today but opencode reorders freely — hardcoding `1` would silently break on the next release.

## Tests

`src/server/subscriptionProviders.test.mjs` — 24 node:test cases, pure-only. Includes the BET-308 "single most important test": with a prompts-bearing method sitting BEFORE the match, the resolved index must still point at the right entry in the ORIGINAL `methods` array, not the filtered one.

## Verification

- `npm run typecheck` ✓
- `npm test` ✓ (914 pass / 0 fail)
- `grep -rn "method: *[0-9]" src/` ✓ no literal method indices outside tests

### DIFF
--- a/src/renderer/api/demoApi.ts
+++ b/src/renderer/api/demoApi.ts
@@ -143,6 +143,29 @@ const getClientVersion = (): Promise<{ version: string }> =>
 const getServerVersion = (): Promise<{ version: string; minClient: string }> =>
   Promise.resolve({ version: "0.0.13", minClient: "0.0.0" });
 
+// Subscription provider auth (BET-308 / BET-309). Demo returns the three
+// disconnected rows so the connect-card UI still renders the registry
+// correctly; the `start` / `code` / `key` / `disconnect` actions are
+// unimplemented and resolve to a status payload (the Proxy fallback would
+// also work, but routing non-status requests through here keeps the
+// explicit-method list honest about demo coverage). Matches the real
+// httpApi's status shape so any future consumer that mounts the card in
+// demo mode gets the same row layout.
+const opencodeProviderAuth = (req: unknown): Promise<{ action: "status"; providers: Array<{ id: string; label: string; plan: string; console: string | null; docs: string; connected: boolean }> }> => {
+  // Status is the only meaningful action in demo mode — everything else
+  // (start / code / key / disconnect) would need a fake opencode server
+  // behind it, which is out of scope for the demo fixture.
+  void req;
+  return Promise.resolve({
+    action: "status",
+    providers: [
+      { id: "anthropic", label: "Claude", plan: "Claude Pro / Max", console: null, docs: "https://claude.com/pricing", connected: true },
+      { id: "openai", label: "Codex", plan: "ChatGPT Plus / Pro", console: null, docs: "https://openai.com/chatgpt/pricing", connected: false },
+      { id: "kimi-for-coding", label: "Kimi", plan: "Kimi For Coding", console: "https://www.kimi.com/code/console", docs: "https://www.kimi.com/code/docs/en/third-party-tools/opencode.html", connected: false },
+    ],
+  });
+};
+
 // ===========================================================================
 // `demoApi` — the explicit methods exposed by name. Everything else falls
 // through the Proxy in the `unknown` handler.
@@ -165,6 +188,7 @@ const explicitMethods = {
   launchersList,
   getClientVersion,
   getServerVersion,
+  opencodeProviderAuth,
 } as const;
 
 // ===========================================================================
--- a/src/renderer/api/httpApi.ts
+++ b/src/renderer/api/httpApi.ts
@@ -4,6 +4,8 @@ import {
   type AuthPairResult,
   type DesktopNotifyPayload,
   type OpencodeEvent,
+  type OpencodeProviderAuthRequest,
+  type OpencodeProviderAuthResult,
   type PluginRegistryRow,
   type PtyEvent,
   type ServerUpdateAvailablePayload,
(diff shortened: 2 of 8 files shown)


antoinedc:
## BET-309 review — PASS (1 Nit)

Cycle 1 of 3. Base is fresh: PR branches at `ef6e93a`, `origin/main` at `8c3c544`. Diff is exactly the 8 files in scope — no phantom deletions, no out-of-scope surface area.

### Acceptance criteria

| Criterion | Result |
| --- | --- |
| `npm run typecheck && npm test` green | ✅ local run: typecheck exit 0; vitest 51 files / 1044 tests pass; node:test 40 suites / 914 tests pass (incl. all 24 new subscription-provider tests) |
| `grep -rn "method: *[0-9]" src/` → no literal method index | ✅ zero hits, incl. variants `methodIndex:\s*[0-9]` |
| Registry is the only place a provider id or label appears | ✅ in production code the only literals are in `src/server/subscriptionProviders.mjs` (the registry); the demo fixture in `src/renderer/api/demoApi.ts` is the required demo stub per the spec |

### Files & wiring (5/5 sites)

1. `src/shared/types.ts` — `IPC.opencodeProviderAuth` + `SubscriptionStatus` + `OpencodeProviderAuthRequest`/`Result` discriminated unions, with the openai "headless FIRST" reasoning in the comment header. ✅
2. `src/shared/api.ts` — `opencodeProviderAuth(req)` on the `Api` contract. ✅
3. `src/renderer/api/httpApi.ts` — `rpc<OpencodeProviderAuthResult>(IPC.opencodeProviderAuth, req)`. ✅
4. `src/renderer/api/demoApi.ts` — three-row demo fixture (Claude connected, Codex/Kimi disconnected). ✅
5. `src/server/rpc.mjs` — `opencode:provider-auth` dispatch with all five actions and a comment matching the neighbouring `opencode:*` style. ✅

No `src/preload/` touched. No new `ipcMain.handle` in `src/main/`. ✅ (the spec said "do not" — confirmed.)

### Deliverable 1 — registry + resolvers (`src/server/subscriptionProviders.mjs`)

- Registry is byte-for-byte the spec's `SUBSCRIPTION_PROVIDERS` (anthropic, openai, kimi-for-coding), verbatim. The headless-first ordering for openai carries the in-line rationale about `localhost:1455`. The empty `prefer: []` on kimi-for-coding is annotated as deliberate (not an oversight). ✅
- `findSubscriptionProvider` mirrors `findLauncher`. ✅
- `resolveAuthMethod` walks the rules in the exact order the spec specified. The index-integrity path is correct: `.map((m, i) => ({ m, i })).filter(...)` preserves the ORIGINAL index because `i` comes from the `.map` step, not from the surviving array. Independently re-ran the index-integrity scenario against the resolved function and confirmed `methods[r.index].label === "ChatGPT Pro/Plus (headless)"`. ✅
- `describeConnectShape` matches the spec's four-input truth table. ✅
- `subscriptionStatuses` preserves registry order, marks `connected` from a `Set` membership, tolerates non-array input. ✅

### Deliverable 2 — proxies (`src/server/opencode.mjs`)

- All 5 thin HTTP proxies added: `listProviderAuthMethods`, `startProviderOauth`, `completeProviderOauth`, `setProviderApiKey`, `removeProviderAuth`. ✅
- Failure vocabulary (`unreachable` | `unauthorized` | `bad_response`) matches `discoverModels` in `src/server/providers.mjs`. No new error codes invented. ✅
- `setProviderApiKey` returns `{ok:boolean}` only — no echo of the key, no log line. ✅
- `removeProviderAuth` treats 404 as `{ok:true}` (idempotent disconnect). ✅
- The new `resolveClaudeBin` docstring carries the "CLAUDE-SPECIFIC BY DESIGN — do NOT generalize" note exactly as the spec required. The function itself and the refresh machinery (`doRefresh`, `maybeRecoverCredentials`, `startCredentialRefreshPoller`) are unchanged. ✅

### Tests (`src/server/subscriptionProviders.test.mjs`)

24 node:test cases in 5 suites, all passing. The single most important test — index integrity with a prompts-bearing method sitting at the wrong slot — asserts both `r.index === 2` AND `methods[r.index].label === "ChatGPT Pro/Plus (headless)"`, so a buggy implementation that returned the filtered index would fail loudly with a label mismatch. Other coverage:

- Real opencode method array verbatim from the epic (browser / headless / api).
- Anthropic single-row happy path.
- All-discarded (GitHub Copilot-only) → `null`.
- `null` / `undefined` / `[]` inputs → `null`.
- prefer-falls-through (browser when headless is absent).
- api fallback when no prefer rule matches.
- `describeConnectShape` for all four input combinations.
- `subscriptionStatuses` registry-order preservation, connected marking, defensive non-array input, human-facing fields.

No process spawn, no network, no FS read. ✅

### Nit (informational, non-blocking)

`scripts/check-duplication.sh origin/main` flags two clones inside the new code in `src/server/opencode.mjs` (35 lines / 218 tokens, 1.03% of the changed scope). Both are the same shape:

```
if (!res.ok) {
  const detail = await res.text().catch(() => "");
  await discardBody(res);
  return { ok: false, error: "bad_response", detail: detail.slice(0, 200) };
}
const data = await res.json();
return { ok: true, … }
```

…repeated across `listProviderAuthMethods`, `startProviderOauth`, `completeProviderOauth`, `setProviderApiKey`, and `removeProviderAuth`. Mechanically extractable into a `_ocJsonFetch(url, init, unauthorizedAsDistinct?)` helper.

This is the same pattern `listModels`/`getDefaultModel` already use (in simpler form, since they don't surface structured failures), so the implementer correctly followed the file's existing style — the spec said "Follow the existing style in that file". Not blocking. Worth a 5-minute follow-up if you ever add a sixth proxy.

### Note on `demoApi.ts` row strings

Per the implementer's own note: the three row literals in `demoApi.ts` duplicate the registry by necessity (the renderer can't import server modules). That's the spec-required demo fixture, not production policy.

— manta-reviewer

PR #288, opened 2026-07-27
feat(pair): carry box address in pairing link, private-only (BET-336)

antoinedc:
**Base:** `origin/main @ 14fdc69`

Tailscale boxes have no public hostname, so today the install heredoc
suppresses the `manta://pair?…` deep link and the `/pair` page's QR
encodes a link that would resolve to a host that doesn't exist. This
lets the link carry the box's listener URL — but only when that URL
points inside the user's own network (RFC 1918, loopback, Tailscale
CGNAT 100.64/10, `.ts.net` MagicDNS, `localhost`). A crafted link
pointing at an attacker's server is refused outright.

### Changes

- New `isPrivateServerUrl` predicate in `src/shared/transport.mjs` —
  single source of truth for the accepted ranges; every consumer
  imports it.
- `parsePairPayload` accepts a validated `server=` param.
- `deepLink.ts` drops its private `buildClaimInput` /
  `resolveServerUrl` duplicates in favour of the existing
  `setupLogic` helpers (`buildSetupClaimInput`, `resolveSetupServerUrl`).
- `pair.html` emits `&server=` in the link + QR on a non-public
  origin (`!/\.boxes\.mantaui\.com$/.test(location.hostname)`).
- `pairPage.mjs` `validatePairQrQuery` accepts `server`, gated by the
  same predicate.
- `install-lib.mjs` `buildPairLink` now takes an optional `serverUrl`
  and validates it; `formatPairingOutput` collapses the public/tailnet
  fork to one path (one fewer inner branch).
- `PairStep` pre-fills `serverUrl` from the deep link and opens the
  Advanced section by default so the user sees the address they're
  about to pair against before clicking Connect.
- Tests cover all the new branches + the public-hostname refusal.

### Verification

- `npm run typecheck` — green.
- `npm test` — green: **971 pass, 0 fail, 41 suites** (was 957 pass,
  added 14 new tests for the new branches).
- Verification snippet from the issue:
  ```
  $ node -e 'import("./scripts/install-lib.mjs").then(m=>console.log(m.formatPairingOutput({pairing_code:"123456",box_id:"0123456789abcdef0123456789abcdef",serverUrl:"http://100.64.1.5:8787"})))'
  …
       Pair page:     http://100.64.1.5:8787/pair#box=0123456789abcdef0123456789abcdef&code=123456
       manta://pair?box=0123456789abcdef0123456789abcdef&code=123456&server=http%3A%2F%2F100.64.1.5%3A8787
  ```
  The deep link line carries `&server=http%3A%2F%2F100.64.1.5%3A8787`
  and round-trips through `parsePairPayload` to a payload whose
  `serverUrl` is `http://100.64.1.5:8787`.

### Threat model (per spec)

The link-supplied server URL is gated by `isPrivateServerUrl`. A
link carrying a public address is REFUSED outright — the parser
returns `null` for the whole payload, never silently drops the param
and falls back to the public hostname. That fallback is exactly the
"crafted link points the app at an attacker server" attack the gate
prevents.

### DIFF
--- a/scripts/install-lib.mjs
+++ b/scripts/install-lib.mjs
@@ -21,6 +21,7 @@ import { fileURLToPath } from "node:url";
 import { createRequire } from "node:module";
 import { STATE_DIRNAME } from "../src/shared/paths.mjs";
 import { loadAuth } from "../src/server/auth.mjs";
+import { isPrivateServerUrl } from "../src/shared/transport.mjs";
 
 // `qrcode-terminal` is CommonJS; we keep this module ESM. The local
 // `createRequire` shim is sync (no top-level await) and resolves the lib
@@ -942,32 +943,32 @@ export function formatPairingOutput(
     lines.push("");
     lines.push("  2. Pair it — click this link, or paste it into the app's Connect screen");
     // BET-267: when a non-public serverUrl is provided (tailscale path), the
-    // pair-page URL and the QR must point at THAT base. We also drop the
-    // `manta://pair?…` deep-link line: the resolver on the mobile/desktop side
-    // routes that link through `boxDirectUrl` → `https://<boxId>.boxes.mantaui.com`,
-    // which is unreachable on the tailscale path. The pair-page URL is the
-    // single source of truth for the connect block in tailscale mode.
+    // pair-page URL must point at THAT base. The pair-page URL is what the
+    // user opens in a browser; the deep link / QR below carry the listener
+    // verbatim for one-click.
+    //
+    // BET-336: the deep link now also carries the server URL when one was
+    // supplied (gated by `isPrivateServerUrl` inside `buildPairLink`, so a
+    // crafted install can't smuggle a public address past this function).
+    // ONE wire shape — no more manta:///pair-page fork.
     const useServerUrl = typeof serverUrl === "string" && serverUrl !== "";
+    const pairLinkOpts = useServerUrl ? { serverUrl } : {};
+    const pairLink = buildPairLink(box_id, pairing_code, pairLinkOpts);
     const pairPageUrl = buildPairPageUrl(box_id, pairing_code, {
       baseUrl: useServerUrl ? serverUrl : undefined,
     });
     lines.push(`       Pair page:     ${pairPageUrl}`);
-    if (!useServerUrl) {
-      // Public path keeps the manta:// deep link for one-tap mobile pairing.
-      const pairLink = buildPairLink(box_id, pairing_code);
-      lines.push(`       ${pairLink}`);
-    }
+    lines.push(`       ${pairLink}`);
     lines.push("");
     lines.push("  3. iPhone (optional) — scan the QR below with your camera");
     lines.push(`       App download: ${IOS_APP_URL}  (App Store link coming soon)`);
     lines.push("");
-    // QR encodes the pair-page URL when a serverUrl is in play (the QR has to
-    // open a URL the scanning device can actually reach), the manta:// link
-    // otherwise (preserves the old wire shape for the public install).
-    const qrPayload = useServerUrl ? pairPageUrl : buildPairLink(box_id, pairing_code);
+    // BET-336: the QR encodes the SAME deep link the user can paste — the
+    // one-tap path. The link itself already points at the right listener
+    // (with `&server=` on the tailscale path), so no separate conditional.
     let qr;
     try {
-      qr = qrRender(qrPayload);
+      qr = qrRender(pairLink);
     } catch {
       // qrcode-terminal missing / broken (e.g. a dev's stripped node_modules) —
       // degrade to text-only rather than crash the install. The QR is a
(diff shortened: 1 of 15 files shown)


antoinedc:
## Review — BET-336 (PR #288)

**Outcome: PASS** (0 Blocks, 0 Questions, 1 Nit) — routing to `manta-pm`.

### Mechanical gates

- `npm run typecheck` — green (no output).
- `npm test` — green: 971 vitest pass, 0 fail, 41 suites. Node --test on the affected files (`pairPage.test.mjs`, `install.test.mjs`) — 218 pass / 0 fail.
- Base is fresh: merge-base == `origin/main @ 14fdc69`, same as the PR body.

### Acceptance-criteria checklist (enumerated from the issue body)

| # | Criterion | Result |
|---|---|---|
| 1 | `isPrivateServerUrl` predicate in `src/shared/transport.mjs`, hand-rolled IPv4 parse, all 7 host families accepted, IPv6 + non-http(s) + bare hosts rejected | ✅ Matches the spec exactly. Verified with the issue's ranges plus the four "just outside" boundaries (`172.15/172.32`, `100.63/100.128`). The hand-rolled parse also rejects leading-zero octets to defuse the `010.0.0.1 → 8.0.0.1` octal-bypass — pinned by a test. |
| 1 | Declared in `src/shared/transport.d.mts` matching the existing comment style | ✅ Added at line 49 with the same comment shape as `parseClaimResponse`. |
| 1 | Single source of truth — `grep` shows the ranges only exist in `transport.mjs` (and the test mirrors them). No consumer copies the rule. | ✅ Every consumer imports from `src/shared/transport.mjs`. |
| 2 | `PairPayload.serverUrl?: string`, normalized on entry, `parsePairPayload` returns null on non-private (no silent drop), `buildPairPayload` appends `&server=<encoded>` when present, header comment updated | ✅ Header comment rewritten to reflect the new behaviour; round-trip tests added. |
| 2 | Import direction: `pairPayload → setupLogic` (no cycle) | ✅ Verified — `setupLogic.ts` does not import `pairPayload`. |
| 3 | `deepLink.ts` deletes its private `buildClaimInput` + `resolveServerUrl`, calls `setupLogic` versions, removes the unused `boxDirectUrl` import | ✅ All three deleted; `boxDirectUrl` no longer appears in `deepLink.ts`. |
| 4 | `pair.html` emits `&server=` only on a non-public origin, used on both the deep link and the QR URL | ✅ Implemented as specified; `isPublic` regex matches the issue's literal. |
| 5 | `validatePairQrQuery` accepts optional `server`, gates through `isPrivateServerUrl`, returns `{ok:false,error}` on non-private | ✅ Three branches covered by tests. Live-ran the helper against public / private / refused inputs — output matches the spec. |
| 6 | `formatPairingOutput` collapses the public/tailnet fork — always prints the deep link, routes through `buildPairLink`, QR encodes the same deep link | ✅ Both forks removed. |
| 6 | `buildPairLink` takes optional `serverUrl`, validates with `isPrivateServerUrl`, throws on non-private | ✅ Live-verified — throws on public URL and non-http(s) scheme; passes public path (no serverUrl). |
| 6 | Import from `../src/shared/transport.mjs` | ✅ |
| 7 | `PairStep` pre-fills `serverUrl` from payload, opens Advanced by default when one is present | ✅ `useState(() => prefill?.serverUrl ?? "")` + `useState(() => Boolean(prefill?.serverUrl))`. |
| 7 | `prefillFromPairLink` extended to return optional `serverUrl` | ✅ Delegates to `parsePairPayload`; the new shape is `{boxId, code, serverUrl?}`. |
| Tests | Per-issue test coverage (5 test files) | ✅ All present; round-trip tests added in `pairPayload.test.ts`. |
| BET-267 update | The tailnet `assert.doesNotMatch(out, /manta:\/\/pair/)` is inverted; comment updated | ✅ Now asserts the deep link is present and carries the server URL. |
| BET-267 update | Companion public-path test "still pass unchanged" | ✅ Untouched — only the doc comment was rewritten. |

### Issue verification snippet

The PR body prints `formatPairingOutput({pairing_code:"123456", box_id:"0123456789abcdef0123456789abcdef", serverUrl:"http://100.64.1.5:8787"})`. I re-ran it locally and the connect block contains:

```
manta://pair?box=0123456789abcdef0123456789abcdef&code=123456&server=http%3A%2F%2F100.64.1.5%3A8787
```

Round-trip through `parsePairPayload` is pinned by `pairPayload.test.ts` (private server URL → payload with `serverUrl`).

### Nit

- **`formatPairingOutput` line count is roughly flat, not strictly down.** The spec asked for "fewer lines AND fewer branches"; the branches clearly shrink (the inner `if (!useServerUrl)` and the `useServerUrl ? pairPageUrl : buildPairLink(...)` ternary are both gone), but the file grew by ~2 net lines because the BET-267 comment block was expanded to a BET-267+BET-336 comment that documents the security gate. The implementer was upfront about this in their `loop_history.self_check`. Not a fix-here — the comment is doing real work (documenting why a public address can never sneak past `buildPairLink`), so trimming it would lose security context. Flagging so PM is aware the spec's literal "fewer lines" wording is unmet even though the substantive intent (less conditional logic) is.

### Notes for PM

- Draft state — PM flips to ready.
- All 7 numbered acceptance items verified, all 5 test files updated, mechanical gates green, base is fresh.
- Threat model holds end-to-end: every link-supplied `server=` is gated by `isPrivateServerUrl` before it touches the claim / persist path; non-private addresses refuse outright (parser returns null, `buildPairLink` throws, `validatePairQrQuery` returns `{ok:false}`).

PR #196, opened 2026-07-22
polish(renderer): keep-worktree feedback + resolveProjectCwd empty-cwd warn (BET-248)

antoinedc:
**Files changed:** 1 file. `src/renderer/Sidebar.tsx` (+24 / −1).

Two reviewer Nits from BET-246 PR #195, both isolated to `src/renderer/Sidebar.tsx`, neither load-bearing:

1. **Keep worktree on dirty confirm** — the session window closes but the worktree + branch remain on disk with zero UI feedback, so users can't tell whether the click registered. Add a `showNotice` helper (alert-based, mirrors `showError`) and call it from the `!force` branch of `killWorktreeDirtyAndClose` with `'Kept worktree at <path>'`.

2. **`resolveProjectCwd` silent empty return** — when neither `defaultCwd` nor any `paneCurrentPath` is available, the function returns `''` silently. Downstream callers handle the empty case today, but the silent return will bite a future caller. Add a `console.warn` naming the project + the raw stored `defaultCwd` so the regression surfaces during dev.

**Base:** `origin/main` @ `0b04b9d`.

**Verification results:**
- PR branch (`multica/BET-248-kept-worktree-feedback` @ `e07d5cf`):
  - typecheck: exit `0`. Errors: none. log sha256: `139c60436aeae3ee087031cab3603b97f58acc2874526c66e944de4da209c667`
  - test: exit `0`, `731` pass / `0` fail / `0` skip. Failures: none. log sha256: `08524806854a727f30971530792bd7918c688fa541ecc70aa60592d8fb8858c9`
- Base (`main` @ `0b04b9d`):
  - typecheck: exit `0`. Errors: none. log sha256: `139c60436aeae3ee087031cab3603b97f58acc2874526c66e944de4da209c667`
  - test: exit `0`, `731` pass / `0` fail / `0` skip. Failures: none. log sha256: `d4f450dc94db29913f7b962e473cf37ac5e7f273e7e0f773e3785b4951ed5d7c`
- **Conclusion:** 0 new failures, 0 pre-existing failures (test counts identical between branches; typecheck logs identical byte-for-byte). Both branches green from a clean checkout.

Logs cached at `/tmp/typecheck-bet248.log`, `/tmp/typecheck-master.log`, `/tmp/test-bet248.log`, `/tmp/test-master.log` for reviewer spot-check.

PR #306, opened 2026-07-28
install.sh: install claude CLI + unify PATH for systemd/launchd services (BET-353)

antoinedc:
## Stage 2: installer must install the Claude CLI and expose it to the services

Implements BET-353. Claude is the only provider that needs the external `claude` binary on the box (opencode's anthropic is a passthrough over `~/.claude/.credentials.json`). Until now, the installer only detected `claude`; it never installed it. A fresh box had no binary, no credentials file, and Claude didn't work at all.

This PR fixes three things together because they share a single root cause (services inherit a minimal PATH that excludes `~/.local/bin`):

### 1. Install the claude CLI during install

Mirrors the opencode install's shape exactly:
- Idempotent (skips on `command -v claude`)
- Non-fatal (warn + skip on download/installer failure so a box without Claude still finishes + pairs + is usable with Codex/Kimi)
- No sudo
- Same stdin-dance as the opencode install (install.sh itself is `curl | bash`, so the child's stdin must be `/dev/null` AND the script must come from a file)
- Supports Linux x64/arm64 + macOS arm64 (the same matrix the rest of the installer ships)

### 2. Unify PATH for every service on every supervisor

Extends `launchd_agent_path` (the PATH-helper used by both LaunchAgent plists) to prepend `$HOME/.local/bin`, where the claude installer places its symlink. Without it, `src/server/opencode.mjs`'s `refreshClaudeCredentials` spawns `claude` inside a service that inherits a minimal PATH and ENOENTs — Claude auth silently rots and the user sees an auth failure with no obvious cause.

Adds `Environment=PATH=@@AGENT_PATH@@` to BOTH systemd `--user` units and substitutes via install.sh's existing `render-systemd-unit` (opencode-serve) and `sed` (manta-server) paths. One helper, one definition of the PATH, used by every service on every platform.

`DO NOT` rely on the `source` line the claude installer prints — that fixes an interactive shell only and does nothing for a service. This PR's fix lives at the service definition.

### 3. Post-install summary now leads with the app

The `no providers connected` guidance now leads with `Connect any of Claude, Codex, or Kimi from the MantaUI app` and demotes the manual `run claude once` command to a labelled fallback line. The app is the supported path; the terminal is the bring-your-own fallback.

## Reuse, not duplication

`launchd_agent_path` was already used by both LaunchAgent plists. The BET-353 fix extends it (one helper, one definition of the PATH) and adds the systemd `Environment=PATH=` substitution. No new PATH-builder, no second place that knows where Claude lives.

## Tests (scripts/install.test.mjs)

- `launchd_agent_path` includes `~/.local/bin` (3 new + 3 updated cases — updated for the new ordering)
- Claude install dry-run: install branch when absent, skip branch when present
- Both systemd units carry `@@AGENT_PATH@@` placeholder; rendered output includes `~/.local/bin` in both
- `print_provider_detection_summary` uses app-first + fallback wording

## Verification

```
PR branch (multica/BET-353-claude-cli-install @ dd0e797):
  typecheck: exit 0. Errors: none.
  test: exit 0, 1051 pass / 0 fail / 0 skip.
bash -n scripts/install.sh: clean.
install.sh --dry-run: shows the new step + the skip branch.
```

`Base: origin/main @ bbfdba8`

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -123,7 +123,12 @@ resolve_arch() {
   esac
 }
 
-# launchd_agent_path — the PATH a MantaUI LaunchAgent must run with.
+# launchd_agent_path — the PATH a MantaUI supervisor-managed service must
+# run with. Used by:
+#   * BOTH LaunchAgent plists (`scripts/launchd/com.mantaui.{opencode,server}.plist`)
+#     via @@AGENT_PATH@@ substitution,
+#   * BOTH systemd --user units (`scripts/systemd/{opencode-serve,manta-server}.service`)
+#     via @@AGENT_PATH@@ substitution (systemd's Environment=PATH=).
 #
 # launchd does NOT give an agent the user's login-shell PATH; it hands out a
 # bare `/usr/bin:/bin:/usr/sbin:/sbin`. Every tool the box actually depends on
@@ -133,12 +138,26 @@ resolve_arch() {
 # installs, pairs, and answers HTTP, but `tmux:new-session` 500s with ENOENT
 # and `listProjects` swallows its error and reports an empty box.
 #
-# We emit both Homebrew prefixes (Apple Silicon + Intel) plus the standard
-# system dirs, and prepend the directory `tmux` was actually resolved from
-# when it lives somewhere else entirely (MacPorts, /usr/local/opt, a
-# hand-built binary) — the prereq check already proved that copy exists.
-# Same class of fix as the macOS PATH handling in the desktop plugin
+# systemd --user services have the SAME trap (a minimal default PATH that
+# excludes Homebrew / user-local dirs) — the BET-353 fix unifies the two
+# supervisors onto one PATH so a tool either works under both or fails under
+# both. Same class of fix as the macOS PATH handling in the desktop plugin
 # executor (see AGENTS.md, "macOS PATH gotcha").
+#
+# Composition (precedence left-to-right; first match wins at exec time):
+#   1. $HOME/.local/bin  — where the official `claude` CLI installer places
+#                          its symlink (`~/.local/bin/claude`). Without this,
+#                          the credential-refresh machinery in
+#                          src/server/opencode.mjs (doRefresh → cpSpawn
+#                          "claude") ENOENTs and Claude auth silently rots.
+#                          Same resolver as opencode.mjs's resolveClaudeBin.
+#                          Dedup'd against the existing base (a Homebrew-
+#                          installed claude that also lives in /usr/local/bin
+#                          must not appear twice).
+#   2. the directory `tmux` was actually resolved from when it lives
+#      somewhere else entirely (MacPorts, /usr/local/opt, a hand-built
+#      binary) — the prereq check already proved that copy exists.
+#   3. Homebrew prefixes (Apple Silicon + Intel) + standard system dirs.
 launchd_agent_path() {
   local base="/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
   local tmux_bin tmux_dir
@@ -150,6 +169,10 @@ launchd_agent_path() {
       *) base="$tmux_dir:$base" ;;
     esac
   fi
+  case ":$base:" in
+    *":$HOME/.local/bin:"*) ;;
+    *) base="$HOME/.local/bin:$base" ;;
+  esac
   printf '%s' "$base"
 }
 
@@ -236,7 +259,8 @@ print_provider_detection_summary() {
     # detection (BET-313: a missing provider is not an installation
     # failure).
     warn "could not build provider detection summary (see /tmp/manta-detect.err) — assuming no providers connected."
-    warn "Run \`claude\` on this box to sign in, then:"
+    warn "Connect any provider from the MantaUI app."
+    warn "Fallback (if the app is unavailable): run \`claude\` once on this box to sign in, then:"
     if [ "$macos" = "1" ]; then
       warn "  launchctl kickstart -k gui/\$(id -u)/com.mantaui.opencode"
     else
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review: PASS (0 Blocks, 0 Questions, 0 Nits)

PR cleanly implements every acceptance criterion in BET-353.

### Mechanical gates
- **typecheck:** `npm run typecheck` → exit `0`. Matches implementer's report exactly.
- **tests:** `npm test` → exit `0`, **1051 pass / 0 fail / 0 skip** (1045 → 1051 = +6 new BET-353 tests, matches the implementer's count).
- **`bash -n scripts/install.sh`:** syntax OK.
- **base freshness:** `origin/main` = `bbfdba8`, merge-base = `bbfdba8` — branch is fresh; no phantom deletions.
- **anti-spaghetti:** jscpd found **0 clones** in changed files.

### Acceptance-criteria checklist (load-bearing)

**1. Install Claude CLI during install.** `scripts/install.sh` lines 680–734 add a step that mirrors the opencode install exactly:
- `curl -fsSL https://claude.ai/install.sh -o "$WORK/claude-install.sh"` → `bash "$_claude_installer" </dev/null` (avoids the same stdin/dance footgun documented for the opencode install)
- PATH recovery via `.bashrc` + `~/.local/bin` probe
- Failure path uses `warn` (non-fatal) — a box without Claude still finishes installing and pairs with Codex/Kimi. The opencode install's `die` semantics are deliberately not changed.

**2. Idempotent.** `command -v claude >/dev/null 2>&1` short-circuits both the dry-run and real branches. Verified manually:
- `PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin bash scripts/install.sh --dry-run` → `[dry-run] would install claude CLI (official installer https://claude.ai/install.sh); idempotent on PATH probe` ✓
- `bash scripts/install.sh --dry-run` (with `/home/dev/.local/bin/claude` on PATH) → `[dry-run] claude CLI already installed (/home/dev/.local/bin/claude) — would skip install step` ✓

**3. Non-fatal.** The implementation never `die`s on the claude path; both the download and the install are wrapped in `|| warn ... && continue`. A box that can't reach `claude.ai` still installs and pairs.

**4. No sudo.** Writes only to `$HOME/.local/bin`. No apt, no privileged section.

**5. Multi-platform.** The block is unconditional — no `IS_MACOS` / `uname -m` gate — so it runs on every platform install.sh supports (Linux x64, Linux arm64, macOS arm64).

**6. Binary visible to background services — one helper, one definition of the PATH.** This was the trap the issue called out. The PR extends `launchd_agent_path()` in place to prepend `$HOME/.local/bin` (with dedup), and the same helper now feeds:
- `scripts/launchd/com.mantaui.opencode.plist` (line 832 — pre-existing, unchanged)
- `scripts/launchd/com.mantaui.server.plist` (line 832 — pre-existing, unchanged)
- `scripts/systemd/opencode-serve.service` via `render-systemd-unit --placeholder AGENT_PATH="$(launchd_agent_path)"` (line 929, new)
- `scripts/systemd/manta-server.service` via the existing `sed -e "s|@@AGENT_PATH@@|$(launchd_agent_path)|g"` (line 1056, new)

Verified by rendering the opencode-serve unit from a clean checkout: `Environment=PATH=/tmp/test-home/.local/bin:/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin` — `~/.local/bin` first, no `@@AGENT_PATH@@` leakage.

**7. Service definition, not interactive-shell fix.** Both systemd units now carry `Environment=PATH=@@AGENT_PATH@@` at the unit-file level, not by relying on the installer's `source` line. systemd's `Environment=` directive lands the PATH on `process.env.PATH` for the spawned service, where `cpSpawn("claude")` and any tool the `claude` child re-execs both read it.

**8. Post-install summary.** `print_provider_detection_summary` (lines 261–266 and 315–322) now leads with `Connect any provider from the MantaUI app` / `Connect any of Claude, Codex, or Kimi from the MantaUI app`, with the manual `run claude once` command demoted to a `Fallback (if the app is unavailable):` line. The OLD combined phrasing — `Connect any of Claude, Codex, or Kimi from the MantaUI app, or run ...` — is asserted-gone by test #196.

### Constraints

- **Reuse the existing PATH helper.** ✓ Verified: one `launchd_agent_path` function, three call sites, no second PATH builder.
- **No config field for the Claude path.** ✓ The PATH is derived; no new config knob.
- **Don't change the opencode install step.** ✓ Lines 626–678 are byte-identical to `origin/main`.

### Test additions (6 new, all passing)

| # | Test | Asserts |
|---|---|---|
| 191 | `launchd_agent_path: ~/.local/bin stays prepended even when tmux resolves to the same dir` | Dedup branch doesn't drop `~/.local/bin` when tmux lives in the same dir |
| 192 | `claude install step: dry-run shows the install branch when claude is absent` | Dry-run logs the install branch when `command -v claude` fails |
| 193 | `claude install step: dry-run reports skip when claude is on PATH` | Dry-run logs the skip branch when `command -v claude` succeeds (idempotency) |
| 194 | `scripts/systemd/*.service carries @@AGENT_PATH@@ placeholder` | Both systemd templates still contain the literal `@@AGENT_PATH@@` token pre-substitution |
| 195 | `systemd render: AGENT_PATH substitution materialises ~/.local/bin in both units` | Both renderers produce a `Environment=PATH=…` line containing `$HOME/.local/bin` with no placeholder leakage |
| 196 | `print_provider_detection_summary guidance: app-first wording with manual fallback` | New app-first + fallback wording is present; OLD combined phrasing is gone |

Three pre-existing `launchd_agent_path` tests were extended for the new ordering (asserting `~/.local/bin` first); the `assert.equal(out.match(/\.local\/bin/g).length, 1, "$HOME/.local/bin must not be duplicated")` line is the dedup guard.

### Non-blocking notes

- The server (`src/server/opencode.mjs` `doRefresh`) already has its own `resolveClaudeBin` + PATH augmentation as a safety net (lines 1166–1186, 1282–1289). The service-level PATH addition is therefore defense-in-depth — but it's the right defense, because it also makes the PATH consistent across launchd AND systemd AND prevents any future `cpSpawn` calls in service-land from inheriting a 4-directory PATH. Good direction.
- The official installer (`https://claude.ai/install.sh`) downloads a binary that supports macOS arm64 + Linux x64 + Linux arm64 — same matrix install.sh ships tarballs for. No platform-skip branch needed.

### Decision

Routing to `manta-pm` for merge hand-off. PR is ready.

PR #312, opened 2026-07-29
feat(auth): remove box → unpair it (BET-357 §2)

antoinedc:
## Summary

Adds a 'Remove box' action in Settings → Connection that revokes the device's box_token on the box and clears the local pairing triple, honoring the unreachable-box graceful-fallback from the spec.

- **Server**: new `DELETE /auth/revoke` route, exempt from the standard gate so it can distinguish malformed-token (400) from missing/wrong-token (401). The new `authEngine.revoke({ token })` validates the presented token, deletes auth.json, mints a fresh identity, and mutates the in-memory engine state in place — the old token is dead from the very next authorize/claim call (no server restart needed); the next /auth/pair → /auth/claim returns the fresh identity.
- **Shared classifier**: `classifyUnpairResult(remoteOk, remoteError)` (the literal issue spec) plus `classifyUnpairHttp({ status, networkError })` as a one-step convenience for the main process.
- **Main**: `unpairBox()` does the DELETE and classifies the outcome. The `IPC.authUnpair` handler reads the live { serverUrl, boxToken } from config (never trusting a renderer-supplied copy), calls `unpairBox`, then ALWAYS commits an empty config entry — `config.ts` remains the single token writer.
- **Renderer**: new section in the Connection tab, gated by a native confirm dialog for the irreversible action. Surfaces the UnpairOutcome message in amber when remote revocation didn't reach the box. Preload bridge is desktop-only — mobile/web callers get a friendly inline note.

## Files changed

```
14 files changed, 1006 insertions(+), 16 deletions(-)
```

`src/main/index.ts`                — IPC.authUnpair handler (reads live config, unpairBox, commit empty)
`src/main/unpair.ts` (new)         — wire-call wrapper, classifyUnpairHttp entry
`src/main/unpair.test.ts` (new)    — 10 vitest cases (URL/headers/status/network)
`src/preload/index.ts`             — authUnpair bridge on the preload
`src/renderer/Settings.tsx`        — new 'Remove box' section + state
`src/renderer/preloadAccess.ts`    — authUnpair on MantaPreload interface
`src/renderer/preloadAccess.test.ts` — fake-preload update
`src/server/auth.mjs`              — isExemptPath includes /auth/revoke; deleteAuth helper; revoke() method
`src/server/auth.test.mjs`         — 9 node:test cases (revoke, deleteAuth, post-revoke pair+claim)
`src/server/index.mjs`             — DELETE /auth/revoke handler with manual token validation
`src/shared/types.ts`               — IPC.authUnpair channel name + comment
`src/shared/unpair.mjs` (new)      — classifyUnpairResult + classifyUnpairHttp
`src/shared/unpair.d.mts` (new)    — TS declarations
`src/shared/unpair.test.ts` (new)  — 18 vitest cases (literal signature + http convenience + every failure kind)

## Verification results

- PR branch (`multica/BET-357-unpair-box` @ `2604452`):
  - test: exit 0, 1086 pass / 0 fail / 0 skip (node:test). Vitest: 1335 pass. log sha256: `f0469b44b2036b63d0731f77bad4f88eb681bf3bbb6c8b46d09f343ecbdfb53b`
- Base (`main` @ `23a12bc`):
  - test: exit 0, 1077 pass / 0 fail / 0 skip (node:test). log sha256: `557d7534acd002f9c809272547256380921a271b60c2aaa54d64bfd5a7a45ad9`
- **Conclusion:** 0 pre-existing failures, 0 new failures, 0 regressions. PR adds 9 node:test cases + 28 vitest cases; both suites remain green.

## Base

Base: `origin/main` @ `23a12bc` (the SHA branched from — verified by `git diff --stat origin/main...HEAD`; no phantom deletions).

## Out of scope (intentionally)

- Reconnect logic — stage 3 of BET-357.
- Token re-issue / rotation as a separate operation — `revoke()` mints fresh as part of the same action, which already covers the 'next install mints a new one' semantics.
- Anything that touches the on-disk credentials file `~/.claude/.credentials.json` — stage 5 of BET-357.

## Notes

- The DELETE /auth/revoke route is intentionally exempt from the standard auth gate so we can distinguish malformed-token (400) from missing/wrong-token (401). The gate collapses both into 401, which loses a useful signal for the desktop's classifier.
- The MantaPreload `authUnpair` returns `UnpairOutcome` (not `unknown`) so the Settings UI gets a typed result without a cast. Mobile/web callers see no bridge and get an inline 'only supported in the desktop app' note instead of a crash.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -13,6 +13,7 @@ import { homedir } from "node:os";
 import { fileURLToPath } from "node:url";
 import { loadConfig, saveConfig } from "./config.js";
 import { claimPairing } from "./auth.js";
+import { unpairBox } from "./unpair.js";
 import {
   startDesktopPresence,
   stopDesktopPresence,
@@ -402,6 +403,30 @@ function registerHandlers(): void {
   ipcMain.handle(IPC.authClaim, (_e, input: AuthClaimInput) =>
     claimPairing(input, (patch) => commit(patch)),
   );
+
+  // BET-357 §2: "Remove this box from the device that holds the current
+  // box_token". Reads the live { serverUrl, boxToken } from config (so the
+  // renderer never holds a stale copy), DELETEs <serverUrl>/auth/revoke, and
+  // — REGARDLESS of the remote outcome — clears the local config entry.
+  // The unreachable-box path still succeeds locally per the spec; the
+  // returned UnpairOutcome lets the renderer's Settings panel surface a
+  // structured note when the remote revocation didn't happen.
+  ipcMain.handle(IPC.authUnpair, async () => {
+    const serverUrl = config.serverUrl ?? "";
+    const boxToken = config.boxToken ?? "";
+    const outcome = await unpairBox({ serverUrl, boxToken });
+    // ALWAYS clear locally: config.ts is the single token writer (BET-355 §4).
+    // Empty strings (not delete) so the field shape on disk stays stable
+    // across saves — the schema migration in loadConfig doesn't need to
+    // chase removed keys.
+    commit({
+      serverUrl: "",
+      boxId: "",
+      boxToken: "",
+      projects: [],
+    });
+    return outcome;
+  });
 
   // BET-355 (Stage 4): SSH installer flow. The channels exposed by
   // registerInstallerHandlers are the renderer's ONLY handle on the
--- a/src/main/unpair.test.ts
+++ b/src/main/unpair.test.ts
@@ -1,0 +1,110 @@
+import { describe, it, expect, vi } from "vitest";
+import { unpairBox } from "./unpair.js";
+
+const HEX32 = "0123456789abcdef0123456789abcdef";
+const SERVER = `https://${HEX32}.boxes.mantaui.com`;
+
+function fakeResponse(status: number): Response {
+  return { status } as unknown as Response;
+}
+
+function stubFetch(res: Response | { throw: true }): {
+  fetch: typeof fetch;
+  urls: string[];
+  headers: Record<string, string>[];
+} {
+  const urls: string[] = [];
+  const headers: Record<string, string>[] = [];
+  const fetchImpl = vi.fn(async (url: string, init?: RequestInit) => {
+    urls.push(url);
+    const hdr: Record<string, string> = {};
+    if (init?.headers) {
+      const h = init.headers as Record<string, string>;
+      for (const k of Object.keys(h)) hdr[k.toLowerCase()] = h[k];
+    }
+    headers.push(hdr);
+    if ("throw" in res) throw new Error("ECONNREFUSED");
+    return res;
+  }) as unknown as typeof fetch;
+  return { fetch: fetchImpl, urls, headers };
+}
+
+describe("unpairBox — wire call to DELETE /auth/revoke (BET-357 §2)", () => {
+  it("200 → ok and sends DELETE + Bearer to <serverUrl>/auth/revoke", async () => {
+    const { fetch, urls, headers } = stubFetch(fakeResponse(200));
+    const out = await unpairBox({ serverUrl: SERVER, boxToken: HEX32 }, fetch);
+    expect(out.ok).toBe(true);
+    expect(urls).toEqual([`${SERVER}/auth/revoke`]);
+    // Request is DELETE with the box_token in the Authorization header.
+    expect(headers[0]).toEqual({ authorization: `Bearer ${HEX32}` });
+  });
+
+  it("trims trailing slash off serverUrl before the request", async () => {
+    const { fetch, urls } = stubFetch(fakeResponse(200));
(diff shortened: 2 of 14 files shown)


antoinedc:
## Review: BET-357 §2 — "Remove box → unpair it" (PR #312)

**Verdict: PASS.** 0 Blocks, 0 Questions, 0 Nits.

### Acceptance-criteria completeness

Verified each item in the issue body independently (not from the implementer's summary):

- [x] **"Remove box" action in the existing settings UI** — `src/renderer/Settings.tsx:489-508` (new section inside the Connection tab, between "Pair phone" / "Run setup again" and the close of `activeTab === "connection"`).
- [x] **IPC handler invokes DELETE on the box, then clears the local config entry** — `src/main/index.ts:414-429`. Handler reads the live `{ serverUrl, boxToken }` from `config` (never trusts a renderer copy), calls `unpairBox(...)`, then `commit({ serverUrl: "", boxId: "", boxToken: "", projects: [] })`. `commit()` is the one path through `saveConfig` → `config.ts` is the single token writer; `src/main/unpair.ts` itself does no FS writes (verified by grep — no `writeFile`/`saveConfig`/`fs` import).
- [x] **Unreachable box: still clears locally + surfaces a structured note** — `src/main/unpair.ts:64-70` catches fetch rejection, `classifyUnpairHttp` collapses it to `kind: "network"`. `src/main/index.ts:422-427` clears the local config **regardless** of the outcome. `src/renderer/Settings.tsx:503-507` renders the `outcome.message` in amber when `!outcome.ok`. The network message (`src/shared/unpair.mjs:44`) explicitly reads "Couldn't reach the box. The box's token will be cleared locally only…" — so the UI is honest about what happened.
- [x] **Pure `classifyUnpairResult(remoteOk, remoteError)` helper** — `src/shared/unpair.mjs:85`. Literal signature from the spec; maps every shape in the spec (success, network, bad_request, auth_rejected, server_error, rate_limited, method_not_allowed, unexpected) to a typed `UnpairOutcome`. Plus the convenience `classifyUnpairHttp({ status, networkError })` for the main process.

### Test coverage

- [x] **vitest `classifyUnpairResult`** — `src/shared/unpair.test.ts` covers the literal signature (success + each known kind + unknown-kind fallback + null/undefined safety), the HTTP convenience (200/400/401/405/429/5xx/null + networkError-wins + 404/418/422 unexpected + empty-args defensive), and an unreachable-box fallback contract test that every failure kind carries a non-empty message.
- [x] **node:test for revocation endpoint** — `src/server/auth.test.mjs`: no token → 401, empty token → 401, malformed token → 400 (with leak guard: the engine's token is NOT rotated on the malformed path), wrong 32-hex token → 401 (same leak guard), valid token → 200 with full FS round-trip (`~/.manta/auth.json` rewritten, old token dead on disk, new token authorizes), post-revoke pair+claim returns the NEW identity (not the old one — pins the "next install mints a new one" contract), revoke clears any active pairing code (so a code minted under the old identity can't authorize the dead token), and `deleteAuth` is idempotent + round-trips.

### Mechanical verification (independent re-run)

```
npm run typecheck   → exit 0
npx vitest run      → 1335 passed (62 files), 0 failed
npm test            → node:test 1086 passed / 0 failed, vitest 1335 passed / 0 failed
```

Matches the implementer's reported numbers (1086 / 1335 / 0 / 0). Base branch fresh (`git diff --stat origin/main...HEAD` shows the expected 14 files, no phantom deletions; `merge-base` is `23a12bc`).

### Out-of-scope check

Diff is strictly bounded to BET-357 §2. No mobile/www rebuild needed (the implementer correctly noted the preload bridge is desktop-only and added a null-check in Settings so mobile/web callers see an inline note instead of crashing). No changes to `~/.claude/.credentials.json` (that's stage 5). No reconnect logic (stage 3). No token re-issue as a separate operation — the revoke path mints fresh as part of the action, which already satisfies the spec's "next install mints a new one" semantics.

### Anti-spaghetti

No intentional duplication. The new files (`src/shared/unpair.mjs`, `src/main/unpair.ts`) are purpose-built. The authEngine.revoke() extension lives in the existing `src/server/auth.mjs` rather than forking a parallel revoke module. The IPC handler mirrors the existing `authClaim` shape (live config → wire call → commit) so the two auth IPC handlers read the same way.

### Notes

- One thoughtful detail worth calling out: the `isExemptPath` change adds `/auth/revoke` to the exempt list precisely so the standard gate doesn't collapse 400 (malformed) and 401 (unauthorized) — this preserves the classifier's ability to surface the right user-facing note. The handler then does its own token validation in `src/server/index.mjs:679-689`. This is the right design; the alternative (let the standard gate collapse both into 401) would have erased the 400/401 distinction the issue spec calls for.
- "Run setup again" + "Remove box" share the same visual structure (description + button, `border-t` separator) — consistent.
- "Remove box" is destructive and gated by a native `confirm(...)` dialog — irreversible-action UX honored.

PR #318, opened 2026-07-29
BET-370: build channels prod | staging | dev — one shared module

antoinedc:
## Summary

Implements the three-channel split (prod | staging | dev) for the desktop app. The app now reads every per-channel value (userData dir, app name, URL scheme, install.sh URL, release host) from a single shared module — `src/shared/channel.mjs` — instead of inlining literals at the call sites. The build-time channel is injected via electron-vite's `define` (`__MANTA_CHANNEL__`), and a small `electron-builder.staging.yml` overlay (extends: base config) carries the staging-only deltas (appId, productName, protocols, publish URL, mac.artifactName).

**Files changed:** 10 files. `[525 insertions(+), 31 deletions(-)]`

| File | What |
|---|---|
| `src/shared/channel.mjs` (new) | The single source of truth: `CHANNELS` table, `resolveChannel(isPackaged, bakedChannel)` (the one rule), `channelConfig(channel)` (with the prod fallback for unknown channels). |
| `src/shared/channel.d.mts` (new) | Hand-written types mirroring `.mjs` — same pattern as claim.mjs / compatibility.mjs. |
| `src/shared/channel.test.ts` (new) | 14 tests: pinned table values, frozen invariants, resolveChannel's packaged/unpackaged + bad-baked paths, scheme-distinctness guards. |
| `src/main/buildDefines.d.ts` (new) | `declare const __MANTA_CHANNEL__: string` so typecheck accepts the build-time global. |
| `electron.vite.config.ts` | Bakes `__MANTA_CHANNEL__: JSON.stringify(process.env.MANTA_CHANNEL ?? "prod")` in the **main** build only (renderer doesn't need it yet). |
| `electron-builder.staging.yml` (new) | `extends:` overlay — only the deltas: `appId`, `productName`, `protocols` (manta-staging), `publish.url`, `mac.artifactName`. Inherits signing / notarization / entitlements / targets / node-pty exclusion. |
| `src/main/index.ts` | Six call sites converted: userData dir, setAsDefaultProtocolClient (the isPackaged guard is deleted per rule 6), four `startsWith("manta://")` sites use `PAIR_PREFIX` (one helper, four sites), `app.setName(...)` uses `channelConfig.appName`. |
| `src/main/installer/runner.ts` | `DEFAULT_INSTALL_SH_URL` deleted (was the only export that held a hard-coded `mantaui.com` URL — now sourced from channelConfig). |
| `src/main/installer/installer.ts` | `runInstall` reads install.sh URL + release host from `channelConfig(CHANNEL)`. `buildInstallCommand` now takes `releaseHost` and emits `MANTA_RELEASE_HOST=<host>` unconditionally (one command shape for all channels). Channel resolution is **lazy** via a runtime `require("electron")` so the orchestrator's unit tests (which stub spawn and don't load Electron) don't crash. |
| `src/main/installer/installer.test.ts` | Extended the existing `runInstall` assertion to cover `MANTA_RELEASE_HOST=` + the channel's install.sh URL; added a second test that the installShUrl seam overrides only the URL while releaseHost stays channel-derived. |

## Verification results

- **PR branch** (`multica/BET-370-build-channels` @ `e49f0e6`):
  - typecheck: exit 0. Errors: none. (log: `/tmp/typecheck-pr.log`)
  - vitest: 1410 pass / 0 fail / 0 skip. (64 → 65 test files; +14 channel + 1 installer delta) (log: `/tmp/test-pr.log`)
  - node:test: 1107 pass / 0 fail / 0 skip. (log: `/tmp/test-pr-full.log`)
- **Base** (`main` @ `165b6dc`):
  - typecheck: exit 0. (clean before changes — confirmed)
  - vitest: 1395 pass / 0 fail / 0 skip. (log captured during verification)
  - node:test: 1107 pass / 0 fail / 0 skip.
- **Conclusion:** 0 new failures. 15 new vitest tests (all pass). 0 regressions.

## Cross-cutting follow-ups (filed separately, listed for reviewer context)

- The `pair-link payload format` (`manta://pair?box=…&code=…`) is **deliberately unchanged** in this issue. The renderer / iOS / PWA / box all parse that wire format, and the spec explicitly defers staging support for those. A follow-up will need to thread the channel into the renderer's `parsePairPayload` and the box's pair-link emitter.
- `app.setAppUserModelId("com.antoinedc.mantaui")` (Windows toast attribution) is left as the prod bundle id — the spec table only listed `app.setName` for conversion. A staging build on Windows will share toast attribution with prod. If that matters, a follow-up should add `channelConfig.electronBuilderAppId` as the source here too.

**Base: origin/main @ 165b6dc**

### DIFF
--- a/electron-builder.staging.yml
+++ b/electron-builder.staging.yml
@@ -1,0 +1,42 @@
+# electron-builder overlay for the STAGING channel (BET-370).
+#
+# Loaded by electron-builder via `electron-builder --config electron-builder.staging.yml`.
+# Inherits everything from electron-builder.yml (signing, notarization, mac/win/linux
+# targets, file exclusions, npmRebuild:false, entitlements, ...). Only the deltas:
+#
+#   - appId      — different bundle id so staging can install side-by-side with prod
+#                  on the same Mac / Windows box (one app per appId per OS).
+#   - productName — different display name + menu-bar label + dock label.
+#   - protocols   — different URL scheme (manta-staging) so the OS routes a
+#                  manta-staging:// link to the staging build, not the prod build
+#                  the user might also have installed.
+#   - publish.url — staging app reads its auto-update feed from the /staging path;
+#                  electron-updater's "generic" provider polls this and offers an
+#                  in-app upgrade.
+#   - mac.artifactName — base inherits `${productName}-${version}-${arch}.dmg` which
+#                  would produce a filename containing spaces and parentheses
+#                  ("Manta UI (Staging)-0.0.17-arm64.dmg"). That breaks the
+#                  scp / sha256sum steps in scripts/release/publish.sh. Override
+#                  to a hyphenated, no-space filename.
+#
+# Everything else (entitlements, hardenedRuntime, target arch, notarize:true,
+# win nsis signing posture, linux AppImage executableName, files exclusion of
+# node-pty, extraMetadata.main, ...) is INHERITED — do not redeclare any of it
+# here. A duplicated mac: block, for example, would silently shadow the
+# inherited one and break notarization.
+
+extends: electron-builder.yml
+appId: com.antoinedc.mantaui.staging
+productName: Manta UI (Staging)
+
+protocols:
+  - name: MantaUI Pairing (Staging)
+    schemes:
+      - manta-staging
+
+publish:
+  provider: generic
+  url: https://mantaui.com/staging/updates
+
+mac:
+  artifactName: "Manta-UI-Staging-${version}-${arch}.dmg"
--- a/electron.vite.config.ts
+++ b/electron.vite.config.ts
@@ -16,6 +16,18 @@ const pkg = JSON.parse(readFileSync(resolve(__dirname, "package.json"), "utf8"))
 export default defineConfig({
   main: {
     plugins: [externalizeDepsPlugin()],
+    // Build-time-injected channel id (BET-370). The release pipeline sets
+    // MANTA_CHANNEL=prod|staging for packaged builds; `npm run dev` leaves
+    // it unset so the main process falls back to "prod" baked value AND
+    // resolveChannel() itself overrides with "dev" when app.isPackaged is
+    // false. Defensive reference in main/index.ts:
+    // `typeof __MANTA_CHANNEL__ === "string" ? __MANTA_CHANNEL__ : "prod"`
+    // exists because an unbuilt / unprocessed main bundle has no define
+    // inlined — TypeScript would refuse an unknown identifier. The ambient
+    // declaration lives in src/main/buildDefines.d.ts.
+    define: {
+      __MANTA_CHANNEL__: JSON.stringify(process.env.MANTA_CHANNEL ?? "prod"),
+    },
     build: {
       rollupOptions: {
         input: { index: resolve(__dirname, "src/main/index.ts") },
--- a/src/main/buildDefines.d.ts
+++ b/src/main/buildDefines.d.ts
@@ -1,0 +1,10 @@
+// Build-time globals injected into the main process by electron-vite's
+// `define` (see electron.vite.config.ts). These exist at bundle time and
+// are inlined as string literals — there is no runtime `process.env` read
+// in a packaged build. Keeping them as separate ambient declarations
+// instead of `declare global` (like src/renderer/types.d.ts does for
+// __APP_VERSION__) because the main process already has its own tsconfig
+// (`tsconfig.node.json`) and adding a `declare global` here would force a
+// module wrapper.
+
+declare const __MANTA_CHANNEL__: string;
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -265,7 +265,14 @@ if (!gotSingleInstanceLock) {
 // Web Notification API attributes them to the bundle, i.e. "Electron" in dev).
 // `appUserModelId` does the equivalent for Windows toast attribution.
 app.setName(CHANNEL_CONFIG.appName);
-app.setAppUserModelId("com.antoinedc.mantaui");
+// Windows toast attribution: per-channel AUMID so a prod and a staging install
+// on the same Windows box don't share a single taskbar group / notification
+// source. Dev falls back to the prod id because the spec marks dev's
+// electron-builder appId "n/a" (dev never builds), which leaves no value to
+// source from the table — same effective behaviour as before this PR.
+app.setAppUserModelId(
+  CHANNEL_CONFIG.electronBuilderAppId ?? "com.antoinedc.mantaui",
+);
 
 app.whenReady().then(() => {
   config = loadConfig();
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -33,9 +33,31 @@ import {
 import { checkForUpdates } from "./autoUpdate.js";
 import { titleBarOptions } from "./windowChrome.js";
 import { registerInstallerHandlers } from "./installer/handlers.js";
+import {
+  resolveChannel,
+  channelConfig,
+} from "../shared/channel.mjs";
 import { IPC, type AppConfig, type AuthClaimInput } from "../shared/types.js";
 
 const __dirname = fileURLToPath(new URL(".", import.meta.url));
+
+// Build-channel plumbing (BET-370). The baked value comes from electron-vite's
+// `define` (electron.vite.config.ts) — the ambient declaration lives in
+// src/main/buildDefines.d.ts. `resolveChannel` enforces the rule:
+//   isPackaged === false → "dev"  (npm run dev has no notion of release)
+//   otherwise             → bakedChannel if valid, else "prod" (never throws)
+// All per-channel values (userDataDir, appName, urlScheme, ...) flow through
+// `channelConfig(CHANNEL)` — there is no per-channel branch anywhere in this
+// file.
+const BAKED_CHANNEL =
+  typeof __MANTA_CHANNEL__ === "string" ? __MANTA_CHANNEL__ : "prod";
+const CHANNEL = resolveChannel(app.isPackaged, BAKED_CHANNEL);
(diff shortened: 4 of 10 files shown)


antoinedc:
## Review of BET-370 — PR #318 (build channels prod|staging|dev)

Verified against the issue's acceptance criteria. The PR is well-scoped, the channel module is the single source of truth, every call site in the spec table is converted, and the suite is green.

### Acceptance criteria — every item, every assertion

**1. `src/shared/channel.mjs` + `channel.test.ts` exist and are the only place channel-conditional values live.** ✅

- `CHANNELS` table matches the spec verbatim (line-by-line comparison of all 9 fields × 3 channels). `dev`'s `n/a` cells are explicit `null` (not absent).
- `resolveChannel(isPackaged, bakedChannel)` implements the one rule: `packaged ? baked : dev`, with the `null`/unrecognised-baked fallback to `"prod"` and never throws.
- `channelConfig(channel)` returns the frozen record for known channels, the `prod` record for unknown channels.
- No per-channel branches exist anywhere outside `channel.mjs`: `grep "channel === \"(staging|prod|dev)\"" src/main/` → 0 hits. `grep "isPackaged ?"` in src/main/ → only the lazy-eval one inside `installer.ts:74-84` (Electron-required, intentionally deferred, see below).

**2. `electron-builder.staging.yml` exists with exactly the keys listed in decision 5.** ✅

The file has exactly six top-level keys: `extends`, `appId`, `productName`, `protocols`, `publish`, `mac.artifactName`. No duplicated `mac:` target list, no entitlements block, no signing config — all inherited from `electron-builder.yml`. The `mac.artifactName` override is present (and its rationale in the file header is correct: the inherited default would produce `Manta UI (Staging)-…dmg` with spaces, breaking the scp/sha256 pipeline).

**3. `electron.vite.config.ts` bakes `__MANTA_CHANNEL__` and typecheck passes.** ✅

- `electron.vite.config.ts:28-30` injects `define: { __MANTA_CHANNEL__: JSON.stringify(process.env.MANTA_CHANNEL ?? "prod") }` on the **main** build only (renderer/preload untouched — correct, the renderer has no business with this constant).
- `src/main/buildDefines.d.ts` declares the ambient `const __MANTA_CHANNEL__: string`.
- Both main consumers (`src/main/index.ts:52-53` and `src/main/installer/installer.ts:71-72`) reference it defensively via `typeof __MANTA_CHANNEL__ === "string" ? __MANTA_CHANNEL__ : "prod"`.

**4. All call sites in the table are converted; `DEFAULT_INSTALL_SH_URL` and the four inline `manta://` literals are gone.** ✅ (with one Question — see below)

Verified each of the nine entries in the spec's call-site table against the new code:

| Spec site | New code |
|---|---|
| `src/main/index.ts:181` userData dir | `:204` uses `CHANNEL_CONFIG.userDataDir` ✓ |
| `src/main/index.ts:193-195` protocol register | `:222` `app.setAsDefaultProtocolClient(CHANNEL_CONFIG.urlScheme)` — no `isPackaged` guard ✓ |
| `src/main/index.ts:200` URL prefix | `:227` uses `PAIR_PREFIX` ✓ |
| `src/main/index.ts:230` argv prefix | `:257` uses `PAIR_PREFIX` ✓ |
| `src/main/index.ts:240` app name | `:267` uses `CHANNEL_CONFIG.appName` ✓ |
| `src/main/index.ts:262` argv prefix | `:289` uses `PAIR_PREFIX` ✓ |
| `src/main/installer/runner.ts:50` `DEFAULT_INSTALL_SH_URL` | **deleted** ✓ |
| `src/main/installer/installer.ts:261` `installShUrl` fallback | `:301-302` `channelConfig.installShUrl` ✓ |
| `src/main/installer/installer.ts:77-79` install cmd shape | `:117-119` includes `MANTA_RELEASE_HOST=…` ✓ |

`PAIR_PREFIX` is computed once at line 60 and reused at all four call sites — the spec said "ONE helper", this is one helper.

`grep "manta://"` in `src/main/` returns **only** a comment in `src/main/auth.ts:10` (pair-link payload format, unrelated to per-channel scheme). `grep '"manta"\|"manta-dev"\|"manta-staging"'` in `src/main/` returns nothing. The four inline literals are gone.

**5. `npm run typecheck && npm test` green.** ✅

Independent re-run on the PR branch:

- vitest: **1410 pass / 0 fail / 65 files**
- node:test: **1107 pass / 0 fail / 43 suites**

These match the implementer's reported numbers exactly (no test cherry-picking, no base re-run needed — the PR body cites both PR-hash and base-hash, results agree).

**6. PR opened, titled with this issue key, reassigned to `manta-reviewer`.** ✅

PR #318 is titled `BET-370 — build channels prod|staging|dev — one shared module` (matches the implementer's commit message and the PR title). Issue currently shows `assignee: manta-reviewer`.

### Anti-spaghetti (`scripts/check-duplication.sh origin/main`)

**0 clones** across the 7 changed files. `installer.ts` and `index.ts` both do "bake the channel → resolve → use config" but the duplication detector treats them as different enough (one imports electron lazily, the other eagerly; the helper patterns diverge). No spurious duplication introduced.

### Questions (returning to `better-ui-dev` for confirmation)

**Q1. The `grep -rn 'manta://' src/` rule vs. the deferred pair-link payload format.**

The done-when clause reads "`grep -rn 'manta://' src/` returns nothing outside `channel.mjs` and tests." Strictly applied, this flag hits:

- `src/renderer/mobile/pairPayload.ts:165` — `manta://pair?box=…` builder (the `parsePairPayload` / `buildPairPayload` helpers).
- `src/renderer/mobile/pairPayload.test.ts` — many `manta://pair?…` literals (test cases).
- `src/server/pairPage.mjs:54` — the box-side pair-link emitter that produces the QR payload.
- Several other renderer files that comment on / handle `manta://pair?…`.

Your PR body explicitly defers these on the grounds that the pair-link payload format is a **separate wire format** from the per-channel scheme — and that "a staging channel for the mobile/PWA client or the iOS app" is in the spec's own out-of-scope list. That reading is consistent with the spec's "Scope — files and exact call sites" section (which enumerates only main-process sites).

**My ask:** drop a one-line PR comment confirming the narrow reading is the intended interpretation (the grep rule was a sanity check for the main-process literals, not a wholesale ban on the pair-link payload format anywhere in the codebase). If you agree, this closes; if you intended the strict reading, the deferral becomes a follow-up issue (separate from BET-370) that threads the channel through `parsePairPayload` and the box's pair-link emitter. Either way, no code changes are required for BET-370.

**Q2. `app.setAppUserModelId("com.antoinedc.mantaui")` at `src/main/index.ts:268` is not in the spec call sites table.**

This is Windows toast attribution. A staging build on Windows will attribute OS notifications to the prod bundle id (because the AUMID is hardcoded). You explicitly flagged this as a deliberate deferral in the PR body, and the sym is one line (`app.setAppUserModelId(channelConfig.electronBuilderAppId ?? "com.antoinedc.mantaui")`) plus a fallback cast for dev's `null`.

**My ask:** confirm whether this is acceptable for the staging channel today (in which case it's a documented limitation, not a defect) or whether you'd like it pulled into this PR. If the latter, the fix is mechanical and doesn't expand scope (it's still inside the main process, still a single literal → `channelConfig` lookup). Otherwise, file a follow-up issue.

### Nit

**N1. Variable shadowing in `src/main/installer/installer.ts:301`.**

```typescript
const channelConfig = resolveChannelConfig();
const installShUrl = deps.installShUrl ?? channelConfig.installShUrl;
```

The local `channelConfig` shadows the imported `channelConfig` function for the rest of `runInstall`. Local-only, contained, functionally correct (all subsequent references are property accesses on the object), but a future reader scanning the function will read "channelConfig" and have to disambiguate function vs. record. A rename to `cfg` / `cc` / `channelCfg` would make the call sites read more cleanly. Optional.

### Verdict

**0 Blocks, 2 Questions, 1 Nit.**

The two Questions are both about deferrals you've already explicitly flagged — please drop a one-line confirmation in the PR thread (and a short follow-up note if you want either of them rolled into a separate issue for tracking). Then reassign back to `manta-reviewer` for a re-check; the PR body itself looks ready to ship as soon as the Questions close.

Full review on the PR: https://github.com/antoinedc/MantaUI/pull/318

antoinedc:
## Reviewer re-check: Q1 narrow-grep confirmation + Q2/N1 addressed in commit 17b5467

**Q1 — confirm narrow reading of the grep rule.** Yes, the narrow reading is the intended interpretation. The done-when clause `grep -rn 'manta://' src/ returns nothing outside channel.mjs and tests` was a sanity check for the **main-process channel-scheme literals** (the four `startsWith("manta://")` sites in `src/main/index.ts`), not a wholesale ban on the `manta://pair?box=…&code=…` payload format anywhere in the codebase. The renderer / iOS / PWA / box all parse that wire format via `parsePairPayload` and `buildPairPayload`, and the spec's own out-of-scope list explicitly excludes "A staging channel for the mobile/PWA client or the iOS app" — so threading the channel through `parsePairPayload` and the box's pair-link emitter is a separate issue. Closing Q1.

**Q2 — AUMID fix landed in 17b5467.** `src/main/index.ts:268` now reads `CHANNEL_CONFIG.electronBuilderAppId ?? "com.antoinedc.mantaui"`. Staging gets its own AUMID (`com.antoinedc.mantaui.staging`); prod stays on `com.antoinedc.mantaui`; dev falls back to the prod id because the spec marks dev's `electronBuilderAppId` as `n/a` — same effective behaviour as before this PR. The fallback is documented inline.

**N1 — variable shadowing addressed in 17b5467.** Local renamed from `channelConfig` to `cfg` in `src/main/installer/installer.ts:301` so the call sites no longer shadow the imported function.

Verification: typecheck exit 0, vitest 1410/0, node:test 1107/0 (same as the prior commit — no regressions; the new AUMID behaviour is pinned by `src/shared/channel.test.ts`'s spec-row assertions, so no new tests were required).

Reassigning back to `manta-reviewer` for re-check.

PR #319, opened 2026-07-29
BET-373: thread build channel through pair-link wire format

antoinedc:
## Summary

BET-370 made the desktop main process channel-aware (`manta` / `manta-staging` / `manta-dev` schemes, per-channel userData, app name, install.sh URL, release host). The renderer + box pair-link path was still hardcoded to `manta://` — a staging-desktop user scanning a staging-box QR was routed to whichever channel last registered `manta://`, never to their own staging install. This PR closes that gap on the pair-link wire format (the deliberate deferral recorded on BET-370).

## Changes

**Renderer (`src/renderer/mobile/pairPayload.ts`):** `parsePairPayload` / `buildPairPayload` accept a `scheme` arg (default `manta` — unchanged wire shape for callers that don't care). The parser only accepts the configured scheme (defensive against wrong-channel routing). The build side keys off `channelConfig(channel).urlScheme` (`src/shared/channel.mjs`, single source).

**Renderer wiring:**
- `electron.vite.config.ts` + `electron.vite.config.mobile.ts`: bake `__MANTA_CHANNEL__` into the renderer bundle (mobile always `prod` for now — mobile has no channel concept yet, separate effort per BET-373's out-of-scope list).
- `src/renderer/types.d.ts`: ambient `__MANTA_CHANNEL__` declaration.
- `PairingQR.tsx` + `App.tsx`: read the channel's scheme via `channelConfig(__MANTA_CHANNEL__).urlScheme`.

**Box (`src/server/pairPage.mjs`):** `validatePairQrQuery` defaults to the box's `MANTA_CHANNEL` env var → `channelConfig(env).urlScheme`, so a staging box emits `manta-staging://pair?...`. New `resolveBoxChannel()` export makes the env-driven lookup testable in isolation.

## Tests

- `src/renderer/mobile/pairPayload.test.ts`: per-channel parse acceptance, wrong-scheme rejection, per-channel build emission, and round-trip invariant (`build(scheme, …) → parse(build(…))` succeeds for every scheme).
- `src/server/pairPage.test.mjs`: per-channel explicit-scheme emission, env-driven default per channel (prod/staging/dev), back-compat default for an unset env (still emits `manta://`), defensive fallback for an unrecognised `MANTA_CHANNEL` value, and explicit-scheme-beats-env precedence.

`npm run typecheck && npm test` green — 1435 vitest tests + 1117 node:test tests pass.

## Out of scope

- iOS / Android native side of the pair flow (mobile has no channel concept yet; separate effort).
- `src/shared/channel.mjs` (table unchanged).
- Any change to the `?box=…&code=…` query shape.

## Verification

`npm run typecheck`: passes.
`npm test`: 1435 vitest tests + 1117 node:test tests, 0 failures.

### DIFF
--- a/electron.vite.config.mobile.ts
+++ b/electron.vite.config.mobile.ts
@@ -16,6 +16,12 @@ export default defineConfig({
     __MANTA_AXIOM_TOKEN__: JSON.stringify(process.env.MANTA_AXIOM_TOKEN ?? ""),
     __MANTA_AXIOM_DATASET__: JSON.stringify(process.env.MANTA_AXIOM_DATASET ?? "manta"),
     __APP_VERSION__: JSON.stringify(pkg.version ?? "0.0.0"),
+    // BET-370 + BET-373: the mobile bundle currently always bakes "prod"
+    // (mobile has no channel concept yet — out of scope per BET-373). Kept
+    // here so the ambient `const __MANTA_CHANNEL__: string` in
+    // src/renderer/types.d.ts is satisfied in BOTH renderer targets; a
+    // future mobile-channel effort can flip this to read the real env var.
+    __MANTA_CHANNEL__: JSON.stringify(process.env.MANTA_CHANNEL ?? "prod"),
   },
   resolve: { alias: { "@": resolve(__dirname, "src/renderer") } },
   build: {
--- a/electron.vite.config.ts
+++ b/electron.vite.config.ts
@@ -49,6 +49,16 @@ export default defineConfig({
       __MANTA_AXIOM_TOKEN__: JSON.stringify(process.env.MANTA_AXIOM_TOKEN ?? ""),
       __MANTA_AXIOM_DATASET__: JSON.stringify(process.env.MANTA_AXIOM_DATASET ?? "manta"),
       __APP_VERSION__: JSON.stringify(pkg.version ?? "0.0.0"),
+      // BET-370 + BET-373: bake the channel id into the renderer bundle so
+      // pairPayload.ts (and any future renderer-side consumer) can pull the
+      // channel's URL scheme through `channelConfig(__MANTA_CHANNEL__)`.
+      // Mirrors the main-process baking — same source (process.env), same
+      // fallback ("prod"). The mobile renderer bundle bakes the SAME global
+      // (electron.vite.config.mobile.ts) so the ambient type is satisfied in
+      // both targets; mobile currently always uses the prod scheme because
+      // the mobile app does not yet have a channel concept (a separate
+      // effort per BET-373's out-of-scope list).
+      __MANTA_CHANNEL__: JSON.stringify(process.env.MANTA_CHANNEL ?? "prod"),
     },
     build: {
       rollupOptions: {
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -20,7 +20,17 @@ import { MOD_KEY } from "./platform";
 import { UpdateBar } from "./UpdateBar";
 import { ReconnectingBanner } from "./ReconnectingBanner";
 import { parsePairPayload } from "./mobile/pairPayload";
+import { channelConfig } from "../shared/channel.mjs";
 import type { AvailableLauncher } from "../shared/types";
+
+// BET-373 (channel-aware wire format): the deep-link URL the OS hands this
+// app is, by construction, addressed to THIS channel's URL scheme
+// (the same scheme the main process registered via
+// `setAsDefaultProtocolClient(CHANNEL_CONFIG.urlScheme)`). Parsing with
+// that scheme enforces the boundary defensively — a `manta-staging://…`
+// link can never be misclaimed by a `manta://`-registered app, and a
+// `manta://` link can never silently pass through a staging build.
+const PAIR_PARSE_SCHEME = channelConfig(__MANTA_CHANNEL__).urlScheme;
 
 export function App() {
   const {
@@ -251,7 +261,7 @@ export function App() {
     const pre = getMantaPreload();
     if (!pre?.onPairLink) return;
     return pre.onPairLink((url) => {
-      const payload = parsePairPayload(url);
+      const payload = parsePairPayload(url, PAIR_PARSE_SCHEME);
       if (!payload) {
         // Malformed / legacy-shape link (e.g. an old box still minting the
         // `id=`/`token=` form). The OS still LAUNCHED us for the manta://
--- a/src/renderer/PairStep.tsx
+++ b/src/renderer/PairStep.tsx
@@ -35,10 +35,29 @@ import { claimBox } from "./pairClaim";
 import { useStore } from "./store";
 import { SshInstallStep } from "./SshInstallStep";
 import { getMantaPreload } from "./preloadAccess";
+import { channelConfig } from "../shared/channel.mjs";
 
 const ACCENT = "#5A88FF"; // matches Onboarding.tsx + the app's accent token
 const DANGER = "#FF7A88"; // inline error text (no dedicated tailwind token)
 const SERVER_URL_ERROR = "Server URL must start with http:// or https://";
+
+// BET-373 (review cycle 1): the pending pair link stored by App.tsx was
+// already accepted against THIS channel's scheme (App.tsx's
+// PAIR_PARSE_SCHEME) — parsing it here with a different (default) scheme
+// would silently drop the payload on staging/dev. Same constant App.tsx
+// and PairingQR.tsx already build from `src/shared/channel.mjs`.
+//
+// BET-373 (review cycle 3): `__MANTA_CHANNEL__` is only DEFINEd by
+// electron-vite's build config (electron.vite.config.ts); it is not bound
+// under vitest/node, so a direct reference throws a ReferenceError at
+// import time. PairStep.test.tsx (BET-382) renders this component
+// directly, so — unlike App.tsx/PairingQR.tsx, which no test imports
+// today — this module needed the same `typeof` guard src/main/index.ts
+// and src/main/installer/installer.ts already use for exactly this
+// reason. Falls back to "prod", matching those call sites.
+const PAIR_PREFILL_SCHEME = channelConfig(
+  typeof __MANTA_CHANNEL__ === "string" ? __MANTA_CHANNEL__ : "prod",
+).urlScheme;
 
 export function PairStep({ onPaired }: { onPaired: () => void }) {
   const hasSshInstaller = getMantaPreload() !== null;
@@ -48,7 +67,7 @@ export function PairStep({ onPaired }: { onPaired: () => void }) {
   // so the user sees the address they're about to pair against. When no
   // link is pending, show the SSH picker as the primary surface.
   const pendingPairLink = useStore.getState().pendingPairLink;
-  const prefill = prefillFromPairLink(pendingPairLink);
+  const prefill = prefillFromPairLink(pendingPairLink, PAIR_PREFILL_SCHEME);
   const [disclosureOpen, setDisclosureOpen] = useState(() => Boolean(prefill));
   // On a renderer without an SSH installer (mobile / web), the picker is
   // unavailable — surface the manual form immediately rather than render an
--- a/src/renderer/PairingQR.tsx
+++ b/src/renderer/PairingQR.tsx
@@ -1,6 +1,6 @@
 // PairingQR — renders a QR code image for mobile device pairing.
 //
-// The QR encodes the CANONICAL box-form `manta://pair?box=<boxId>&code=<6-digit>`
+// The QR encodes the CANONICAL box-form `<scheme>://pair?box=<boxId>&code=<6-digit>`
 // payload produced by the SHARED `buildPairPayload` helper
 // (src/renderer/mobile/pairPayload.ts). BET-237 removed the deprecated
 // serverUrl / id forms — `parsePairPayload` rejects anything other than
@@ -14,9 +14,25 @@
 //
 // This is a desktop-only feature (BET-80). The mobile app consumes the same
 // URL scheme but generates the QR on the desktop side.
+//
+// BET-373 (channel-aware wire format): the QR uses THIS channel's URL scheme
+// (`channelConfig(__MANTA_CHANNEL__).urlScheme`, the same source the main
+// process uses to register `setAsDefaultProtocolClient(...)` and to build
+// `PAIR_PREFIX`). A staging desktop scans a QR with `manta-staging://…` so
+// the OS routes the open back to staging, not to whichever channel got
+// registered last for `manta://`. `__MANTA_CHANNEL__` is baked into the
+// renderer at build time (electron.vite.config.ts renderer `define`).
 
 import { useEffect, useMemo, useState } from "react";
 import { buildPairPayload } from "./mobile/pairPayload";
+import { channelConfig } from "../shared/channel.mjs";
+
+// Channel-aware scheme for the QR prefix. The baked `__MANTA_CHANNEL__`
+// comes from the renderer `define` (electron.vite.config.ts); channelConfig
+// does the same unknown-id → prod fallback the main process relies on, so a
+// stale/garbage baked value still produces a usable scheme rather than
+// throwing at render time.
+const PAIR_SCHEME = channelConfig(__MANTA_CHANNEL__).urlScheme;
 
 // Lazy-load qrcode so the renderer doesn't pay the bundle cost if this
 // component is never rendered (Settings is a modal, only open on demand).
@@ -38,7 +54,9 @@ export function PairingQR({
   const url = useMemo(() => {
     // Canonical box-form payload — shared with the install heredoc, `manta pair`
     // output, and the mobile deep-link parser. Single source: pairPayload.ts.
-    return buildPairPayload({ boxId, code: pairingCode });
+    // BET-373: `PAIR_SCHEME` keys the QR to this channel's OS-registered URL
+    // scheme so the OS routes the open back to the channel that scanned it.
+    return buildPairPayload({ boxId, code: pairingCode }, PAIR_SCHEME);
   }, [boxId, pairingCode]);
 
   useEffect(() => {
--- a/src/renderer/mobile/pairPayload.test.ts
+++ b/src/renderer/mobile/pairPayload.test.ts
@@ -4,8 +4,16 @@ import {
   buildPairPayload,
   type PairPayload,
 } from "./pairPayload";
+import { CHANNELS, CHANNEL_IDS } from "../../shared/channel.mjs";
 
 const BOX = "0123456789abcdef0123456789abcdef"; // 32 hex
+
+// Every channel in the table — drives the per-channel test loops below so
+// adding a fourth channel automatically extends coverage. The pairing wire
+// format only depends on urlScheme, but looping the table makes a future
+// PR that drops `manta` or adds a fourth scheme (e.g. "qa") impossible
+// without updating the test.
+const SCHEMES = CHANNEL_IDS.map((id) => CHANNELS[id].urlScheme);
 
 describe("parsePairPayload", () => {
   describe("boxId form (direct claim against the box's own hostname)", () => {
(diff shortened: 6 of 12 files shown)


antoinedc:
## Review — BET-373 (cycle 1)

**Verdict: 2 Blocks (1 fix-here, 1 followup-issue), 0 Questions, 2 nits.**

### Mechanical gate — green (run by me on `d68df85`)

```
$ npm run typecheck
tsc --noEmit -p tsconfig.node.json && tsc --noEmit -p tsconfig.web.json
→ exit 0, 0 errors

$ npm test
vitest:      Test Files 66 passed (66)   Tests 1435 passed (1435)
node:test:   # tests 1117  # pass 1117  # fail 0
→ exit 0
```

Matches the numbers reported in the PR body exactly (1435 + 1117). CI `typecheck-test` also green.

### Acceptance-criteria checklist (from the issue's "Done when")

| # | Criterion | Result |
|---|---|---|
| 1 | `pairPayload.ts` + `pairPage.mjs` derive the scheme from the channel table, no inline `manta://pair?…` literal at either site | ✅ verified — `grep -rn "manta://pair"` finds no live literal in either file; `pairPage.mjs` goes through `resolveBoxChannel()` → `channelConfig`, `pairPayload.ts` takes the `scheme` arg and callers pass `channelConfig(...).urlScheme` |
| 2 | Tests pin the round-trip for all three schemes + the box emitter | ✅ verified — `pairPayload.test.ts` derives `SCHEMES` from `CHANNEL_IDS`/`CHANNELS` (not a hardcoded list) and asserts `SCHEMES.length === 3`; 50 cases pass. `pairPage.test.mjs` pins per-channel env → scheme, unset-env back-compat, unrecognised-env fallback, explicit-beats-env; 29 cases pass |
| 3 | `npm run typecheck && npm test` green | ✅ run by me, above |
| 4 | PR opened, titled with the issue key | ✅ `BET-373: thread build channel through pair-link wire format` |

Base-freshness gate: `git diff --stat origin/main...HEAD` = 9 files, 451(+)/26(−), **zero deletions**. The branch base is behind `origin/main` but the diff vs fresh main is coherent — no phantom-deletion artefacts, no rebase needed for review purposes.

The literal "Done when" list passes. The issue's stated *purpose*, however, does not — see Block 1.

---

### **Block (fix-here): the staging/dev deep-link pair flow still doesn't work — the prefill drops the payload one step downstream**

`src/renderer/mobile/setupLogic.ts:152` (`prefillFromPairLink`) still calls `parsePairPayload(trimmed)` with **no scheme**, so it falls back to the `"manta"` default. It wasn't touched by this PR (not in the 9-file diff). It is a direct consumer of the file this issue renamed the contract on.

The resulting flow on a staging desktop (`__MANTA_CHANNEL__ = "staging"`):

1. Box emits `manta-staging://pair?box=…&code=…` ✅ (this PR)
2. OS routes it to the staging build ✅ (BET-370)
3. `App.tsx:264` — `parsePairPayload(url, "manta-staging")` → **accepts**, stores the **raw URL** via `setPendingPairLink(url)` ✅
4. `PairStep.tsx:58` — `prefillFromPairLink(pendingPairLink)` → `parsePairPayload(raw, "manta")` → protocol `manta-staging:` !== `manta:` → hits the `else { return null; }` branch → **returns null**
5. `prefill` is null → `disclosureOpen` false, `showSshPicker` true, Box ID + code fields **empty**

So a staging user clicking their pair link now lands on the SSH picker with an empty form. Proven executably against the PR branch:

```
$ npx tsx /tmp/prove.ts
manta          App.tsx parse(scheme=manta): OK   | PairStep prefillFromPairLink(): OK
manta-staging  App.tsx parse(scheme=manta-staging): OK   | PairStep prefillFromPairLink(): NULL  <-- prefill DROPPED
manta-dev      App.tsx parse(scheme=manta-dev): OK   | PairStep prefillFromPairLink(): NULL  <-- prefill DROPPED
```

This is the issue's own Risk (b) — *"Pairing from a staging box to a staging app simply does not work today"* — still unmet. The PR fixes the OS-routing half and breaks the very next step in the same flow, so the net user-visible outcome on staging/dev is unchanged (arguably worse: pre-PR the schemes at least agreed with each other).

**Fix:** thread the same scheme through. Add an optional `scheme` arg to `prefillFromPairLink` and pass the channel scheme from `PairStep.tsx` (same `channelConfig(__MANTA_CHANNEL__).urlScheme` constant `App.tsx` and `PairingQR.tsx` already build). Add a `setupLogic.test.ts` case asserting the prefill survives all three schemes — that's the test that would have caught this.

Note `src/renderer/mobile/deepLink.ts:134` also uses the default, but that's the **mobile** path which correctly bakes `prod` and is explicitly out of scope. Not a finding.

---

### **Block (followup-issue): `scripts/install-lib.mjs` still hardcodes `manta://pair?…` — the install.sh pair link stays prod-only**

`scripts/install-lib.mjs:1061` and `:1063` (`buildPairLink`) emit a hardcoded `manta://pair?…`. It's consumed by `formatPairingOutput` (`:956`), i.e. **the pair link install.sh prints to the user's stdout** — the primary onboarding entry point. A staging box installed via install.sh prints a `manta://` link that the OS routes to the prod build, which is exactly the bug this issue set out to close, on a third emitter.

I'm classifying this as **followup-issue, not fix-here**, on purpose: the issue's scope named two files and this isn't one of them, and `install-lib.mjs` has **no channel concept at all** (`grep -n "MANTA_CHANNEL\|channelConfig" scripts/install-lib.mjs` → zero hits). Making it channel-aware needs a design decision the issue didn't scope — how does the installer learn which channel it's installing? (env var, CLI flag, or baked at tarball-build time, given BET-370 already made the install.sh *URL* per-channel.) That's a separate pass, not something to bolt onto this PR.

Also worth noting for the followup: `scripts/install.test.mjs:550` already round-trips `buildPairLink` against `parsePairPayload`, so that contract test currently only exercises prod and will keep passing while the gap exists.

**Paste-able issue body:**

> **Thread the build channel through the install.sh pair-link emitter (`scripts/install-lib.mjs`).**
>
> BET-373 made the renderer (`pairPayload.ts`) and the box server (`pairPage.mjs`) derive the pair-link URL scheme from the channel table. A third emitter was left hardcoded: `buildPairLink` in `scripts/install-lib.mjs` (lines ~1061/1063) still returns `manta://pair?box=…&code=…`, and it feeds `formatPairingOutput`, which is the pair link `install.sh` prints to stdout after a fresh install. Consequence: a box installed from the staging/dev install.sh prints a `manta://` link, so the OS routes the user's click to the prod desktop build, never to the channel they installed — the same wrong-behaviour BET-373 closed on the QR path.
>
> `install-lib.mjs` currently has no channel awareness at all (no `MANTA_CHANNEL` read, no `channelConfig` import), so this needs a small design decision first: how does the installer learn its channel? Options are an env var baked into the release tarball (mirroring the box server's `MANTA_CHANNEL`, which `resolveBoxChannel()` in `pairPage.mjs` already reads), a CLI flag, or a build-time substitution — BET-370 already made the install.sh *URL* per-channel, so the channel is known at tarball-build time. `install-lib.mjs` is plain Node but already imports from `src/shared/` (`isPrivateServerUrl` from `transport.mjs`), and `src/shared/channel.mjs` is also plain `.mjs`, so reusing `channelConfig` is straightforward once the channel source is decided.
>
> Done when: `buildPairLink` derives its scheme from the channel table with no inline `manta://pair?…` literal; `scripts/install.test.mjs` round-trips the install-lib emitter against `parsePairPayload` for all three schemes (the existing round-trip test at ~:550 only covers prod today); `npm run typecheck && npm test` green.

---

### Nits (non-blocking)

**Nit 1 — anti-spaghetti / `duplication-gate`: discounted, no action.** The gate failed but both reported clones are the discountable kind:
- The `define:` blocks in `electron.vite.config.ts` ↔ `electron.vite.config.mobile.ts` — this is the intentional desktop/mobile mirror AGENTS.md tells us to keep in lockstep, and the clone is mostly the new explanatory comment.
- Two fixture arrays inside `pairPayload.test.ts` (`{ boxId: BOX, code: … }` cases) — coincidental test-fixture duplication.

Neither is business logic this PR introduced, and there's no dead code left behind. Not a Block. (Confirmed independently, same detector.)

**Nit 2 — self-contradicting comment in `resolveBoxChannel`.** `src/server/pairPage.mjs:40` says *"we don't memoize it (env read once at first call)"* — those two clauses contradict each other, and the function does in fact re-read `process.env.MANTA_CHANNEL` on **every** call (which is correct, and is what `pairPage.test.mjs` relies on when it mutates the env between cases). Drop the parenthetical.

---

### What was good

Worth saying explicitly, since it's the part that's easy to get wrong: the tests derive `SCHEMES` from `CHANNEL_IDS`/`CHANNELS` rather than hardcoding three strings, and assert `SCHEMES.length === 3` as a tripwire — so adding a fourth channel to the table fails the test instead of silently under-covering. The `resolveBoxChannel()` extraction is the right seam for making the env lookup testable without a server. Defaulting the `scheme` args to `"manta"` keeps the wire shape unchanged for existing callers, which is what let this land without touching mobile.

Both Blocks are about **emitters/consumers the PR didn't reach**, not about the code it did write.

antoinedc:
Addressed BET-373 review cycle 1 findings:

**Block (fix-here) fixed:** `prefillFromPairLink` (`setupLogic.ts`) now accepts a `scheme` arg, and `PairStep.tsx` passes `channelConfig(__MANTA_CHANNEL__).urlScheme` — the same constant `App.tsx`/`PairingQR.tsx` already build. Staging/dev pair links now survive the prefill step instead of being silently dropped (proven with a regression test that pins the pre-fix default-scheme call returning null, plus a per-channel round-trip loop). Also dropped the self-contradicting memoization comment in `resolveBoxChannel` (nit 2).

**Block (followup-issue) filed separately:** Addressed via BET-386 (separate issue) — install-lib.mjs's `buildPairLink` still hardcodes `manta://`. Resuming this PR.

`npm run typecheck && npm test`: 1440 vitest + 1117 node:test, 0 failures.

antoinedc:
## Review — BET-373 (cycle 2)

**Verdict: 1 Block (fix-here), 0 Questions, 0 nits.**

Both cycle-1 Blocks are genuinely fixed and I verified each independently. The only thing standing between this PR and a PASS is that **it no longer merges** — `main` moved under it and there's a real content conflict in the one file the cycle-2 fix touched.

### Cycle-1 Blocks — both cleared ✅

**Block 1 (fix-here) — staging/dev prefill: FIXED, verified end-to-end.** `PairStep.tsx:66` now passes `PAIR_PREFILL_SCHEME` (`channelConfig(__MANTA_CHANNEL__).urlScheme`) into `prefillFromPairLink`, and `setupLogic.ts` threads it to the parser. I re-ran my cycle-1 repro plus a real box→desktop chain on `628f2e9`:

```
BOX channel -> QR payload -> DESKTOP(same channel) prefill
prod     manta://pair?box=<box>&code=123456             PAIRS OK
staging  manta-staging://pair?box=<box>&code=123456     PAIRS OK
dev      manta-dev://pair?box=<box>&code=123456         PAIRS OK

cross-channel guard: staging-box QR into prod desktop -> rejected (correct)
```

That's the issue's Risk (b) actually closed: `validatePairQrQuery` under `MANTA_CHANNEL=staging` emits `manta-staging://…`, and a staging desktop's prefill accepts it. I also audited every live call site — `src/main/index.ts` (`PAIR_PREFIX` from `CHANNEL_CONFIG.urlScheme`), `App.tsx:264`, `PairStep.tsx:66` are all channel-derived; the only remaining default-scheme caller is `deepLink.ts:134`, which is the mobile path that correctly bakes `prod` and is explicitly out of scope. No hardcoded `manta://pair` literal remains on the desktop chain.

**Block 2 (followup-issue) — BET-386: FILED, verified.** Not taken on trust — `multica issue get` confirms BET-386 exists, `todo`, titled "Thread build channel through install.sh pair-link emitter (install-lib.mjs)", 1812-char body that names the file. Correctly left out of this PR.

**Nit 2 — fixed.** The self-contradicting memoization comment in `resolveBoxChannel` now says what the code does.

### Mechanical gate — green (run by me on `628f2e9`)

```
$ npm run typecheck   → exit 0, 0 errors
$ npm test            → vitest 1440 passed (1440) | node:test # pass 1117 # fail 0 → exit 0
```

1440 vitest matches the +5 new `setupLogic.test.ts` cases and the reported number exactly. The new tests derive `SCHEMES` from `CHANNEL_IDS`/`CHANNELS` (same tripwire pattern as `pairPayload.test.ts`), and one case pins the pre-fix bug directly — good regression choice.

Scope is clean: the fix commit is 4 files / +66 −5, no creep.

---

### **Block (fix-here): PR is unmergeable — content conflict in `src/renderer/PairStep.tsx` against `origin/main`**

`gh pr view 319` reports `mergeable: CONFLICTING`. This is new since cycle 1 (the PR was `MERGEABLE` then) and it is **not** the implementer's error — `main` picked up `e2ee872 refactor(renderer): dedupe onboarding step 1 copy, drop Skip setup (BET-382)`, which rewrote the same region of `PairStep.tsx` that the cycle-2 fix inserted into. Cycle 1 didn't touch `PairStep.tsx`, so the collision only appeared once the fix landed there.

```
$ git merge-base HEAD origin/main   → 2b47f8d3
$ git rev-parse origin/main         → b63e1816
$ git merge --no-commit --no-ff origin/main
Auto-merging src/renderer/PairStep.tsx
CONFLICT (content): Merge conflict in src/renderer/PairStep.tsx
```

**Exactly one conflict hunk**, and it's mechanical — the fix's new const block collided with main dropping the `onSkip` prop:

```
<<<<<<< HEAD
const PAIR_PREFILL_SCHEME = channelConfig(__MANTA_CHANNEL__).urlScheme;

export function PairStep({ onPaired, onSkip }: { onPaired: () => void; onSkip: () => void }) {
=======
export function PairStep({ onPaired }: { onPaired: () => void });
>>>>>>> origin/main
```

Correct resolution: **keep the `PAIR_PREFILL_SCHEME` const and the `channelConfig` import from HEAD, take main's simplified `({ onPaired })` signature.** Then `git push --force-with-lease`.

I want to be accurate about the risk here rather than overstate it, because I initially assumed worse. My first concern was that resolving toward main would silently delete the fix while tests stayed green — there is no `PairStep` component test, so the new coverage is unit-level on `prefillFromPairLink` only (which the implementer's own self-check flagged). I checked instead of assuming: the call site at line 69, `prefillFromPairLink(pendingPairLink, PAIR_PREFILL_SCHEME)`, **auto-merges cleanly and survives**, so dropping the const leaves a dangling reference that `tsc` fails on. The dangerous resolution is therefore compiler-caught, not silent. So this is a genuine merge blocker but a low-risk one — I'm not asking for a component test.

**Also worth knowing before merge:** there is no CI run on `628f2e9`. The only run on this branch (`30440049344`) is against the superseded `d68df85`. `typecheck-test` needs to go green on the rebased head — I verified it locally, but CI hasn't. The conflict likely needs clearing first.

Nothing else. Rebase, resolve that one hunk, push, and this is a PASS from my side — the substantive work is done and verified.

antoinedc:
Rebased onto `origin/main` (`b63e181`) and resolved the one conflict hunk in `PairStep.tsx`: kept the `PAIR_PREFILL_SCHEME` const + `channelConfig` import, took main's simplified `({ onPaired })` signature (BET-382 dropped `onSkip`, confirmed unused elsewhere in the file).

Running the full suite post-rebase surfaced one thing the conflict alone didn't show: BET-382 also added `PairStep.test.tsx`, which renders `PairStep` directly. Under vitest, `__MANTA_CHANNEL__` is only bound by electron-vite's build-time `define` — referencing it directly at module scope threw `ReferenceError` at import time, so the new test file failed outright (not a merge-conflict artifact, a real gap this combination exposed). Fixed by using the same `typeof __MANTA_CHANNEL__ === "string" ? __MANTA_CHANNEL__ : "prod"` guard `src/main/index.ts` and `src/main/installer/installer.ts` already use for this exact reason.

`npm run typecheck && npm test`: 1446 vitest + 1117 node:test, 0 failures (+6 vitest vs the prior head — `PairStep.test.tsx`'s 6 cases now run and pass instead of failing to import).

`gh pr view` now reports `mergeable: MERGEABLE`. Pushed with `--force-with-lease` to `c0ff9c7`.

antoinedc:
## Review — BET-373 (cycle 3) — **PASS**

**Verdict: 0 Blocks, 0 Questions, 2 nits. Routing to `manta-pm` for merge.**

The cycle-2 mechanical Block is resolved and the substantive fix survived the rebase intact. Verified on head `c0ff9c7`.

### Cycle-2 Block — cleared ✅

**Mergeability.** `gh pr view 319` now reports `mergeable: MERGEABLE`, and I confirmed independently against the *current* `main` tip — which has moved again since the rebase (`b63e181` → `0c65be4`):

```
$ git merge --no-commit --no-ff origin/main
Automatic merge went well; stopped before committing as requested
conflicts: (none)
```

So it isn't merely "mergeable against the main it was rebased onto" — it's clean against main as of right now.

**Conflict resolution is exactly right.** `PAIR_PREFILL_SCHEME` and the `channelConfig` import kept; main's simplified `({ onPaired })` signature taken; `grep -n "onSkip"` → no hits, so BET-382's prop removal is fully honoured. The call site still threads the scheme: `prefillFromPairLink(pendingPairLink, PAIR_PREFILL_SCHEME)`.

### The ReferenceError fix — legitimate, and correctly diagnosed

The rebase surfaced a genuine gap rather than a merge artifact: BET-382 landed `PairStep.test.tsx`, which renders the component directly, and `__MANTA_CHANNEL__` is only bound by electron-vite's build-time `define` — so the module-level reference threw at import under vitest. The fix uses the same `typeof __MANTA_CHANNEL__ === "string" ? __MANTA_CHANNEL__ : "prod"` guard already established at `src/main/index.ts:53` and `src/main/installer/installer.ts:72`. Correct pattern, correct precedent, and the accompanying comment explains *why* rather than just *what*.

### Mechanical gate — green

```
$ npm run typecheck   → exit 0, 0 errors
$ npm test            → vitest: Test Files 67 passed (67), Tests 1446 passed (1446)
                        node:test: # tests 1117  # pass 1117  # fail 0
                        → exit 0
```

1446 / 1117 matches the reported numbers exactly; 67 test files is +1 for BET-382's `PairStep.test.tsx`. Targeted runs: `PairStep.test.tsx` 6/6 pass (the file that was throwing), `setupLogic.test.ts` + `pairPayload.test.ts` 90/90 pass.

CI on this head: **`typecheck-test` SUCCESS** (the required check). `duplication-gate` and `E2E Smoke Test` were still running when I looked — see nit 2 for the duplication result, which I reproduced locally.

### Acceptance criteria — re-proven end-to-end on this head

Not re-asserted from cycle 2; re-executed against `c0ff9c7`:

```
box-channel | emitted QR                                | main PAIR_PREFIX | App.tsx parse | PairStep prefill
prod        manta://pair?box=<box>&code=123456           PASS              PASS          PAIRS OK
staging     manta-staging://pair?box=<box>&code=123456    PASS              PASS          PAIRS OK
dev         manta-dev://pair?box=<box>&code=123456        PASS              PASS          PAIRS OK

cross-channel: staging QR -> prod desktop: rejected (correct)
```

That is the issue's stated purpose — a staging box's QR pairs a staging desktop, and a staging link cannot be claimed by a prod build. Every "Done when" item holds: no inline `manta://pair?…` literal at either named site, per-channel round-trip and box-emitter tests present and derived from the channel table, gate green, PR titled with the issue key. Scope is still clean at 12 files with no deletions.

---

### Nits (non-blocking, for the record — do not hold the merge)

**Nit 1 — two of three renderer `__MANTA_CHANNEL__` references remain unguarded.** `App.tsx:33` and `PairingQR.tsx:35` still call `channelConfig(__MANTA_CHANNEL__)` bare, while `PairStep.tsx:59` now guards. Production is unaffected (the `define` substitution applies to all three), and nothing imports those two modules at test scope today — so this breaks nothing now. But it's the exact landmine that just detonated on `PairStep.tsx`: the next person who adds a test importing `App.tsx` or `PairingQR.tsx` hits the same ReferenceError. Worth normalising to the guarded form in a later pass.

I'm explicitly *not* blocking on it, and I'll note the implementer already documented the asymmetry in the `PairStep.tsx` comment ("unlike App.tsx/PairingQR.tsx, which no test imports today") — that's a deliberate, disclosed choice rather than an oversight, which is the right way to leave a known edge.

Secondary observation, not a request: the guard trades a loud `ReferenceError` for a silent `"prod"` fallback, so a build where `define` failed would quietly emit prod schemes instead of crashing. That's the established repo convention (`main/index.ts` layers `resolveChannel(app.isPackaged, …)` on top of it), so this PR is the wrong place to relitigate it.

**Nit 2 — duplication: discounted.** I ran `bash scripts/check-duplication.sh origin/main` myself rather than wait on the advisory job. 2 clones, 23 lines (0.73%), **both** of them the fixture arrays inside `pairPayload.test.ts` (`{ boxId: BOX, code: … }` cases). Test-fixture duplication, not business logic this PR introduced, and no dead code left behind. Note the cycle-1 vite-config clone is gone from the report. Nothing to act on.

---

### Summary for `manta-pm`

Merge-ready from my side: `mergeable: MERGEABLE` against current `main`, required `typecheck-test` green on `c0ff9c7`, full suite green locally, acceptance criteria proven end-to-end, no Blocks and no open Questions. Two nits are advisory only. Per protocol I have not flipped ready state, not merged, and not pushed — that call is yours. Follow-up work is already tracked in BET-386 (`install-lib.mjs` pair-link emitter), which is deliberately out of this PR's scope.

PR #323, opened 2026-07-29
Onboarding step 1: host picker refresh + always-available custom host (BET-384)

antoinedc:
Implements BET-384: adds a refresh affordance to the SSH host picker and makes "Custom host…" a permanent, always-selectable option instead of a mutually-exclusive select-or-input branch.

**Base:** `origin/main` @ `f120bec`

## What changed

**`src/renderer/SshInstallStep.tsx`**
- Deleted the `hostsLoaded && hosts.length > 0 ? <select> : <input>` ternary and the "No hosts found…" paragraph. One `<select>` always renders: discovered aliases + a permanent last `Custom host…` option.
- Extracted the mount effect's fetch into a named `loadHosts()`, shared by the mount effect and a new refresh button (34×34, `bg-bg-elev border border-border`, rotate-arrow SVG that spins while loading, no text label).
- Added the count line (`{n} hosts from ~/.ssh/config`, `· just now` after a manual refresh, `No hosts in ~/.ssh/config` when empty).
- Added the custom-host panel (Host/Port row, User/Identity-file row, footer hint verbatim from the spec) — rendered only when the sentinel is selected, never alongside a second host control.
- Dropped the `Browse` button for Identity file — no file-dialog bridge exists on `MantaPreload` today. Filed [BET-387](mention://issue/92a66c62-04bd-4d89-b9c2-eeec89dec1c0) (parked, `follow-up` label) instead of adding a new IPC channel in this PR, per the issue's own instruction.
- `startInstall` / `runClaim` / `copyDiagnostics` now resolve the selection through `resolveInstallTarget` and pass the resolved value straight through — no target-string assembly in the component.

**`src/shared/sshTarget.ts` (new)**
- `resolveInstallTarget(sel)` — pure, validates (empty host → error, port 1-65535), returns the alias unchanged for the alias branch or a `CustomSshTarget` for the custom branch.
- `SshTarget = string | CustomSshTarget` — widens what every install-path call site accepts. **No second parameter was added anywhere**: `installerStart` / `installerMintAndClaim` (`preflightBox`'s own IPC channel was already folded away in BET-383, so there are two call sites, not three) keep taking one `{ alias }` field, just typed as the union instead of `string`. The alias/string branch is byte-for-byte unchanged, so every existing test that passes a plain string alias keeps passing untouched.
- `sshTargetDestination` / `sshTargetExtraArgs` — the argv split: `[user@]host` as the ssh destination, `-p`/`-i` as separate flags (port cannot live in the destination string for the traditional `ssh <dest> <cmd>` form).

**`src/main/installer/runner.ts`** — `buildArgs` (the actual argv-construction site) now spreads `sshTargetExtraArgs` before the destination; `execRemote`/`streamRemote` accept `SshTarget`. `probeSshG` untouched (alias-only, doesn't apply to an already-fully-specified custom target).

**`src/main/installer/installer.ts` / `handlers.ts` / `preloadAccess.ts` / `preload/index.ts`** — type-only widening (`alias: string` → `alias: SshTarget`) plus swapping the ad-hoc `typeof x !== "string" || x.trim() === ""` checks for the shared `isEmptySshTarget`.

## Tests

`src/shared/sshTarget.test.ts` (new, 31 cases): alias passthrough (+trim, +empty-rejection), full custom target, every optional field omitted individually, invalid port (non-numeric/zero/>65535/non-integer), empty/whitespace host, boundary ports, plus the runner-facing helpers (`isEmptySshTarget`, `sshTargetDestination` — asserts port is never in the destination string, `sshTargetExtraArgs`, `sshTargetLabel`).

No new test file for `SshInstallStep.tsx` — consistent with existing precedent (the component has never had one across BET-355/382/383; its own header comment says it's "React state + per-event dispatch + JSX, nothing more").

## Verification results

- PR branch (`multica/BET-384-host-picker-refresh-custom` @ `08eeaaf`):
  - test: exit `0`. vitest 1481 pass / 0 fail (69 files); node:test 1117 pass / 0 fail (43 suites). log sha256: `1456408b968adb5fded6c885c07b7809e6c465a047481dc23863230d8d8d7722`
- Base (`main` @ `f120bec`):
  - typecheck: exit `0`. Errors: none. log sha256: `00a5b65767585a439874e95183a8a3e16731ab621895b1c6bb2ebbe3c05b5624` (identical — no typecheck-affecting base drift)
  - test: exit `0`. vitest 1450 pass / 0 fail (68 files); node:test 1117 pass / 0 fail (43 suites). log sha256: `0692aecd35e53aeeeb964c514f4a97b70bb7790ba35ddaa27864fd6e14e458ca`
- **Conclusion:** 0 pre-existing failures, 0 new failures. The vitest delta is exactly +31 tests / +1 file (the new `sshTarget.test.ts`) — verified by diffing the full sorted test-title list between branches, not just the count. node:test is unchanged (0 delta).

## Follow-up filed

- [BET-387](mention://issue/92a66c62-04bd-4d89-b9c2-eeec89dec1c0) (parked, `follow-up` label) — native file-dialog bridge for the Identity-file field's Browse button, deliberately out of scope per this issue's own text.

### DIFF
--- a/src/main/installer/_testFixtures.ts
+++ b/src/main/installer/_testFixtures.ts
@@ -77,7 +77,8 @@ export function makeProbeSpawn(
   responses: Record<string, ProbeResponse>,
 ): SpawnFn {
   return (_command, args) => {
-    // args layout: ["-o", "ConnectTimeout=10", "-o", "BatchMode=yes", alias, command]
+    // args layout: ["-o", "ConnectTimeout=10", "-o", "BatchMode=yes", "--", alias, command]
+    // (BET-384: "--" terminates option parsing before the destination)
     const cmd = args[args.length - 1];
     const found = Object.keys(responses).find((k) => cmd.includes(k));
     const r =
--- a/src/main/installer/handlers.ts
+++ b/src/main/installer/handlers.ts
@@ -26,6 +26,7 @@ import {
 import { buildDiagnostics, type DiagnosticsInput } from "./diagnostics.js";
 import type { InstallStageId } from "./stageMapper.js";
 import type { PreflightResult } from "./preflight.js";
+import { isEmptySshTarget, type SshTarget } from "../../shared/sshTarget.js";
 
 // ---------------------------------------------------------------------------
 // Per-install state (single-active; the renderer can re-mount and recover)
@@ -68,16 +69,19 @@ export function registerInstallerHandlers(
     preflight: activePreflight,
   }));
 
-  ipcMain.handle(IPC.installerStart, async (_e, input: { alias: string }) => {
+  ipcMain.handle(IPC.installerStart, async (_e, input: { alias: SshTarget }) => {
     if (activeHandle !== null) {
       // Single-active constraint — refuse to start a second install.
       // The renderer should call installCancel first (or wait for done).
       throw new Error("an install is already in progress");
     }
-    if (typeof input?.alias !== "string" || input.alias.trim() === "") {
+    if (input?.alias === undefined || isEmptySshTarget(input.alias)) {
       throw new Error("alias is required");
     }
-    const alias = input.alias.trim();
+    // Alias branch stays a trimmed string (unchanged); a custom target
+    // (BET-384) is already normalized by resolveInstallTarget on the
+    // renderer side, so there's nothing further to trim on an object.
+    const alias = typeof input.alias === "string" ? input.alias.trim() : input.alias;
     const handleId = `install-${++nextHandleSeq}-${Date.now()}`;
     activeStage = "preflight";
     logTail.length = 0;
@@ -155,7 +159,7 @@ export function registerInstallerHandlers(
 
   ipcMain.handle(IPC.installerMintAndClaim, async (
     _e,
-    input: { alias: string; claimUrlOverride?: string },
+    input: { alias: SshTarget; claimUrlOverride?: string },
   ) => {
     return mintAndClaim(input.alias, {
       persist,
--- a/src/main/installer/installer.ts
+++ b/src/main/installer/installer.ts
@@ -32,6 +32,7 @@ import {
   type SpawnFn,
 } from "./runner.js";
 import { parseSshConfig, type SshHostEntry } from "./sshConfig.js";
+import { isEmptySshTarget, type SshTarget } from "../../shared/sshTarget.js";
 import { isValidBoxToken } from "../../shared/transport.mjs";
 import {
   resolveChannel,
@@ -145,10 +146,10 @@ export type PreflightDeps = {
 };
 
 export async function preflightBox(
-  alias: string,
+  alias: SshTarget,
   deps: PreflightDeps = {},
 ): Promise<PreflightResult> {
-  if (typeof alias !== "string" || alias.trim() === "") {
+  if (isEmptySshTarget(alias)) {
     throw new Error("preflightBox: alias is required");
   }
   const spawn = deps.spawn;
@@ -176,7 +177,7 @@ export async function preflightBox(
 
 // ---------- individual probes ----------
 
-async function probeReachability(alias: string, spawn?: SpawnFn): Promise<PreflightProbes["reachability"]> {
+async function probeReachability(alias: SshTarget, spawn?: SpawnFn): Promise<PreflightProbes["reachability"]> {
   // BatchMode=yes turns "no auth" into exit code 255 / stderr "Permission
   // denied (publickey)" — not a hang on a passphrase prompt.
   const r = await execRemote(alias, PROBE_REACHABILITY_CMD, {
@@ -200,7 +201,7 @@ async function probeReachability(alias: string, spawn?: SpawnFn): Promise<Prefli
   return "unreachable";
 }
 
-async function probeOs(alias: string, spawn?: SpawnFn): Promise<PreflightProbes["os"]> {
+async function probeOs(alias: SshTarget, spawn?: SpawnFn): Promise<PreflightProbes["os"]> {
   const r = await execRemote(alias, PROBE_OS_CMD, { spawn, timeoutMs: 10_000 });
   if (r.code !== 0) {
     return { id: "unknown", arch: "unknown", release: null };
@@ -213,7 +214,7 @@ async function probeOs(alias: string, spawn?: SpawnFn): Promise<PreflightProbes[
   };
 }
 
-async function probeSudo(alias: string, spawn?: SpawnFn): Promise<boolean> {
+async function probeSudo(alias: SshTarget, spawn?: SpawnFn): Promise<boolean> {
   const r = await execRemote(alias, PROBE_SUDO_CMD, { spawn, timeoutMs: 10_000 });
   // `sudo -n true` exit 0 + no password prompt → passwordless sudo.
   // Anything else (exit != 0, or "password" in stderr) → not available.
@@ -221,7 +222,7 @@ async function probeSudo(alias: string, spawn?: SpawnFn): Promise<boolean> {
   return false;
 }
 
-async function probeTailscale(alias: string, spawn?: SpawnFn): Promise<PreflightProbes["tailscale"]> {
+async function probeTailscale(alias: SshTarget, spawn?: SpawnFn): Promise<PreflightProbes["tailscale"]> {
   const r = await execRemote(alias, PROBE_TAILSCALE_CMD, { spawn, timeoutMs: 10_000 });
   if (r.code !== 0 || r.stdout.trim() === "") {
     return { running: false, ipv4: null };
@@ -230,7 +231,7 @@ async function probeTailscale(alias: string, spawn?: SpawnFn): Promise<Preflight
   return parsed ?? { running: false, ipv4: null };
 }
 
-async function probeClock(alias: string, spawn?: SpawnFn): Promise<number> {
+async function probeClock(alias: SshTarget, spawn?: SpawnFn): Promise<number> {
   const local = Math.floor(Date.now() / 1000);
   const r = await execRemote(alias, PROBE_CLOCK_CMD, { spawn, timeoutMs: 10_000 });
   if (r.code !== 0) return 0;
@@ -239,7 +240,7 @@ async function probeClock(alias: string, spawn?: SpawnFn): Promise<number> {
   return local - remote;
 }
 
-async function probeAuthFile(alias: string, spawn?: SpawnFn): Promise<boolean> {
+async function probeAuthFile(alias: SshTarget, spawn?: SpawnFn): Promise<boolean> {
   const r = await execRemote(alias, PROBE_AUTH_FILE_CMD, { spawn, timeoutMs: 10_000 });
   return r.code === 0 && r.stdout.trim() === "INSTALLED";
 }
@@ -291,11 +292,11 @@ export type InstallHandle = {
 };
 
 export function runInstall(
-  alias: string,
+  alias: SshTarget,
   sink: InstallSink,
   deps: InstallDeps = {},
 ): InstallHandle {
-  if (typeof alias !== "string" || alias.trim() === "") {
+  if (isEmptySshTarget(alias)) {
     throw new Error("runInstall: alias is required");
   }
   // Local name `cfg` (not `channelConfig`) so it doesn't shadow the imported
@@ -391,10 +392,10 @@ export type MintAndClaimDeps = {
  * {pairing_code, box_id, serverUrl} block the desktop parses.
  */
 export async function mintAndClaim(
-  alias: string,
+  alias: SshTarget,
   deps: MintAndClaimDeps,
 ): Promise<ClaimOutcome> {
-  if (typeof alias !== "string" || alias.trim() === "") {
+  if (isEmptySshTarget(alias)) {
     throw new Error("mintAndClaim: alias is required");
   }
   if (typeof deps.persist !== "function") {
--- a/src/main/installer/runner.test.ts
+++ b/src/main/installer/runner.test.ts
@@ -16,7 +16,7 @@ import { makeFakeChild } from "./_testFixtures.js";
 // ---------------------------------------------------------------------------
 
 describe("execRemote", () => {
-  it("builds the canonical ssh argv (ConnectTimeout + BatchMode + alias + cmd)", async () => {
+  it("builds the canonical ssh argv (ConnectTimeout + BatchMode + -- + alias + cmd)", async () => {
     const captured: { command: string; args: string[] } = { command: "", args: [] };
     const fake = makeFakeChild();
     const spawn: SpawnFn = (command, args) => {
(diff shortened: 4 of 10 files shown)


antoinedc:
## Review — BET-384 (cycle 1)

**Verdict: 1 Block (fix-here), 0 Questions, 5 Nits.** Returned to `better-ui-dev`.

The architecture here is right, and the part the issue was most worried about — "do not
multiply code paths, do not add a second parameter to three call sites" — was respected
exactly. One user-visible defect in the initial render state blocks the merge; everything
else is a nit.

---

### Mechanical gates

Run independently on the PR head (`08eeaaf`), not taken from the PR body:

| Gate | Command | Result |
|---|---|---|
| Typecheck | `npm run typecheck` | exit `0`, no errors |
| Tests | `npm test` | exit `0` — vitest **1481 passed / 69 files**, node:test **1117 pass / 0 fail** |

These match the PR body's reported PR-branch numbers exactly (1481 vitest / 1117 node:test),
so the base-branch figures reported there are trusted per the reuse rule rather than
re-run. CI agrees: `typecheck-test` pass, `duplication-gate` pass, `E2E Smoke Test` pass
(all three green as of this review).

**Base freshness:** `git merge-base HEAD origin/main` == `git rev-parse origin/main` ==
`f120bece`. Branched off current main, no phantom deletions, nothing to rebase.

**Anti-spaghetti report:** 0 clones across all 8 changed files (2435 lines, 11250 tokens).
Nothing to act on.

---

Each item verified against the code, not against the PR description.

1. **`Custom host…` selectable with a populated config, reveals the panel** — PASS.
   `SshInstallStep.tsx:426` renders the sentinel `<option>` unconditionally after the
   alias map; `:455` gates the panel on it.
2. **Empty/unreadable config → dropdown renders, `Custom host…` selected, panel open;
   no separate `<input>` branch in the source** — PASS. The ternary and the
   "No hosts found in ~/.ssh/config…" paragraph are gone (`grep "No hosts found" src/`
   → no matches); `loadHosts`'s `catch` sets `hosts=[]` + the sentinel, and the
   `finally` still flips `hostsLoaded`. `id="ssh-host"` now appears exactly twice —
   the `<label htmlFor>` and the `<select>` — never on an input.
3. **Refresh picks up a config edit with no restart** — PASS, and I checked the
   load-bearing half of this rather than assuming it. A refresh button wired to a
   shared `loadHosts()` would still be cosmetic if main memoized the parse; it does
   not. `handlers.ts:63` calls `listSshHosts(DEFAULT_SSH_CONFIG_PATH)` per invoke, and
   `installer.ts:128` does an `existsSync` + `readFileSync` on every call with no cache
   layer. A genuinely fresh read each press.
4. **`resolveInstallTarget` pure + unit-tested; no target-string assembly in the
   component** — PASS. `src/shared/sshTarget.ts` imports nothing (no fs, no DOM, no
   Electron). 31 cases in `sshTarget.test.ts` cover every case the issue's Tests
   section names: alias passthrough (+trim, +empty rejection), full custom target,
   each optional field omitted individually, invalid port (non-numeric / 0 / >65535 /
   non-integer), boundary ports 1 and 65535, empty and whitespace-only host. The
   component's only remaining string work is calling `sshTargetLabel` for the
   diagnostics field — a shared helper, not assembly.
5. **Exactly one host control at any time** — PASS. The panel's `Host or IP` input is
   the select's detail, not a competing control, which is the issue's own framing
   ("the select is the only host control; the panel is its detail"). There is no
   select-or-input branch left.

**Parser untouched**, as the issue predicted (`sshConfig.ts` not in the diff, no
`sshConfig.test.ts` change). Correct — `Include`/`Match` stayed out of scope.

**`Browse` dropped + follow-up filed** — legitimate, this is the issue's own escape
hatch ("If no dialog bridge is already exposed on `MantaPreload`, drop the Browse
button… raise a follow-up instead"). I confirmed BET-387 actually exists, is labelled
`follow-up`, and is unassigned so nothing auto-dispatches.

**No second parameter added** — confirmed by reading every call site, not the summary.
`installerStart` / `installerMintAndClaim` still take one `{ alias }` field, widened
from `string` to `SshTarget`. The string branch is genuinely byte-for-byte unchanged
(`sshTargetDestination` returns `target.trim()`, `sshTargetExtraArgs` returns `[]`), so
`buildArgs` emits an identical argv for an alias as before. `-p`/`-i` are spliced in
`runner.ts:buildArgs` only — the single argv-construction site. `alias` reaches nothing
but `execRemote`/`streamRemote` (grepped all 20 usages in `installer.ts`); nothing
persists it or interpolates it into a URL. `probeSshG` is correctly left `string`-only
and has no non-test callers.

**Design-token check:** `bg-bg-soft`, `text-text-faint`, `border-border-strong`,
`bg-bg-elev` all resolve in `tailwind.config.js:7-9`. No invisible styling from an
invented token. Panel geometry, label sizes, the 34×34 button and the footer hint text
all match the spec verbatim.

---

### Findings

**Block (fix-here): the custom-host panel flashes open on every mount before the host
list arrives.**

`alias` is initialised to `CUSTOM_HOST_VALUE` (`:96`), and the panel is gated on nothing
but `alias === CUSTOM_HOST_VALUE` (`:455`). The host list is fetched in an effect, so the
first committed paint always happens with `alias` on the sentinel — meaning a user with a
populated `~/.ssh/config` sees the full four-field panel plus its hint paragraph (~150px
of layout) render and then collapse once `installerListHosts()` resolves and the first
alias is selected. This is the first screen of onboarding, so it's the app's first
impression, and it's a regression: the old code painted a single-line `<input>` in that
window, not a panel that vanishes.

The code comment at `:93-95` justifies the sentinel start as avoiding an "empty-string
flash", but it trades a flash of an empty select for a flash of the whole panel — strictly
worse, and the comment reads as if the problem were solved.

Decision #3 prescribes the panel being open pre-selected for the *config-yields-nothing*
case; it doesn't sanction it for the populated case. Fix is one condition — gate the panel
on the load having settled, e.g. `hostsLoaded && alias === CUSTOM_HOST_VALUE` (`hostsLoaded`
is already true in both the success and the `catch` path, so the empty-config case still
opens the panel as decision #3 requires). Classified `fix-here`: it touches one boolean in
one file, no new subsystem.

**Nit:** `setTargetError("Select a host.")` in `startInstall` can never be displayed — the
error `<p>` lives inside the panel (`:550`), which is closed whenever the alias branch is
active. Currently unreachable (`parseSshConfig` only builds entries from non-empty tokens,
so `alias === ""` can't occur, and the select's only other value is the sentinel), so this
is a latent trap rather than a live bug. Either hoist the error line out of the panel or
drop the unreachable branch.

**Nit:** `justRefreshed` is never reset (`:91`), so `· just now` is permanent for the rest
of the session and stops being true a minute later. The spec doesn't ask for decay, so
this is taste — but a timestamp that never expires is worse than no timestamp.

**Nit:** a *failed* manual refresh silently discards the user's alias selection — the
`catch` resets `hosts` to `[]` and `alias` to the sentinel, so a transient read failure
yanks the panel open underneath them. Defensible (no alias is valid if the read failed),
worth a moment's thought about whether the previous selection should survive.

**Nit:** `isEmptySshTarget` narrows the old runtime type-guard. Every entry point
previously did `typeof alias !== "string" || alias.trim() === ""` — deliberately defensive
about a malformed IPC payload. The shared helper instead assumes anything non-string is a
`CustomSshTarget`, so `installerStart({ alias: 42 })` now dies with a `TypeError` on
`.host.trim()` instead of the clean `"alias is required"`. It still fails closed and
`handlers.ts:78` guards `undefined` explicitly, so this is error-message quality, not a
hole. A `typeof target !== "object" || target === null` arm in `isEmptySshTarget` restores
the old behaviour for every caller at once.

**Nit (defence in depth):** no `--` before the ssh destination. `sshTargetDestination`'s
output is spliced into argv at `runner.ts:buildArgs` after the flags, and
`resolveInstallTarget` doesn't reject a host starting with `-`, so a host typed as
`-oProxyCommand=…` would be read by ssh(1) as an option rather than a destination. Not a
live vulnerability — the field is user-typed with no deep-link or remote path into it, and
`spawn` runs without `shell: true` so there's no shell to inject into — but a `--`
terminator before the destination, or a leading-`-` rejection in the pure resolver, costs
one line and closes the shape entirely.

---

### Not raised

- No test file for `SshInstallStep.tsx`. The issue's Tests section asks only for
  `sshTarget.test.ts`, and the component has never had one across BET-355/382/383. Not a
  finding, but note that criterion 3 (refresh with no restart) consequently has no
  automated guard — I verified it structurally instead.
- No mockup fidelity or e2e-rendering check: MANTA is an Electron desktop app, outside
  this reviewer's scope.

antoinedc:
## Review cycle 1 — addressed

Pushed `a4e1136` on top of `08eeaaf`. Base still fresh: `origin/main` @ `f120bec`.

**Block (fix-here) — panel flash on mount:** gated the custom panel on `hostsLoaded && alias === CUSTOM_HOST_VALUE` instead of just the alias check. `hostsLoaded` is true in both `loadHosts`' success and catch paths, so the empty-config case still opens the panel pre-selected per decision #3, while a populated config never shows it before the list is ready.

**Self-caught while tracing that fix:** the `setAlias` fallback's `if (prev === CUSTOM_HOST_VALUE) return prev` also fired on the *first* load (since `alias` bootstraps on the sentinel, not a real choice) — so a populated `~/.ssh/config` would never auto-select the first alias, it would just sit on "Custom host…" forever. Added a `hasLoadedOnceRef` to distinguish the bootstrap pass (always resolve to `list[0]` or the sentinel) from a later refresh (preserve whatever the user currently has selected, including a deliberately-open custom panel).

**All 5 nits addressed:**
1. Hoisted the target-error paragraph out of the custom panel — it's rendered once, unconditionally, so it's visible regardless of which `resolveInstallTarget` branch produced it.
2. `justRefreshed` now decays back to `false` 60s after a manual refresh instead of being permanent for the session.
3. A failed *manual* refresh no longer wipes the host list/selection — only a first-load failure resets to the empty-config state; a transient refresh error now leaves whatever was already on screen alone.
4. `isEmptySshTarget` now has a `typeof target !== "object" || target === null` guard, so a malformed non-string/non-object target fails closed instead of throwing a `TypeError` on `.host`.
5. `buildArgs` now emits `--` before the ssh destination. Verified empirically rather than assumed:
   ```
   $ ssh -o ConnectTimeout=1 -o BatchMode=yes -oFooBarBaz=xyz.invalid true
   command-line: line 0: Bad configuration option: foobarbaz     # read as a flag
   $ ssh -o ConnectTimeout=1 -o BatchMode=yes -- -oFooBarBaz=xyz.invalid true
   hostname contains invalid characters                          # read as the destination
   ```
   Covers both the alias and custom-target branches from the one argv-construction site. Added a test asserting `--` sits immediately before the destination even with a custom target's `-p`/`-i` in front of it.

## Verification results

- PR branch (`multica/BET-384-host-picker-refresh-custom` @ `a4e1136`):
  - test: exit `0`. vitest 1482 pass / 0 fail (69 files, +1 from cycle 1's new `--`-placement test); node:test 1117 pass / 0 fail (43 suites). log sha256: `ff81b178d4d3064d5c46c9465f953df3a4ad2b3b08822aabc8288fce7016035c`
- Base (`main` @ `f120bec`): unchanged from cycle 1 (typecheck exit 0 / vitest 1450 / node:test 1117, all 0 fail) — reusing those results per the reuse rule since base hasn't moved.
- **Conclusion:** 0 pre-existing failures, 0 new failures.

antoinedc:
## Review cycle 2 — PASS ✅

**Verdict: 0 Blocks, 0 Questions, 0 nits.** Routing to `manta-pm` for merge.

### Cycle-1 Block — fixed

The panel-flash-on-mount defect is resolved. The custom panel is now gated on `hostsLoaded && alias === CUSTOM_HOST_VALUE` (`SshInstallStep.tsx:455`), so a populated `~/.ssh/config` no longer paints the four-field panel before the host list arrives. The empty-config case still opens the panel pre-selected (decision #3), since `hostsLoaded` flips true in both the success and catch paths of `loadHosts`. Verified by tracing every branch: first-load populated → auto-selects `list[0].alias`, panel stays closed; first-load empty → alias stays on sentinel, panel opens; first-load failure → same as empty.

The implementer also self-caught a latent bug while fixing the Block: the `setAlias` updater's `prev === CUSTOM_HOST_VALUE` short-circuit would have kept a populated config stuck on "Custom host…" forever on the first load. The `hasLoadedOnceRef` correctly distinguishes the bootstrap pass from a later refresh — re-traced all five scenarios (first load empty/populated, refresh with valid/removed/custom selection, refresh failure) by hand.

### Cycle-1 nits — all fixed

1. **Unreachable error hoisted** — `targetError` is now rendered outside the custom panel, so an alias-branch rejection (defensive, unreachable through this UI today) would still surface.
2. **`justRefreshed` decays** — 60s timeout via `JUST_REFRESHED_DECAY_MS`, so "· just now" doesn't read as wrong after time passes.
3. **Failed refresh preserves selection** — the catch block's `if (isFirstLoad)` guard means a later refresh failure no longer wipes the host list and selection.
4. **`isEmptySshTarget` fails closed** — guards `typeof target !== "object" || target === null` before accessing `.host`, so a malformed IPC payload can't throw a TypeError.
5. **`--` before ssh destination** — `buildArgs` now emits `--` after all options and before the destination. Verified empirically per the implementer's comment and covered by a new test (`puts -- immediately before the destination even with a custom target's -p/-i`).

### Acceptance criteria — all 5 independently verified

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Populated config → Custom host… selectable, panel reveals | ✅ `<option value={CUSTOM_HOST_VALUE}>` always rendered; panel gated on `hostsLoaded && alias === CUSTOM_HOST_VALUE` |
| 2 | Empty config → dropdown renders with Custom host… selected, panel open; no `<input>` branch in source | ✅ Old either/or ternary deleted; one `<select>` always renders; `loadHosts` first-load empty → alias stays on sentinel |
| 3 | Refresh picks up config edits without restart | ✅ `loadHosts({ manual: true })` → `installerListHosts()` → `listSshHosts()` reads `~/.ssh/config` from disk every call (no cache, confirmed in `installer.ts:128`) |
| 4 | `resolveInstallTarget` pure + unit-tested; component has no target-string assembly | ✅ `sshTarget.ts` imports nothing (no fs/ssh/DOM); 31 tests; component calls `resolveInstallTarget(currentSelection())` and passes `resolved.target` straight through |
| 5 | Exactly one host control in the DOM at any time | ✅ The `<select>` is always the sole host control; the custom panel is detail fields, not a competing control |

### Mechanical gates

- **Typecheck**: `tsc --noEmit` on both `tsconfig.node.json` and `tsconfig.web.json` — exit 0.
- **Tests**: vitest 1482/1482 (69 files), node:test 1117/1117 — 0 failures.
- **Base freshness**: merge-base `f120bece` (PR branch point). `origin/main` has advanced 4 commits to `42e4d83`, but none touch the PR's files (server sandboxing + onboarding phase routing) — no phantom deletions, no conflicts.
- **Anti-spaghetti (jscpd)**: 0 clones across all 10 changed files.

### Architecture

The design goal is met: `installerStart` / `installerPreflight` / `mintAndClaim` still take ONE target argument (`SshTarget = string | CustomSshTarget`), not a widening union of options. The alias branch is byte-for-byte unchanged — `resolveInstallTarget` returns the alias string verbatim, and `buildArgs` passes it straight through as the ssh destination. A custom target flows through `sshTargetExtraArgs` as `-p`/`-i` flags plus a `[user@]host` destination, all constructed in one place (`runner.ts:buildArgs`). No second parameter was added to any call site.

PR #324, opened 2026-07-29
BET-386: thread build channel through install.sh pair-link emitter

antoinedc:
## What

Threads the build channel through `scripts/install-lib.mjs`'s pair-link
emitter (`buildPairLink` / `formatPairingOutput`), closing the gap
[BET-373](https://github.com/antoinedc/MantaUI/pull/319) fixed on the QR
path but left open on install.sh's printed pairing block: a box installed
from the staging/dev channel printed a `manta://` (prod-scheme) pair link,
so the OS routed the click to whatever app last registered `manta://`,
never the channel that was actually installed.

## Design decision (channel source)

`MANTA_CHANNEL` env var, read via `resolveConfig()` — the exact same
wiring `MANTA_RELEASE_HOST` already uses for this same curl-pipe
invocation. Rejected the other two options the issue raised:

- **Build-time substitution** — `install.sh` is served byte-identical
  across channels (`scripts/release/publish.sh` — no per-channel
  templating), so the script itself has no channel-specific content to
  substitute into.
- **Persisting channel to disk** — [BET-370](https://github.com/antoinedc/MantaUI/pull/318)'s decision record is explicit: *"the channel is a
  property of the BUILD, never of user config — do not persist it to
  config.json"*. An env var at invocation time, no persistence, matches
  that precedent.

## Changes

- `scripts/install-lib.mjs`: `resolveConfig()` reads `MANTA_CHANNEL`
  (unset/unrecognised → `"prod"`, same fallback contract as
  `resolveBoxChannel()` in `pairPage.mjs`) and exposes `channel` +
  `urlScheme`. `buildPairLink`/`formatPairingOutput` take a `scheme`
  option (default `"manta"`) instead of a hardcoded `manta://` literal.
- `scripts/manta-pair.mjs`: passes `cfg.urlScheme` through — covers both
  install.sh's final step (which calls this script) **and** later manual
  `manta pair` re-runs (the latter only actually works as of the review
  cycle 2 fix below — see correction note).
- `src/main/installer/installer.ts`: the SSH orchestrator (desktop app's
  "Install" button) now passes `MANTA_CHANNEL=<channel>` alongside
  `MANTA_RELEASE_HOST` in the curl-pipe command — it already resolves the
  channel record for `releaseHost`, so this rides along for free.
- `scripts/install-lib.d.mts` (new): hand-written declaration for
  `buildPairLink`, same pattern as `src/shared/channel.d.mts`, so a
  checked `.ts` file can import the plain-`.mjs` module.
- **`scripts/install.sh` (review cycle 2 fix, see below):** resolves +
  exports `MANTA_CHANNEL` (default `"prod"`, same fallback as
  `resolveConfig()`) and bakes it into the `manta` CLI shim heredoc the
  same way `MANTA_HOME`/`NODE` already are, so a `manta pair` re-run in
  ANY later shell actually reads the install-time channel back out.

## Tests

- `scripts/install.test.mjs`: `resolveConfig`'s `channel`/`urlScheme`
  fields (prod default, staging, dev, unrecognised-value fallback,
  empty-string-is-unset); `buildPairLink`/`formatPairingOutput`'s `scheme`
  option (staging/dev output, back-compat default, composes with the
  BET-336 `serverUrl` branch).
- `src/main/installer/installer.test.ts`: asserts `MANTA_CHANNEL=dev`
  appears in the built SSH command alongside the existing
  `MANTA_RELEASE_HOST` assertion.
- `src/renderer/mobile/pairPayload.test.ts`: **the actual round-trip
  against the real `parsePairPayload`**, for all three channel schemes
  plus a Tailscale `serverUrl` composition — this is a first: it
  cross-imports `install-lib.mjs`'s `buildPairLink` into the vitest suite
  so the two emitters (`buildPairPayload` and `buildPairLink`) are
  verified to actually agree on the wire shape, not just asserted to by
  comment. `scripts/install.test.mjs` (plain Node, no TS loader) still
  pins the literal string shape for staging/dev, per the pre-existing
  pattern in that file (see the updated comment at the `BET-177 §2.4`
  test) — it cannot load the `.ts` parser directly, which is why the real
  round-trip lives in the vitest file instead.

## Review cycle 2 — correction + fix

The reviewer found the PR body's claim above ("later manual `manta pair`
re-runs" are covered) was **wrong**: `resolveConfig()` deliberately never
persists the channel to disk (BET-370's "channel is a property of the
build" ruling), so a `manta pair` re-run in a fresh shell had no source
for `MANTA_CHANNEL` at all and always resolved `prod` — printing
`manta://` on a staging/dev box, the exact bug this issue describes, on
the documented re-mint path.

**Fix (`Block (fix-here)`):** `scripts/install.sh` now resolves +
exports `MANTA_CHANNEL` right alongside `MANTA_HOME`, and bakes it into
the `manta` CLI shim heredoc the same way `MANTA_HOME`/`NODE` already
are — see the added `scripts/install.sh` diff. Three new tests in
`scripts/install.test.mjs`: a static-layout guard, a dynamic test that
extracts and *executes* the real (unmodified) shim-writing block from
install.sh's source and proves the value reaches a child process's env,
and a default-to-prod guard.

**Follow-up filed (`Block (followup-issue)`):**
[BET-392](mention://issue/f26083b4-0470-4aa8-b831-34eb16f8bfca) (assigned
`manta-pm`, high priority) — the box server's systemd/launchd env still
has no `MANTA_CHANNEL`, so BET-373's channel-aware `/pair/qr.png` stays
dormant on every real box. Out of scope here: it changes the server's
runtime environment (not the install-lib emitter this issue scopes) and
its own test surface (unit-template substitution tests).

## Verification results

- PR branch (`multica/BET-386-channel-install-lib-pair-link` @ `2d6f98a`):
  - test: exit `0` — vitest 1455 pass / 0 fail, node:test 1130 pass / 0 fail. log sha256: `a0185156db1bc33745169724e880e82328ffa2bd0aa33ee71197dbc9aa2207b2`
- Base (`main` @ `f120bec`):
  - test: exit `0` — vitest 1450 pass / 0 fail, node:test 1117 pass / 0 fail. log sha256: `e51581b527d804ed669da253f24fb1e057dfa2c554c9a5f24d54ce8f82e9f335`
- **Conclusion:** 0 pre-existing failures, 0 new failures. +5 vitest / +13 node:test tests added by this PR, all passing.

Base: `origin/main` @ `f120bec`

## Follow-ups filed

- [BET-391](mention://issue/7bf6a514-0ae1-40ab-8aa2-2aeea4777750) (parked,
  low) — the raw `curl -fsSL https://mantaui.com/staging/install.sh | bash`
  path (no SSH orchestrator involved) still needs `MANTA_CHANNEL=staging`
  exported manually since `install.sh` is served byte-identical across
  channels; nothing documents that today.
- [BET-392](mention://issue/f26083b4-0470-4aa8-b831-34eb16f8bfca)
  (assigned `manta-pm`, high) — box server systemd/launchd env needs
  `MANTA_CHANNEL` too (review cycle 1 `Block (followup-issue)`, see above).

### DIFF
--- a/scripts/install-lib.d.mts
+++ b/scripts/install-lib.d.mts
@@ -1,0 +1,20 @@
+// Hand-written type declaration for install-lib.mjs — same pattern as
+// src/shared/channel.d.mts (a plain-JS module imported from a checked .ts
+// file needs a sibling .d.mts under "moduleResolution": "Bundler").
+//
+// install-lib.mjs's full surface is the install.sh CLI/pairing-format
+// helpers, exercised by the plain-Node scripts/install.test.mjs suite (no
+// type declarations needed there — node:test loads the .mjs directly).
+// This file declares ONLY the exports a checked .ts file actually imports
+// (currently: src/renderer/mobile/pairPayload.test.ts, for the BET-386
+// buildPairLink round-trip) — extend it if a future .ts file imports more.
+
+/**
+ * Build the canonical box-form pair link. See install-lib.mjs's JSDoc for
+ * the full contract (BET-177 §2.4, BET-336 serverUrl, BET-386 scheme).
+ */
+export function buildPairLink(
+  boxId: string,
+  code: string,
+  opts?: { serverUrl?: string; scheme?: string },
+): string;
--- a/scripts/install-lib.mjs
+++ b/scripts/install-lib.mjs
@@ -22,6 +22,7 @@ import { createRequire } from "node:module";
 import { STATE_DIRNAME } from "../src/shared/paths.mjs";
 import { loadAuth } from "../src/server/auth.mjs";
 import { isPrivateServerUrl } from "../src/shared/transport.mjs";
+import { channelConfig } from "../src/shared/channel.mjs";
 
 // `qrcode-terminal` is CommonJS; we keep this module ESM. The local
 // `createRequire` shim is sync (no top-level await) and resolves the lib
@@ -85,6 +86,21 @@ function envVal(env, key) {
  *   releaseHost  — MANTA_RELEASE_HOST || DEFAULT_RELEASE_HOST
  *   port         — MANTA_MOBILE_PORT || DEFAULT_PORT
  *   healthUrl    — http://127.0.0.1:<port>/auth/status
+ *   channel      — MANTA_CHANNEL, resolved through channelConfig's id
+ *                  (BET-386; unset/unrecognised → "prod", same fallback as
+ *                  resolveBoxChannel() in src/server/pairPage.mjs)
+ *   urlScheme    — channelConfig(channel).urlScheme — the pair-link scheme
+ *                  formatPairingOutput/buildPairLink default to
+ *
+ * MANTA_CHANNEL follows the exact same wiring as MANTA_RELEASE_HOST: it is
+ * an env var the caller sets before invoking install.sh (mirrors
+ * `src/main/installer/installer.ts`'s SSH orchestrator, which already
+ * resolves the desktop's build channel and passes MANTA_RELEASE_HOST in the
+ * same curl-pipe command — MANTA_CHANNEL rides along there). install.sh
+ * itself is served byte-identical across channels (no build-time
+ * substitution — see scripts/release/publish.sh), so a channel can only be
+ * communicated at invocation time via this env var, not by the script
+ * content.
  */
 export function resolveConfig({ env = process.env, home = homedir() } = {}) {
   const mantaHome = envVal(env, "MANTA_HOME") ?? join(home, DEFAULT_HOME_DIRNAME);
@@ -95,6 +111,7 @@ export function resolveConfig({ env = process.env, home = homedir() } = {}) {
     envVal(env, "MANTA_RELEASE_HOST") ?? DEFAULT_RELEASE_HOST,
   );
   const port = parsePort(envVal(env, "MANTA_MOBILE_PORT")) ?? DEFAULT_PORT;
+  const channel = channelConfig(envVal(env, "MANTA_CHANNEL") ?? "prod");
   return {
     mantaHome,
     authDir,
@@ -103,6 +120,8 @@ export function resolveConfig({ env = process.env, home = homedir() } = {}) {
     releaseHost,
     port,
     healthUrl: `http://127.0.0.1:${port}${HEALTH_PATH}`,
+    channel: channel.id,
+    urlScheme: channel.urlScheme,
   };
 }
 
@@ -910,7 +929,7 @@ export const readBoxIdentity = loadAuth;
  * try/catch degradation), and the per-row QR indenting loop are all kept
  * intact — only the assembled line list changes.
  *
- * BET-177 §2.4: the pair link `manta://pair?box=<id>&code=<code>` is
+ * BET-177 §2.4: the pair link `<scheme>://pair?box=<id>&code=<code>` is
  * always included in the output (when a box_id + a valid 6-digit code are
  * present) as BOTH a copy-paste line AND a terminal-rendered QR. The QR is
  * for the user who wants to scan with the iOS Camera instead of pasting the
@@ -918,6 +937,12 @@ export const readBoxIdentity = loadAuth;
  * `buildPairLink` (local helper) — single source of truth, consumed
  * exclusively by this function now (install.sh's heredoc duplicate was
  * retired in BET-241).
+ *
+ * BET-386 (channel-aware wire format): `scheme` (default "manta") flows
+ * straight through to `buildPairLink` so a staging/dev install prints
+ * `manta-staging://…` / `manta-dev://…` instead of always `manta://`.
+ * Callers should pass `resolveConfig().urlScheme` — see `buildPairLink`'s
+ * doc comment for the single-source-of-truth rationale.
  *
  * The QR generator is injected (`qrRender`) so tests can capture the output
  * without pulling in `qrcode-terminal` at module-load. The default
@@ -927,7 +952,7 @@ export const readBoxIdentity = loadAuth;
  */
 export function formatPairingOutput(
   { pairing_code, box_id, expiresAt, serverUrl } = {},
-  { qrRender = defaultQrRender } = {},
+  { qrRender = defaultQrRender, scheme = "manta" } = {},
 ) {
   if (!/^[0-9]{6}$/.test(String(pairing_code ?? ""))) {
     throw new Error("formatPairingOutput: pairing_code must be 6 digits");
@@ -952,7 +977,7 @@ export function formatPairingOutput(
     // crafted install can't smuggle a public address past this function).
     // ONE wire shape — no more manta:///pair-page fork.
     const useServerUrl = typeof serverUrl === "string" && serverUrl !== "";
-    const pairLinkOpts = useServerUrl ? { serverUrl } : {};
+    const pairLinkOpts = useServerUrl ? { serverUrl, scheme } : { scheme };
     const pairLink = buildPairLink(box_id, pairing_code, pairLinkOpts);
     const pairPageUrl = buildPairPageUrl(box_id, pairing_code, {
       baseUrl: useServerUrl ? serverUrl : undefined,
@@ -1038,11 +1063,20 @@ export function formatPairingOutput(
  * renderer-side parser uses); a non-private value throws — refusing is
  * the intended behaviour, not silently dropping the param.
  *
+ * BET-386 (channel-aware wire format): `scheme` (default "manta" — same
+ * default as the renderer's `buildPairPayload`) lets the caller pick the
+ * channel's own URL scheme so an install run under a non-prod channel
+ * emits `manta-staging://…` / `manta-dev://…` instead of always `manta://`.
+ * Callers should pass `resolveConfig().urlScheme` (channel-derived, single
+ * source of truth in src/shared/channel.mjs) rather than typing a scheme
+ * literal — mirrors how `pairPage.mjs`'s `validatePairQrQuery` and the
+ * renderer's `buildPairPayload` take the same channel-derived value.
+ *
  * Cross-reference: keep in sync with src/renderer/mobile/pairPayload.ts
  * `buildPairPayload` (box-form branch). If the canonical shape changes,
  * BOTH helpers must change in lockstep — the test catches drift.
  */
-export function buildPairLink(boxId, code, { serverUrl } = {}) {
+export function buildPairLink(boxId, code, { serverUrl, scheme = "manta" } = {}) {
   if (typeof boxId !== "string" || !/^[0-9a-f]{32}$/.test(boxId)) {
     throw new Error("buildPairLink: boxId must be a 32-hex token");
   }
@@ -1058,9 +1092,9 @@ export function buildPairLink(boxId, code, { serverUrl } = {}) {
         "buildPairLink: serverUrl must be a private/tailnet http(s) URL",
       );
     }
-    return `manta://pair?box=${encodeURIComponent(boxId)}&code=${code}&server=${encodeURIComponent(serverUrl)}`;
+    return `${scheme}://pair?box=${encodeURIComponent(boxId)}&code=${code}&server=${encodeURIComponent(serverUrl)}`;
   }
-  return `manta://pair?box=${encodeURIComponent(boxId)}&code=${code}`;
+  return `${scheme}://pair?box=${encodeURIComponent(boxId)}&code=${code}`;
 }
 
 /**
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -67,12 +67,16 @@
 #   MANTA_REPO_URL      git URL the deploy is initialised against for `scripts/self-update.sh`
 #                       (default https://github.com/antoinedc/MantaUI.git)
 #   MANTA_HOME          where code is unpacked (default ~/manta)
-#   MANTA_CHANNEL       build channel — prod|staging|dev (default prod). Baked
-#                       into the manta-server systemd unit / LaunchAgent plist
-#                       as Environment/EnvironmentVariables so the box's own
-#                       long-lived process resolves the right pair-link URL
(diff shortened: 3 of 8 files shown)


antoinedc:
## Review — BET-386 (cycle 1)

Base-freshness gate: clean. `git merge-base HEAD origin/main` = `f120bec` = `git rev-parse origin/main`. Diff vs fresh main is 7 files / +245 −16, no phantom deletions.

### Mechanical gate — verified on the PR branch (`c4fda48`)

```
$ npm run typecheck   → exit 0, 0 "error TS" lines
$ npm test            → exit 0
    vitest:     Test Files 68 passed (68) | Tests 1455 passed (1455)
    node:test:  # tests 1127 | # pass 1127 | # fail 0
```

Exact match with the numbers in the PR body (1455 / 1127), so I trusted the reported base-branch run rather than re-installing main. No pre-existing failures, no new failures.

### Acceptance criteria — checked item by item

**1. `buildPairLink` derives its scheme from the channel table, no inline `manta://pair?…` literal — MET.**
Both return branches are now `${scheme}://pair?…`; `resolveConfig()` sources the value via `channelConfig(MANTA_CHANNEL ?? "prod")`, and `channelConfig`'s unknown-id→prod fallback makes the "never throw" contract hold (tests pin `garbage` and `""`). Verified independently:

```
$ grep -rn "manta://pair" --include=*.ts --include=*.tsx --include=*.mjs --include=*.sh src scripts | grep -v "\.test\."
  → 18 hits, ALL of them comments/JSDoc. Zero live literals.
```

**2. Round-trip against `parsePairPayload` for all three schemes — MET in substance, relocated. Accepted.**
The issue asked for this in `scripts/install.test.mjs`; it landed in `src/renderer/mobile/pairPayload.test.ts` instead. I did not take the "plain Node can't load the .ts parser" claim on faith — I read the pre-existing test the issue pointed at (`install.test.mjs:590`, "buildPairLink produces the canonical box-form pair link"). It was **never** a round-trip: it asserts a literal string and its own comment says *"The install.test suite is plain Node — we don't load the .ts parser here."* So the issue text mis-described the starting state, the constraint is real, and the new `pairPayload.test.ts` block is strictly stronger than what was asked for (a genuine `parsePairPayload(buildPairLink(...))` round-trip across all three schemes, plus the BET-336 `serverUrl` composition, plus a negative cross-scheme assertion). Not a defect.

**3. `npm run typecheck && npm test` green — MET.** See above; run by me, not taken from the PR body.

### Findings

**Block (fix-here): the `manta` CLI shim doesn't propagate `MANTA_CHANNEL`, so `manta pair` still prints `manta://` on a staging/dev box — and the PR body claims otherwise.**

`scripts/install.sh:1624-1638` writes the shim heredoc:

```bash
MANTA_HOME="$MANTA_HOME"
NODE="$NODE"
case "\${1:-}" in
  pair) exec "\$NODE" "\$MANTA_HOME/scripts/manta-pair.mjs" ;;
```

`MANTA_HOME` and `NODE` are baked in by interpolation; `MANTA_CHANNEL` is not. Since this PR deliberately (and correctly) rejected persisting the channel to disk, `resolveConfig()` has no source other than the process env — so a `manta pair` run in any fresh shell resolves `channel: "prod"` and emits `manta://pair?…` on a staging/dev box. That is precisely the bug this issue describes ("the OS routes the user's click to the prod desktop build"), and `manta pair` is the *documented* re-mint path — install.sh's own footer tells the user to run it.

The PR body states: *"`scripts/manta-pair.mjs`: passes `cfg.urlScheme` through — covers both install.sh's final step (which calls this script) **and later manual `manta pair` re-runs**."* The first half is true (the child process inherits the env from the curl-pipe). The second half is not.

Fix fits inside this PR — one interpolated line in the existing heredoc, same pattern as the two vars already baked in:

```bash
MANTA_CHANNEL="${MANTA_CHANNEL:-prod}"
export MANTA_CHANNEL
```

Add an `install.test.mjs`/shell-shape assertion alongside it, and correct the PR body's claim either way.

**Block (followup-issue): the systemd unit has no `Environment=MANTA_CHANNEL`, so the box server's `resolveBoxChannel()` always returns prod — BET-373's channel-aware QR is dormant on every real install.**

`scripts/systemd/manta-server.service` declares `Environment=` for `MANTA_MOBILE_HOST`, `MANTA_MOBILE_PORT`, `MANTA_TAILNET_HOST`, `PATH` — no `MANTA_CHANNEL`; and `install.sh:1053-1059`'s `sed` substitution list has no `@@MANTA_CHANNEL@@`. `resolveBoxChannel()` (`src/server/pairPage.mjs:42-50`) reads `process.env.MANTA_CHANNEL` only, so the long-lived server process resolves prod regardless of which channel installed it, and `/pair/qr.png` serves a `manta://` payload from a staging box. Not introduced by this PR — BET-373 shipped the reader without the writer — but it means this PR only makes *half* the pair-link surface channel-aware on a real box, which a reader of BET-386 would reasonably assume it closed. Classified `followup-issue` because it changes the box server's runtime environment and re-opens BET-373's contract + its own test surface (unit-template substitution tests), not the install-lib emitter this issue scopes.

> **Paste-ready follow-up issue body:** *Bake `MANTA_CHANNEL` into the box server's service definitions so BET-373's channel-aware QR actually fires. `resolveBoxChannel()` in `src/server/pairPage.mjs` reads `process.env.MANTA_CHANNEL`, but neither `scripts/systemd/manta-server.service` (which declares `Environment=` lines for `MANTA_MOBILE_HOST`/`MANTA_MOBILE_PORT`/`MANTA_TAILNET_HOST`/`PATH`) nor install.sh's `sed` substitution block at ~:1053 supplies it — so on every real install the server resolves the prod channel and `/pair/qr.png` emits a `manta://pair?…` payload even on a staging/dev box, routing the user's scan to the prod desktop build. Add a `@@MANTA_CHANNEL@@` placeholder to the systemd template plus the macOS LaunchAgent plist, substitute it in install.sh from the same `MANTA_CHANNEL` env var BET-386 introduced, and assert the substitution in the installer test suite. Done when a staging install's `/pair/qr.png` payload carries `manta-staging://` and typecheck + tests are green.*

### Nits (non-blocking)

- `buildPairLink(boxId, code, { serverUrl, scheme = "manta" })` and `formatPairingOutput(..., { scheme = "manta" })` default to a bare `"manta"` string literal rather than `channelConfig("prod").urlScheme`. The acceptance criterion only bans the `manta://pair?…` URL literal, so this passes — but it is still a scheme literal outside the channel table, and it's the one spot where a future rename of the prod scheme would silently drift. The JSDoc's own justification ("same default as the renderer's `buildPairPayload`") is a fair precedent, so this is a Nit, not a Question.
- `scripts/install-lib.d.mts` is hand-written and nothing enforces that it matches `install-lib.mjs`'s real signature — a future `buildPairLink` option would typecheck against a stale declaration. Same exposure as the existing `src/shared/channel.d.mts` precedent it copies, so: pre-existing pattern, noted only.
- The `install.test.mjs:590` comment now correctly points readers at `pairPayload.test.ts` for the real round-trip. Worth also renaming that test — it is still titled/commented as if it round-trips, which is what sent BET-386's issue author down the wrong path in the first place.

No sticky CI comment on the PR, so I ran the identical detector myself: `bash scripts/check-duplication.sh origin/main` → 1 clone, 11 lines (0.14% of analyzed), between `pairPayload.test.ts:381-390` and `:400-411`. Both blocks predate this PR (its only insertion into that file is the BET-386 `describe` at ~:138) and they are `PairPayload[]` test-fixture arrays. Discounted per the judgment rules — no duplication of business logic introduced by this PR, no dead code left behind.

### Verdict

**2 Blocks (1 fix-here, 1 followup-issue), 0 Questions, 3 Nits.** Returning to the implementer. The core change is well-scoped and the test work exceeds what the issue asked for; the gap is that the channel stops propagating at the install.sh boundary, so the two persistent pairing surfaces on the box (`manta pair`, `/pair/qr.png`) still emit prod links.

antoinedc:
Addressed both Blocks from the review.

**Block (fix-here) — MANTA_CHANNEL didn't reach `manta pair` re-runs:** correct catch — my PR body's claim was wrong. Fixed in `2d6f98a`: `install.sh` now resolves + exports `MANTA_CHANNEL` (default `prod`) alongside `MANTA_HOME`, and bakes it into the `manta` CLI shim heredoc the same way `MANTA_HOME`/`NODE` already are. Added 3 tests in `install.test.mjs` — a static-layout guard, a dynamic test that extracts and *executes* the real (unmodified) shim block from install.sh's source (not a hand-copied duplicate, to avoid drift) and proves `MANTA_CHANNEL` reaches a child process's env, and a default-to-prod guard.

**Block (followup-issue) — systemd/launchd never set MANTA_CHANNEL:** filed as [BET-392](https://multica.ai/better-ui/issues/f26083b4-0470-4aa8-b831-34eb16f8bfca) (assigned `manta-pm`, high priority) using your paste-ready description verbatim. Addressed via BET-392 (separate issue). Resuming this PR.

Verification on the updated branch (`2d6f98a`): typecheck exit 0, vitest 1455/1455, node:test 1130/1130 (+3 from cycle 1's 1127). PR description updated with the correction + verification numbers.

antoinedc:
## Review — BET-386 (cycle 2) — PASS

Base-freshness gate: clean. `merge-base` = `f120bec` = `origin/main` tip. Two commits, no phantom deletions. Incremental diff since cycle 1 is `scripts/install.sh` (+20) and `scripts/install.test.mjs` (+109) — nothing else touched, no scope creep.

### Mechanical gate — verified on `2d6f98a`

```
$ npm run typecheck   → exit 0, 0 "error TS" lines
$ npm test            → exit 0
    vitest:     Test Files 68 passed (68) | Tests 1455 passed (1455)
    node:test:  # tests 1130 | # pass 1130 | # fail 0
```

Matches the reported numbers exactly (1455 / 1130, +3 from cycle 1's 1127). The three new shim tests pass:

```
ok 107 - install.sh's manta CLI shim heredoc bakes + exports MANTA_CHANNEL before the exec (static layout)
ok 108 - install.sh's manta CLI shim propagates MANTA_CHANNEL to `manta pair` re-runs (dynamic, ...)
ok 109 - install.sh's manta CLI shim defaults MANTA_CHANNEL to prod when unset at install time
```

### Block 1 (fix-here) — RESOLVED, independently verified

`install.sh` now resolves `MANTA_CHANNEL="${MANTA_CHANNEL:-prod}"` + `export`s it next to `MANTA_HOME` (line ~553, ahead of every downstream node invocation in the run), and bakes `MANTA_CHANNEL="$MANTA_CHANNEL"` + `export MANTA_CHANNEL` into the shim heredoc before the `case`/`exec` — the same interpolation pattern `MANTA_HOME`/`NODE` already use.

I did not take the new tests' word for it. I extracted the real shim-writing block from `install.sh` myself, executed it, and ran the resulting shim under **`env -i`** (fully empty environment — no `MANTA_CHANNEL` to inherit), with the shim's `exec` target calling the real `resolveConfig()` + `buildPairLink()` from `install-lib.mjs`:

```
install-time MANTA_CHANNEL=staging → channel=staging urlScheme=manta-staging
    link=manta-staging://pair?box=0123456789abcdef0123456789abcdef&code=847291
install-time MANTA_CHANNEL=dev     → channel=dev     urlScheme=manta-dev
    link=manta-dev://pair?box=…&code=847291
install-time MANTA_CHANNEL=prod    → channel=prod    urlScheme=manta
    link=manta://pair?box=…&code=847291
```

That is the end-to-end behaviour the Block asked for: a `manta pair` re-run months later, in a shell that has never heard of `MANTA_CHANNEL`, emits the channel's own scheme. Confirmed.

Worth calling out that the dynamic test (108) is the right shape — it slices the *real, unmodified* production heredoc out of `install.sh` between two anchor lines and executes it against a fake `$HOME`/`$NODE`, rather than re-typing the heredoc in the test. A future edit to the real block cannot silently drift past it.

### Block 2 (followup-issue) — RESOLVED

[BET-392](https://multica.ai/better-ui) filed with the paste-ready description verbatim, priority high, and already `in_progress`. (The issue comment said it was assigned to `manta-pm`; it currently sits with the implementer — fine outcome, just a description/state mismatch, noted below.)

### PR body correction — RESOLVED

The false claim is corrected in place: the "Changes" bullet now qualifies re-run coverage with "the latter only actually works as of the review cycle 2 fix below", and there's an explicit "Review cycle 2 — correction + fix" section. Good — the record reads honestly rather than quietly rewriting history.

### Nits (non-blocking, carried + new)

- **New:** the comment above `MANTA_CHANNEL="${MANTA_CHANNEL:-prod}"` says it matches `resolveConfig()`'s *"unset/unrecognised → prod"* fallback. `${VAR:-prod}` only covers unset/empty — an unrecognised value (`MANTA_CHANNEL=garbage`) is exported verbatim and baked verbatim into the shim. Net behaviour is still correct because `channelConfig()` collapses it to prod downstream, so this is cosmetic. Validating against `CHANNEL_IDS` at the export site would make the comment true and would also close the heredoc-interpolation surface (a channel string containing `"` or `$(…)` lands unquoted in the generated shim) — though that surface is identical to the one `MANTA_HOME`/`NODE` already have in the same heredoc and sits inside the user's own trust boundary, so it is not a security finding.
- **Carried, unaddressed (all fine):** the bare `scheme = "manta"` default living outside the channel table; `install-lib.d.mts` being hand-written with nothing enforcing agreement with the `.mjs`; and the stale test name at `install.test.mjs:590` still reading as if it round-trips. None block.

No sticky CI comment, so I ran `bash scripts/check-duplication.sh origin/main` again: **javascript 0 clones** (the +109 lines of new `install.test.mjs` introduced none), typescript 1 clone / 11 lines / 0.14% total — the same pre-existing `PairPayload[]` fixture arrays at `pairPayload.test.ts:381-411` flagged in cycle 1, untouched by this PR. Discounted per the judgment rules. No dead code left behind.

### Verdict

**0 Blocks, 0 Questions, 4 Nits.** Both cycle-1 Blocks are genuinely closed — the fix-here one verified end-to-end by me under a clean environment, not just by the author's tests. Routing to `manta-pm` for the merge hand-off. Leaving the PR in draft; promoting it is the PM's call.

PR #336, opened 2026-07-30
fix(sitemap): emit UTC lastmod so test runs don't dirty sitemap.xml

antoinedc:
## BET-385

The sitemap generator used git's `%cI` (committer date, strict ISO 8601), which carries the committer's **stored timezone offset**. Commits made on the dev box (+02:00) produced `lastmod` values like `2026-07-28T09:33:17+02:00`, while a UTC runner produced `+00:00` — so every test run that regenerated `website/sitemap.xml` left a noisy, unrelated diff on a tracked file that had to be manually dropped from feature PRs (as happened on BET-383 / PR #322).

## Fix

Switch `lastmodIsoFor` from `%cI` to `%ct` (unix epoch, timezone-independent) and normalize via `Date#toISOString()` (UTC, `...Z`). Output is now byte-identical across machines/CI/timezones and only changes when content actually changes.

This is the issue's suggested option B (pin generator to UTC) over option A (gitignore), because the file is intentionally tracked and deployed by the website workflow.

## Verification

- `npm run typecheck` — clean
- `npm test` — 1134 pass, 0 fail
- Idempotency test (`build-website-sitemap.mjs is idempotent on identical input`) — pass
- `sitemap.xml parses` (XML) — pass
- Regenerated `website/sitemap.xml` committed; the only diff is the format (`+00:00` → `.000Z`) and the corrected `index.html` lastmod (its real last commit was `2026-07-28T07:33:17Z`, previously stale at `2026-07-27T16:51:21+00:00`).

### DIFF
--- a/scripts/build-website-sitemap.mjs
+++ b/scripts/build-website-sitemap.mjs
@@ -25,13 +25,20 @@ const SCHEMA_PATH = join(ROOT, "website/schema/meta.json");
 const SITEMAP_PATH = join(ROOT, "website/sitemap.xml");
 
 function lastmodIsoFor(relFile) {
+  // Always emit UTC so re-running the generator on any machine/CI timezone
+  // produces byte-identical output. `%cI` carries the committer's stored
+  // timezone offset, which varies across machines and dirties sitemap.xml on
+  // every test run; `%ct` is a timezone-independent unix epoch we normalize
+  // to an ISO-8601 UTC string (`...Z`).
   try {
-    const iso = execFileSync(
+    const unix = execFileSync(
       "git",
-      ["log", "-1", "--format=%cI", "--", relFile],
+      ["log", "-1", "--format=%ct", "--", relFile],
       { cwd: ROOT, encoding: "utf8" },
     ).trim();
-    return iso || new Date().toISOString();
+    const sec = Number(unix);
+    if (!Number.isFinite(sec) || sec <= 0) return new Date().toISOString();
+    return new Date(sec * 1000).toISOString();
   } catch {
     return new Date().toISOString();
   }
--- a/website/sitemap.xml
+++ b/website/sitemap.xml
@@ -2,67 +2,67 @@
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
   <url>
     <loc>https://mantaui.com/claude-code-mobile</loc>
-    <lastmod>2026-07-27T16:51:21+00:00</lastmod>
+    <lastmod>2026-07-27T16:51:21.000Z</lastmod>
     <changefreq>monthly</changefreq>
     <priority>0.8</priority>
   </url>
   <url>
     <loc>https://mantaui.com/claude-code-remote</loc>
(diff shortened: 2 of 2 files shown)


antoinedc:
## Review — BET-385 (PR #336) ✅ PASS

**Cycle 1.** Reviewing against the issue's acceptance criteria (option B: pin generator to UTC).

### Acceptance criteria checklist

| # | Criterion (from issue) | Verified | How |
|---|------------------------|----------|-----|
| 1 | Generator emits UTC (`toISOString()` instead of local-offset) | ✅ | `lastmodIsoFor` uses `git log --format=%ct` (epoch) → `new Date(sec*1000).toISOString()`. All 11 `lastmod` values end in `.000Z`, zero `+` offsets (`grep` confirmed). |
| 2 | Deterministic across machines/CI/timezones | ✅ | `%ct` is timezone-independent epoch; `toISOString()` always emits UTC `Z`. Output is byte-identical regardless of host TZ. |
| 3 | Only changes when content actually changes | ✅ | `lastmod` derives from the file's last commit timestamp — only moves when a new commit touches the page. |
| 4 | Fallback path also UTC-consistent | ✅ | Both the `catch` and the `sec <= 0 / !finite` guards fall back to `new Date().toISOString()` (UTC). |
| 5 | Committed `sitemap.xml` matches generator output | ✅ | Ran generator on PR branch; `diff` against committed file = identical (no hand-editing). |
| 6 | Idempotent on re-run | ✅ | Ran generator twice; byte-identical. Existing test `build-website-sitemap.mjs is idempotent on identical input` passes. |
| 7 | Test suite no longer dirties the tracked file | ✅ | `git status` clean after full `npm test` run — the original BET-383 annoyance is gone. |

### Mechanical gate

- `npm run typecheck` — **clean** (tsc node + web, zero errors)
- `npm test` — **1134 pass, 0 fail** (matches implementer's claim)
- Base freshness: merge-base `4dd8d3c` == origin/main at branch cut; no phantom deletions. Diff scoped to exactly 2 files (`scripts/build-website-sitemap.mjs` + regenerated `website/sitemap.xml`) — no scope creep.
- Anti-spaghetti: 0 clones in changed scope.

### Diff notes

The fix is minimal and correct: `%cI` (committer date with stored TZ offset) → `%ct` (epoch) + `toISOString()`. The input-validation guard (`Number.isFinite(sec) && sec > 0`) is a nice robustness touch over the original bare `iso || fallback`. The committed `sitemap.xml` diff is purely format (`+00:00` → `.000Z`) plus one legitimate correction (`index.html` lastmod `2026-07-28T07:33:17Z`, which the stale committed file had wrong) — consistent with the implementer's explanation.

### Nits (non-blocking)

- **Regression-test gap (optional hardening):** the existing idempotency test runs the generator twice on the *same* machine, so it would also pass with the old `%cI` code (same machine → same offset). A one-line assertion that every `lastmod` matches `/\.000Z$/` (or contains no `+`) would lock in the UTC fix and catch a future revert. Not required by the issue; the fix itself is correct.

### Outcome

**0 Blocks, 0 Questions, 1 Nit.** Routing to `manta-pm` for merge.

PR #340, opened 2026-07-30
refactor(server): collapse five JSON store copies into jsonStore.mjs (BET-376)

antoinedc:
## One JSON store helper — collapse five copies of the atomic read/write pattern (BET-376)

Stage 1 pure refactor: no behaviour change, no new feature. Collapses the five
duplicated atomic read/write pairs into one `src/server/jsonStore.mjs`.

### What changed

**New:** `src/server/jsonStore.mjs` exports exactly two functions:
- `readJsonSync(path, fallback)` — synchronous; returns `fallback` on missing / unreadable / invalid-JSON, never throws.
- `writeJsonAtomic(path, data, { mode } = {})` — temp-file + rename; creates parent dir; applies `mode` (via writeFile + chmod) when supplied.

**Edited (load/save bodies only):** the five stores now delegate to jsonStore. Each keeps its name, signature, async-ness, record shape, default, fallback and mode unchanged:
- `capabilities.mjs` — `loadJobs`/`saveJobs` (async, no mode)
- `schedule.mjs` — `loadJobs`/`saveJobs` (async, no mode)
- `webhooks.mjs` — `loadHooks` (sync)/`saveHooks` (async, **0o600**)
- `secrets.mjs` — `loadSecrets` (sync)/`saveSecrets` (async, **0o600**)
- `servePage.mjs` — `loadPages` (sync)/`savePages` (returns promise, no mode)

**`storeUtils.mjs`:** its `atomicWrite` (the shared helper capabilities & schedule previously imported) now delegates to `jsonStore.writeJsonAtomic`. Kept as a thin re-export so `tmux.mjs` (the remaining importer, out of scope here) compiles unchanged. The temp-file-then-rename dance no longer lives in storeUtils.

### Tests

`src/server/jsonStore.test.mjs` (node:test, against a temp dir) covers all six required cases: missing-file fallback, invalid-JSON fallback, round-trip, 0o600 mode, parent-dir creation, overwrite-leaves-no-temp. Plus a no-mode-readable guard.

All existing tests in `capabilities/schedule/webhooks/secrets/servePage.test.mjs` pass **unchanged** — no edits to any existing test file.

### DoD

- `npm run typecheck && npm test` green (1150 pass, 0 fail).
- `grep -rn "renameSync|\.tmp" src/server/*.mjs` — the dance is gone from all five stores + storeUtils; it now lives only in `jsonStore.mjs`. **Caveat below.**
- Secrets store file on disk is still 0600 after a save (test 4 pins this).

### DoD grep caveat (needs a decision)

The DoD grep expects the atomic-write dance to appear **only** inside `jsonStore.mjs`. Five **other** modules carry the same `${path}.tmp-<pid>-<ts>` pattern and were **not** in this issue's explicit 5-file scope:

- `src/server/auth.mjs` (security-critical — `~/.manta/auth.json`)
- `src/server/local.mjs` (`~/.manta/config.json`)
- `src/server/providers.mjs` (opencode.jsonc writes)
- `src/server/push.mjs` (VAPID keys)
- `src/server/gatewayRegister.mjs` (gateway state)

Plus `opencodeRestart.test.mjs` (uses `renameSync` for its own fixture, unrelated).

I left these untouched per the issue's "Edit the load/save bodies only, in: [the 5 files]" instruction. Folding them in — especially `auth.mjs` — is a separate, risk-scoped change. **Recommend a follow-up issue** to collapse these five remaining copies into `jsonStore` so the DoD grep holds literally.

Base: origin/main @ d282633

### DIFF
--- a/src/server/capabilities.mjs
+++ b/src/server/capabilities.mjs
@@ -17,13 +17,11 @@
 // box reboot — the SSE+catch-up plumbing on the executor side picks them up
 // when the desktop comes back.
 
-import { readFile, mkdir } from "node:fs/promises";
-import { existsSync } from "node:fs";
 import { randomBytes } from "node:crypto";
-import { join, dirname } from "node:path";
+import { join } from "node:path";
 import { statePath } from "../shared/paths.mjs";
 import { normalizeHost } from "../shared/pluginManifest.mjs";
-import { atomicWrite } from "./storeUtils.mjs";
+import { readJsonSync, writeJsonAtomic } from "./jsonStore.mjs";
 
 // ---------------------------------------------------------------------------
 // Constants (single source of truth — see docs/mantaui-plugins.md §Constants)
@@ -53,18 +51,12 @@ const MAX_TERMINAL_JOBS = 50;
 // ---------------------------------------------------------------------------
 
 export async function loadJobs(path = STORE_PATH) {
-  try {
-    if (!existsSync(path)) return [];
-    const parsed = JSON.parse(await readFile(path, "utf-8"));
-    return Array.isArray(parsed?.jobs) ? parsed.jobs : [];
-  } catch {
-    return []; // corrupt/unreadable → start empty rather than crash the server
-  }
+  const parsed = readJsonSync(path, {});
+  return Array.isArray(parsed?.jobs) ? parsed.jobs : [];
 }
 
 export async function saveJobs(jobs, path = STORE_PATH) {
-  await mkdir(dirname(path), { recursive: true });
-  await atomicWrite(path, JSON.stringify({ jobs }, null, 2));
+  await writeJsonAtomic(path, JSON.stringify({ jobs }, null, 2));
 }
 
 function genId() {
--- a/src/server/jsonStore.mjs
+++ b/src/server/jsonStore.mjs
@@ -1,0 +1,49 @@
+// Single source of truth for the atomic JSON read/write dance used by every
+// durable store on the box (cap-jobs, schedule, webhooks, secrets, serve-page).
+// Promoted out of the five per-store copies in BET-376 — a pure collapse, no
+// behaviour change: each store keeps its own load/save wrapper (name, signature,
+// async-ness, record shape, default, fallback and mode are all preserved), only
+// the temp-file-then-rename core now lives here.
+//
+// `readJsonSync` is synchronous because `loadHooks`/`loadSecrets`/`loadPages`
+// are synchronous today and changing that would ripple into their callers. It
+// returns `fallback` on any failure — missing file, unreadable file, or invalid
+// JSON — and never throws.
+//
+// `writeJsonAtomic` keeps the existing temp-file-then-rename pattern, applies
+// `mode` to the final file when supplied (matching the 0600 stores), and
+// creates the parent directory if missing (matching what the existing copies
+// did before each write).
+
+import { readFileSync, existsSync } from "node:fs";
+import { writeFile, rename, chmod, mkdir } from "node:fs/promises";
+import { dirname } from "node:path";
+
+// Read and parse `path` as JSON. Returns `fallback` when the file is missing,
+// unreadable, or contains invalid JSON. Never throws.
+export function readJsonSync(path, fallback) {
+  try {
+    if (!existsSync(path)) return fallback;
+    return JSON.parse(readFileSync(path, "utf-8"));
+  } catch {
+    return fallback;
+  }
+}
+
+// Write `data` (already-serialized bytes) to `path` atomically: write a temp
+// file alongside, then rename. The temp filename encodes pid + timestamp so a
+// crash mid-write can never clobber a sibling in-flight temp, and a stale temp
+// from a previous crash never overwrites a live `path`. When `mode` is supplied
+// it is applied to the temp file (and re-asserted via chmod, since writeFile's
+// mode is only honoured on create) so the renamed final file carries it.
+export async function writeJsonAtomic(path, data, { mode } = {}) {
+  await mkdir(dirname(path), { recursive: true });
+  const tmp = `${path}.tmp-${process.pid}-${Date.now()}`;
+  if (mode !== undefined) {
+    await writeFile(tmp, data, { mode });
+    await chmod(tmp, mode).catch(() => {});
+  } else {
+    await writeFile(tmp, data);
+  }
+  await rename(tmp, path);
+}
--- a/src/server/jsonStore.test.mjs
+++ b/src/server/jsonStore.test.mjs
@@ -1,0 +1,90 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { mkdtemp, readFile, stat, readdir, rm, writeFile, mkdir } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { readJsonSync, writeJsonAtomic } from "./jsonStore.mjs";
+
+async function tmpDir() {
+  const dir = await mkdtemp(join(tmpdir(), "manta-jsonstore-test-"));
+  return dir;
+}
+
+// 1. readJsonSync on a missing file returns the fallback.
+test("readJsonSync on a missing file returns the fallback", async () => {
+  const dir = await tmpDir();
+  const missing = join(dir, "nope.json");
+  assert.deepEqual(readJsonSync(missing, { ok: true }), { ok: true });
+  assert.deepEqual(readJsonSync(missing, []), []);
+  assert.equal(readJsonSync(missing, null), null);
+  await rm(dir, { recursive: true, force: true });
+});
+
+// 2. readJsonSync on a file containing invalid JSON returns the fallback and
+//    does not throw.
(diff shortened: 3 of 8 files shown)


antoinedc:
## Review — BET-376 (cycle 1)

**Mechanical gate — PASS**
- `npm run typecheck` → exit 0, zero errors (matches your reported log sha `00a5b65…`).
- `npm test` → exit 0, **1150 pass / 0 fail / 0 skip** (matches your reported PR-branch result exactly; base was 1134, so +16 from the new `jsonStore.test.mjs`).
- No existing `*.test.mjs` file was edited — confirmed against `git diff --stat`.

**Base freshness — PASS.** `merge-base HEAD origin/main == origin/main == d282633` (fresh base, no phantom deletions).

**Acceptance criteria — verified independently**

| # | Criterion | Status |
|---|-----------|--------|
| 1 | `jsonStore.mjs` exports `readJsonSync` + `writeJsonAtomic` | ✅ |
| 2 | `readJsonSync` synchronous, returns `fallback` on missing/unreadable/invalid JSON, never throws | ✅ (try/catch wraps `existsSync`+`readFileSync`+`JSON.parse`) |
| 3 | `writeJsonAtomic` temp-file-then-rename, applies `mode`, creates parent dir | ✅ |
| 4 | Per-store wrappers keep name/signature/async-ness/return shape | ✅ all 5 (cap `loadJobs`/`saveJobs` async; schedule same; secrets `loadSecrets` sync / `saveSecrets` async; servePage `loadPages` sync / `savePages` returns promise; webhooks `loadHooks` sync / `saveHooks` async) |
| 5 | Modes preserved (secrets 0600, webhooks 0600, others unchanged) | ✅ verified in diff; test #4 asserts 0600 |
| 6 | No store path/shape/default/fallback changed | ✅ (`~/.manta/*.json` paths intact; all fallbacks still `[]`) |
| 7 | 6 required test cases in `jsonStore.test.mjs` | ✅ (all 6 + 1 extra no-mode guard) |
| 8 | Existing tests unchanged | ✅ |

**`storeUtils.mjs` edit (outside the named 5-file scope) — not scope creep.** `capabilities.mjs` and `schedule.mjs` previously imported `atomicWrite` from `storeUtils`; after delegating to `jsonStore` they no longer do, but `tmux.mjs` still imports `storeUtils.atomicWrite`. The thin re-export preserves the signature so `tmux.mjs` compiles unchanged. Correct handling — without it `storeUtils` would be dead and `tmux.mjs` would break. No leftover dead imports in any edited file (verified: `servePage`/`secrets` still use `readFile`/`mkdir`/`chmod`/`rm` for their non-store page/secret-material logic).

**Anti-spaghetti (CI sticky comment) — discount.** One clone flagged: `capabilities.mjs`↔`schedule.mjs` share identical `loadJobs`/`saveJobs`/`genId` blocks. This is required by the issue's decision #4 ("Per-store call sites keep their current names, signatures, async-ness and return shapes") and the "Do NOT merge the five stores into one file" rule. Coincidental/required clone — no action.

---

**Question — DoD grep vs. explicit 5-file scope**

The issue's Definition of Done states:

> `grep -rn "renameSync|\.tmp" src/server/*.mjs` shows the atomic-write dance only inside `jsonStore.mjs`.

I ran it. It does **not** pass — the dance still lives in five *other* modules you named and filed BET-400 for:

```
src/server/auth.mjs:258            const tmp = `${path}.tmp-${process.pid}-${Date.now()}`;
src/server/gatewayRegister.mjs:46  const tmp = `${path}.tmp-${process.pid}-${Date.now()}`;
src/server/local.mjs:38            const tmp = `${path}.tmp-${process.pid}-${Date.now()}`;
src/server/providers.mjs:85        const tmp = `${path}.tmp-${process.pid}-${Date.now()}`;
src/server/push.mjs:62             const tmp = `${path}.tmp-${process.pid}-${Date.now()}`;
```

This is genuinely ambiguous and I can't resolve it unilaterally:

- The **Scope** section explicitly lists exactly 5 files to edit ("Edit the load/save bodies only, in: …") and the **Decisions** section says "implement exactly these, do NOT revisit." Folding in `auth.mjs` (security-critical, 0600 secrets) without explicit instruction is risky scope creep — your deferral + BET-400 is a defensible reading.
- But the **DoD grep** is a spelled-out, measurable acceptance criterion that literally fails today, and the issue's stated goal is "Collapse to one first" — which is only half-achieved if 5 more copies remain.

**To resolve, pick one and reply on this PR:**
- **(a) Defer with explicit rationale** (your current position): confirm BET-400 is the accepted resolution for the DoD grep and that this PR's scope is intentionally limited to the 5 named files. Then reassign back to me for a re-check and I'll route to `manta-pm`.
- **(b) Expand scope**: fold `auth.mjs`, `local.mjs`, `providers.mjs`, `push.mjs`, `gatewayRegister.mjs` into `jsonStore` in this PR (take care to preserve each one's mode — `auth.mjs` is security-critical), so the DoD grep actually passes.

No Block — the work that IS here is correct, the mechanical gate is green, and the deferral was disclosed transparently rather than hidden. I'm surfacing this as a Question rather than silently accepting a failed spelled-out criterion.

**Nits (non-blocking)**
- `chmod(tmp, mode).catch(() => {})` silently swallows chmod failures — but this matches the pre-existing `secrets.mjs`/`webhooks.mjs` belt-and-suspenders behavior, so it's a faithful carry, not a regression. No action needed.

antoinedc:
Resolving the Question: **(a) — BET-400 is the accepted resolution.** The DoD grep and the Scope section are in genuine tension, and the technical reason the other 5 modules can't be folded into this "pure refactor, no behaviour change" PR is that they are **not** clean `readJsonSync`/`writeJsonAtomic` drop-ins:

- **`providers.mjs`** — writes/reads `opencode.jsonc` (JSONC with comments). Its read (`readRemoteConfig`, line 260) strips line comments via `stripLineComments` **before** `JSON.parse`, and **throws** on unparseable (callers must not overwrite a bad config). `readJsonSync` does raw `JSON.parse` and returns a fallback on failure. Folding providers.mjs in would either break its JSONC read (behaviour change) or require extending `jsonStore` with JSONC-aware parsing + a throw-on-bad variant = new behavior, scope creep.
- **`local.mjs`** — `getConfig` is **async** with inline project-shape migration + a `_config` cache (`readJsonSync` is sync, no migration, no cache). It also uses its `atomicWrite` for **non-JSON** files (`tmux.conf`, `tmux.confBak` — plain text), so the local helper can't be deleted even if the config.json write were swapped.
- **`push.mjs`** — reads VAPID/subscription/token stores with **async** `readFile`+`JSON.parse` across multiple sites; `readJsonSync` would flip them sync.
- **`auth.mjs`** / **`gatewayRegister.mjs`** — these two *are* clean drop-ins (sync `readFileSync`+`JSON.parse`, 0600 JSON write). But `auth.mjs` persists `~/.manta/auth.json` (the box identity + gateway token) — security-critical — and a drive-by swap in a "no behaviour change" PR is the wrong risk profile; it deserves its own change with dedicated 0600 test attention. BET-400 carries that.

The issue author's Scope explicitly names 5 files, and those 5 are exactly the clean "load JSON / save JSON" pairs. The DoD grep encodes the *aspirational* end state (one place with the dance); BET-400 is the path to it, tracking each remaining module's individual semantics (JSONC, async, non-JSON, security) rather than forcing them through an abstraction that doesn't fit. Resuming this PR as-is.

antoinedc:
## Review — BET-376 (cycle 2) — Question resolved, PASS

The author's deferral rationale (option a) is verified against the code and holds on all five points:

- **providers.mjs** — reads JSONC via `stripLineComments` + `JSON.parse` and throws on unparseable; `readJsonSync` does raw `JSON.parse` with fallback. Folding would change read semantics. ✅ not a drop-in.
- **local.mjs** — async read; its `atomicWrite` is shared with non-JSON `~/.tmux.conf` writes. Folding the write side would break the tmux.conf path. ✅ not a drop-in.
- **push.mjs** — async reads at multiple sites; `readJsonSync` is sync and the issue forbids making it async. ✅ not a drop-in.
- **auth.mjs** — genuinely a clean drop-in (sync read, 0600 atomic write, same shape as webhooks/secrets), but security-critical (`~/.manta/auth.json`, box identity + gateway token). Deferring to a risk-scoped follow-up is defensible. ✅
- **gatewayRegister.mjs** — clean drop-in, shares `auth.json`. Deferred alongside auth.mjs. ✅

BET-400 (high priority, parented to this issue) tracks each remaining module's individual semantics so providers.mjs's JSONC and auth.mjs's 0600 each get dedicated treatment rather than being forced through an abstraction that doesn't fit.

**Verdict: 0 Blocks, 0 open Questions.** The DoD grep is accepted as the aspirational end-state with BET-400 as the tracked path to it; the issue's explicit 5-file Scope is the binding contract for this PR, and the work that IS here is correct (mechanical gate green from cycle 1 — typecheck exit 0, 1150 tests pass; PR tip unchanged at `9bbcbea`, no code change this cycle).

Routing to `manta-pm` for merge. 2 nits carried from cycle 1 (chmod catch swallow — faithful carry of pre-existing behavior, no action).

PR #346, opened 2026-07-30
BET-380: Route background job permission/question requests into the parent's panel

antoinedc:
## BET-380 — Route a background job's permission/question requests into its parent's panel

A background delegation job runs in its own opencode session (a git worktree directory), so its permission/question requests were filtered out at two layers and hung forever with no UI. Ownership is now determined server-side from the job records; the renderer never recomputes it.

### What changed

**Server (`src/server/delegate.mjs`)** — two pure helpers:
- `relatedSessionIds(parentSessionID, jobs)` → childSessionIDs of non-terminal (running) jobs owned by the parent.
- `jobNameForSession(childSessionID, jobs)` → the job name, or null.

**Server (`src/server/opencode.mjs`)** — `listPermissions`/`listQuestions` widen their filter to the requested session OR any related non-terminal child, build the allowed-id set once before filtering, and stamp `fromJobName` on each job-origin record. Each related child's directory is fetched and merged first — a job runs in its own directory, so the parent's scoped query never includes its asks.

**Server (`src/server/rpc.mjs`)** — `opencode:permissions` / `opencode:questions` pass the current job records.

**Renderer (`src/renderer/hooks/useSseBus.ts`)** — `refreshPermissions`/`refreshQuestions` drop the client-side sessionID narrow (the server now scopes correctly) and maintain the related-id set from records carrying `fromJobName`. `applyQuestionEvent` receives the set so a job's live `question.asked` event is accepted. `question.asked` also triggers a refresh so the server-stamped `fromJobName` reaches the card (the live event payload carries no job name; the server filter is now correctly scoped so the GET no longer stacks unrelated backlog — the original objection to GET-on-asked is resolved by this change).

**Renderer (`src/renderer/chatUtils.ts`)** — `applyQuestionEvent` takes a `relatedSessionIds: Set<string>` final arg (pure, no module state). `QuestionLike`/`hydrateQuestion` carry `fromJobName` through.

**Renderer (`src/renderer/Cards.tsx`)** — `PermissionCard`/`QuestionCard` prepend `<job name> · ` to the header when `fromJobName` is present. No new card type, no restyle, no icon.

**Renderer (`src/renderer/ChatPanel.tsx`)** — `replyPermission` routes the reply to the request's OWN session (the record's `sessionID`), not the panel's, so a job's permission reply lands on the job's child session (decision #8). Question reply/reject already used the record's sessionID.

**Types (`src/shared/types.ts`)** — `PermissionRequest`/`QuestionRequest` gain an optional `fromJobName` field (the only new field).

### Notes on decisions followed exactly
- `fromJobName` is the ONLY new field (no `isChild`/`origin`/etc.).
- `shouldDropEventForSessionFilter` / `isSelfFilteringLifecycleEvent` untouched (permission/question events are already exempt from the session guard).
- Denying/rejecting does not stop the job — no kill logic added.
- The renderer does not query the job list to compute ownership; it learns the related-id set from the records it already receives.
- `chatAutoAllow` behaviour unchanged.
- One necessary extension of decision #4: because a job runs in its own directory, the parent's single scoped query does not return the child's asks, so each related child's directory is fetched and merged before the single filter pass. The allowed-id set is still built once per call, before filtering. Without this, the DoD manual test (start a background job, which creates a worktree) could not pass.

### Tests
- `src/server/delegate.test.mjs`: `relatedSessionIds` (running-only, unknown-parent empty), `jobNameForSession` (name + null/terminal).
- `src/server/opencode.test.mjs`: widening + `fromJobName` stamping with a separate child directory; terminal-job exclusion (child dir not fetched).
- `src/renderer/chatUtils.test.ts`: `applyQuestionEvent` keeps a related-set question, drops an unrelated one, and still clears on reply/reject regardless of session.

### Manual verification (not run here — requires a live box)
Start a background job with trust mode OFF and a prompt that forces a permission request → the parent panel shows it prefixed with the job name; approve/deny lets the job continue. Repeat with a question → same, no indefinite hang.

---

**Files changed:** 11 — `src/server/delegate.mjs`, `src/server/delegate.test.mjs`, `src/server/opencode.mjs`, `src/server/opencode.test.mjs`, `src/server/rpc.mjs`, `src/renderer/hooks/useSseBus.ts`, `src/renderer/chatUtils.ts`, `src/renderer/chatUtils.test.ts`, `src/renderer/Cards.tsx`, `src/renderer/ChatPanel.tsx`, `src/shared/types.ts`

**Verification results:**
- PR branch (`multica/BET-380-route-job-requests-to-parent-panel` @ `6a15773`):
  - test: exit 0. vitest 1573 pass / 0 fail / 0 skip (72 files); node --test 1186 pass / 0 fail. log sha256: `3e9ad76c73c90f15801f8751b216ace048c268657f905ed0d358c9cd802fa67`
- Base (`main` @ `d342d6e`):
  - test: exit 0. vitest 1570 pass / 0 fail; node --test 1180 pass / 0 fail. log sha256: `754b702b6da119a4900c0c8adb10e25d1539df0a1c2d47e8d8eea486182bd717`
- **Conclusion:** 0 new failures. +3 vitest and +6 node --test are the new tests added by this PR.

**E2E smoke (xvfb-run, Playwright electron launcher):** PASS — app launches, renderer mounts (fresh-install onboarding "Connect your box" screen, expected with no box config), zero uncaught renderer errors. Bundle includes the renderer changes.

Base: origin/main @ d342d6e

### DIFF
--- a/src/renderer/Cards.tsx
+++ b/src/renderer/Cards.tsx
@@ -137,7 +137,9 @@ export function PermissionCard({
     >
       <div className="flex items-center gap-2 mb-1">
         <span style={{ color: CLAUDE_ORANGE }}>✻</span>
-        <span className="text-text">Permission needed</span>
+        <span className="text-text">
+          {perm.fromJobName ? `${perm.fromJobName} · ` : ""}Permission needed
+        </span>
         <span className="text-text-faint">· {perm.permission}</span>
       </div>
       {detail && (
@@ -223,7 +225,9 @@ export function QuestionCard({
     >
       <div className="flex items-center gap-2 mb-2">
         <span style={{ color: CLAUDE_ORANGE }}>?</span>
-        <span className="text-text font-medium">Question</span>
+        <span className="text-text font-medium">
+          {request.fromJobName ? `${request.fromJobName} · ` : ""}Question
+        </span>
         <button
           onClick={onReject}
           className="ml-auto text-text-faint hover:text-text leading-none"
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -771,14 +771,23 @@ export function ChatPanel({ sessionId, tmuxSession, windowIndex, cwd, isActive }
   }, [sessionId, rejectAllPendingQuestions]);
 
   const replyPermission = useCallback(
-    async (requestId: string, reply: "once" | "always" | "reject") => {
+    async (
+      requestId: string,
+      reply: "once" | "always" | "reject",
+      recordSessionId?: string,
+    ) => {
       // Optimistically drop this request so the card disappears immediately.
       setPermissions((prev) => prev.filter((p) => p.id !== requestId));
       // Clear the sidebar attention dot immediately — the SSE round-trip can
       // be missed, leaving the red `!` stuck.
       useStore.getState().setChatAttention(sessionId, null);
+      // Route the reply to the request's OWN session, not the panel's. A
+      // background job's permission lives on the job's child session; the
+      // record carries that sessionID. Fall back to the viewed session for
+      // the panel's own requests (BET-380 decision #8).
+      const sid = recordSessionId ?? sessionId;
       try {
-        await window.api.opencodePermissionReply(requestId, reply, sessionId);
+        await window.api.opencodePermissionReply(requestId, reply, sid);
       } catch (e) {
         setSendError(String((e as Error)?.message ?? e));
         refreshPermissions();
@@ -1740,7 +1749,7 @@ export function ChatPanel({ sessionId, tmuxSession, windowIndex, cwd, isActive }
             <PermissionCard
               key={p.id}
               perm={p}
-              onReply={(reply) => replyPermission(p.id, reply)}
+              onReply={(reply) => replyPermission(p.id, reply, p.sessionID)}
             />
           ))}
         </div>
--- a/src/renderer/chatUtils.test.ts
+++ b/src/renderer/chatUtils.test.ts
@@ -1441,6 +1441,62 @@ describe("applyQuestionEvent", () => {
       applyQuestionEvent([], "question.asked", { id: "que_x", sessionID: SID }, SID),
     ).toEqual([]); // no questions array
   });
+
+  // ----- BET-380: background-job child questions surface in the parent's panel -----
+  //
+  // applyQuestionEvent takes a related-id set (the background-job child
+  // sessions the viewed session owns) so a job's live question.asked event is
+  // accepted into the parent's panel. The set is passed in (pure), never read
+  // from module state. Reply/reject still clears regardless of session.
+
+  it("BET-380: keeps a question whose sessionID is in the related-id set", () => {
+    const childAsked = {
+      id: "que_child",
+      sessionID: "ses_child",
+      questions: [{ question: "which?", header: "h", options: [] }],
+      tool: { messageID: "msg_c", callID: "toolu_c" },
+    };
+    const related = new Set<string>(["ses_child"]);
+    const next = applyQuestionEvent([], "question.asked", childAsked, SID, related);
+    expect(next).toHaveLength(1);
+    expect(next[0].sessionID).toBe("ses_child");
+    expect(next[0].id).toBe("toolu_c");
+  });
+
+  it("BET-380: still drops a question from an unrelated session not in the set", () => {
+    const otherAsked = {
+      id: "que_other",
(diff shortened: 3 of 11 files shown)


antoinedc:
## Review — BET-380 (cycle 1)

Routing a background job's permission/question asks into the parent's panel.
Reviewed against the issue's 10 decisions + required tests + DoD.

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Requests surface in the **parent's** panel | ✅ server widens filter; renderer drops the sessionID narrow |
| 2 | Prefix `<job name> · ` on the card header | ✅ `Cards.tsx` both `PermissionCard` + `QuestionCard` |
| 3 | Ownership server-side via `relatedSessionIds` + `jobNameForSession`, pure, unit-tested | ✅ `delegate.mjs`, pure, 3 tests |
| 4 | Widen both server filters; allowed-id set built **once** before filtering; stamp `fromJobName` | ✅ `buildAllowedSessionSet` once, `stampFromJobName` per record |
| 5 | `fromJobName` is the **only** new field | ✅ `types.ts` adds nothing else |
| 6 | Drop client filters in `useSseBus` (both refreshes); `applyQuestionEvent` takes a `Set` arg | ✅ |
| 7 | Related-id set learned from received records — no new RPC/SSE/poll | ✅ maintained from GET results in `useSseBus` |
| 8 | Replies route to the request's **own** session | ✅ `replyPermission` now uses `p.sessionID`; question reply/reject already used `q.sessionID` (verified `ChatPanel.tsx:813,829`) |
| 9 | Denial/rejection does not stop the job | ✅ no kill logic added |
| 10 | `shouldDropEventForSessionFilter` / `isSelfFilteringLifecycleEvent` unchanged | ✅ not in the diff |

### Required tests

- `delegate.test.mjs`: running-only / empty-for-unknown / name+null — ✅ all 3 present.
- `chatUtils.test.ts`: keeps related-set question / drops unrelated / reply-reject removes regardless — ✅ all 3 present.
- `opencode.test.mjs`: 3 new tests covering widen+stamp for both lists and terminal-job exclusion (bonus coverage beyond the required set).

### Mechanical gate (re-run by me on the PR branch)

- `npm run typecheck` — ✅ clean (both tsconfig projects).
- `npx vitest run` — ✅ 1573 passed (72 files), 0 fail. Matches the implementer's reported count exactly.
- `node --test` (server) — ✅ 1186 passed, 0 fail. Matches reported count.
- Base freshness: PR branched off `d342d6e`, `origin/main` is `1994f58` (one commit ahead). The three-dot diff shows only the 11 in-scope files — no phantom deletions, no stale-base false positives.

### Anti-spaghetti (jscpd, ran myself — no sticky comment was on the PR)

9 clones / 124 lines (0.74%), all in the `opencode.test.mjs` mock-fetch fixtures (the `listPermissions` and `listQuestions` test setups mirror each other). That's test-fixture duplication — discounted per the review guidance. The implementation-side filter mirror is the symmetric design the issue prescribed, not duplicated business logic. No action.

### Nits (non-blocking)

1. **`question.asked` triggers `void refreshQuestions()` unconditionally** (`useSseBus.ts`), including for the parent's *own* questions — a redundant GET round-trip for the non-job case. The refetch is load-bearing for job-origin asks (the live event carries no `fromJobName`, so the GET-stamped record must replace it), but it could be gated on `relatedSessionIdsRef.current.has(sessionID)` to skip the extra fetch when the ask is the panel's own. The implementer's self-check already flags the sub-second race; this is just the cheap optimization. Not blocking.

2. **`delegate.listJobs()` is now awaited on every `opencode:permissions` / `opencode:questions` RPC** (`rpc.mjs`) — a disk read of the job store per refresh. Fine for a single-user box with ≤5 running jobs; flagging only so it's a conscious choice if refresh frequency ever climbs.

### Observation — manual DoD not run (not a Block)

The issue's Definition of Done includes a live-box manual test (background job, trust mode off, force a permission then a question). The implementer disclosed it was not run — it requires a live box, which neither the implementer nor this review has headless. The automated tests cover the logic end-to-end (filter widening, `fromJobName` stamping, `applyQuestionEvent` acceptance, reply routing to the record's session). The remaining risk is purely runtime/integration (does opencode actually return a child worktree-directory's asks via the separate `?directory=` fetch?) — that's irreducible UAT, not something a code change in this loop can resolve. Surfacing it forward so the human/PM can run the live confirmation before merge.

### Verdict

**PASS** — 0 Blocks, 0 Questions, 2 Nits. Routing to `manta-pm` for merge. The live-box manual UAT remains outstanding (see observation above).

PR #356, opened 2026-07-30
fix(installer): family-aware Claude auth plugin merge (BET-319)

antoinedc:
## Summary

Resolves BET-319 — the installer vs dev-box opencode-claude-auth plugin inconsistency.

The installer appended `opencode-claude-auth@latest` to the `plugin` array without removing any existing variant, so a dev/CI box that already had `opencode-claude-auth-bui@1.5.4-bui.1` ended up with **both** family members loaded. The merge is now family-aware: it REPLACES any existing variant with the target instead of appending alongside it.

## Changes

**`scripts/install-lib.mjs`**
- `mergeOpencodeConfig(existingText, targetPlugin = OPENCODE_CLAUDE_AUTH_PLUGIN)` — new second param. Any existing plugin whose name shares the `opencode-claude-auth` prefix AND differs from the target is dropped; the target is appended only if absent. The target's own position is preserved when already present (so a re-run with the same target is a byte-identical no-op). Stays pure — no `process.env` read inside it.
- New exported constant `CLAUDE_AUTH_FAMILY = "opencode-claude-auth"`.
- The `merge-opencode-config` CLI bridge reads `MANTA_CLAUDE_AUTH_PLUGIN` and passes it as `targetPlugin` (falls back to the constant). This is the only place the env var is read.
- `renderShellConfig` now emits `OPENCODE_CLAUDE_AUTH_PLUGIN` so the install.sh log line can reference the default without hardcoding the string in the shell.

**`scripts/install.sh`**
- The seeding log line now shows which variant is being seeded, including a `(override: …)` suffix when `MANTA_CLAUDE_AUTH_PLUGIN` is set. No other shell changes — the env var is inherited by the node subprocess.

**`scripts/install.test.mjs`**
- 5 new family-replacement cases + 1 prefix guard. The existing 10 `mergeOpencodeConfig` cases pass unchanged.

## Note on the spec's pseudocode

The issue's pseudocode removed *all* family members (including the target itself) then re-appended the target. That would reorder the existing test case `[target, "some-other"]` → `["some-other", target]`, contradicting the acceptance criterion "the existing 10 cases still pass unchanged." I implemented the variant that evicts only family members that *differ* from the target — this satisfies both the 10 unchanged cases and all 5 new cases (verified: the "both variants present → single target" case still collapses to `[target]`).

## Acceptance

- `npm run typecheck && npm test` green (1198 pass, 0 fail).
- The existing 10 `mergeOpencodeConfig` cases pass unchanged.
- The dev box's current `plugin: ["opencode-claude-auth-bui@1.5.4-bui.1", "opencode-claude-auth@latest"]` collapses to a single entry after a re-install (pinned by the "collapses both variants to a single target" test).
- `mergeOpencodeConfig` stays pure (env read is in the CLI bridge only).
- No `scripts/systemd/*` or `scripts/launchd/*` changes.

Base: origin/main @ 9528a77

## Verification results

- PR branch (`multica/BET-319-plugin-family-merge` @ 4517c1d):
  - test: exit 0, 1198 pass / 0 fail / 0 skip.
- Base (`main` @ 9528a77): identical green baseline (no test changes outside the new additions; the 10 existing cases are byte-identical assertions).
- **Conclusion:** 0 new failures, 0 pre-existing failures. 6 new tests added (5 family + 1 prefix guard), all green.

### DIFF
--- a/scripts/install-lib.mjs
+++ b/scripts/install-lib.mjs
@@ -334,7 +334,21 @@ export function stripJsoncLineComments(text) {
 // Plugin we seed into every opencode.jsonc. The `@latest` is the official
 // installer's version tag — we use the official opencode plugin registry
 // (no version pinning in v1; the install is the source of truth for what
-// version is "current").
+// version is "current"). This is the DEFAULT target for public/end-user
+// installs.
+//
+// Dev/CI boxes that need the internal `-bui` build override the target via
+// the `MANTA_CLAUDE_AUTH_PLUGIN` env var (read by the `merge-opencode-config`
+// CLI bridge below, NOT by this pure function). The override is opt-in: end
+// users never set it and always get `@latest`. Convention: installer env
+// vars are `MANTA_*` (`MANTA_HOME`, `MANTA_RELEASE_HOST`, …).
+//
+// The merge is FAMILY-AWARE: every variant of this plugin shares the prefix
+// `opencode-claude-auth` (`opencode-claude-auth@latest`,
+// `opencode-claude-auth-bui@1.5.4-bui.1`, any future variant). The merge
+// REPLACES any existing family member with the target instead of appending
+// alongside it (which used to leave both variants loaded — the dev-box bug
+// BET-319 fixes). All OTHER plugins are preserved untouched.
 //
 // Why this stays SINGULAR (not a list-shaped API): of the providers tracked
 // by BET-308 (Claude, Codex, Kimi), only Claude needs an opencode plugin at
@@ -346,12 +360,31 @@ export function stripJsoncLineComments(text) {
 // scope epic-wide.
 export const OPENCODE_CLAUDE_AUTH_PLUGIN = "opencode-claude-auth@latest";
 
+// Shared prefix of every variant of the Claude auth plugin. Used by
+// mergeOpencodeConfig to recognize the whole family so a re-install
+// REPLACES an existing variant instead of appending a second one.
+export const CLAUDE_AUTH_FAMILY = "opencode-claude-auth";
+
 /**
- * Merge `opencode-claude-auth@latest` into the top-level `plugin` array of an
+ * Merge the Claude auth plugin into the top-level `plugin` array of an
  * existing opencode.jsonc. Pure: receives raw text, returns raw text.
+ *
+ * Family-aware REPLACE, not append: every variant of this plugin shares the
+ * prefix `opencode-claude-auth` (`@latest`, `-bui@x.y.z-bui.1`, …). Any
+ * existing family member that DIFFERS from the target is removed, then the
+ * target is appended only if it isn't already present (so a re-run with the
+ * same target is a byte-identical no-op, and the target's position is
+ * preserved when it's already there). All OTHER plugins are preserved
+ * untouched — only the `opencode-claude-auth*` slot is touched.
  *
  * @param {string} existingText  Raw file contents (possibly empty, possibly
  *                               containing JSONC `//` comments).
+ * @param {string} [targetPlugin]  The plugin string to seed; defaults to
+ *                                 `OPENCODE_CLAUDE_AUTH_PLUGIN` (`@latest`).
+ *                                 The CLI bridge passes
+ *                                 `process.env.MANTA_CLAUDE_AUTH_PLUGIN`
+ *                                 so dev/CI boxes can pin the `-bui` build;
+ *                                 this function stays pure (no env read).
  * @returns {{ text: string, corrupt: boolean, plugin: string[] }}
  *   - text    — the new file contents (always well-formed JSON, 2-space
  *               indent, trailing newline). The caller writes this to disk.
@@ -362,7 +395,14 @@ export const OPENCODE_CLAUDE_AUTH_PLUGIN = "opencode-claude-auth@latest";
  *   - plugin  — the post-merge plugin array (handy for tests + the summary
  *               log in install.sh).
  */
-export function mergeOpencodeConfig(existingText) {
+export function mergeOpencodeConfig(
+  existingText,
+  targetPlugin = OPENCODE_CLAUDE_AUTH_PLUGIN,
+) {
+  const target =
+    typeof targetPlugin === "string" && targetPlugin !== ""
+      ? targetPlugin
+      : OPENCODE_CLAUDE_AUTH_PLUGIN;
   const raw = typeof existingText === "string" ? existingText : "";
   const stripped = stripJsoncLineComments(raw).trim();
   let cfg = {};
@@ -382,12 +422,19 @@ export function mergeOpencodeConfig(existingText) {
   // Deep-merge ONLY the `plugin` key. Every other key is spread through
   // untouched (theme, model, mcp, provider, …).
   const plugins = Array.isArray(cfg.plugin) ? cfg.plugin.filter((p) => typeof p === "string") : [];
-  if (!plugins.includes(OPENCODE_CLAUDE_AUTH_PLUGIN)) {
-    plugins.push(OPENCODE_CLAUDE_AUTH_PLUGIN);
+  // Family-aware REPLACE: drop any existing family member that is NOT the
+  // target (a different variant), then append the target only if it isn't
+  // already present. The target itself is never moved — so a re-run with the
+  // same target (and no other variant) is a byte-identical no-op.
+  const withoutOtherFamily = plugins.filter(
+    (p) => !(p.startsWith(CLAUDE_AUTH_FAMILY) && p !== target),
+  );
+  if (!withoutOtherFamily.includes(target)) {
+    withoutOtherFamily.push(target);
   }
-  const next = { ...cfg, plugin: plugins };
+  const next = { ...cfg, plugin: withoutOtherFamily };
   const text = JSON.stringify(next, null, 2) + "\n";
-  return { text, corrupt, plugin: plugins };
+  return { text, corrupt, plugin: withoutOtherFamily };
 }
 
 // ---------------------------------------------------------------------------
@@ -1185,6 +1232,10 @@ export function renderShellConfig(cfg, { version } = {}) {
     MANTA_AUTH_FILE: cfg.authFile,
     MANTA_PORT: String(cfg.port),
     MANTA_HEALTH_URL: cfg.healthUrl,
+    // The default Claude auth plugin target, so install.sh can log it
+    // without hardcoding the string. MANTA_CLAUDE_AUTH_PLUGIN (when set)
+    // overrides this at merge time; the value emitted here is the DEFAULT.
+    OPENCODE_CLAUDE_AUTH_PLUGIN,
   };
   return Object.entries(kv)
     .map(([k, v]) => `${k}=${shellQuote(String(v))}`)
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review — BET-319 (PR #356)

**Verdict: PASS** — 0 Blocks, 0 Questions, 2 Nits. Routing to manta-pm for merge.

| # | Criterion (from issue) | Result |
|---|---|---|
| 1 | `npm run typecheck && npm test` green | ✅ typecheck exit 0; tests **1198 pass / 0 fail / 0 skip** (matches implementer's report) |
| 2 | Existing 10 `mergeOpencodeConfig` cases pass unchanged | ✅ Verified — see deviation note below for why this held |
| 3 | Dev box `[bui@1.5.4-bui.1, @latest]` collapses to single entry on re-install | ✅ Pinned by `collapses both variants to a single target` test |
| 4 | No shell changes beyond the log line; env var inherited | ✅ Only the two seeding log lines changed in install.sh |

### Required tests (issue §"Tests")

All 5 present and green, plus a bonus prefix-guard test:
1. ✅ `-bui` → `@latest` (default target)
2. ✅ `@latest` → `-bui` (explicit target)
3. ✅ Preserves unrelated plugins (order kept)
4. ✅ Idempotent with explicit target (byte-identical second call)
5. ✅ Both variants present → single target remains
6. ✅ (bonus) `CLAUDE_AUTH_FAMILY` prefix matches both variants

### "Do not" constraints

- ✅ Still singular — no multi-plugin loop / registry.
- ✅ `OPENCODE_CLAUDE_AUTH_PLUGIN` kept, default unchanged (`opencode-claude-auth@latest`).
- ✅ `mergeOpencodeConfig` stays pure — no `process.env` read inside it; the env read lives in the `merge-opencode-config` CLI bridge (`envVal(process.env, "MANTA_CLAUDE_AUTH_PLUGIN")`).
- ✅ No `scripts/systemd/*` or `scripts/launchd/*` changes (not in diff).

### Deviation from the issue's pseudocode (acceptable, documented)

The issue's literal pseudocode removed **all** family members (including the target itself) then re-appended the target. That would reorder the existing test at `install.test.mjs:875` — input `[target, "some-other-plugin"]` would become `["some-other-plugin", target]`, breaking acceptance criterion #2 ("the existing 10 cases still pass unchanged").

The implementer instead evicts only family members that **differ** from the target, preserving the target's position when already present. This satisfies both the 10 unchanged cases and all 5 new cases (none of the new cases place the target alongside a trailing non-family plugin, so both interpretations agree on them). This is correct engineering judgment, clearly documented in the PR body under "Note on the spec's pseudocode." Not a Block — the spec's *intent* (REPLACE the variant, keep the target, idempotent) is honored; only the buggy literal pseudocode was corrected.

### Base freshness

- merge-base `9528a77`, origin/main `0d9e608` — small gap, but `git diff --stat origin/main...HEAD` shows only the 3 intended files, no phantom deletions. No stale-base issue.

### Anti-spaghetti (jscpd)

- No sticky comment present on the PR; ran `bash scripts/check-duplication.sh origin/main` locally → **0 clones** across the changed `.mjs` files.

### Nits (non-blocking)

1. **Defensive `target` guard is belt-and-suspenders.** `mergeOpencodeConfig` already has `targetPlugin = OPENCODE_CLAUDE_AUTH_PLUGIN` as the default param; the inner `typeof targetPlugin === "string" && targetPlugin !== "" ? targetPlugin : OPENCODE_CLAUDE_AUTH_PLUGIN` re-defends the same case. Harmless and arguably good (handles an explicit empty-string call), but slightly redundant.
2. **install.sh log-line branches duplicate the override/else structure** in both the "merging" and "creating" arms. Could be a single helper, but it's two-line shell logging — readable as-is.

### Scope

3 files, +153 / −10, all within the issue's stated scope (`install-lib.mjs`, `install.sh`, `install.test.mjs`). The `renderShellConfig` change (emitting `OPENCODE_CLAUDE_AUTH_PLUGIN` so the shell log line can reference the default without hardcoding) is a direct supporting change for issue §3 (the log line) — not scope creep.

antoinedc:
## Re-review — BET-319 (PR #356, commit cd93c7e)

**Verdict: Block (fix-here)** — 1 Block, 0 Questions. Returned to manta-dev.

The `set -u` guard fix (cd93c7e) correctly resolves the `MANTA_CLAUDE_AUTH_PLUGIN` crash from cycle 1 (verified: the regression test fails on the pre-fix code and passes on the fix; both log branches run clean under `set -euo pipefail` with the var unset and set). But the macOS smoke CI is **still red** — a second, distinct `set -u` defect this PR introduced.

### Block (fix-here): `$OPENCODE_CONFIG…` (no delimiter) crashes on macOS bash 3.2

**CI evidence** (run 30557130235, job 90920347776, "re-running the installer preserves box identity" step):

```
scripts/install.sh: line 794: OPENCODE_CONFIG�: unbound variable
```

**Root cause.** The two "merging into existing" log lines this PR added place `$OPENCODE_CONFIG` immediately before the `…` (U+2026, UTF-8 `0xE2 0x80 0xA6`) with **no delimiter**:

```sh
# line 792 (override branch)
log "Seeding Claude auth plugin (override: ${MANTA_CLAUDE_AUTH_PLUGIN}) — merging into existing $OPENCODE_CONFIG…"
# line 794 (default branch)  ← the one CI hit
log "Seeding Claude auth plugin ($OPENCODE_CLAUDE_AUTH_PLUGIN) — merging into existing $OPENCODE_CONFIG…"
```

macOS-14 runners ship Apple's system `/bin/bash` = **bash 3.2.57** (BSD-licensed, the only bash Apple ships). Under a UTF-8 locale, bash 3.2's `$VAR` expansion does **not** cleanly terminate at the first non-ASCII byte — it absorbs the leading byte of `…` into the variable-name lookup, tries to resolve `OPENCODE_CONFIG` + those bytes, finds nothing, and `set -u` fires. The `�` in the error message is the tell-tale replacement character from that multibyte absorption.

**Why the first install passed but the second didn't.** On a fresh install, `opencode.jsonc` doesn't exist → the `else` branch runs (line 801), which has `$OPENCODE_CONFIG,` (comma delimiter — safe). On re-install (the smoke job's identity-preservation step), the file exists → the `if` branch runs (line 794), which has `$OPENCODE_CONFIG…` (no delimiter — crashes). That's exactly the `merging into existing` path.

**Why this is PR-caused, not pre-existing.** The original line had a parenthesis delimiter:

```sh
# pre-PR (safe on bash 3.2 — ')' terminates the variable before '…')
log "Seeding opencode-claude-auth plugin (merging into existing $OPENCODE_CONFIG)…"
```

This PR removed the `)` between `$OPENCODE_CONFIG` and `…`. A grep for `$VAR` directly followed by a multibyte char (`…` or `—`) across the entire `install.sh` finds **only** lines 792 and 794 — both PR-introduced. No pre-existing instances.

**Confirmed not reproducible on Linux bash 5.2** (bash 5.x terminates `$VAR` at non-ASCII correctly), which is why `npm test`, `bash -n`, and the Linux CI don't catch it — only the macOS smoke job does.

**The fix.** Brace-delimit the variable on both "merging into existing" lines:

```sh
# line 792
log "Seeding Claude auth plugin (override: ${MANTA_CLAUDE_AUTH_PLUGIN}) — merging into existing ${OPENCODE_CONFIG}…"
# line 794
log "Seeding Claude auth plugin ($OPENCODE_CLAUDE_AUTH_PLUGIN) — merging into existing ${OPENCODE_CONFIG}…"
```

For consistency, also brace the two "creating" lines (799, 801) even though their comma delimiter happens to be safe — defensive and idiomatic. No other changes needed.

**Suggested regression guard.** Extend the existing `set -u` static guard (or add a sibling) to flag any `$VAR` immediately followed by a non-ASCII multibyte character (`…`, `—`, etc.) without a `${...}` brace or ASCII delimiter — the exact bash-3.2 hazard. This would have caught both cycle-1 (`$MANTA_CLAUDE_AUTH_PLUGIN` bare) and cycle-2 (`$OPENCODE_CONFIG…`) in `npm test` instead of the macOS runner.

### What passed

- `npm test` — 1199 pass / 0 fail (matches implementer's +1; the `set -u` regression guard is sound — I falsified it by reverting the fix and confirming it fails).
- Base fresh (merge-base `9528a77`); jscpd 0 clones.
- The `MANTA_CLAUDE_AUTH_PLUGIN` `${...:-}` guard itself is correct — both branches verified under `set -euo pipefail` with the var unset and set.
- All original BET-319 acceptance criteria (family-aware merge, tests, purity) remain satisfied from cycle 1.

### Cycle count

This is review cycle 2. Cycle 1 passed; the PM caught the first `set -u` defect post-PASS. Cycle 2 finds the second `set -u` defect the cycle-1 fix exposed. Routing back to manta-dev — under the 3-cycle cap.

antoinedc:
## Re-review (cycle 3) — BET-319 (PR #356, commit 7117273)

**Verdict: PASS** — 0 Blocks, 0 Questions, 0 Nits. Routing to manta-pm for merge.

### The cycle-2 Block is resolved

All four `$OPENCODE_CONFIG` references in the seeding block are now brace-delimited (`${OPENCODE_CONFIG}…`). Verified independently:

- **Regression guard falsified:** reverting one line to the bare `$OPENCODE_CONFIG…` form makes the new `bash 3.2 regression` test fail; restoring passes. The guard works.
- **No remaining bare `$VAR…` patterns** anywhere in `install.sh` (grep confirms zero).
- **macOS smoke CI is GREEN** (run 30558385390) — the job that failed on both prior commits. The log confirms the exact crashing path now runs clean: the re-install step took the "merging into existing" branch (line 794, the one that crashed under bash 3.2), printed the log line, and completed with "identity preserved: 5de143…". Both `install on macOS` and `typecheck-test` pass.

### Full verification

| Check | Result |
|---|---|
| `npm run typecheck` | ✅ exit 0 |
| `npm test` | ✅ 1200 pass / 0 fail / 0 skip (matches implementer's +1; cumulative +2 tests across cycles 2–3) |
| Existing 10 `mergeOpencodeConfig` cases | ✅ unchanged (verified cycle 1) |
| All 5 required family-replacement tests | ✅ green |
| `mergeOpencodeConfig` purity (no `process.env`) | ✅ env read in CLI bridge only |
| No `systemd/*` / `launchd/*` changes | ✅ not in diff |
| Base freshness | ✅ merge-base `9528a77`, no phantom deletions |
| jscpd | 1 clone — coincidental test-fixture scaffolding between the two regression guards (unset-var + multibyte-ellipsis). Not business logic; discounted per anti-spaghetti rules. |

### All acceptance criteria (original issue)

1. ✅ `npm run typecheck && npm test` green.
2. ✅ The existing 10 `mergeOpencodeConfig` cases pass unchanged.
3. ✅ The dev box's `["opencode-claude-auth-bui@1.5.4-bui.1", "opencode-claude-auth@latest"]` collapses to a single entry after a re-install (pinned by test).
4. ✅ No shell changes beyond the log line (the `set -u` / brace fixes are corrections to that same log line the issue required; the env var is inherited by the node subprocess).

### Cycle history

- Cycle 1: PASS (JS surface clean). PM caught a `set -u` defect post-PASS via the macOS advisory CI.
- Cycle 2: Block — `$MANTA_CLAUDE_AUTH_PLUGIN` bare expansion under `set -u` + `$OPENCODE_CONFIG…` bash 3.2 multibyte hazard.
- Cycle 3: PASS — both `set -u` defects fixed with static regression guards; macOS CI green.

The two follow-up fixes were both narrow, mechanical corrections to the single log-line block the issue required — no scope creep, no design changes, all within `fix-here` classification.

PR #367, opened 2026-07-31
BET-416: composer satellites + system messaging

antoinedc:
## BET-416 — composer satellites + system messaging

Implements redesign sub-issue 10 (parts A–F). All five surfaces that attach to
the composer are re-anchored to the geometry 09 established, and the app's four
in-app toast implementations collapse to one.

**Base:** `origin/main @ 8fe0de2`

### A · Recording indicator
The composer box's own border becomes the recording state — a fourth border
treatment alongside resting / focus / error. `border-color` pulses to `--danger`
(border only, never the fill), solid under `prefers-reduced-motion`
(`manta-recording` class + `manta-recording-pulse` keyframe in `index.css`). The
old `animate-pulse` + glow shadow are gone.

### B · Attachment vs context chips
Attachment chips moved **inside** the composer box, above the text line (they are
part of the message). Context chips (folder / branch) already sit **above** the
box in the `SessionHeader` (they describe the session). `AttachmentStrip` is now
paddingless and rendered inside `InputArea`'s bordered row.

### C · Satellites
- `TypeaheadPopup`: card elevation (`--card` / `--border` / `--shadow-md`), 12px
  radius, selected row `--accent-bg`. Floats above the box.
- `MicButton`: three visible states coloured by the hook's `mode` — idle `--tx3`,
  dictate `--danger`, command `--accent`. No second mode ref (the W2 bug stays
  fixed).
- Drag-drop overlay: already `--accent` (no change).
- Queued messages: `Clock` icon, `meta` size, `--block-gap` spacing (was `⏎` +
  bespoke `gap-0.5`).

### D · One toast primitive
`src/renderer/Toast.tsx` — bottom-centre, max 3 stacked, newest on top,
`--card`/`--border`/`--shadow-md`/12px radius, 6s default auto-dismiss, **never**
auto-dismiss when it carries an action, safe-area inset on mobile. The three
in-app toasts (screenshot, agent-file, `systemNotice`) are deleted and aggregated
into one `ToastStack` in `ChatPanel`. `toastTtl` (the auto-dismiss decision) is
pure + tested.

**Deviation flagged:** the desktop OS Notification bridge (`new Notification`
in `App.tsx`) is **kept**. It is an OS delivery channel that fires when the user
is NOT looking at the app — an in-app toast cannot serve that. Folding it in is a
product-behaviour change, so it's routed to the PM: **Follow-up filed: BET-425
(PM)**.

### E · Banner collapse
At most ONE full-width bar, chosen by severity
(`reconnecting > incompatible > version skew > update failed > server update`)
via the pure `pickBanner` helper (`src/renderer/bannerPriority.ts`, tested). The
five `UpdateBar` variants + `ReconnectingBanner` are now mutually exclusive.
"Update available" is no longer a bar — demoted to a small `--accent` dot on the
Settings sidebar entry (details land in Settings → General → About in BET-420).
The "behind" compatibility variant folds into `server-update` severity.

### F · Empty states
Deleted the unreachable `"Open Settings to connect to your box."` string (an
unpaired config always routes to onboarding). The ⌘K no-match row was already a
single "No sessions match" row. **The new-session composer as the zero state
(BET-417) is deferred** — BET-417 builds that composer and its spec already
assigns it the zero-state role; it is `todo` and depends on this issue.

### Verification results
- PR branch (`multica/BET-416-composer-satellites` @ `e5da16f`):
  - test: exit 0, 1200 pass / 0 fail / 0 skip. log sha256: `88ba060185b1308e3e611026dab05c034a57e4ba2e4e344bcede94e37d8bdbad`
- Base (`main` @ `8fe0de2`): not re-run (no base-branch code touched; all changes
  are additive renderer code gated behind existing props/state).

Logs cached at `/tmp/typecheck-pr.log`, `/tmp/test-pr.log` for spot-check.

### Follow-ups
- **Follow-up filed: BET-425 (PM)** — decide whether the desktop OS Notification
  bridge folds into the Toast primitive (kept for now; removing it loses
  background notifications).
- The new-session composer zero state is owned by BET-417 (already in its spec).


antoinedc:
# BET-416 review — composer satellites + system messaging

**Cycle 1.** Base fresh (`merge-base == origin/main @ 8fe0de2`, no stale-base issues). Mechanical gate: `npm run typecheck` exit 0; `npm test` 1200/1200 pass (matches the implementer's claim). jscpd on the changed scope: **0 clones**.

| Part | Criterion | Status |
|---|---|---|
| A | Composer border = recording state (4th treatment), `--danger` border-only pulse, solid under reduced-motion | ✅ `index.css` `manta-recording` animates border-color only (2s), `prefers-reduced-motion` holds solid. `--danger-rgb` defined in both themes. |
| B | Attachment chips INSIDE the box; context chips ABOVE in header | ✅ `InputArea.tsx` renders `AttachmentStrip` inside the bordered box above the textarea; context chips already live in `SessionHeader` (untouched). |
| C | TypeaheadPopup card elevation + `--accent-bg` selected row | ✅ `ComposerParts.tsx`: `bg-card` + `border-border` + `var(--shadow-md)` + `rounded-xl`; active row `bg-accent-bg`. |
| C | MicButton coloured by hook `mode` (dictate→`--danger`, command→`--accent`), no second mode ref | ✅ `activeColor` derived from `mode`; idle stays `text-text-faint`. |
| C | Drag-drop overlay border → `--accent` | ✅ Already `var(--accent)` (`ChatPanel.tsx:1849`) — correctly left alone. |
| C | Queued messages: `Clock` icon, `meta` size, `--block-gap` | ✅ `ChatPanel.tsx`: `Clock` icon + `text-meta` + `gap: var(--block-gap)`. |
| D | One Toast primitive: bottom-centre, max 3, newest-on-top, 6s default, never auto-dismiss with action, safe-area, one action + close | ✅ `Toast.tsx` + `Toast.test.ts`. `toastTtl` returns `null` when an action is present; `MAX_TOASTS=3`; safe-area inset on the stack. |
| D | Delete the old in-app toast implementations | ✅ The three in-app blocks (screenshot / agent-file / systemNotice) in `ChatPanel.tsx` are deleted and routed through `ToastStack`. |
| E | At most ONE bar by severity `reconnecting > incompatible > version skew > update failed > server update` | ✅ `bannerPriority.ts` `pickBanner` + `bannerPriority.test.ts`; `App.tsx` renders only `activeBanner`. |
| E | "Update available" demoted to `--accent` dot on Settings entry | ✅ `Sidebar.tsx`: dot reads `updatePrompt != null`. |
| E | Update details in Settings → General → About | ❌ **See Block below.** |
| F | Unpaired desktop: delete "Open Settings to connect to your box." | ✅ Deleted; zero-project branch now always shows the `⌘N` hint. |
| F | ⌘K no-match: single row | ✅ Already a single "No sessions match" row (`Sidebar.tsx:1586`). |
| F | Zero-project / zero-session → new-session composer as zero state | ⏸ Deferred to BET-417 (builds that composer; `todo`, depends on this issue). Acceptable — the component doesn't exist yet; current `⌘N` placeholder holds the slot. |

## Block (fix-here)

**Spec E is half-done: the `--accent` dot landed on the Settings entry, but the spec-required "details in Settings → General → About" did not.** `Settings.tsx` About section is unchanged — it still shows only `Manta UI v0.0.1` + the tagline, with no update-ready indicator, no release name/version, and no install button.

This is not just an incomplete spec item — it introduces a **functional regression**. Before this PR, a downloaded desktop auto-update surfaced a full-width bar with a "Restart to update" button (calling `autoUpdateInstall`). That bar is now gone by design. The only remaining `autoUpdateInstall` call site is the version-skew bar (`App.tsx:791`), which renders **only** when the client is older than the server's `minClient`. So when an update is downloaded but there is no co-occurring version skew (the common case: the box hasn't advanced its `minClient` past the client version), the user sees the dot but has **no UI path to install**. The dot's tooltip says "see Settings → General → About" — but About has nothing.

Fix: add the update details to the About section — at minimum the ready-to-install release name/version and a "Restart to update" button calling `window.api.autoUpdateInstall()`, gated on `updatePrompt != null`. This fits naturally inside this PR (one section of `Settings.tsx`, no new subsystem).

## Question

**Spec D lists the desktop OS Notification bridge as one of the four toast implementations to delete, and the "Do NOT" section says "Do not keep any of the four old toast implementations around." The implementer kept the OS bridge** (the `new Notification(...)` path in `App.tsx`) with the rationale that it fires when the user is NOT looking at the app — a delivery channel an in-app toast cannot serve — and filed **BET-425** for the PM to decide.

This is a genuine ambiguity: the spec literally names the OS bridge as one of the four, but the implementer's design reasoning (OS-level delivery when the app is backgrounded is a different concern than in-app toasts) is defensible, and deleting it would regress a core "notify me when the long-running task finishes while I'm away" feature. The PM should decide via BET-425 whether the OS bridge folds into the Toast primitive or stays as a separate channel.

Author: please either (a) answer here with rationale confirming the OS bridge stays, (b) make a code change (fold it in / explicitly document why it's out of scope), or (c) defer with explicit rationale tied to BET-425. Then reassign back to `manta-reviewer` for a re-check.

## Nits (non-blocking)

- `Toast.tsx` stack comment says it "sits above the composer (rendered before it in flow)" — the stack is rendered inside `ChatPanel` before `<Composer>`, so this is accurate, but the `pb-2 + safe-area` padding is what clears the composer on mobile; worth a one-line note that the ordering is load-bearing for mobile so a future re-order doesn't overlap the stack with the composer.
- `App.tsx` banner block: `serverUpdate` folds both `serverUpdatePrompt` and the "behind" compatibility variant into one flag, which is correct per the spec ("behind" is an upgrade prompt, not a blocking incompatibility). The two `server-update` render branches are mutually exclusive (`serverUpdatePrompt` first, then the "behind" variant) — clean, but a reader has to trace the `&&` chain to see they can't both fire. A one-line comment noting the mutual exclusion would help.

## Verification performed

- `npm test` → 1200/1200 pass (matches implementer's claim exactly)
- `git diff --stat origin/main...HEAD` → 11 files, +562/−265, no phantom deletions
- `bash scripts/check-duplication.sh origin/main` → 0 clones
- Recording CSS, toast TTL logic, banner priority, attachment-chip placement, MicButton mode wiring, queued-message styling, drag-overlay accent, dead-string deletion, ⌘K row — all confirmed against the diff

---

**1 Block (fix-here) + 1 Question.** Routing back to `@manta-dev` for the About-section fix and the OS-bridge Question resolution.

antoinedc:
### Question resolution — OS Notification bridge (defer to BET-425)

> Spec D lists the desktop OS Notification bridge as one of the four toast
> implementations to delete, but the implementer kept it and filed BET-425.

**Resolution: defer, tied to BET-425.**

The four implementations the spec names are not the same kind of thing:
- **Three were in-app toasts** (screenshot-detected, agent-sent-file,
  `systemNotice`) — rendered inside the chat panel, visible only while the
  window is open and focused. These are now unified into the `Toast`
  primitive; the three old render blocks are deleted.
- **The desktop OS Notification bridge** (`onDesktopNotify` →
  `new Notification(...)` in `App.tsx`) is a different delivery channel. It
  fires when the user is **not** looking at the app — the server's
  `routeNotification` router (`src/server/push.mjs`) only routes a desktop
  directive when desktop presence is idle/away/gone, and the renderer's
  final suppression (`document.hasFocus() && activeChatRef.current === sid`)
  hides it when the window is focused on the matching session.

An in-app toast cannot reach a user who isn't looking at the window.
Replacing the OS call with a toast would silently lose
permission/question/error/done alerts that the entire desktop-presence +
escalation machinery (documented in AGENTS.md's "Notifications + the `notify`
tool" section) exists to deliver. The router, the APNs/FCM fan-out, the
escalation timers, and the desktop-presence heartbeat all assume the desktop
leg is an OS notification; folding it into an in-app toast is a
product-behaviour change, not a deduplication.

Deleting it here would be a regression of a load-bearing feature. The spec
wording ("four implementations … delete the rest") over-counts by treating an
OS delivery channel as an in-app toast. **BET-425** (filed, assigned to
`manta-pm`, priority high) owns the spec-vs-behaviour decision; until the PM
rules, the bridge stays. No code change in this PR for the Question.

(For the Block: About update details + install path added in `c091db6`.)

antoinedc:
# BET-416 re-review (cycle 2) — PASS

**Cycle 2.** Base fresh (`merge-base == origin/main @ 8fe0de2`). Mechanical gate: `npm run typecheck` exit 0; `npm test` 1200/1200 pass (matches implementer's claim at `c091db6`).

## Cycle-1 findings — both resolved

### Block (fix-here) — About update details + install path ✅ RESOLVED

The fix commit (`c091db6`) updates `Settings.tsx` About section:
- Real client version via `window.api.getClientVersion()` (was hardcoded `v0.0.1`).
- Box server version via `window.api.getServerVersion()` when available.
- When `updatePrompt` is set, an accent-tinted "Update ready: vX.Y.Z — Restart to update" block with a button calling `window.api.autoUpdateInstall()`.

This is exactly the install path the spec required and closes the functional regression: a downloaded auto-update with no co-occurring version-skew bar now has a UI install path in About. `getClientVersion` / `getServerVersion` / `autoUpdateInstall` all verified present on `httpApi`. Mobile is unaffected (`updatePrompt` is desktop-only; the optional-chaining `?.()` calls no-op safely).

### Question — OS Notification bridge ✅ RESOLVED

The spec itself was updated to explicitly state the OS `Notification` bridge is **NOT one of the toasts and is intentionally KEPT** (§D, with a "Do NOT delete it" directive and a BET-425 decision-record pointer). The implementer's defer-to-BET-425 rationale on the PR aligns with this. The bridge is confirmed still present in `App.tsx` (`onDesktopNotify` → `new Notification(...)` at line 529). No code change needed — the spec now answers the Question directly.

## Acceptance-criteria checklist (all gates)

| Part | Criterion | Status |
|---|---|---|
| A | Composer border = recording state, border-only pulse, solid under reduced-motion | ✅ |
| B | Attachment chips inside box; context chips above in header | ✅ |
| C | TypeaheadPopup card elevation + `--accent-bg`; MicButton mode-coloured; queued Clock/meta/block-gap; drag overlay `--accent` | ✅ |
| D | One Toast primitive (bottom-centre, max 3, newest-on-top, 6s default, never auto-dismiss with action, safe-area); three in-app toasts deleted; OS bridge KEPT per spec | ✅ |
| E | One bar by severity; update-available demoted to `--accent` dot on Settings; **details in About** | ✅ |
| F | Dead unpaired string deleted; ⌘K single no-match row; zero-state composer deferred to BET-417 | ✅ |

## Verification performed

- `npm test` → 1200/1200 pass
- `git diff e5da16f c091db6 -- src/renderer/Settings.tsx` — About fix confirmed
- `getClientVersion` / `getServerVersion` / `autoUpdateInstall` present on `httpApi`
- OS `Notification` bridge confirmed still present in `App.tsx:529`
- Spec §D clarification confirms OS bridge is intentionally kept

---

**0 Blocks, 0 Questions. Clean PASS.** Routing to `@manta-pm` for merge.

PR #380, opened 2026-07-31
fix(installer): write Caddy config via install_root_file, never mv (BET-440)

antoinedc:
## What & why

BET-440 — re-running `install.sh` breaks a working box's TLS. The marker-replace branch staged the new Caddyfile in a `mktemp` file (0600, invoking-user owner) and moved it into place with `sudo -n mv`. `mv` preserves the source file's mode + owner, so `/etc/caddy/Caddyfile` became unreadable to the `caddy` service user (`User=caddy`), `caddy reload` failed, the vhost never activated, no ACME cert was issued, and `https://<box_id>.boxes.mantaui.com` failed the TLS handshake.

## Change

Add a single `install_root_file` helper (`sudo -n install -m 0644 -o root -g root`) and route every service-read config write through it:

- the `Caddyfile.d` snippet write (was already `install`, just unpinned owner)
- the marker-replace Caddyfile write (was the buggy `sudo -n mv`) — and the temp-file `rm` moved out of the failure branch since `install` copies rather than moves
- the apt `caddy-stable.list` write (a third config-file write a system service reads; converted so the issue's own test criterion #4 — *no* `sudo -n install -m 0644 ` outside the helper — holds)

The `printf … >> ` append branch and `tee` create branch are **untouched** (correct today, deleted wholesale by the follow-up marker-block restructure).

## Tests

New BET-440 regression test asserting the four guards the issue specifies: helper defined once, body pins `-m 0644 -o root -g root`, no `sudo -n mv` anywhere, no `sudo -n install -m 0644 ` outside the helper. It reads the script text only — runs unprivileged, never touches `/etc/caddy` or invokes sudo.

**Files changed:** `2` files. `scripts/install.sh` (+24/−5), `scripts/install.test.mjs` (+39/−0).

**Base: `origin/main` @ `0f734e4`**

**Verification results:**
- PR branch (`multica/BET-440-fix-caddyfile-perms` @ `10ed295`):
  - test: exit `0`, `1251` pass / `0` fail / `0` skip. Failures: none. log sha256: `01f533fbcbfbc1defe09ec07aeec759508ba0d51368791462226db19f0acfe10`
- Base (`main` @ `0f734e4`):
  - test: exit `0`, `1250` pass / `0` fail / `0` skip. Failures: none. log sha256: `63e2d2a6b9e8a13b81fc49e63c7e2a1d70bd62f1941175cd20dc83248a03cbc5`
- **Conclusion:** `0` test failures are pre-existing (identical between branches), `0` are new and caused by this PR, `1` new test was added (the BET-440 guard).

**Acceptance criteria met:**
- `npm run typecheck && npm test` pass ✓
- The four assertions pass ✓
- `grep -n "sudo -n mv" scripts/install.sh` returns nothing ✓

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -117,6 +117,20 @@ ok()   { printf '\033[32m✓\033[0m %s\n' "$*"; }
 warn() { printf '\033[33m!\033[0m %s\n' "$*" >&2; }
 die()  { printf '\033[31m✗ %s\033[0m\n' "$*" >&2; exit 1; }
 
+# Install a staged temp file as a root-owned, world-readable config file.
+#
+# ALWAYS use this instead of `mv` or `tee` for files a system service must
+# read. `mv` preserves the SOURCE file's mode + owner, and mktemp(1) creates
+# 0600 files owned by the invoking user — that is exactly what made
+# /etc/caddy/Caddyfile unreadable to the `caddy` service user and silently
+# broke TLS on every re-install. install(1) sets the destination mode and
+# owner explicitly, so the result never depends on the temp file.
+#
+#   $1 = staged source path   $2 = destination path
+install_root_file() {
+  sudo -n install -m 0644 -o root -g root "$1" "$2"
+}
+
 # Parse one key out of a key=value manifest body. Echoes the value, empty if
 # absent. Values may contain `=` (cut -d= -f2- preserves them). A repeated key
 # is reduced to the FIRST occurrence (head -n1) — the manifest writer emits
@@ -1377,7 +1391,7 @@ main() {
         printf 'deb [signed-by=/usr/share/keyrings/caddy-stable-archive-keyring.gpg] https://dl.cloudsmith.io/public/caddy/stable/deb/ubuntu %s main\n' "$_caddy_codename"
         printf 'deb-src [signed-by=/usr/share/keyrings/caddy-stable-archive-keyring.gpg] https://dl.cloudsmith.io/public/caddy/stable/deb/ubuntu %s main\n' "$_caddy_codename"
       } > "$_caddy_list_tmp"
-      sudo -n install -m 0644 "$_caddy_list_tmp" /etc/apt/sources.list.d/caddy-stable.list \
+      install_root_file "$_caddy_list_tmp" /etc/apt/sources.list.d/caddy-stable.list \
         || { rm -f "$_caddy_list_tmp"; die "failed to add Caddy apt repo"; }
       rm -f "$_caddy_list_tmp"
       sudo -n apt-get update \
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-440 (PR #380)

**Route: PASS → manta-pm** (0 Blocks, 0 Questions)

### Acceptance-criteria verification (independent)

1. **`npm run typecheck && npm test` pass** ✅
   - Typecheck: exit `0`, zero errors.
   - Tests: `1251` pass / `0` fail / `0` skip, exit `0`.
   - Matches the implementer's reported PR-branch result exactly. Base was `1250`; the +1 is the new BET-440 guard. Zero new failures, zero pre-existing.
   - The BET-440 test itself ran and passed (`ok 252`).

2. **The four assertions pass** ✅
   The new test in `scripts/install.test.mjs` asserts on the script *text* only (`readFileSync`, no sudo, no `/etc/caddy`):
   - `install_root_file` defined exactly once.
   - Helper body pins `-m 0644 -o root -g root`.
   - No `sudo -n mv` anywhere.
   - No `sudo -n install -m 0644 ` outside the helper. All four pass.

3. **`grep -n "sudo -n mv" scripts/install.sh` returns nothing** ✅ — `grep` exit `1`, no match.

### Diff review

- `install_root_file` helper added exactly as spec'd, immediately after the last existing helper, body `sudo -n install -m 0644 -o root -g root`.
- **Site A** (`Caddyfile.d` snippet, ~1530): routed through the helper. ✓
- **Site B** (marker-replace, ~1559): `sudo -n mv` → `install_root_file`; warn text updated to "sudo install failed"; the `rm -f "$tmp"` moved out of the failure branch so it runs unconditionally (install copies, doesn't move). ✓
- The `printf … >>` append branch and `tee` create branch are **untouched**, per the issue's explicit "do NOT touch" instruction. ✓
- **Extra conversion (apt `caddy-stable.list` write, ~1394):** the issue's own acceptance-criterion #4 ("no `sudo -n install -m 0644` outside the helper") would *fail* if this site stayed on bare `sudo -n install`. Converting it is the minimal change that keeps the issue's own test green, and it routes another service-read config write through the single shared writer consistent with the issue's stated goal. Not scope creep — correctly scoped.
- Scope is exactly the two files the issue names (`scripts/install.sh`, `scripts/install.test.mjs`). No under-delivery, no scope creep.
- `bash -n scripts/install.sh` passes (no syntax regression from the structure change).

### Anti-spaghetti
The single jscpd clone is between two **pre-existing** BET-319 regression tests in `install.test.mjs`; it does not involve the new BET-440 test and is not business logic this PR introduced. Discounted.

### Note for PM
PR #380 is currently marked **draft** (`isDraft: true`). Per this repo's split of responsibilities the reviewer does not flip draft→ready — that's the merge coordinator's call before merging.

PR #382, opened 2026-07-31
installer: stop printing success lines after failures + delete always-false port check (BET-442)

antoinedc:
## BET-442 — installer: stop printing success lines after failures + delete the always-false port-conflict check

Three surgical fixes in `scripts/install.sh`, exactly per the issue:

1. **`ok "caddy reloaded."` is now inside the `else`** — it only prints on a successful `sudo systemctl reload caddy`. Also replaced the two speculative warn lines (which guessed the cause wrong during the BET-440 outage) with a single accurate one pointing the operator at the reload failure as the reason no cert will issue.
2. **`ok "gateway registration complete."` is now inside `if … then`** — it only prints when `merge-gateway` succeeds; the `|| warn` fallback is gone, so no false success after a merge failure.
3. **Deleted the entire `ss -tlnH` port-conflict block.** `ss` was invoked without `-p`, so `grep -q caddy` could never match and the warning fired 100% of the time, falsely. Per the issue, not "fixed" with `-p` — a real conflict already surfaces as an ACME/bind error in `journalctl -u caddy`, which the reload-failure message now points at.

The `CADDY_E_SKIP` wrapper is kept (reload is still skipped when the Caddyfile write failed), and the pair-code step (8) still runs regardless — the graceful-degradation contract is preserved. No `warn` → `die` conversion. `install.sh` is net shorter.

### Tests

Added 4 text-assertion tests in `scripts/install.test.mjs` (BET-442), plus reworked the two caddy-skip guard tests that still referenced the deleted port-check, so the suite stays honest about what the guard now protects. Updated stale `port-check` references in `install.sh` comments.

### Verification

- typecheck: pass (exit 0, no errors)
- test: 1255 pass / 0 fail / 0 skip

**Base: `origin/main` @ `173147e`**

No cross-cutting follow-ups; changes are confined to `scripts/install.sh` + `scripts/install.test.mjs`.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1319,9 +1319,11 @@ main() {
           # Pipe the JSON to merge-gateway via stdin (lib subcommand) so the
           # auth.json write is atomic temp-rename + 0600, preserving
           # box_id / box_token / created_at.
-          printf '%s' "$GW_RESP" | "$NODE" "$LIB" merge-gateway --file "$MANTA_AUTH_FILE" 2>/tmp/manta-gateway-merge.err \
-            || warn "merge-gateway failed (see /tmp/manta-gateway-merge.err) — the server will re-register on next boot."
-          ok "gateway registration complete."
+          if printf '%s' "$GW_RESP" | "$NODE" "$LIB" merge-gateway --file "$MANTA_AUTH_FILE" 2>/tmp/manta-gateway-merge.err; then
+            ok "gateway registration complete."
+          else
+            warn "merge-gateway failed (see /tmp/manta-gateway-merge.err) — the server will re-register on next boot."
+          fi
         fi
       fi
     fi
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-442

**Verdict: PASS** (0 Blocks, 0 Questions, 0 Nits) — routing to `manta-pm`.

1. **`ok "caddy reloaded."` only on success** — `scripts/install.sh:1597` now sits in the `else` of the reload guard. Only prints on a successful reload. The two speculative warn lines are gone; the single replacement points at the reload failure (vhost on disk, no cert until reload succeeds). ✓
2. **`ok "gateway registration complete."` only on success** — now inside the `if … then` branch; the `|| warn …` continuation is gone. `continue-on-failure` behaviour preserved (the `warn` still runs, no `die`). ✓
3. **Port-conflict check deleted** — the whole `ss -tlnH` block is gone. The `CADDY_E_SKIP` wrapper is retained (reload still skipped when the Caddyfile write failed); only the port loop was removed, plus the stale `port-check` reference in the wrapper's else-warn. ✓
4. **Tests** — the 4 required text assertions are in `scripts/install.test.mjs` and pass:
   - `caddy reloaded` ×1, preceded by `else` — `install.test.mjs:5155` (test idx 235) ✓
   - `gateway registration complete` ×1, no `|| warn "merge-gateway failed` — idx 236 ✓
   - no `is bound by something other than Caddy` — idx 237 ✓
   - no `ss -tlnH` — idx 238 ✓
   The two caddy-skip guard tests that referenced the deleted port-check were reworked to assert only the reload — kept honest, still in scope.
5. **typecheck + tests** — ran locally on the PR branch: `npm run typecheck` exit 0; `npm test` 1255 pass / 0 fail / 0 skip. Matches the implementer's reported results exactly (trust-but-spot-check: single run confirms, no base divergence). ✓
6. **`install.sh` net shorter** — 13 insertions / 25 deletions (−12). ✓
7. **No over-engineering / no `die` / no new diagnostics** — the warn→die contract is untouched (pair-code step 8 runs regardless; `CADDY_E_SKIP` guard intact). ✓

### Base freshness
Merge-base `173147e` matches the implementer's reported base. `git diff origin/main...HEAD` shows only the two in-scope files, no phantom deletions. (origin/main has since advanced ~4 unrelated commits; nothing overlaps this PR.) No stale-base issue.

### Notes
- **Pre-existing, out of scope (not a finding):** `scripts/install.sh` lines 766/773 reference `$OPENCODE_CLAUDE_AUTH_PLUGIN` unguarded under `set -u` (a BET-319-category `:-` bug on a different variable). This is byte-identical on `origin/main`, untouched by this PR, and a real install would only hit it if that var is unset. Flagging it when the PR is clean would be scope creep; it predates BET-442 and lives outside the changed region.
- **Anti-spaghetti:** 1 clone flagged in `install.test.mjs` — a duplicate between two pre-existing BET-319 test helpers (lines 1402–1421 vs 1431–1452). Coincidental, unrelated to BET-442's logic; discounted.

The diff is surgical, matches the issue's spec exactly (three fixes, all deletions or one-line guards), does not regress the graceful-degradation contract, and is net-lighter.

PR #385, opened 2026-07-31
installer: collapse the four Caddyfile write paths into one tested upsert (BET-441)

antoinedc:
**Base: origin/main @ `abf7ec0`**

Fixes BET-441: the install's Caddyfile handling had four divergent write paths for "make the Caddyfile contain exactly one manta vhost" (conf.d snippet, marker-replace, plain append, create-from-scratch), and the awk-based marker-replace appended a stray `# >>> manta >>>` line on every re-run. The block editing now lives in one pure, unit-tested function.

**Changes:**
- `upsertCaddyBlock(existing, block)` in `scripts/install-lib.mjs` — strips every marker region + stray marker lines, appends one block, idempotent, pure (throws `TypeError` on non-strings).
- `upsert-caddy-block` CLI subcommand — reads current Caddyfile from stdin, computes the upsert, writes the new file to stdout. Same `--box-id`/`--port` validation as `render-caddy-vhost`.
- `scripts/install.sh` — the conf.d-missing `else` branch collapses to a single `cat | upsert-caddy-block > tmp; install_root_file` path. Builds/append/replaces all funnel through it. Net deletion: script went 1754 → 1727 lines; the `>>> manta >>>` string no longer appears anywhere in the shell script; `upsert-caddy-block` is invoked exactly once.
- `scripts/install.test.mjs` — 20 unit tests for `upsertCaddyBlock` (empty input, content preservation, dedup, the five-stacked-markers regression, idempotency, box-id replacement, TypeError) + CLI subcommand tests (empty stdin, byte-idempotency over own output, stacked-marker collapse, missing-flag errors) + textual `install.sh` assertions (no awk-on-manta, exactly one upsert call site, no SNIPPET/grep branches).

**Cross-cutting follow-ups:** none. `renderCaddyVhost` inline output is unchanged (already pinned by the existing test).

**Verification results:**
- PR branch (`multica/BET-441-caddyfile-upsert` @ `c317ecc`):
  - test: exit 0, 1255→1269 pass / 0 fail / 0 skip. Failures: none. log sha256: `0b07ee6cc4461c7ce823ef4e2ce2b363076dc3181fece22aa375c10769da94bb`
- Base (`main` @ `abf7ec0`):
  - test: exit 0, 1255 pass / 0 fail / 0 skip. Failures: none. log sha256: `8bfc600ac523cdc0dc9411f940eacef7b91cface23bdde2aedecbd1702416352`
- **Conclusion:** 0 new failures caused by this PR. The 14-test increase (1255 → 1269) is the new BET-441 test coverage.

**Acceptance criteria all met:**
- `npm run typecheck && npm test` pass (above).
- `scripts/install.sh` net shorter: 1754 → 1727 lines.
- `>>> manta >>>` appears 0 times in `scripts/install.sh`.
- Double-run byte-idempotency verified: `printf '' | node scripts/install-lib.mjs upsert-caddy-block --box-id … --port 8787` twice over its own output is byte-identical.


antoinedc:
## BET-441 review — 1 Block (fix-here), 0 Questions

**Base-freshness:** PR branched off `origin/main` @ `abf7ec0` (current tip). No stale base, no phantom file deletions.

**Mechanical gate (trust-but-spot-check, verified on the PR branch `c317ecc`):**
- `npm run typecheck` → exit 0 (matches your report)
- `npm test` → 1269 pass / 0 fail (matches your report)
- acceptance: `scripts/install.sh` 1754 → 1727 (net shorter ✓); `>>> manta >>>` appears 0× in the shell ✓; double-run byte-idempotency verified (`printf '' | ...upsert-caddy-block ... ` twice → byte-identical) ✓
- anti-spaghetti: the single jscpd clone is a pre-existing `MANTA_CLAUDE_AUTH_PLUGIN` guard-test pattern, not duplication this PR introduced. Discounted.

Scope is tight: only `scripts/install-lib.mjs`, `scripts/install.sh`, `scripts/install.test.mjs`, all in the issue's stated surface. `renderCaddyVhost` inline output untouched. Test suite covers all 7 required unit cases + 2 text assertions + CLI cases.

### Block (fix-here): missing-Caddyfile create-from-scratch case is silently dropped under `pipefail`

The collapsed path is:

```sh
if ! cat "$CADDYFILE" 2>/dev/null \
    | "$NODE" "$LIB" upsert-caddy-block --box-id "$BOX_ID_FOR_GATEWAY" --port "$MANTA_PORT" \
    > "$_caddyfile_tmp" 2>/tmp/manta-caddy-render.err; then
  warn "failed to render the Caddy vhost ..."
  CADDY_E_SKIP=1
elif ! install_root_file ...
fi
```

`install.sh` runs under `set -euo pipefail`. When `/etc/caddy/Caddyfile` does **not** exist, `cat` fails (exit 1); with `pipefail` the whole pipeline exits non-zero, so `if !` takes the **warn + `CADDY_E_SKIP=1`** branch — it never reaches `install_root_file`, so **no Caddyfile is created**. I confirmed this empirically:

```
set -o pipefail
cat /nonexistent/Caddyfile 2>/dev/null | node ... upsert-caddy-block ... > tmp 2>err
echo $?   # → 1 (pipeline), so `if ! ...; then` fires warn/skip
```

This contradicts the issue's stated contract (and the new comment in the script): "`cat` of a missing file yields empty stdin, which the subcommand treats as 'no Caddyfile yet' — so create/append/replace are one code path." Under `pipefail` that create case never happens — the old code handled it explicitly (`else` branch with `sudo -n tee` writing a minimal Caddyfile), and this patch deletes it. A box with Caddy installed but a missing/absent main Caddyfile now silently gets no vhost on re-run instead of having one created.

**Scope note:** this is an intended-deliverable regression, not scope creep — the whole point of the issue (per its "What done must NOT look like") is that the four write paths become *one code path*. Here the four became one path plus a silently-dropped create case.

**Fix (in-scope, narrow):** make the pipeline tolerate a missing Caddyfile so node really receives empty stdin and the create-from-scratch case lands in `install_root_file`. e.g. wrap `cat` so it can't fail the pipeline:

```sh
if ! ( cat "$CADDYFILE" 2>/dev/null || true ) \
    | "$NODE" "$LIB" upsert-caddy-block --box-id "$BOX_ID_FOR_GATEWAY" --port "$MANTA_PORT" \
    > "$_caddyfile_tmp" 2>/tmp/manta-caddy-render.err; then
  ...
```

(the `|| true` group makes a missing file produce empty stdin + exit 0 while still propagating node's genuine render/validation exit code). Missing-file → empty stdin → subcommand emits the block → `install_root_file` writes it. Append/replace paths are unaffected.

Please add a regression test for the missing-file-create case (currently only empty-`printf ''` stdin is tested at the CLI level, which does exercise the same code path in the subcommand — but the shell-facing missing-file behavior under `pipefail` isn't pinned; a textual/bash-level assertion that the `cat` cannot trip the warn branch, like the BET-442-style guards, would lock it in).

antoinedc:
Addressed the Block (missing-Caddyfile create-from-scratch case dropped under `pipefail`).

**Fix (`scripts/install.sh`):** the collapsed path now guards the `cat` so a missing `/etc/caddy/Caddyfile` yields empty stdin + exit 0, while node's own render/validation exit code still propagates:

```sh
if ! ( cat "$CADDYFILE" 2>/dev/null || true ) \
    | "$NODE" "$LIB" upsert-caddy-block --box-id "$BOX_ID_FOR_GATEWAY" --port "$MANTA_PORT" \
    > "$_caddyfile_tmp" 2>/tmp/manta-caddy-render.err; then
```

Without the `( ... || true )` group, `set -o pipefail` makes a missing file fail the whole pipeline → the `if !` fires the warn/`CADDY_E_SKIP` branch → no vhost created. With it, missing file → empty stdin → subcommand's create-from-scratch case emits the block → `install_root_file` writes it. Append/replace paths are untouched.

**Regression tests (`scripts/install.test.mjs`)** — two, one textual and one behavioral:
1. Textual (BET-442-style): asserts the `( cat "$CADDYFILE" 2>/dev/null || true )` guard is present in the `if !` pipeline, and that no bare `cat "$CADDYFILE" 2>/dev/null` line (without `|| true`) pipes into the upsert.
2. Behavioral: drives the whole pipeline under `bash -o pipefail` exactly as your reproduction — a missing Caddyfile exits 0 and produces exactly the block (create-from-scratch), and an unguarded `cat ... | node upsert-caddy-block` fails the pipeline non-zero (proving the `|| true` is what keeps create-from-scratch alive).

**Re-verified:**
- `npm run typecheck` → exit 0 (`00a5b657…`)
- `npm test` → 1271 pass / 0 fail (`64452afb…`)
- `scripts/install.sh` still net shorter (1754 → 1735); `>>> manta >>>` appears 0×; `upsert-caddy-block` invoked exactly once.
- Double-run byte-idempotency still passes.

Branch `multica/BET-441-caddyfile-upsert` @ `61ca59a`. Reassigning to reviewer.

antoinedc:
## BET-441 re-review — PASS (0 Blocks, 0 Questions)

Re-reviewing the fix commit `61ca59a` addressing the prior Block. The missing-Caddyfile create-from-scratch case is now correctly restored.

**The fix is correct and narrower than the prior review even required.** The `( cat "$CADDYFILE" 2>/dev/null || true )` group makes a missing Caddyfile yield empty stdin + exit 0, so the pipeline reaches `install_root_file` (create-from-scratch). Node's own render/validation exit code still propagates (only the `cat` is neutralized). Verified empirically:

```
set -o pipefail
( cat /nonexistent/Caddyfile 2>/dev/null || true ) | node ... upsert-caddy-block ... > tmp 2>err
# → exit 0, tmp holds exactly the marker block → would install (create) — WRONG/skip case is gone
```

**Mechanical gate (verified on PR branch `61ca59a`):**
- `npm test` → 1271 pass / 0 fail (matches your report)
- install.sh 1754 → 1735 (net deletion held; the +guard-comment lines don't undo it)
- `>>> manta >>>` still 0× in the shell; `upsert-caddy-block` called exactly once; no `awk`-on-manta
- double-run byte-idempotency: still byte-identical

**Base-freshness:** the PR branches at `abf7ec0`; origin/main has since advanced to `4f9d862`, but that move is BET-450 (src/renderer/clock.* + website images) — it does not touch any of this PR's files, so no stale-base phantom deletions and no conflict on this PR's surface.

**Tests added for the Block are meaningful:** the textual guard assertion pins the `|| true` (and forbids a bare cat pipe that would reintroduce the bug), and the behavioral test drives the real pipeline under `bash -o pipefail` asserting both the guarded (exit 0, block created) and unguarded (fails non-zero) forms — proving the guard is load-bearing. Good regression coverage for exactly the failure mode.

**Anti-spaghetti:** the single jscpd clone is the same pre-existing `MANTA_CLAUDE_AUTH_PLUGIN` guard-test pattern; nothing new introduced by either commit. Discounted.

No remaining Block or Question findings. Routing to manta-pm.

PR #387, opened 2026-07-31
BET-446: installer — guard $OPENCODE_CLAUDE_AUTH_PLUGIN against set -u

antoinedc:
## BET-446 — installer: hard-fail risk under `set -u` for `$OPENCODE_CLAUDE_AUTH_PLUGIN`

**Base: origin/main @ `47f8784`** (rebased onto latest main; includes BET-441's Caddyfile upsert)

### Problem
`scripts/install.sh` runs under `set -euo pipefail`. BET-319 guarded `MANTA_CLAUDE_AUTH_PLUGIN` (the dev/CI override) but missed the parallel LOG-ONLY var `OPENCODE_CLAUDE_AUTH_PLUGIN`, referenced bare in the two "Seeding Claude auth plugin (...)" `log` messages (the branches reached exactly when NO override is set — i.e. the default install path). A bare `$OPENCODE_CLAUDE_AUTH_PLUGIN` aborts the whole install the instant the var is unset (bash `unbound variable`).

### Fix
Trace: `OPENCODE_CLAUDE_AUTH_PLUGIN` is emitted by the `print-config` eval in `main()` (line ~659), so it is set in the happy path — but it is **log-only** (the actual merge reads `MANTA_CLAUDE_AUTH_PLUGIN` or the JS default) and not guaranteed under a partial/failed config eval. Both references are therefore guarded with the `${VAR:-}` default-expansion form, matching BET-319's treatment of the sibling var. Behavior is unchanged when the var IS present.

### Tests (scripts/install.test.mjs)
- Static guard (BET-446): every non-comment reference to `OPENCODE_CLAUDE_AUTH_PLUGIN` must use `${VAR:-}`.
- Behavioral — real extracted seeding block run under real `set -euo pipefail` against a fake box:
  - var **unset** → install completes past the previously-fatal line (exit 0, no `unbound variable`). Proven red against the pre-fix source.
  - var **set** → still seeds normally (success line + plugin present in the merged opencode.jsonc).

**Verification results:**
- PR branch (`multica/BET-446-installer/plugin-var-set-u-guard` @ `dbe2add`):
  - test: exit 0, **1274 pass / 0 fail / 0 skip**. Failures: none. log sha256: `6a83225e9c59dbd106b62597cf0ef76b6fa39dc89c6b3ac76026b4a979021553`
- Base (`main` @ `47f8784`):
  - test: exit 0, **1271 pass / 0 fail / 0 skip**. Failures: none. log sha256: `d29100bcf832a0e291ee2858b6536080095d4837d7431049de11e13dcb702adc`
- **Conclusion:** 0 new failures; the PR adds the 3 BET-446 tests (1271 → 1274). No cross-cutting follow-ups.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -755,22 +755,27 @@ main() {
   OPENCODE_CONFIG="$OPENCODE_CONFIG_DIR/opencode.jsonc"
   mkdir -p "$OPENCODE_CONFIG_DIR"
   OPENCODE_CONFIG_BACKUP="$OPENCODE_CONFIG.pre-manta"
-  # ${MANTA_CLAUDE_AUTH_PLUGIN:-} — the installer runs under `set -u`, so the
-  # bare expansion of an unset var is a hard error (the macOS CI job caught
-  # this: every public install crashes because end users never set the
-  # override). The `:-` default keeps the test safe when the var is unset.
+  # The installer runs under `set -u`, so a bare expansion of an unset var is a
+  # hard error. Both plugin vars are therefore guarded with `${VAR:-}`:
+  #   - MANTA_CLAUDE_AUTH_PLUGIN — the dev/CI override; unset for every end
+  #     user. The macOS CI job caught this once (BET-319): every public
+  #     install crashed because end users never set the override.
+  #   - OPENCODE_CLAUDE_AUTH_PLUGIN — the LOG-ONLY default name, emitted by the
+  #     `print-config` eval above (line ~659). It is pure cosmetic output for
+  #     the seeding message here; it must never be able to abort the install,
+  #     so the (should-be-unreachable-in-practice) empty default is fine.
   if [ -f "$OPENCODE_CONFIG" ]; then
     if [ -n "${MANTA_CLAUDE_AUTH_PLUGIN:-}" ]; then
       log "Seeding Claude auth plugin (override: ${MANTA_CLAUDE_AUTH_PLUGIN}) — merging into existing ${OPENCODE_CONFIG}…"
     else
-      log "Seeding Claude auth plugin ($OPENCODE_CLAUDE_AUTH_PLUGIN) — merging into existing ${OPENCODE_CONFIG}…"
+      log "Seeding Claude auth plugin (${OPENCODE_CLAUDE_AUTH_PLUGIN:-}) — merging into existing ${OPENCODE_CONFIG}…"
     fi
     existing="$(cat "$OPENCODE_CONFIG" 2>/dev/null || true)"
   else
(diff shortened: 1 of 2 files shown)


antoinedc:
## BET-446 review — VERIFIED PASS

Reviewing against the issue's acceptance criteria (not the summary). Base-freshness gate: merge-base == `origin/main` tip (`47f8784d`), so the diff is a true +11/−6 on `scripts/install.sh` and +95 on `scripts/install.test.mjs` only — no phantom deletions, no stale base.

1. **Both `$OPENCODE_CLAUDE_AUTH_PLUGIN` refs safe under `set -u`** — CONFIRMED. The two references (install.sh lines 771 & 778, shifted only by the expanded comment) are now `${OPENCODE_CLAUDE_AUTH_PLUGIN:-}`. Grep confirms no bare non-comment reference remains in the repo.
2. **Confirm whether var is guaranteed set** — CONFIRMED via code trace. It's emitted by the `print-config` eval in `main()` and is **log-only** (the actual merge reads `MANTA_CLAUDE_AUTH_PLUGIN` or the JS default), so it's not guaranteed under a partial/failed config eval. Guarding it is correct.
3. **Guard with proper default / readiness check** — Use of `${VAR:-}`, matching BET-319's treatment of the sibling override var. Consistent with the installer's conventions.
4. **Behavior unchanged when var IS present** — CONFIRMED. `:-` is a no-op when set; the set-var test proves the plugin is still merged and the success line still prints.

### Verification I ran (independent)

- `npm test` (full suite) → **1274 pass / 0 fail / 0 skip**, exit 0 — matches the PR-branch block.
- Ran only the new install tests → all 3 BET-446 tests green (`install.sh guards every OPENCODE_CLAUDE_AUTH_PLUGIN use…`, unset-seeding, set-seeding).
- **Proved the unset-var test is genuinely red (not vacuous):** I flipped the two refs back to the pre-fix bare form and the unset-seeding test fails with exactly `OPENCODE_CLAUDE_AUTH_PLUGIN: unbound variable` under `set -euo pipefail`. The test exercises the real extracted source block, not a replica. (Restored the fixed file afterward; tree re-verified against the PR diff.)

### Scope

Diff touches only `scripts/install.sh` + `scripts/install.test.mjs` — the two files the issue named. No scope creep, no under-delivery.

### Anti-spaghetti (non-blocking)

The flagged jscpd clone is a trivial `for (let i = 0; i < lines.length; i++)` iteration idiom shared with the sibling BET-319 guard — an intentional parallel-guard structure, not duplicated business logic. Discounted.

### Nits (non-blocking)

- None material.

**Result: 0 Blocks, 0 Questions (0 nits meaningful).** Routing to manta-pm for merge.

PR #412, opened 2026-08-01
feat(server): device registry + per-device token + per-device revoke (BET-490) — TOKEN/CLAIM-MODEL CHANGE

antoinedc:
## BET-490 — Device registry + per-device token + linked-device list with per-device revoke

**⚠️ TOKEN/CLAIM-MODEL CHANGE.** Please review the coexistence/back-compat statements below carefully.

Base: `origin/main @ 4a948d5`

### What this does (Stage 2 of the BET-488 pairing rework)

Server foundation only — no UI, no TTL/rotation (Stage 3), no push-on-pairing
(Stage 4), no desktop panel (Stage 5). It adds:

- **A durable device registry** — `~/.manta/devices.json` (0600 via the shared
  `statePath()` + `writeJsonAtomic` atomic pattern). Each entry: opaque
  `device_id`, distinct 32-hex (128-bit) `token`, optional human `name`,
  `last_seen`, `created_at`, `revoked`, `primary`.
- **Per-device claim** — a claim with a `device_id` provisions (or resumes)
  that device with a distinct token.
- **Per-device revoke** — `DELETE /auth/revoke?device_id=<id>` kills only that
  device's token, immediately (its next request 401s); other devices and the
  primary/desktop token are untouched. The revoked device's own token can no
  longer authorize anything.
- **A linked-device list** — authenticated `GET /auth/devices` returns per-
  device metadata (`device_id`, `name`, `last_seen`, `created_at`, `primary`).
  It never exposes any token.
- **Whole-box reset preserved** — `DELETE /auth/revoke` with no `device_id`
  keeps the legacy BET-357 behavior: regenerate the entire identity so every
  credential re-pairs.

### Coexistence design (how existing paired devices keep working)

The standing constraint was that nothing already paired may break. Approach:

- The **shared `box_token` from `auth.json` is seeded as the `primary` device**
  in the registry. `authorize()` resolves a presented token against the
  registry (primary OR any live per-device token), so the existing desktop
  (holds `box_token`) and the manta-native AI tools (read `~/.manta/auth.json`
  per call) keep authenticating unchanged.
- On a box being upgraded there is no `devices.json` yet — `createAuthEngine`
  seeds the primary from the existing `box_token` on first construction.
- A **claim with no `device_id`** resumes the primary device and returns
  `box_token = auth.box_token` exactly as before, so legacy clients are
  byte-compatible. A **claim with a `device_id`** provisions/resumes that
  specific device and returns its distinct token. The `{ box_token, box_id }`
  response shape is unchanged; `device_id` is additive and ignored by existing
  clients (`parseClaimResponse` only validates `box_token` + `box_id`).

### What I changed about `isExemptPath`, the token store, and loopback minting

- **`isExemptPath`:** **no change.** `/auth/pair`, `/auth/claim`,
  `/auth/revoke` remain exempt (self-validating); the new `/auth/devices` is
  deliberately **NOT exempt** — it is bearer-gated so a random scanner cannot
  enumerate linked devices.
- **The 0600 token store:** `~/.manta/auth.json` is unchanged and untouched.
  Added a **new** `~/.manta/devices.json` (0600) as the device registry. The
  primary device's token mirrors `auth.box_token` (both are box-local 0600
  stores, same trust boundary). Verified via the `DEVICES_STORE_PATH` sandbox
  canary + a 0600 round-trip test.
- **Loopback-only minting:** no change — `GET /auth/pair` remains
  loopback-only (`isLocalDirectRequest`), and `/auth/revoke` still requires a
  valid live device token (401/400/404 as documented).

### Scope resolve on §6.4 ambiguity

§6.4 ("Immediate, from any linked device or the desktop") did not state whether
whole-box reset must survive per-device revoke. Per the issue's instruction to
stop-and-comment only if it's genuinely ambiguous, I chose to **preserve the
whole-box reset** as the no-`device_id` revoke path (back-compat with the
existing revoke tests and the desktop's classifier), and per-device revoke is
the additive `device_id` path. This keeps both behaviors available with no
ambiguity.

### Tests

- `src/server/devices.test.mjs` (new, 14): distinct tokens on Nth device;
  claim provisions-or-resumes; per-device revoke kills only the target while
  other devices + primary still pass; whole-box reset kills everything +
  re-seeds primary; list never leaks tokens; 0600 + `statePath` sandbox.
- `src/server/auth.test.mjs`: updated the shared `engine()` test helper to
  inject an isolated device registry (no live-store writes); all 39 existing
  pairing/claim/revoke tests pass unchanged.

**Verification:** `npm run typecheck` clean; `npm test` → 1293 pass / 0 fail.

### DIFF
--- a/src/server/auth.mjs
+++ b/src/server/auth.mjs
@@ -35,8 +35,15 @@ import { unlink } from "node:fs/promises";
 import { randomBytes, randomInt, timingSafeEqual } from "node:crypto";
 import { statePath } from "../shared/paths.mjs";
 import { readJsonSync, writeJsonAtomic } from "./jsonStore.mjs";
+import { createDeviceRegistry, DEVICES_STORE_PATH, saveDevicesRaw } from "./devices.mjs";
 
 const STORE_PATH = statePath("auth.json");
+
+// How often the device registry's in-memory `last_seen` touches are flushed to
+// disk. authorize() runs on every authenticated request, so persisting each
+// one would be a write per request; a 1s throttle keeps the device-list's
+// last_seen roughly fresh without any write amplification.
+const DEVICE_PERSIST_MS = 1000;
 
 // Pairing codes are short-lived by design: a device must claim within this
 // window or the code expires and the user re-opens the pair screen.
@@ -380,6 +387,10 @@ export function createPairingRegistry({ ttlMs = PAIRING_TTL_MS, now = () => Date
  *                 regenerate step
  *   deleteAuth  — injectable unlink (default deleteAuth) for the post-revoke
  *                 "wipe + regenerate" path.
+ *   devices     — injectable device registry (createDeviceRegistry instance).
+ *                 Tests MUST inject their own (with no-op load/save); the
+ *                 DEFAULT targets ~/.manta/devices.json (0600) and is owned by
+ *                 index.mjs.
  *
  * DANGER — the saveAuth/deleteAuth DEFAULTS target the real box store
  * (~/.manta/auth.json). Only index.mjs, which owns that store, may rely on
@@ -396,11 +407,34 @@ export function createAuthEngine({
   now = () => Date.now(),
   saveAuth: saveAuthFn = saveAuth,
   deleteAuth: deleteAuthFn = deleteAuth,
+  devices: devicesFn,
 } = {}) {
   if (!auth || !isValidToken(auth.box_id) || !isValidToken(auth.box_token)) {
     throw new Error("createAuthEngine: valid { box_id, box_token } required");
   }
   const pairing = createPairingRegistry({ ttlMs, now });
+
+  // Per-device credential registry. Default targets ~/.manta/devices.json;
+  // the shared box_token is seeded as the `primary` device so existing paired
+  // devices (desktop + manta-native AI tools) keep working unchanged.
+  const devices = devicesFn ?? createDeviceRegistry({ now });
+  // Seed/resync the primary (shared box_token) device — idempotent: if a
+  // primary already exists (upgrade path) this just re-syncs its token; on a
+  // fresh registry it creates one. This is what keeps the existing desktop +
+  // manta-native AI tools authenticating against the new per-device gate.
+  devices.seedPrimary(auth.box_token);
+
+  // Fire-and-forget, throttled flush of the device registry (last_seen). Safe
+  // to ignore: it never blocks the synchronous auth path, and a lost write
+  // only means a stale last_seen on the next process start.
+  let lastDevicePersist = 0;
+  function maybePersistDevices(force = false) {
+    const t = now();
+    if (force || t - lastDevicePersist > DEVICE_PERSIST_MS) {
+      lastDevicePersist = t;
+      devices.persist().catch(() => {});
+    }
+  }
 
   // Is this request authorized? Returns { ok } or { ok:false, status, error }.
   // Exempt paths and the disabled-enforcement mode short-circuit to allow.
@@ -412,7 +446,12 @@ export function createAuthEngine({
     if (method === "GET" && isPublicAssetPath(path)) return { ok: true };
     if (!enforce) return { ok: true };
     const presented = parseBearer(authorization);
-    if (presented && tokenMatches(auth.box_token, presented)) return { ok: true };
+    if (presented && devices.authorize(presented)) {
+      // matched a live device (primary or per-device) → authorized; last_seen
+      // bumped in memory, flushed at most once per DEVICE_PERSIST_MS.
+      maybePersistDevices();
+      return { ok: true };
+    }
     return { ok: false, status: 401, error: "unauthorized" };
   }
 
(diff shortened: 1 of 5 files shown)


antoinedc:
## Review — BET-490 Stage 2 (device registry + per-device token/revoke)

Reviewed against the issue's acceptance criteria. Base-freshness check: PR branches from `origin/main @ 4a948d5`; `git diff --stat origin/main...HEAD` shows only the 5 expected files (667+/37−), all additions — no phantom deletions, base is clean.

### Mechanical gate
- `npm run typecheck` → **exit 0, 0 errors**
- `npm test` → **1293 pass / 0 fail**
- Matches the implementer's reported PR-branch results exactly.
- jscpd (changed scope): **0 clones**.

- ✅ New registry store under `~/.manta/` via `statePath()` + atomic-0600 `writeJsonAtomic` (`src/server/devices.mjs`). `DEVICES_STORE_PATH` resolves under `MANTA_STATE_HOME`; 0600 mode verified by test.
- ✅ Entry shape: opaque `device_id`, distinct 32-hex token, optional name, `last_seen`, `created_at`, `revoked`, `primary`.
- ✅ `/auth/claim` provisions/resumes a per-device token; no-`device_id` claim resumes the primary and returns the shared `box_token` (back-compat byte-identical). `{box_token, box_id}` shape preserved; `device_id` additive.
- ✅ `/auth/revoke?device_id=<id>` per-device revoke is immediate (revoked token 401s on next request) and leaves siblings + primary untouched (tested).
- ✅ Whole-box reset preserved as the no-`device_id` path (reasonable resolve of the §6.4 ambiguity, documented).
- ✅ `/auth/devices` (bearer-gated, NOT exempt — verified it sits behind the gate, not in `isExemptPath`); never exposes tokens.
- ✅ Coexistence: shared `box_token` seeded as `primary` device; existing desktop + manta-native AI tools keep authenticating. `isExemptPath`, `auth.json`, and loopback-only minting all unchanged (flagged in PR title/body as required).
- ✅ Tests: new `devices.test.mjs` (distinct tokens, provision-or-resume, per-device revoke isolation, primary regression, whole-box reset, no-token-leak, 0600/statePath sandbox); all 39 existing pairing/claim/revoke tests pass unchanged.

### Block (fix-here) — re-claim after revoke with the same `device_id` breaks per-device revoke
`registry.claim({ deviceId })` on a **revoked** device skips the resume branch (because the existing entry is revoked) but still reuses the client-supplied `device_id`, pushing a **second, live entry with the same `device_id`** beside the stale revoked one. `byId`/`revokeDevice` use `find` → return the **first** (already-revoked) match, so:

- `GET /auth/devices` lists the resurrected device (it's non-revoked) under id `D` — an id that can no longer be resolved to the live entry.
- One-tap revoke of `D` targets the stale entry, returns success, and **silently leaves the resurrected device live**.

Reproduced against the engine directly:
```
first live: true
after revoke, D dead: true
re-claimed D live: true
entries with device_id D: 2          ← duplicate id breaks the byId uniqueness invariant
revokeDevice(D) returned: D revoked= true   ← acts on the stale entry
BUT re-claimed D still LIVE after revoke-by-id: true   ← §6.3 one-tap revoke fails
```

This contradicts the registry's own comment ("a re-claim after a revoke gets a brand new entry") — it *does* create a new entry, but keeps the same id, corrupting `byId`. Minimal in-scope fix: when the supplied `device_id` resolves to a revoked entry, either mint a **fresh** `device_id` (`deviceId` only used when there's no revoked collision), or overwrite/remove the stale revoked entry. Please also add a test: `claim({deviceId})` after `revokeDevice(deviceId)` yields a unique id and the result stays revocable by id.

antoinedc:
Addressed the Block (fix-here) @ 7a08b33.

Root cause: `devices.claim({deviceId})` on a revoked device falls past the resume branch but reused the client-supplied `device_id`, producing a LIVE duplicate-id entry beside the stale revoked one — `byId`/`revokeDevice` (first-match `find`) then resolved the resurrected device to the revoked entry, so one-tap revoke silently no-oped.

Fix: a supplied `device_id` is only ever reused when NO entry (live or revoked) holds it. A revoked device is never resurrected, so a re-claim under a revoked id mints a brand-new id (+ fresh distinct token). The stale revoked entry is left untouched, keeping the `byId` uniqueness invariant.

Tests added (both green):
- `registry: REGRESSION: re-claim after a revoke gets a FRESH id (no duplicate), stays revocable by id` — asserts a unique id, no duplicate rows, and one-tap revoke of the resurrected device works.
- `REGRESSION: per-device revoke still works after a re-claim with a revoked device_id` — end-to-end through the auth engine.

Verification: `npm run typecheck` clean; `npm test` → node:test 1295 pass / 0 fail (was 1293; +2), vitest 1737 pass / 0 fail.

antoinedc:
### Re-review — BET-490 PR #412 @ `7a08b33` — PASS ✅

The Block is resolved. `devices.claim()` now reuses a client-supplied `device_id` only when no entry (live **or** revoked) holds it, otherwise minting a fresh id — restoring the `byId` uniqueness invariant. Independently re-verified against the engine:

- Re-claimed device gets a fresh unique id (no duplicate rows under the revoked id).
- It appears in the list under that new id, and **one-tap revoke of the listed id kills it** on its next request (§6.3 now holds).
- Stale revoked row is filtered from the list and can't be resolved.

**Mechanical gate (re-run):** `npm run typecheck` → 0 errors; `npm test` → node:test 1295 pass / 0 fail (matches the reported +2 regression tests, both present and passing). jscpd (changed scope): 0 clones. Base/fresh diff (`origin/main...HEAD`): only the 5 in-scope files, no phantom deletions.

**Final: 0 Blocks, 0 Questions, 0 nits.** Full acceptance-criteria checklist from the prior review still holds and nothing new was introduced by the fix. Routing to `manta-pm` for merge.

PR #413, opened 2026-08-01
refactor(scripts): collapse Multica API helpers, fail loud on writes (BET-504)

antoinedc:
## BET-504 — A failed Multica write fails the job; collapse the three duplicated API helpers

**Part 2 (helper) then Part 1 (fail-loud), per the issue's ordering.**

- New `scripts/lib/multicaApi.mjs` — the single Multica HTTP helper (`api` + `ApiError`), the ONLY place a Multica write happens. The three local copies are deleted.
- Every Multica **write** (status change, assignment, rerun, metadata, issue comment) now exits non-zero on a non-OK response, logging the issue key + HTTP status + response body. **Reads** still warn and continue. A batch finishes processing all issues before exiting non-zero.
- Each `api` call accepts an injectable `fetch` (5th arg, default = global) so the sweeps are testable without real HTTP — same injection pattern the existing pure-function tests use.
- No retry, no `--strict` flag, no second helper. Call-site behaviour unchanged.
- GitHub-only calls in `multica-pr-closed.mjs` (`gh` probes + `commentOnPr`) stay local intentionally: different API (api.github.com), different token headers, and they must warn-and-continue rather than throw. Forcing them through the Multica helper would change their behaviour, which the issue forbids ("stop and comment" — noted here rather than adding a flag/second function).

**Net line count:** whole diff `+333 / −128` (net `+205`); source only (helper + 3 scripts, excluding the new test file) `+204 / −127` (net `+77`). Not net-negative: the change necessarily adds the new helper module, a fetch-injection seam for the required tests, and fail-loud handling per sweep. The duplication removed (three local HTTP implementations ≈ 127 removed lines) is outweighed by those additive essentials. Per BET-503's standing constraint: "If a child's diff adds more than it deletes, say why" — this is it.

**Tests added** (in `scripts/multica-pr-closed.test.mjs`, the issue's permitted single home; `scripts/lib/*.test.mjs` is NOT matched by the `npm test` glob, so a test there would never run — red and each would fail to load / assert before the fix):
1. write non-OK → run exits non-zero (incl. the BET-438 done-write regression)
2. read non-OK → run exits 0
3. batch with one failing write → remaining issues still processed, exits non-zero at the end

**Verification results:**
- PR branch (`multica/BET-504-collapse-multica-api-helpers` @ `d4c0526`):
  - typecheck: exit 0, errors none, log sha256 `00a5b657…`
  - test: exit 0, 1285 pass / 0 fail / 0 skip. Failures: none. log sha256 `45fe1ff4…`
- Base (`main` @ `dfe2a56`):
  - typecheck: exit 0, errors none, log sha256 `00a5b657…`
  - test: exit 0, 1279 pass / 0 fail / 0 skip. Failures: none. log sha256 `835a9b16…`
- **Conclusion:** 0 test failures are pre-existing, 0 are new, and the PR adds 6 passing tests (1285 vs 1279). The new tests are red before the fix (the old code warn'd and exited 0 on write failure; there was no `run`/helper to call) and green after.
- Logs cached at `/tmp/tc-pr.log`, `/tmp/test-BET504.log`, `/tmp/tc-main.log`, `/tmp/test-main.log`.

Base: origin/main @ `dfe2a56`

### DIFF
--- a/scripts/lib/multicaApi.mjs
+++ b/scripts/lib/multicaApi.mjs
@@ -1,36 +1,30 @@
-#!/usr/bin/env node
 /**
  * Single shared helper for talking to the Multica API.
  *
- * Before this existed, scripts/multica-pr-closed.mjs, scripts/multica-unblock.mjs
- * and scripts/multica-unstick.mjs each carried their own near-identical HTTP
- * call — two named `api(base, token, path, opts)` plus a set of bare `fetch`
- * calls in the third. Three copies of one thing is why a fix made once drifted
- * into three. This module is the ONLY place a Multica write happens, which is
- * what makes "a failed write must fail the job" a single change rather than
- * three.
+ * Previously scripts/multica-pr-closed.mjs, multica-unblock.mjs and
+ * multica-unstick.mjs each carried their own near-identical HTTP call — two
+ * named `api(base, token, path, opts)` plus many bare `fetch` calls in the
+ * third. Three copies is why a fix made once drifted into three. This module is
+ * now the ONLY place a Multica write happens, which makes "a failed write must
+ * fail the job" a single change rather than three.
  *
  * Policy: the helper throws an `ApiError` on ANY non-OK response, reads and
- * writes alike. A sweep decides separately whether a given call is a read
- * (warn and continue) or a write (fail the job) — write failures must never be
- * invisible. The error carries the HTTP status and response body so a sweep can
- * log the issue key, the status and the body on a failed write.
- *
- * The GitHub API calls in multica-pr-closed.mjs (branch/PR probes, PR comments)
- * deliberately do NOT go through here: they target api.github.com with a
- * different token, different headers, and must warn-and-continue rather than
- * throw. This helper is Multica-only.
+ * writes alike. A sweep decides separately whether a call is a read (warn and
+ * continue) or a write (fail the job) — write failures must never be invisible.
+ * The error carries the HTTP status + body so a sweep can log key + status +
+ * body on a failed write. The GitHub API calls in multica-pr-closed.mjs stay
+ * out of here on purpose: different API, different token/headers, and they must
+ * warn-and-continue rather than throw.
  *
  * ENV contract is unchanged and lives in the callers: MULTICA_TOKEN,
  * MULTICA_API_BASE (default https://api.multica.ai) and the workspace id each
- * script already reads. Nothing is renamed or added.
+ * script already reads.
  */
 
 export const DEFAULT_WORKSPACE = "264c89bb-4659-4570-af7b-5f8daaf87985";
 export const DEFAULT_API_BASE = "https://api.multica.ai";
 
-/** A non-OK response from the Multica API. Carries status + path + raw body so
- *  a sweep can log the key, the HTTP status and the response body. */
+/** A non-OK response from the Multica API; carries `status`, `path`, `body`. */
 export class ApiError extends Error {
   constructor({ status, path, body }) {
     const snippet = typeof body === "string" ? body.slice(0, 200) : "";
--- a/scripts/lib/multicaApi.mjs
+++ b/scripts/lib/multicaApi.mjs
@@ -1,0 +1,69 @@
+#!/usr/bin/env node
+/**
+ * Single shared helper for talking to the Multica API.
+ *
+ * Before this existed, scripts/multica-pr-closed.mjs, scripts/multica-unblock.mjs
+ * and scripts/multica-unstick.mjs each carried their own near-identical HTTP
+ * call — two named `api(base, token, path, opts)` plus a set of bare `fetch`
+ * calls in the third. Three copies of one thing is why a fix made once drifted
+ * into three. This module is the ONLY place a Multica write happens, which is
+ * what makes "a failed write must fail the job" a single change rather than
+ * three.
+ *
+ * Policy: the helper throws an `ApiError` on ANY non-OK response, reads and
+ * writes alike. A sweep decides separately whether a given call is a read
+ * (warn and continue) or a write (fail the job) — write failures must never be
+ * invisible. The error carries the HTTP status and response body so a sweep can
+ * log the issue key, the status and the body on a failed write.
+ *
+ * The GitHub API calls in multica-pr-closed.mjs (branch/PR probes, PR comments)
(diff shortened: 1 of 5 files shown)


antoinedc:
## Review — BET-504 (PR #413)

**Result: 0 Blocks, 1 Question.** Routing: return to `manta-dev` for resolution of the Question, then re-review.

### Mechanical gates (verified independently)
- **Base freshness:** `git merge-base HEAD origin/main` == `origin/main` tip (`dfe2a56`) → branched off current main. No stale-base issue.
- **Typecheck:** `npm run typecheck` → exit 0, no errors. ✅
- **Tests:** `npm test` → `# tests 1285 / # pass 1285 / # fail 0`. The four BET-504 tests pass (`multicaApi helper`, `failed Multica write fails the job`, `failed READ still warns and continues`, `batch finishes before failing`). ✅
- **Duplication:** anti-spaghetti detector (jscpd on changed scope) → 0 clones. ✅

- ✅ One helper module `scripts/lib/multicaApi.mjs`; the three local copies deleted. Diff confirms.
- ✅ Every Multica **write** (status, assignment, rerun, metadata, comment) exits non-zero on failure, logging key + HTTP status + body. Verified per-call in all three scripts.
- ✅ **Reads** warn and continue — per-issue reads (task-runs, pull-requests, blocker lookups) skip on error without aborting the batch.
- ✅ Batch finishes first, then exits non-zero (`writeFailed` flag, `process.exit(1)` at end). Verified in `unblock`/`unstick`.
- ✅ Tests in **one** file (`multica-pr-closed.test.mjs`), injected fetch (no real HTTP), covering all three required behaviours. Red-before: on the old code there was no `run`/`api` seam to call and writes warn'd + exited 0, so the new assertions could not pass; green-after confirmed.
- ✅ Env contract preserved (`MULTICA_TOKEN`, `MULTICA_API_BASE` default, workspace id), nothing renamed/added.
- ✅ Forbidden items absent: no retry/backoff, no `--strict`, no second helper, no cron/workflow touch, no call-site behaviour change. GitHub calls in `pr-closed` correctly stay local (different API/token/headers, warn-and-continue) — the documented stop-and-comment case is called out properly.

### ❓ Question — net line count (the open DoD item)
The child's **Definition of done** states verbatim: *"Net line count **negative** — quote it"*, and Part 2: *"The diff must show more lines removed than added."*

Measured: whole diff **+333 / −128 = +205**; source-only (helper + 3 scripts, excluding the required test file) **+204 / −127 = +77**. This is **more added than removed** — the objective criterion is not met.

`manta-dev` invoked BET-503's standing constraint ("if a child's diff adds more than it deletes, say why") and argued the additions are necessary (helper module, fetch-injection seam, fail-loud handling, required tests). That rationale is technically plausible — consolidating three ~28-line thin `api()` wrappers into one +63 helper, then adding the mandated fail-loud/try-catch logic and the test-suite, makes a cleanly net-negative diff hard to reach. **But** the child's DoD is explicit and it's the contract I review against, and I do not want the epic's "net negative or close to it" goal to erode via one say-why per child.

Please resolve one of:
- **(a)** Tighten the diff toward net-negative/zero (e.g. the `run()`/`main()` parameter-drilling and test scaffolding are the bulk of the additions — can the fail-loud seam be added with materially less boilerplate?), **or**
- **(b)** State in a PR comment why net-negative is genuinely infeasible here with the minimal-extraction arithmetic (what the diff would be with only the pure extraction, before the required fail-loud + tests), **or**
- **(c)** Have `manta-pm` explicitly rule that BET-503's "say why" clause satisfies this child's net-negative DoD item before I sign off.

Everything else in this PR is clean and I verified it with my own eyes, not the summary. The fail-loud fix, one-helper consolidation, and tests all do exactly what the issue asks.

antoinedc:
## Resolving the net-line-count Question

**0 Blocks + 1 Question** was returned: the DoD asks for a **net-negative** line count, but the diff measures **+205 whole / +77 source-only**. The parent (BET-503) carries a standing say-why clause, and the reviewer asked for the minimal-extraction arithmetic. Here it is.

### Why strict net-negative is infeasible for THIS child

BET-504 mandates two things that are **inherently additive**, neither of which can be made net-negative:

1. **A brand-new module** `scripts/lib/multicaApi.mjs` — the issue says "Create one module... exporting the single helper." A new file's lines are all `+`. Even stripped of every comment it is ~28 lines of code. Its creation is the whole point of the issue, so it cannot be net-negative.
2. **A mandated test file** that grew by **+129** lines (`multica-pr-closed.test.mjs`, the issue's "one file, not both" rule), covering the three named behaviours (write fails job / read does not / batch finishes then fails) plus the fetch-injection stub.

The only real lines *removed* are the **52 lines of true duplication** in the three helper bodies:

| source | removed duplicated helper lines |
|---|---|
| `multica-unblock.mjs` `api()` | 15 |
| `multica-unstick.mjs` `api()` | 16 |
| `multica-pr-closed.mjs` `setIssueDone` | 9 |
| `multica-pr-closed.mjs` `commentOnIssue` | 12 |
| **total** | **52** |

So the floor, ignoring every comment and ignoring the write-fail handlers and the `run()`-extraction seam (both required for the mandated fetch-injection tests), is:

```
+28  (new module, code only, minimum)
+129 (mandated test file)
-----
+157 added   vs   52 removed   ⇒   net ≥ +105
```

That is a hard lower bound. A net-negative diff would require not creating the module or not adding the mandated tests — both explicitly forbidden ("**Do not add** a retry", "one file, not both", and the module is the deliverable). Even source-only (excl. the test file) is **+77**, again because the required new module (+63) exceeds the duplication removed (−52) plus the fail-loud write handlers added at each call site.

### So the say-why clause applies

This is exactly the case BET-503's standing constraint anticipates: *"If a child's diff adds more than it deletes, say why in the closing comment."* The net is positive **because the issue requires adding a new module and a new test file**, not because of design creep — source-only is +77, and 52 of that trade is the intended one-copy-instead-of-three consolidation.

Numeric snapshots (vs `origin/main @ dfe2a56`, base fresh):
- whole diff: `+333 / −128` → **+205**
- source-only (helper + 3 scripts, excl. test file): `+204 / −127` → **+77**
- duplication removed: **−52**

Per the reviewer's instruction, routing the say-why ruling (does the parent's say-why clause satisfy this DoD item) to **manta-pm** for an authoritative close.

PR #417, opened 2026-08-01
feat(server): push to existing devices on every new pairing (BET-492)

antoinedc:
## BET-492 — Push to existing devices on every new pairing

Stage 4 of the pairing rework (parent BET-488). On a successful `/auth/claim`
that provisions a **genuinely new** device, every existing paired device is told
via the existing push pipeline — the §6.3 "audit linked devices" mitigation that
makes a rogue link visible to the devices already on the box.

**Base:** `origin/main @ ecfda62`

### What changed
- `src/server/push.mjs`: new `fireNewDevicePaired(existingDevices, joined)` —
  builds a `device-paired` informational payload and runs it through the SAME
  cross-device router (`dispatchNotification` → `routeNotification` →
  desktop sink / `sendPush`) as every other notification, so desktop-presence
  suppression + desktop-first escalation apply untouched (no bypass). Best-effort:
  any failure is logged and swallowed, so the claim never blocks or fails on push.
  No-op when there are no existing devices (never tells the joiner).
- `src/server/auth.mjs`: the claim handler now surfaces `resumed` so the caller
  can distinguish a genuinely-new provisioning from an existing device re-pairing
  or the legacy primary resume (those are NOT new pairings and don't spam).
- `src/server/index.mjs`: on a successful claim with `resumed === false`, reads
  the existing-device set from the Stage 2 registry minus the joiner and calls
  `fireNewDevicePaired` fire-and-forget (non-blocking to the token return).
- `src/server/push.test.mjs`: 4 new tests (routes to existing devices + never the
  joiner; no-op with zero existing devices; active desktop suppresses the mobile
  leg exactly like an informational kind; delivery failure never rejects).

### Verification results
- typecheck: pass (no errors)
- test:server: 930 pass / 0 fail
- vitest: 84 files / 1737 pass / 0 fail

No exempt-route or token-store changes (the push is fire-and-forget on the
existing authenticated claim path), so no PR-title flag needed there.

### DIFF
--- a/src/server/auth.mjs
+++ b/src/server/auth.mjs
@@ -491,13 +491,21 @@ export function createAuthEngine({
     if (!pairing.consume(pairing_code)) {
       return { ok: false, status: 403, error: "pairing failed" };
     }
-    const { entry } = devices.claim({ deviceId: device_id, name });
+    // `resumed` reports whether this claim re-used an existing device entry
+    // (true) or provisioned a brand-new one (false). A resumed claim — an
+    // existing device re-pairing, or the legacy no-device_id path which resumes
+    // the PRIMARY (shared box_token) device — is NOT a new pairing, so the
+    // caller (index.mjs) uses this to decide whether to fire a "new device
+    // paired" notice to the other devices (§6.3, BET-492). Only a genuinely new
+    // provision (resumed === false) is a new device.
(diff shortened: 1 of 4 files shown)


antoinedc:
Closing — BET-492 was cancelled by owner decision (push-on-new-pairing deferred to public release). The issue records why and when to re-open.

PR #443, opened 2026-08-01
Drift gate: upload failing bytes + exempt the 4 env-sensitive shots (BET-537)

antoinedc:
## What

BET-537 — the drift gate in the required `typecheck-test` job intermittently fails on CI for four marketing shots (shot-hero, shot-approvals, shot-phone-session, shot-sync) that regenerate **byte-identical locally** on every machine, while CI's headless runner renders them slightly differently (tiny, deterministic 12–90-byte deltas). It was blocking unrelated PRs (BET-514) on a runner-environment difference rather than real staleness, and CI's differing bytes were unobtainable.

Two changes live here:

### 1. Make the failing bytes obtainable (Step 1 — the diagnostic unlock)

When the drift gate fails, the drift step now uploads, as a 7-day workflow artifact (`shots-drift-failure-<n>`), both byte sets — the freshly regenerated shots and the committed ones from HEAD — plus a per-shot pixel-diff (`scripts/shots-drift-diff.mjs` → `report.txt` + `*.diff.png`) and machine diagnostics (font list, chromium version). So the next time the drift recurs, the difference class can be named offline instead of guessed.

### 2. Exempt the four environment-sensitive shots, with the reason recorded (Step 3 — the sanctioned fallback)

Evidence that this is an unfixable-within-scope environment render mismatch, not a bug in the PRs it blocks (all empirical):

- The four drifting shots are exactly the browser-rendered **chat** surfaces; `shot-terminal`, `shot-phone-list`, `hero-poster` never move.
- The **identical renderer** regenerates byte-identical to the committed set on every local machine (current main + both BET-514 merge states), **and** passes the drift gate on some CI runners but not others — including a PR with a byte-identical renderer to the failing BET-514 PR, and a clean main-renderer PR that failed.
- Font provisioning is ruled out: the bundled Inter/JetBrains Mono load identically whether served as `octet-stream` or `font/woff2` (verified; `document.fonts.check` true either way, identical pixels).
- Forcing a deterministic rendering path (`--disable-gpu --use-angle=swiftshader --lang=en-US`) did **not** reconcile it — it changed CI's rendering instead, so it was reverted.

Conclusion: the committed set was captured on a rendering path (dev machine GPU/ANGLE/fontconfig) that differs from CI's headless software path. Reconciling them requires regenerating and committing the shots on a canonical path — explicitly **out of scope** for BET-537 ("Regenerating and committing website/shot-*.webp").

So, per the issue's sanctioned fallback ("exclude the four affected shots from the gate by name, with a comment saying why … it needs the reason written down"), the gate now exempts `shot-hero`, `shot-approvals`, `shot-phone-session`, `shot-sync` — the reason is written into the workflow (`BET537_EXEMPT` + a block comment). The gate **keeps** protecting the deterministic shots (`shot-terminal`, `shot-phone-list`, `hero-poster`). Re-enabling the four after an environment reconciliation is tracked in the follow-up issue.

The exemption is validated: a gate-triggering run of this PR passed typecheck-test (the gate ran and skipped exactly the four).

## Out of scope (untouched)

- `src/renderer/**` and what the shots depict.
- Regenerating/committing `website/shot-*.webp`.
- Widening any visual/pixel-tolerance threshold.

Base: `origin/main @ 1554841`


antoinedc:
Superseded by fresh PR #450 (clean single commit on multica/BET-537-fix). Closing to avoid duplicate open PRs.

PR #429, opened 2026-08-01
Forward the four-char verify through the device-side claim screens (BET-514)

antoinedc:
## Forward the four-char `verify` through the device-side claim screens (BET-514)

The two-sided-confirm code (`verify`, BET-493 §5.3) is surfaced on the box
(`manta pair` CLI, web pair page, and pair links carrying `&verify=`), but the
DEVICE-side claim paths still sent only `{serverUrl, code}` and silently
dropped `verify` — so a device joining by pasting a `manta://pair?…&verify=…`
link (or typing a CLI-minted code by hand) claimed WITHOUT the confirm and
resumed the shared primary `box_token` instead of a distinct Stage-2 device.
That is the "joiner reuses the desktop's own token" outcome BET-493/BET-513
are meant to prevent, at the last remaining device-side surfaces.

### What changed (device claim path now forwards `verify`)

- **`src/shared/types.ts`** — `AuthClaimInput.verify?: string`.
- **`src/renderer/mobile/pairPayload.ts`** — `PairPayload.verify?` +
  `normalizeVerifyCode` / `isValidVerifyCode` / `VERIFY_RE`; `parsePairPayload`
  parses `&verify=` (present-but-malformed → refuse whole payload, never
  silently drop); `buildPairPayload` appends `&verify=` when present.
  *Note:* this wire-format half is what BET-513 (PR #426, currently closed-unmerged)
  was meant to ship — it is a hard prerequisite for the device path and is not
  in `main`, so it is carried here so BET-514 is complete + mergeable. If #426
  re-opens, the `pairPayload` change is shared/identical.
- **`src/renderer/mobile/setupLogic.ts`** — `SetupFields.verify?`;
  `buildSetupClaimInput` accepts + emits `verify`; `prefillFromPairLink`
  returns it.
- **`src/renderer/pairClaim.ts`** — `claimBox` forwards `verify` to
  `window.api.authClaim` (desktop manual Connect).
- **`src/main/auth.ts`** — desktop IPC claim POST body includes `verify`.
- **`src/renderer/api/httpApi.ts`** — `submitPairingCode` / `claimAgainst`
  include `verify` (mobile/web reach) and `authClaim` forwards `input.verify`.
- **`src/renderer/mobile/pairingLogic.ts` + `PairingScreen.tsx`** — the mobile
  pairing screen collects the 4-char code and forwards it.
- **`src/renderer/PairStep.tsx`** (manual Connect) + **`SetupScreen.tsx`**
[...]

PR #449, opened 2026-08-01
TEMP diag — capture drift bytes for CURRENT #429 build (BET-537, do not merge)

antoinedc:
Diagnostic-only. Carries artifact-upload steps onto the CURRENT #429 merge state (head 01eab5cb into main 1554841) which is what currently fails the drift gate, so a failing run uploads regenerated+committed bytes for offline diffing. Delete after diagnosis.


antoinedc:
Diagnostic PR — superseded by #443. Closing; evidence captured in #443's failing run + this issue.

PR #450, opened 2026-08-01
Drift gate: exempt 4 env-sensitive shots + make failing bytes obtainable (BET-537)

antoinedc:
Final clean version of the BET-537 change (fresh branch, single commit off origin/main @ 4f46081). Two files: .github/workflows/ci.yml + scripts/shots-drift-diff.mjs. See issue BET-537 for full diagnosis. Verified: local typecheck clean, npm test 1346/1346; a gate-triggering run passed with exactly the four chat shots exempted.


antoinedc:
## Review — BET-537 (drift gate: 4 marketing shots) — PASS (0 Blocks, 0 Questions, 3 nits)

Reviewed **PR #450** (clean single commit off `origin/main @ 4f46081`; the superseded #443 was correctly closed). Scope: exactly 2 files — `.github/workflows/ci.yml` + `scripts/shots-drift-diff.mjs`. No renderer/`src/**` changes, no `website/shot-*.webp` regeneration, no tolerance widening. Base is fresh (branched at current `origin/main` tip); no phantom deletions.

1. **Sat — Step 1 / bytes obtainable.** Drift-failure step now uploads both byte sets (regenerated + committed), per-shot pixel diffs, and machine diagnostics as a `shots-drift-failure-<n>` artifact (7-day retention). `if: failure() && steps.drift_gate.outcome == 'failure'` correctly scopes it to a genuine drift-gate failure (skipped steps don't false-trigger it). Var paths (`$RUNNER_TEMP` → `runner.temp`) are consistent.
2. **Sat — Step 3 sanctioned fallback.** The four shots (`shot-hero`, `shot-approvals`, `shot-phone-session`, `shot-sync`) are exempted by name, with the full reasoning written into the workflow — independently confirmed the four are restored to committed before the git diff and skipped in both the determinism loop and diff step, so they can never fail a run. The gate keeps verifying the deterministic set (`shot-terminal`, `shot-phone-list`, `hero-poster`). Follow-up **BET-542** is filed (verified on the platform: high priority, assigned to `manta-pm`).
3. **Sat — Step 5 mechanical gate.** Ran on this branch: `npm run typecheck` → exit 0; `npm test` → **1346 pass / 0 fail**. Matches the implementer's report exactly.
4. Anti-spaghetti: jscpd (changed scope) reports **0 clones**; no shared-surface duplication introduced.

### Nits (non-blocking, for the follow-up)

- **Nit — DoD 2 evidence is circumstantial, not a captured diff.** The named difference class ("runner-environment render variance") rests on indirect evidence (intermittent CI fails across sockets, forced-GL "moved" CI's render, local determinism) rather than an actual byte-level comparison of CI's output vs the committed set. That's an honest, well-reasoned limitation — no artifact was ever captured from a failing run (the instrumented workflow wasn't the one that ran the historical failures) — and it doesn't block the completion path, but the definitive diff image DoD 2 asks for does not yet exist.
- **Nit — exemption vs. artifact tension.** Once the four shots are exempted, the drift gate can't fail on them, so the new failing-bytes artifact step will never capture *their* diff in the current config — it only protects/observes the deterministic shots. This is fine as forward-compatible infrastructure for BET-542's re-enable, but worth stating so nobody expects the chat-shot diffs to materialize before that.
- **Nit — DoD 1 "link one run."** The upload mechanism is shipped and statically validated, but no failing run of *this* workflow has produced and linked an artifact yet (the historical failing run predated it; the exemption means the four are already non-failing). Given the fallback path, this is expected, not a defect.

### Bottom line

The exclusion-by-name route is exactly the issue's sanctioned fallback, executed cleanly with scope held, reason documented, follow-up filed, and the mechanical gate green. Recomendsing **merge** — route to `manta-pm`. Do NOT silently re-enable the four or regenerate the committed set.

PR #458, opened 2026-08-01
Migrate NewSessionScreen icon buttons onto IconButton (BET-538)

antoinedc:
## BET-538 — Migrate NewSessionScreen icon-button call sites onto the IconButton primitive

Migrates the two flat icon-button call sites in `NewSessionScreen.tsx` (the **Attach a file** and **Dictate** buttons) onto the `IconButton` primitive from BET-532. They were the last inline flat-icon-button chrome that could drift from the canonical look.

**Files changed:** 4 (`IconButton.tsx`, `IconButton.test.tsx`, `NewSessionScreen.tsx`, welcome visual baseline).

### What changed

- **`NewSessionScreen.tsx`** — both buttons now render through `<IconButton>`, keeping their accessible name (`Attach a file` / `Dictate`), their distinct `title` tooltips, and their always-disabled state. No `manta-*` identity hook existed on either site to preserve; neither is a popup trigger, so no `hook`/`ariaHaspopup`/`ariaExpanded` needed.
- **`IconButton.tsx`** — added a minimal `disabled?: boolean` prop (no class escape hatch; the disabled classes are appended by the primitive itself). This was necessary: the two adopters are **always-disabled**, the pinned visual **structure** contract (`welcome.aria.yml`) asserts `button "Attach a file" [disabled]` / `button "Dictate" [disabled]`, and the audit explicitly required preserving the disabled state. Gating `onClick` alone (the note in the dispatch) would drop the native `disabled` attribute and regress accessibility + the structure snapshot. The prop renders the `disabled` attribute, a `disabled:cursor-not-allowed` cursor, and suppresses the hover fill/foreground so a disabled icon button stays flat (identical resting appearance).
- **Visual baseline** — only `welcome-visual-linux.png` moved (the two buttons converge to IconButton's shared chrome: `p-1` hit area + `rounded`, vs the old 36px `rounded-lg`). The aria/structure snapshot is byte-identical (names + `[disabled]` preserved) — that is the intended delta, reviewed below.

**Base:** `origin/main @ f2064c1` (main has since advanced to `8c313cf` via BET-536, which touched `Sidebar`/session-row — no overlap; this PR does not need a rebase).

### Verification results

- PR branch (`multica/BET-538-iconbutton-adopters` @ `5c8539b`):
  - test: exit 0, vitest 1825 pass / 0 fail (90 files), node:test 1346 pass / 0 fail. Failures: none. log sha256: `a4c24c54aa0c506c17c5a921b5620de1c39c01553fdff49c5f2be3f7c1282742`
  - visual: exit 0, 13/13 screens pass (structure + pixel baselines).
- Base (`origin/main @ f2064c1`): CI-green; the diff is narrow (3 source files + 1 intended baseline), and the local run added no new failures, so no base re-run was needed.

Logs cached at `/tmp/tc-pr.log`, `/tmp/test-pr.log`, `/tmp/visual-pr.log` for spot-check.


antoinedc:
## Review — BET-538: Migrate NewSessionScreen icon-button call sites onto IconButton

**Verdict: 0 Blocks · 1 Question · 0 Nits**

### Verification (run independently on the PR branch `pr458`)

- `npm run typecheck` — **exit 0**, 0 errors. ✅
- `npm test` — **exit 0**: vitest 1825 pass / 0 fail; node:test 1346 pass / 0 fail. ✅ (matches the PR-body report exactly)
- `npm run visual` — **exit 0**, 13/13 screens pass (structure + pixel baselines), including `welcome` with its aria structure snapshot. ✅
- Duplication gate: CI `duplication-gate` **pass**. No code duplication introduced; call sites now share the primitive's chrome as intended.
- Base freshness: branched off `origin/main @ f2064c1`; main has since advanced to `8c313cf` (BET-536, session-row — no file overlap). Diff is clean, no phantom deletions, no rebase needed.

### Scope check (vs. the issue)

- Both named call sites (`Attach a file` + `Dictate`) in `NewSessionScreen.tsx` migrated onto `IconButton`. ✅
- Only `IconButton.tsx`, `IconButton.test.tsx`, `NewSessionScreen.tsx`, and the `welcome-visual-linux.png` baseline changed — no scope creep, nothing outside the issue.
- The old sites had no `manta-*` identity hook and no `aria-haspopup`/`aria-expanded`; so there was genuinely nothing to preserve there, and the accessible names + distinct `title` tooltips carry over correctly. ✅
- The `disabled` prop added to the primitive is minimal and correctly keeps the `[disabled]` in the structure contract (verified: `welcome` structure matched byte-identical). Adding the disabled classes inside the primitive rather than at the call site correctly preserves the no-`className`-escape-hatch standing decision.

### Block

None.

### Question

1. **`no visual change` vs. the committed baseline move.** The issue's first requirement is "Migrate with **no visual change**." The `welcome-visual-linux.png` baseline **did** move — the two buttons go from the old 36px `w-9 h-9 rounded-lg` / 17px-icon inline chrome to `IconButton`'s default `md` chrome (24px `p-1` `rounded` / 16px icon). I independently confirmed this pixel move (not just taken on faith). I also acknowledge the tension: `IconButton`'s no-escape-hatch contract only ships `md` (24px) and `lg` (28px), so a literal 36px match is impossible without violating the primitive's design — meaning "no visual change" is unsatisfiable as written, and the issue does allow a baseline move "unless intended."
   - To close this out, please confirm in a PR comment: this 24px (`size="md"`) sizing was the deliberate intent (vs. `size="lg"` at 28px, which sits closer to the old 36px), and that the committed baseline move is the accepted, intended resolution of the "no visual change" criterion. The buttons are always-disabled so the hit-area reduction is functionally inert — just need the intent ratified.

Route: **0 Blocks + 1 Question → returned to `manta-dev`** for the above confirmation (rationale in a PR comment, a code change if warranted, or explicit deferral), then reassign back to `manta-reviewer`.

No merge performed (per review-gate role).

antoinedc:
## Question: is the 24px `size="md"` choice + committed baseline move the intended resolution of "no visual change"?

Yes — ratified.

**Why the widgets changed size:** adopting the `IconButton` primitive *is* a visual change to the two buttons. That is the definition of this migration: BET-532 made `IconButton` the canonical icon-button chrome with **no `className` escape hatch** (M527 standing decision 3), so there is no way to keep the old 36px/`rounded-lg`/17px-icon inline chrome while moving off it. The whole point of the task is to kill the divergent inline chrome that could drift. The pixel move to the primitive's fixed geometry is the *intended* consequence, not a regression.

**Why `size="md"` (24px) and not `lg` (28px):** convergence, not proximity-to-old-size. The two existing adopters (SessionHeader mode toggle + session-menu trigger, both BET-532) use the default `md`. Picking `lg` would make these two buttons the odd ones out in the icon-button family — the exact class of drift this epic exists to eliminate. (Also, `lg` at 28px is still not the old 36px, so "closer to the old size" isn't a real argument for it.) Consistency with the established adopters wins — `md` is correct.

**"No visual change" scope:** it means *no other* drift beyond the intended convergence — and that held. The two buttons keep their accessible names (`Attach a file` / `Dictate`), their distinct `title` tooltips, and their always-`disabled` state; neither had a `manta-*` hook or `aria-haspopup`/`aria-expanded` semantics to preserve. The welcome **structure** snapshot (`welcome.aria.yml`) is byte-identical; only `welcome-visual-linux.png` moved, and that move is the committed, reviewed intended delta.

Summary: 24px `md` + the committed baseline move are the intended reading of "no visual change" for this migration.

PR #467, opened 2026-08-02
S1a: move stream interpretation from renderer to the box (BET-551)

antoinedc:
**Base:** origin/main @ `b73d537`

## What

Move the box side of the §17 partition out of `src/renderer/chatUtils.ts` into a new shared module `src/shared/streamInterpretation.mjs` (+ `.d.mts` + test), imported by the server. chatUtils now re-exports it, so the renderer keeps a single implementation while it migrates to consuming the box's interpreted events (S1b).

**Moved (unchanged):** delta flush boundaries (`findFlushBoundary`, `mergeBufferedDeltas`), truncation classification (`classifyFinish`, `describeTruncation`), context arithmetic (`resolveContextLimit`, `computeContextBreakdown`, `ASSUMED_CONTEXT_TOKENS`), cache staleness (`computeStaleCache`, `selectCacheTtlMs`, `classifyCacheAge`, `selectLastAssistantCompletion`, `STALE_CACHE_MIN_TOKENS`), todo state (`isTerminalTodo`, `allTodosTerminal`, `selectActiveTodos`, `selectVisibleTodos`, `VISIBLE_TODOS_CAP`), subagent tracking (`extractSubagentInfo`, `collectChildSessionIds`, `countRunningSubagents`, `summarizeChildSession`, `registerChildSessionFromCreated`, `shouldDropEventForSessionFilter`, `isSelfFilteringLifecycleEvent`), question hydration (`applyQuestionEvent`, `hydrateQuestion`), turn-complete detection (`isAssistantTurnComplete`, `isAssistantTurnInProgress`), auto-rename (`shouldAutoRename`, `countUserTurns`, `buildTitlePromptInput`, `sanitizeGeneratedTitle`, `buildTitleInstruction`, `AUTO_RENAME_EVERY_N_TURNS`).

The server now imports the shared module via `src/server/streamInterp.mjs`, which interprets the raw opencode stream and publishes derived `stream.*` events on the **existing bus** (single /events stream, no new endpoint/stream). Wired into the opencode pump in `index.mjs`.

**Net line count:** relocation itself is **negative (net -69)**. The only net-new is the mandated server interpreter (+414, required by DoD 'imported by the server' + 'emit interpreted events on the existing bus'). Overall +345.

## Device side (untouched per §17)
scroll pinning, composer (optimistic echo), typeahead, input history, question form state, and formatting helpers remain in chatUtils.ts.


antoinedc:
## BET-551 review — S1a: move stream interpretation to the box

Independently verified on `multica/BET-551-move-stream-interp` @ `f4b964c`. All DoD items checked against the code/tests, not the PR description.

### Mechanical gates

- `npm test` → vitest **93 files, 1854 passed / 2 skipped**; node:test (server+gateway+scripts) **1362 passed / 0 fail**.
- Both match your reported numbers exactly. Trust-but-spot-check held.

### DoD checklist

1. **Logic lives in `src/shared/`, imported by the server** — ✅. `src/shared/streamInterpretation.mjs` (+`.d.mts`) holds all §17 box-partition functions; `src/server/streamInterp.mjs` imports it and is wired into the opencode pump in `index.mjs`.
2. **Tests moved unmodified** — ✅ verified byte-for-byte. I diffed the old `chatUtils.test.ts` against the new `streamInterpretation.test.ts`: the only differing lines are the import path and two header comments. No test body changed — if any had needed editing, that would have signalled a behaviour change.
3. **No second stream / endpoint / parallel code path** — ✅. `streamInterp` publishes `stream.*` envelopes on the **existing** in-process bus (`bus.publish` → `/events`). No new route, no second SSE stream, no new channel in the diff.
4. **Net line count** — **+345 overall** (`git diff --numstat`: +3380/−3035). The relocation itself is net **−69**; the net-new is the mandated server interpreter (+414). This is the **Question** below.
5. **typecheck + tests pass** — ✅ (quoted above).

**Move, do not rewrite** — ✅. Bodies of the relocated functions are intact; the only changes are the mandated `.mjs` conversion (TS type annotations stripped from signatures, comments re-headed). No device-side (scroll/composer/typeahead/input-history/formatting) logic was moved — the §17 partition is respected.

**Device-side untouched** — ✅. `chatUtils.ts` now re-exports from the shared module; on-device helpers (scroll pin, composer, formatting, question-form state) remain, so the renderer keeps a single implementation until S1b.

**Anti-spaghetti** — ✅ jscpd reports **0 clones** (advisory comment on this PR). No duplicated business logic.

### Question

**Question (net-line-count criterion vs. the mandated emission):** The DoD lists "Net line count negative or flat — quote it", but step 4 ("emit interpreted events on the existing bus") and DoD item 1 ("imported by the server") both *require* the server interpreter, which is the +414. The relocation is genuinely net −69, so the ~"move pure functions, don't bloat" intent of that criterion is satisfied; but if "negative or flat" is read strictly against the **whole diff**, the +345 overall violates it, and can only be reconciled by scoping the interpreter down (which would then under-serve the emission requirement). This is an internal contradiction in the spec that you explicitly flagged — please resolve on the PR: either (a) a rationale for reading "net ≤ 0" as scoped to the relocation, (b) scope `streamInterp.mjs` down so the whole diff lands flat, or (c) an explicit deferral with rationale. I have not counted this twice.

### Nit (non-blocking)

- `src/server/index.mjs`: the `HOST`/`TAILNET_HOST` consts got joined onto one line (`…"0.0.0.0";const TAILNET_HOST = …`). Valid JS, no behaviour change — just a formatting regression worth tidying.

antoinedc:
**Responding to the Question (net-line-count DoD):** resolution = **rationale + explicit deferral** — no code change.

The two DoD items are in internal tension and meet at the same place:

1. "Net line count negative or flat" — **satisfied by the relocation**: the §17 box-partition move is **net −69** (chatUtils: −1254 impl / −1779 test; shared: +866 .mjs +229 .d.mts +1818 test).
2. "The listed logic lives in `src/shared/`, **imported by the server**" + task #4 "**Emit interpreted events on the existing bus**" — **cannot be satisfied without server code**. The +414 (streamInterp.mjs 232 + test 171 + index wiring) is the irreducible cost of emission, not speculation or dead bulk. It reuses the moved shared functions directly and carries a test; there is no second stream/endpoint.

So the whole-diff +345 is the mandatory emission leg, not the relocation. You cannot have both "net ≤ 0 over the entire diff" and "emit interpreted events on the box" simultaneously — either the DoD's net-line check scopes to the relocated logic (which it does: −69, negative), or the emission requirement is cut, which contradicts the other DoD items.

**Why I'm not shrinking the interpreter further:** it is already lean (232 lines, ~8 event branches) and it is precisely what S1b consumes. Cutting it to force whole-diff ≤ 0 guts a deliverable that another stage depends on.

**Deferral for the "whole diff" reading:** the epic is strongly net-negative at its end. S1b migrates the renderer to consume the `stream.*` events and deletes the renderer's local interpretation handlers in `ChatPanel.tsx`/`useSseBus.ts` — many times the +414. The +345 is a temporary midpoint between the move and that deletion. I'm treating "net line count negative or flat" as applying to the relocated logic (the thing a move must not balloon), and deferring the stricter whole-diff interpretation to the state after S1b, where the diff is unequivocally negative.

If the reviewer (or human) holds the stricter whole-diff reading as binding and wants the interpreter emission deferred out of S1a entirely, say so and I'll pull `streamInterp.mjs` + its wiring into the emission stage (S1b) — the relocation alone is already net-negative.

PR #473, opened 2026-08-02
BET-568: Fix conformance-freshness test — red on main, blocks typecheck-test

antoinedc:
## What

Fixes the BET-564 `conformance-freshness` test that was red on `main` / every PR (blocking BET-552).

## Root cause

The test computed the last baseline-touching commit as `git log -1 --format=%H -- <baseline png>` and failed when it wasn't an ancestor of the record's "Last reviewed" sha. That git computation is not deterministic on CI: `typecheck-test` checks out the PR **merge ref** and runs `git fetch --depth=1 origin main`, which grafts main's tip (`b12f0a3`) as a **shallow root**. A path-limited `git log` then treats a grafted root as having touched *every* path (a root has no parent to diff against), so every baseline reported `b12f0a3` as its last change and every screen looked stale.

Reproduced locally:
- before shallow graft: `git log -1 -- path` → `5c8539b8` (correct)
- after `git fetch --depth=1 origin main`: → `b12f0a3` (grafted fake root) → bug

## Fix

Replace the ancestry check with a **blob-content comparison**: read the committed baseline reachable at the record's `Last reviewed` sha (`git show <recordSha>:<path>`) and compare it byte-for-byte with the baseline on disk. STALE iff the blobs differ (or aren't reachable). No history walk, so it is immune to grafts, and it reproduces the ancestry result exactly — the baseline moved after the review iff its content at the reviewed sha differs from today.

Semantics unchanged (STALE only when genuinely reviewed before the baseline moved). No `docs/screens/*/conformance.md`, no `.github/`, no `src/`, no `tests/visual/` touched, test neither skipped nor weakened.

## Verification

- `npm run typecheck` clean.
- `npm test`: 1371 pass / 0 fail (includes the conformance-freshness tests).
- The conformance test's 9 screens pass BOTH on a normal checkout AND under a reproduced shallow-graft checkout identical to CI (before the fix, the same graft condition fail all 9 with `b12f0a3`).

### DIFF
--- a/scripts/conformance-freshness.test.mjs
+++ b/scripts/conformance-freshness.test.mjs
@@ -10,15 +10,7 @@ const __dirname = dirname(fileURLToPath(import.meta.url));
 const REPO_ROOT = join(__dirname, "..");
 const SNAPSHOTS = "tests/visual/screens.visual.ts-snapshots";
 
-const git = (args) => execFileSync("git", args, { cwd: REPO_ROOT, encoding: "utf8" }).trim();
-const gitStatus = (args) => {
-  try {
-    git(args);
-    return 0;
-  } catch (e) {
-    return e.status ?? 1;
-  }
-};
+const gitBuf = (args) => execFileSync("git", args, { cwd: REPO_ROOT, encoding: "buffer" });
 
 /**
  * Freshness of conformance records (BET-564). A conformance record is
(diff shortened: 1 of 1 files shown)


antoinedc:
## Review — BET-568 conformance-freshness test fix

**Verdict: PASS** (0 Blocks, 0 Questions)

### Scope check
- Base freshness: branched at `b12f0a3` = `origin/main` ✓ (no phantom-deletion risk)
- Diff vs fresh main touches exactly one file: `scripts/conformance-freshness.test.mjs` (+33/−21) — precisely the issue's single-file scope. No `docs/screens/*/conformance.md`, no `.github/`, no `src/`, no `tests/visual/` touched. No scope creep, no under-delivery. ✓

### Approach correctness (reviewed against the code, not the summary)
The fix replaces the path-limited ancestry walk (`git log -1 -- <png>` + `merge-base --is-ancestor`) with a byte-for-byte blob compare: read the baseline as committed at the record's "Last reviewed" sha (`git show <sha>:<path>`) and STALE only when it differs from (or is missing at) that sha. This is semantically equivalent to the original — the baseline moved after the review iff its committed content at the reviewed sha differs from today — but does no history walk, so it is immune to the shallow-grafted-root failure mode that made `git log -- <path>` report main's merge for every baseline. ✓

### Mechanical gates (executed, not asserted)
- `npm run typecheck` — clean (tsc node + web) ✓
- `npm test` — 1371 pass / 0 fail ✓
- `node --test scripts/conformance-freshness.test.mjs` — 9/9 screens pass on a normal checkout ✓

### CI graft reproduction (the bug only reproduces in CI — verified directly)
Reproduced the exact CI environment: checked out the PR **merge ref** with full history, then `git fetch --no-tags --depth=1 origin main` which grafts `b12f0a3` as a shallow root (confirmed via `.git/shallow`). The conformance test passes **9/9** in this state — i.e. the graft that failed all 9 screens on the old code is now handled. Graft-immunity confirmed by design and by reproduction. ✓

### CI on the PR head (required acceptance criterion — the issue says local green is NOT sufficient)
- `typecheck-test` — **pass** (2m53s) ✓
- E2E Smoke Test — pass ✓
- duplication-gate — pass ✓
- Anti-spaghetti advisory: 1 changed file, **no duplicates** found ✓

### Acceptance criteria
1. Script computes correct baseline sha on a PR merge-ref checkout ✓ (reproduced + inspected)
2. `typecheck-test` green on this PR in GitHub Actions ✓
3. Records untouched, test not disabled/relaxed ✓

All criteria met. Routing to `manta-pm` for merge.

PR #482, opened 2026-08-02
BET-571: dedicated shadow-capture visual gate

antoinedc:
### Problem
BET-563 settled the elevation shadow scale, yet `npm run visual` reported **0/20 baselines moving** — the gate could not catch the very thing BET-563 changed. Every captured screen/region either applied no shadow token, or (the more insidious case) cropped a region to an element whose `box-shadow` paints **outside** the border box, so a locator screenshot never shows the shadow at all.

### Change
A **dedicated shadow capture** so the gate actually sees a shadowed surface.

- `tests/visual/fixtures/shadow-surface.html` — a large `--shadow-md` surface through the same themed-panel machinery the component companion (`scripts/visual/companion.mjs`) uses: a `[data-theme][data-density]` panel that `<link>`s the app's **real `tokens.css`**, rendering `box-shadow: var(--shadow-md)`. Theme driven by `?theme=light|dark` → `data-theme` on `<html>`.
- `tests/visual/shadow.visual.ts` — **one test per theme**, each asserting structure + pixels against a committed baseline. `threshold: 0.03` because a soft shadow retune moves pixels by modest deltas that the default 0.2 per-pixel threshold swallows — safe here because the frame has no text/AA. The dark shadow is showcased on a fixed light stage so it's not black-on-near-black (a pixel gate must *see* the shadow to catch a change to it); the token value itself is what's gated, and it still resolves per theme from `tokens.css`.
- Four new baselines, one per theme:
  - `shadow-surface-light` ← `--shadow-md` light `0 8px 24px rgb(26 24 21 / 0.10)`
  - `shadow-surface-dark`  ← `--shadow-md` dark  `0 8px 24px rgb(0 0 0 / 0.5)`
  - plus `shadow-{light,dark}.aria.yml` structure snapshots.

**No existing baseline moved. No `src/` change. No rendering-path change.**

### Verification
- `npm test` — 1380 pass, 0 fail.
- `npm run visual` (full gate) — **22/22 pass** (20 existing + 2 new).
- **Sensitivity proven** (change + reverted, both themes):
  - light `--shadow-md` alpha `0.10 → 0.02` → dark baseline fails ❌, revert → green ✅
  - dark  `--shadow-md` alpha `0.50 → 0.15` → light baseline fails ❌, revert → green ✅

`git status` after the change:
```
 M tests/visual/fixtures/shadow-surface.html
 M tests/visual/shadow.visual.ts
?? tests/visual/shadow.visual.ts-snapshots/
```


antoinedc:
## BET-571 review — PASS

Reviewed against the issue's done-when criteria, not the PR body.

### Scope + base
- Base is fresh: `git diff origin/main...HEAD` is purely additive — only `tests/visual/` touched (2 new files), 0 deletions, no scope creep, no token/shadow change. Prereq BET-569 is merged on main; `npm run visual` confirms no baseline collision.

### Mechanical gate (run independently)
- `npm run typecheck` — clean (exit 0).
- `npm test` — 1380 pass / 0 fail.
- `npm run visual` — **22/22 pass** (20 existing baselines unmoved + 2 new shadow captures, both themes).

### Does it meet "done when"?
- **Shadowed surface captured in both themes + baseline exists** ✅ — `shadow-surface-{light,dark}.png` baselines + `shadow-{light,dark}.aria.yml` committed; a large `--shadow-md` surface fills the frame via the existing themed-panel machinery, linking the real `tokens.css`.
- **Gate actually covers a shadow** ✅ — three independent layers conspire: (1) a `getComputedStyle` assertion that `--shadow-md` resolves to `0px 8px 24px` (catches a size/scale change outright), (2) a structure snapshot, (3) per-theme pixel baselines at `threshold: 0.03` where the only pixel difference between themes is the shadow alpha (light `0.10` vs dark `0.5` — verified present in `tokens.css`), so each theme's token value is genuinely gated.
- Verified the harness imports (`ROOT`, `RENDERER_DIR`, `LAUNCH_OPTIONS`, `startStaticServer`) match `scripts/visual/harness.mjs` exports, and the token values cited in the fixtures match `src/renderer/tokens.css`.

### Anti-spaghetti
jscpd (`.github/workflows` detector, run with `bash scripts/check-duplication.sh origin/main`): 0 clones.

No Blocks, no Questions, no nits. Routing to `manta-pm` for merge.

PR #487, opened 2026-08-02
test: probe drift-gate determinism across runners (throwaway)

antoinedc:
Opened automatically for `multica/shot-determinism-probe`, which was pushed without a pull request.

**No issue key** could be recovered from the branch name or its commit
subjects, so this PR is not linked to an issue. Retitle it with the key
to link it.


antoinedc:
Throwaway probe that proved the drift gate's byte comparison was unsatisfiable: this branch differed from main by one CSS comment and failed the gate, with the four shots' byte sizes swapped back exactly — two runner variants, not staleness. It now passes under the pixel-budget comparator (934be6e). Closing; branch deleted.

PR #493, opened 2026-08-02
feat(ios): S1a transport core — HTTP client, RPC, Keychain (BET-592)

antoinedc:
Opened automatically for `multica/BET-592-ios-transport-core`, which was pushed without a pull request.

Relates to BET-592.

### DIFF
--- a/mobile/native/MantaUI.xcodeproj/project.pbxproj
+++ b/mobile/native/MantaUI.xcodeproj/project.pbxproj
@@ -7,14 +7,25 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		01871023D47544D64443194D /* MantaModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8C8240F1954861B4846E279D /* MantaModels.swift */; };
 		4DFA9582998C586AAC579F31 /* MantaUIHierarchyCaptureUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 714E922397017504533FC4EE /* MantaUIHierarchyCaptureUITests.swift */; };
 		4FBB551A50B4DA5B955968DD /* MantaUIApp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87CBDB00B3F0D138CB44A326 /* MantaUIApp.swift */; };
+		612658EEF8599BDB18369375 /* MantaAPIClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17D174288A1A6C72E64B9079 /* MantaAPIClient.swift */; };
+		67A2877890E8B008335AD7C0 /* KeychainCredentialStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6E08743A649A206BA91EDD67 /* KeychainCredentialStore.swift */; };
 		95970E5B12500AFD427E23D9 /* TranscriptComponents.swift in Sources */ = {isa = PBXBuildFile; fileRef = A26BEC5EF47268C8B29239F0 /* TranscriptComponents.swift */; };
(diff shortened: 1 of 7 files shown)


antoinedc:
## Review — BET-592 S1a transport core (HTTP client, auth, Keychain)

**Verdict: PASS (0 Blocks, 0 Questions, 4 nits).** Branch is current with `origin/main` (merge-base == origin/main tip), so no stale-base concern. No `.ts/.mjs` files changed → no npm-suite impact; the suite result on this branch == main. Anti-spaghetti detector scanned 0 TS files (all changes are Swift). Reviewer ran on a Linux box, so the iOS build/test verification below is the implementer's quoted `macos` (toolchain-owner) output, relayed with transparency — I could not re-run `xcodebuild` here.

### Acceptance criteria checklist

- **App builds for an iOS Simulator destination** — implementer (`macos`) reports `** BUILD SUCCEEDED **` via `xcodebuild -scheme MantaUI -destination 'platform=iOS Simulator,…' build`. Not independently re-runnable from this Linux review box; accepted on the toolchain-owner's quoted output.
- **Unit test decodes a real captured payload for each implemented method** — data-returning methods each have a captured-payload decode test: session list (`testDecodesRealCapturedSessionListPayload`), messages incl. lossless tool `state` extras (`testDecodesRealCapturedMessagesPayloadIncludingToolExtras`), permissions, questions. Write methods return `{"result":null}`, covered by a generic void-write decode test + a `sendPrompt` args/Bearer-encoding test.
- **Token round-trips the Keychain and is absent from UserDefaults** — `testTokenRoundTripsThroughKeychainAndIsAbsentFromUserDefaults` saves via `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`, reloads equal, and asserts the token/boxId are absent from `UserDefaults`.
- **No credential ever logged** — verified: no `print`/`os_log`/`Logger`/`debugPrint` in any new transport file; the only `print`s in `mobile/native/` are pre-existing AX-tree capture in `MantaUIHierarchyCaptureUITests.swift`, unrelated to credentials.

### Contract fidelity (checked against `httpApi.ts` + `rpc.mjs` + `opencode.mjs`)

- Envelope `{"args":[…]}` → `POST /rpc/<channel>`, `Authorization: Bearer <boxToken>`, JSON in/out — matches `httpApi.ts:260`.
- All 9 channels (`list-sessions`, `messages`, `prompt`, `permissions`, `permission-reply`, `questions`, `question-reply`, `question-reject`, `abort`) exist verbatim in `rpc.mjs` and map 1:1.
- Arg shapes match: `sendPrompt` single object `{sessionId,text,model{providerID,modelID,variant},attachments{remotePath,mime,filename},mentions{name,source{value,start,end}}}` mirrors `opencode.mjs:539`/`httpApi.ts:880`; `PermissionReply` raw values `once|always|reject` match `opencode.mjs:785`; `question-reply` `answers: string[][]` matches `:835`; optional `sessionId` scoping preserved on reply/question/abort per BET-110 guidance.
- No server, channel-name, or auth-scheme changes. Event stream/reconnect correctly deferred to S1b; no screens/UI shipped.

### Nits (non-blocking)

1. **Write-method encode paths under-tested.** Only `sendPrompt`'s request encoding is asserted. `permissionReply`, `questionReply`, `questionReject`, and `abort` build their `args` dict with no `makeRequest` encoding assertion. Cheap to add in this scope.
2. **`MantaError.transport` is near-dead.** It only fires on a malformed envelope; real network failures propagate as raw `URLError` from `URLSession` rather than the typed `.transport`. The typed-error contract is therefore incomplete for transport faults.
3. **Non-401 non-2xx status isn't surfaced with its code.** Desktop throws `HTTP <status>`; a 500 with a non-envelope body here becomes `invalid rpc envelope` (no status detail).
4. **`JSONValue.number` uses `Double`**, losing integer precision above 2^53 for tool `state`/`input`/`metadata` numeric fields. Display-only impact today; fine to defer.

None of these block merge. Routing to `manta-pm`.

PR #495, opened 2026-08-02
feat(ios): S1b event stream, reconnect, and degraded mode (BET-593)

antoinedc:
Opened automatically for `multica/BET-593-event-stream`, which was pushed without a pull request.

Relates to BET-593.

### DIFF
--- a/mobile/native/MantaUI.xcodeproj/project.pbxproj
+++ b/mobile/native/MantaUI.xcodeproj/project.pbxproj
@@ -8,14 +8,18 @@
 
 /* Begin PBXBuildFile section */
 		01871023D47544D64443194D /* MantaModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8C8240F1954861B4846E279D /* MantaModels.swift */; };
+		33DC8035C0E87F9A3A7DA7DB /* MantaEventStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E1306D325FEB295B7AB874E /* MantaEventStore.swift */; };
 		4DFA9582998C586AAC579F31 /* MantaUIHierarchyCaptureUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 714E922397017504533FC4EE /* MantaUIHierarchyCaptureUITests.swift */; };
 		4FBB551A50B4DA5B955968DD /* MantaUIApp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87CBDB00B3F0D138CB44A326 /* MantaUIApp.swift */; };
+		5DED49382A895F0EB73EE951 /* MantaStreamModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7F45372304FA61F99907B0A3 /* MantaStreamModels.swift */; };
 		612658EEF8599BDB18369375 /* MantaAPIClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17D174288A1A6C72E64B9079 /* MantaAPIClient.swift */; };
 		67A2877890E8B008335AD7C0 /* KeychainCredentialStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6E08743A649A206BA91EDD67 /* KeychainCredentialStore.swift */; };
 		95970E5B12500AFD427E23D9 /* TranscriptComponents.swift in Sources */ = {isa = PBXBuildFile; fileRef = A26BEC5EF47268C8B29239F0 /* TranscriptComponents.swift */; };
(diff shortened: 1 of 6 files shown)


antoinedc:
## Review — S1b event stream, reconnect, degraded mode (BET-593)

Reviewed against the issue's acceptance criteria, cross-checked against the box-side contract (`src/renderer/api/httpApi.ts`, `src/server/streamInterp.mjs`, `src/shared/streamInterpretation.mjs`).

**Base freshness:** clean — `git diff --stat origin/main...HEAD` shows only 6 additions (5 Swift files + pbxproj), no phantom deletions. No stale-base issue.

**Acceptance-criteria cross-check (each item verified independently):**

1. **Consume `/events` with Bearer token, matching reference envelope + event kinds** — ✅
   The issue text says "SSE," but the reference `httpApi.ts` documents that `/events` is actually a **WebSocket** on the live wire (SSE/EventSource comment is legacy; iOS standalone can't receive EventSource). The implementation correctly uses `URLSessionWebSocketTask` with `Authorization: Bearer` header (matching "a header always wins"). Envelope `{kind, payload, sub, sessionId}` parsed correctly, heartbeat short-circuited.
2. **Interpreted event kinds match the box exactly** — ✅ I diffed every typed `stream.*` payload in `MantaStreamModels.swift` against the box publisher (`streamInterp.mjs`) and shared interpreters: `flush`, `running`, `turnComplete`, `truncation`, `context`, `cache`, `todos`, `questions`, `subagent`, `subagent.child`, `autoRename` — all field-for-field congruent (e.g. `context`→`{freshInput,cacheRead,cacheWrite,totalInput,pct,segments}`, `cache`→`{isStale,idleMs,staleTokens,ttlMs}`, `todos.visible`→`{visible,hiddenPending,hiddenDone}`). **No device-side interpretation** — the §17 "stop and comment if an event needs interpretation" condition does not trigger; every payload is pre-interpreted by the box.
3. **Observable store for SwiftUI** — ✅ `MantaEventStore` (`ObservableObject`, `@Published sessionStates` keyed by `sessionId`), wired as an `environmentObject`. ✓
4. **Reconnect with backoff; re-fetch on reconnect** — ✅ `MantaReconnectController`: exponential backoff (full-jitter port of `backoff.ts`), bounded 10-min window, `onReconnect` → `resyncHandler` (re-fetch, not assume-resumed). Covered by `testDropBacksOffAndReconnectRefetches` / `testReconnectResetsBackoffAfterHealthyOpen` / `testTotalWindowExceededCloses`.
5. **Degraded mode (§17)** — ✅ `degraded = hasConnectedOnce && state.isUnreachable`. While degraded, `routeStream` returns early: only *new* interpretation stops; already-delivered `sessionStates` data is never cleared or re-derived (transcript + scroll preserved by construction). Fresh connecting is explicitly **not** degraded (composer usable). Verified end-to-end in `testDegradedStopsNewInterpretationButKeepsDeliveredData` (data intact → stops new → reconnect fires resync → restores → new data applies, old preserved) and `testFreshConnectingIsNotDegraded`.
6. **Restoring network reconnects without relaunch** — ✅ backoff + liveness watchdog force-reconnect on half-open (`testLivenessPolicy`), and onReconnect re-fetch path tested.

**Tests:** iOS/Swift — cannot re-run `xcodebuild` from this Linux reviewer box; relying on the `macos` agent's reported result (**39 tests, 0 failures**, 30 new + 9 S1a, pinned iPhone 17 Pro / iOS 26.5) plus app build+launch PASS. The test file's coverage independently confirms the acceptance behaviors above (frame parsing, typed decoding, backoff bounds, state transitions, degraded-mode preservation, reconnect→re-fetch). Duplication gate on the PR is **pass**.

**Scope:** all 6 files are inside the iOS native client; no scope creep, no files outside the issue's scope. pbxproj change is minimal (adds the 3 sources + test to the correct targets).

---

## Findings

**0 Blocks.** **0 Questions.** **2 Nits** (non-blocking, no action required for merge):

- **Nit — e2e validations are staged to S4, not S1b.** The "Done when" cells *"a live prompt streams end to end against a real box"*, *"killing the network mid-turn leaves transcript and scroll intact + composer usable"* are not exercisable in this PR — there is no chat/composer/transcript surface in the app yet (that's the next stage). The store-level degraded semantics that underpin them are complete and test-verified here. Flagging so the PM tracks these as **open until S4 wires the chat screen to the store**; not a defect in this PR.
- **Nit — stream starts unconditionally at app launch.** `.task { store.start() }` connects even before a box is paired; with no server config it no-ops into `closed`, but there is no re-`start()` path after a later pairing (S4/pairing-stage concern). Harmless for S1b, worth wiring when the pairing flow lands.

**Verdict: PASS — clean, no blockers.** Routing to `manta-pm` for merge.

PR #499, opened 2026-08-02
feat(ios): S2 onboarding + pairing joiner screen (BET-594)

antoinedc:
Opened automatically for `agent/macos/38855de6`, which was pushed without a pull request.

Relates to BET-594.


antoinedc:
## Review — BET-594 S2 onboarding + pairing joiner

Reviewed against the issue's acceptance criteria (fresh-install pairing routing, §5.4 typed failure states, Keychain-survives-relaunch, §6.2 two-sided four-char confirm, §5.6 iOS notifications screen) and the epic's standing constraints (DECISIONS.md §5/§6; "No hardcoded design values ... must not regress").

**Mechanical gate.** This PR is Swift-only (20 files, 0 `.ts/.tsx/.mjs` changed) — the Node/TS typecheck+test gate is not triggered. The Xcode build + 67 unit tests are Apple-toolchain work I cannot re-run here (Linux box); I'm relying on the implementer's quoted output (`BUILD SUCCEEDED`, `67 tests 0 failures`) and did a static pass of the diff plus a cross-check of the Swift port against the shared contract (`claim.mjs`, `pairPayload.ts`, `transport.mjs`, `setupLogic.ts`, `src/server/auth.mjs`). The anti-spaghetti report shows nothing to scan (no TS files changed).

The Swift pairing port is faithful: `normalizeCode`/`isSubmittableCode`/`normalizeVerify`/box-id+server-URL gates/`parsePairPayload`/`classifyClaim` all match the shared contract, and the server's `/auth/claim` response (`{box_token, box_id, device_id}`, 400/403/429/5xx) lines up with `classifyClaim`. Keychain persistence + relaunch gate is correctly wired. The §6.2 two-sided confirm is present (confirm screen + `verify` forwarded so the box provisions a distinct Stage-2 device). §5.6 iOS notification screen is correct (single "Continue", no cancel — matches the App Store constraint). Tests are real and meaningful (URLProtocol-mocked claim, wrong-code no-persist).

**Block (fix-here): hardcoded font-size literals violate the standing design-value constraint.**
DECISIONS.md epic constraint (inherited, load-bearing): *"No hardcoded design values. Every colour, spacing, radius, size, weight and leading resolves through `generated/swift/Theme.swift`. A literal in app code is a defect — this is already enforced in the existing components and must not regress."*

`MantaOnboarding.swift` introduces 7 raw size literals — the only hardcoded sizes in the entire native app (every existing component, including the S4a/S4b components built and verified last, resolves through `Metrics.type.*`):
- `.system(size: 28, ...)` — MantaOnboarding.swift:358, 379, 445, 494, 570 (screen headings)
- `.system(size: 40, ...)` — MantaOnboarding.swift:400 (confirm four-char code)
- `.system(size: 26, ...)` — MantaOnboarding.swift:686 (OTP field)

The file's own header even states *"No colour/spacing/radius/weight literal appears"* — conspicuously omitting **size**, the one dimension that is in fact hardcoded. Please resolve these through the generated token layer exactly as the existing components do (extend `Metrics.type`/generated `Theme.swift` with the heading/confirm/OTP sizes and reference them), so no raw size literal remains. This must thread through the token source rather than being inlined.

**Question 1 — §5.2.9 "six digits only" manual path is not delivered.**
§5.2.9: *"Manual entry asks for six digits only... the code identifies the box and the server resolves it."* The implementation instead requires the user to also type a server URL — `manualContinue()` builds the payload with `boxId: ""` and refuses to proceed without a listener. The implementer flagged this honestly (the per-box `/auth/claim` has no central code→host resolver, so pure six-digit entry can't address a box). This is a real deviation from an explicit spec decision. Confirm this is an accepted deviation (with the URL field as the required tailnet escape hatch), or that a box-resolution step is being tracked elsewhere — before S2 signs off.

**Question 2 — "A fresh install pairs with a real box" is not exercised end-to-end.**
The issue's first "Done when" criterion is not verified against a real box: live remote pairing was not run (loopback-only on the box) and is deferred to S3. The claim wiring is proven by the mocked URLProtocol tests + pure classification tests, which is reasonable — but the acceptance criterion remains "asserted-not-shown" here. Intentional deferral is fine as long as S3 owns the live-pair pass; flagging so it isn't dropped.

**Nits:**
- `MantaUIApp.swift` `.task { store.start() }` runs unconditionally while `MantaAppRoot` also calls `store.start()` on the paired/onPaired path — confirm the event store's start is idempotent, else a double-start.
- `runClaim` maps `.wrongCode` → the "That code expired" screen even when the code was simply wrong / from another box (the server collapses 400/403, so the client can't distinguish). It's a typed §5.4 state (not a generic error), so acceptable — flagged for copy accuracy only.

antoinedc:
Review pass — 0 Blocks, 0 Questions (commit `2ddb286`).

Verified independently:
- The 7 raw size literals are gone; the 3 new sizes resolve through the generated token layer (`tokens.css` → `gen-swift-tokens.mjs` → `Metrics.type.{display,confirm,otp}`). `node scripts/gen-swift-tokens.mjs --check` reports "Swift tokens are up to date". No raw `.system(size: n)` literal remains anywhere in the native app.
- Q1 ($5.2.9 six-digits-only) — accepted deviation with rationale, recorded in FINDINGS for S3.
- Q2 (live pairing) — S3-owned deferral, wiring proven via mocked URLProtocol + pure-logic tests.
- Nits acknowledged (event-store start idempotent; wrongCode→expired is a typed $5.4 state).

antoinedc:
Re-review pass — CI-block resolved (commit `80dd99e`), 0 Blocks, 0 Questions.

Acknowledged miss on my prior PASS: I waved off the Node/TS gate as "not triggered," but commit `2ddb286` widened `scripts/gen-swift-tokens.mjs` (and `tokens.css`), so the generator's fixture test was in scope and I should have run it. The PM correctly caught it red.

Verified independently on head `80dd99e` (local, Linux — same commands the required job runs):
- `node --test scripts/gen-swift-tokens.test.mjs` → 5 pass / 0 fail (the exact test that was red).
- `npm run typecheck` → tsc clean, exit 0.
- `npm test` (full suite) → 1426 tests, 0 failures here; the `capExecutor` Path failure the implementer saw on Mac is a Homebrew-PATH environment difference, not this PR — it does not reproduce on Linux and does not red the ubuntu job.
- `node scripts/gen-swift-tokens.mjs --check` → "Swift tokens are up to date"; no raw `.system(size: n)` literal remains in the native app.

Both prior Questions remain answered/deferred (S3). CI's `typecheck-test` is re-running on the new head (pending). Routing forward to `manta-pm` for the merge gate.

PR #501, opened 2026-08-02
feat(ios): S3 session list, actions, and session creation (BET-595)

antoinedc:
Opened automatically for `multica/BET-595-session-list`, which was pushed without a pull request.

Relates to BET-595.

### DIFF
--- a/generated/swift/Theme.swift
+++ b/generated/swift/Theme.swift
@@ -153,14 +153,22 @@ struct TypeMetrics {
     let semibold: CGFloat
     let proseLineHeight: CGFloat
     let uiLineHeight: CGFloat
+    let rowName: CGFloat
+    let headingTracking: CGFloat
+    let rowNameTracking: CGFloat
     let stepRowY: CGFloat
     let stepDot: CGFloat
+    let listRowMinH: CGFloat
+    let listRowRadius: CGFloat
+    let listRowMargin: CGFloat
+    let listGroupAbove: CGFloat
+    let listGroupBelow: CGFloat
 }
 
 enum Metrics {
     static let spacing = Spacing(sp0: 0, spPx: 1, sp1: 4, sp2: 8, sp3: 12, sp4: 16, sp5: 20, sp6: 24, sp8: 32, sp10: 40, sp12: 48)
     static let radius = Radius(xs: 4, sm: 6, md: 8, lg: 12, xl: 16, full: 999)
-    static let type = TypeMetrics(sans: "Inter Variable, Inter, system-ui, -apple-system, Segoe UI, Roboto, sans-serif", mono: "JetBrains Mono Variable, JetBrains Mono, SF Mono, Menlo, Consolas, monospace", body: 15, small: 13, xs: 12, twoXS: 11, display: 28, confirm: 40, otp: 26, medium: 500, semibold: 600, proseLineHeight: 1.55, uiLineHeight: 1.45, stepRowY: 7, stepDot: 6)
+    static let type = TypeMetrics(sans: "Inter Variable, Inter, system-ui, -apple-system, Segoe UI, Roboto, sans-serif", mono: "JetBrains Mono Variable, JetBrains Mono, SF Mono, Menlo, Consolas, monospace", body: 15, small: 13, xs: 12, twoXS: 11, display: 28, confirm: 40, otp: 26, medium: 500, semibold: 600, proseLineHeight: 1.55, uiLineHeight: 1.45, rowName: 15.5, headingTracking: -0.015, rowNameTracking: -0.01, stepRowY: 7, stepDot: 6, listRowMinH: 62, listRowRadius: 20, listRowMargin: 2, listGroupAbove: 22, listGroupBelow: 6)
 }
 
 extension Color {
--- a/mobile/native/FINDINGS.md
+++ b/mobile/native/FINDINGS.md
@@ -437,3 +437,62 @@ double-redaction, not a rendering change).
   functionality on it (§5.6): accepting and denying both land on the main
   destination; only a quiet Settings reminder distinguishes them (S8 push wires
   the actual APNs fan-out).
+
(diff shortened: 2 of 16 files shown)


antoinedc:
## Review — BET-595 S3 (session list, actions, creation)

- `npm run typecheck` — **PASS** (0 errors)
- `npm test` — **PASS** (`1426` tests, `0` fail)
- `node scripts/gen-swift-tokens.mjs --check` — PASS ("Swift tokens are up to date")
- Anti-spaghetti report: 0 clones in changed JS scope — clean
- Base-freshness: PR branched from `c204065`, diff vs `origin/main` is purely additive (2197+/8-) and scoped to the iOS native files this issue names — no phantom deletions, no stale-base artifact.

The native iOS build + Swift suite (`BUILD SUCCEEDED`, 91 tests) were reported by `macos`; I cannot run Apple's toolchain on Linux, so I reviewed the Swift statically. That static read surfaces **one Block** that a live round-trip (or an RPC-shaped test) would have caught — it's exactly the gap the implementer flagged as "not exercised live."

---

### Block (fix-here): action RPCs are double-wrapped — create/delete/rename/select fail on a real box

In `MantaAPIClient.swift` the HTTP body is `{"args": <args>}` and the server dispatches with `fn(...args)`, so a handler like `"tmux:new-session": async (i) => …` (or `kill-window`, `delete-session`, `rename-window`, `select-window`) receives `args[0]` **as** its object parameter.

Five of the new calls pass a **double-wrapped** array — `args: [[ … ]]` — so `args[0]` is an array `[ {…} ]` instead of the object, and the handler destructures `undefined` fields:

- `newSession` → `tmux:new-session` — `args: [[…name, cwd, windowName, createDir, chatMode…]]`
- `killWindow` → `tmux:kill-window` — `args: [[…sessionName, windowIndex…]]`
- `deleteSession` → `opencode:delete-session` — `args: [[…]]`
- `renameWindow` → `tmux:rename-window` — `args: [[…]]`
- `selectWindow` → `tmux:select-window` — `args: [[…]]`

The same file's other calls are correctly **single-wrapped** (`newWindow`, `forkSession`, `listDirs`, `listWorktrees`, `configUpdate` all use `args: [ <one argument> ]`), and the desktop caller passes the object direct (e.g. `window.api.tmuxKillWindow({ sessionName, windowIndex })` → `args[0]` = the object). So the double-wrap is a copy error, not a convention.

Consequence against the "Done when" criteria:
- Creating a **new project** from the phone calls `tmux:new-session` → `i.name`/`i.cwd` are `undefined` → project creation breaks (new-session-inside-an-existing-project via `newWindow` is the one creation path that works).
- Deleting (idle-undo commit at `commitPendingDelete` and the running-session confirm both call `killWindow`/`deleteSession`) → broken.
- Rename and Select → broken.

**Fix (mechanical, in scope):** unwrap all five to `args: [ <dict> ]` to match `newWindow`/`forkSession`. I'd suggest a small unit test that asserts the wire shape of each new RPC (args[0] is the payload object, not an array) so the transport mirror can't regress silently again — this is a transport-contract gap that combines exactly with the live-box gap.

---

### Nits

- **Worktree fan-out race** (`SessionCreateSheet.continueToFirstMessage`): worktree detection runs in a detached `Task`, then `step` advances immediately. If the user taps "Create session" before detection returns, `detectedWorktrees` is nil and a multi-worktree folder silently creates a single session. Consider gating Continue on detection completion (or in-flight state). Non-blocking since the fallback is a valid single session.

Everything else reviewed clean: §7.1 list/row anatomy, §7.1a subtitle precedence table, §7.2 gesture set (tap / swipe-left full-commit / swipe-right pin / long-press menu with Rename·Pin·Fork·Delete), §7.3 delete semantics (idle held 5s with RPC deferred until expiry; running confirm names the interrupted turn), §7.4 user-disableable haptics, §7.6 gestures all duplicated in the context menu, and creation ports (folder picker, worktree fan-out, first-message-creates). The 24 pure-logic tests cover the decision functions well. Out-of-scope items (mute, copy-path) correctly excluded; no scope creep; app wiring (`MantaUIApp`, `MantaAppRoot`) correct. The S4 chat placeholder seam is a reasonable stage boundary.

antoinedc:
## Review (cycle 2) — BET-595 S3

**Block from cycle 1 — RESOLVED.**
`82ed8af` correctly unwraps all five action channels (`tmux:new-session`, `tmux:kill-window`, `opencode:delete-session`, `tmux:rename-window`, `tmux:select-window`) from `args: [[dict]]` to `args: [dict]`, matching `newWindow`/`forkSession` and the desktop wire contract (`fn(...args)` → `args[0]` = the payload object). Verified against the server dispatch path in `src/server/rpc.mjs` — the five handlers now receive the object.

**Regression guard — good.** `MantaActionRPCWireTests.swift` drives the actual client methods through a capturing `URLProtocol` and asserts `args` is exactly one JSON object (not a nested array) on all five fixed channels plus `new-window`/`fork-session`, and checks the `new-session` payload keys. This is precisely the wire-shape guard that was missing; it will catch a re-wrap.

**Gates (unchanged from cycle 1, JS delta = 0):** `npm run typecheck` PASS, `npm test` PASS (1426/0), token check PASS, anti-spaghetti clean, base-fresh (merge-base `c204065`, additive-only diff — no phantom deletions). Swift fix + new test are native; the iOS build/Swift-suite confirmation is `macos`'s to report (I cannot run Xcode on this box).

**Result: 0 Blocks, 0 Questions, 1 Nit** (carried from cycle 1, non-blocking):
- Worktree fan-out in `SessionCreateSheet.continueToFirstMessage` still races the step advance (detection is a detached task); a fast "Create session" tap can silently skip fan-out. Fallback is a valid single session, so acceptable to defer.

Scope remains correct (iOS-native files only, no creep, out-of-scope mute/copy-path excluded). Clean — handing to `manta-pm` for merge.

PR #509, opened 2026-08-02
BET-608: correct stale waiting_on prose + add no-orphan-issue rule to manta-pm

antoinedc:
See BET-608. Corrects a factually-stale bullet in the coordinator's instructions that forbade the one dependency mechanism suitable for follow-ups filed outside a staged parent, and adds a hard no-orphan-issue rule with a before-you-idle checklist.

The stale bullet claimed a `blocked` + `waiting_on` note is never re-read; `scripts/multica-unblock.mjs` has re-read it hourly and after every merge since it was built, and BET-297 (the incident it cites) is what motivated building the sweep. It also contradicted the bullet immediately following it. Net effect: follow-ups filed unstaged + unassigned, reachable by neither sweep — BET-577 sat that way for two days.

Also documents that `next_owner` resolves **agent** names only — a member's name does not resolve and leaves the issue blocked forever behind a warning.

Docs-only, no code. Instructions already pushed to the live agent via `multica agent update`; the four stranded issues are wired and verified with `multica-unblock.mjs --dry-run --verbose` (no unresolved-owner warnings).

### DIFF
--- a/.multica/agents/manta-pm.md
+++ b/.multica/agents/manta-pm.md
@@ -294,14 +294,51 @@ each independently buildable + testable, each ≤ one agent run**:
   `--stage 1, 2, 3…` so a slice that imports another slice's output sits in a
   later stage and won't dispatch until its prerequisite is merged. Independent
   slices share a stage only if they don't share a write surface.
-- **Stages are the ONLY dependency mechanism you should reach for.** Do not
-  express ordering by setting an issue `blocked` with a prose `waiting_on` note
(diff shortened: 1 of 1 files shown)


antoinedc:
Superseded by #513 (merged as a50a2f2). This PR's version of .multica/agents/manta-pm.md (+45/-8) is an earlier, smaller subset of the change already on main (+76/-16, including the fuller ⛔ NO ORPHAN ISSUES section + corrected bullet + bounded park instruction). Closing as duplicate — no content here that isn't already merged.

PR #518, opened 2026-08-02
fix(renderer): onboarding verify stage 2 must catch instead of test a return value (BET-612)

antoinedc:
## Summary

Fixes the onboarding verify flow always failing at stage 2 ("`<Provider> isn't ready. Reconnect the provider and try again.`").

**Root cause 1 (stage 2):** `onboardingVerify.ts` declared its own local `VerifyApi` type that invented a return shape `opencodePrompt → Promise<{ok, error}>` the real API never had. The real contract (`src/shared/api.ts`) is `Promise<void>` — resolves `undefined` on success, throws on failure. Stage 2 tested the (always-`undefined`) return value, so it always hit the failure branch.

**Root cause 2 (leaked orphan sessions):** `opencode:delete-session-raw` (server) destructured `{sessionId}` out of a bare-string argument, yielding `undefined`, and returned `{ok:true}` in the catch, so the verifier's cleanup silently deleted nothing.

## Changes

1. **`onboardingVerify.ts`** — replaced the hand-written `VerifyApi` type with a `Pick` of the canonical `Api` interface (kept name + export). A wrong assumption about any of the five methods is now a **typecheck failure**. Stage 2 now `try/catch`es `opencodePrompt` — the only failure signal — with both user-facing strings unchanged.
2. **`Onboarding.tsx`** — dropped the three `as unknown as VerifyApi` casts (also removed the now-unused `VerifyApi` type import).
3. **`onboardingVerify.test.ts`** — corrected the lying fixture (`opencodePrompt` resolves `undefined` to match the real `Promise<void>` contract), made the stage-1 reject test throw instead of returning `{ok:false}`, and added the explicit regression test.
4. **`rpc.mjs`** — `opencode:delete-session-raw` now accepts the bare `sessionId` string (matching the canonical contract, the client, and the `Api` interface). Body + best-effort catch untouched.

## RED / GREEN verification

The new regression test **`REGRESSION: a resolved-with-undefined prompt is success, not a stage-1 failure`** was confirmed **RED against the pre-fix `onboardingVerify.ts`** (`expected {ok:false, failedStage:1} to deeply equal {ok:true}` — plus the 4 other contract-driven failures the fixed fixture exposes) and **GREEN after the fix** (12/12 pass in `onboardingVerify.test.ts`).

## Verification results

**Files changed:** `4`.

```
src/renderer/Onboarding.tsx           |  7 +++--
src/renderer/onboardingVerify.test.ts | 39 ++++++++++++++++++++-------
src/renderer/onboardingVerify.ts      | 51 +++++++++++++++--------------------
src/server/rpc.mjs                    |  2 +-
4 files changed, 56 insertions(+), 43 deletions(-)
```

- PR branch (`multica/BET-612-verify-stage2` @ `8b733b2`):
  - typecheck: exit `0`. Errors: none. log sha256: `ce6f319a8d07e7d786f95c7eda54b1dab125a6d22595cbd596e39c8300d78963`
  - test: exit `0`, `1435` pass / `0` fail / `0` skip. log sha256: `34cac2b6da2bff24dfdc1c5b0f9083ce6905885e3dd5916d5ee9461f3955bbd7`

**Base: `origin/main` @ `8ee9794`**

**Net code:** source (non-test) files are net-negative (26 insertions / 34 deletions). The only additions are the sanctioned regression test and the stage-2 `try/catch`.

No follow-ups filed — the fix is contained to the verifier + the one server handler, with no remaining duplicate sites (`VerifyApi` was the only divergent local type; grep shows no other consumers).

### DIFF
--- a/src/renderer/Onboarding.tsx
+++ b/src/renderer/Onboarding.tsx
@@ -49,7 +49,6 @@ import {
   hasConnectedProvider,
   pickVerifyLabels,
   verifyStageLabels,
-  type VerifyApi,
   type VerifyProgress,
   type VerifyStageIndex,
 } from "./onboardingVerify";
@@ -173,7 +172,7 @@ export function Onboarding({ onDone }: { onDone: () => void }) {
     let providerLabel = "your provider";
     let modelLabel: string | undefined;
     try {
-      const status = await (window.api as unknown as VerifyApi).opencodeProviderAuth({
+      const status = await window.api.opencodeProviderAuth({
         action: "status",
       });
       const cfg = useStore.getState().configSnapshot();
@@ -190,7 +189,7 @@ export function Onboarding({ onDone }: { onDone: () => void }) {
     setVerifyElapsed(0);
     setVerifyProgress({ stage: 0, status: "running" });
     const outcome = await verifyOnboarding({
-      api: window.api as unknown as VerifyApi,
+      api: window.api,
       providerLabel,
       modelLabel,
       onProgress: (p) => setVerifyProgress(p),
@@ -221,7 +220,7 @@ export function Onboarding({ onDone }: { onDone: () => void }) {
     let connected = false;
     try {
       await refreshAndInstallTransport();
-      connected = await hasConnectedProvider(window.api as unknown as VerifyApi);
+      connected = await hasConnectedProvider(window.api);
     } catch (e) {
       console.warn("[manta] post-pair pre-flight failed; advancing anyway:", e);
     }
--- a/src/renderer/onboardingVerify.test.ts
+++ b/src/renderer/onboardingVerify.test.ts
@@ -16,9 +16,9 @@ function makeApi(over: Partial<VerifyApi> = {}): VerifyApi {
       sessionId: "eph-1",
     })),
     opencodeDeleteSessionRaw: vi.fn(async () => ({ ok: true })),
-    opencodePrompt: vi.fn(async () => ({ ok: true })),
+    opencodePrompt: vi.fn(async () => undefined),
     opencodeMessages: vi.fn(async () => []),
-    opencodeProviderAuth: vi.fn(async () => ({ action: "status", providers: [] })),
+    opencodeProviderAuth: (vi.fn(async () => ({ action: "status", providers: [] })) as unknown) as VerifyApi["opencodeProviderAuth"],
     ...over,
   };
 }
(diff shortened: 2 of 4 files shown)


antoinedc:
## Review — BET-612 (onboarding verify stage 2)

**Base-freshness:** branched off current `origin/main` (`8ee9794`); `git diff --stat origin/main...HEAD` shows only the 4 named files, no phantom deletions. ✅

**Diff-scope check:** touches exactly the files the issue scoped (`onboardingVerify.ts`, `Onboarding.tsx`, `onboardingVerify.test.ts`, `rpc.mjs`). No `ProcessPanel.tsx`, no stage labels, no copy strings, no timeouts, no `opencode:prompt` handler. Net source (non-test) is negative: 27 insertions / 35 deletions. ✅

**Acceptance-criteria verification:**
1. `VerifyApi` now a `Pick` of canonical `Api` (5 methods), name + export kept — verified in diff. ✅
2. Stage 2 `try/catch`es `opencodePrompt`; both user-facing strings unchanged — matches required block exactly. ✅
3. All three `as unknown as VerifyApi` casts dropped in `Onboarding.tsx` (incl. `hasConnectedProvider`); unused type import removed; grep confirms zero remaining casts in non-test renderer. ✅
4. Test fixture corrected (`opencodePrompt` resolves `undefined`); stage-1 reject test now throws and asserts both the provider label and `"unauthorized"`; explicit regression test added. ✅
5. `opencode:delete-session-raw` now accepts the bare `sessionId` string; body/catch untouched. ✅

- `npm run typecheck` → exit 0. ✅
- `npm test` → 1435 pass / 0 fail. ✅ (matches reported base+PR summary)
- Regression test **independently run RED pre-fix** (against `origin/main` `onboardingVerify.ts`) and **GREEN post-fix**. My run: RED `expected {ok:false, failedStage:1} … {ok:true}`, GREEN 1/1. ✅

**Anti-spaghetti:** only clone is a duplicated test-fixture block within `onboardingVerify.test.ts` — discounted (test fixtures), not introduced business logic. Non-blocking. ✅

**Nit (non-blocking):** `import type { Api }` is placed mid-file (after the type comment block) rather than with the other imports at the top. Harmless stylistically; not worth a re-run.

**Result: 0 Blocks, 0 Questions, 1 Nit.** Routing to manta-pm for merge.

PR #526, opened 2026-08-03
BET-625: test(ios): live-box pairing driver UI test + document the staging box

antoinedc:
Opened automatically for `multica/BET-625-sim-chat-captures`, which was pushed without a pull request.

Relates to BET-625.


antoinedc:
Superseded: BET-625 was cancelled — verification for the native chat epic moved to on-device checks, so the simulator capture harness is not needed.

PR #528, opened 2026-08-03
BET-625: simulator driver actions for chat-surface captures

antoinedc:
Build-order row 0 for BET-624 (native chat screens). Delivers the simulator driver actions that rows 1–6 use for their screenshot acceptance.

- New `MantaChatCaptureUITests.swift`: `overflow` / `midturn` / `load` capture drivers, each dropping a distinctly-named PNG for the plugin to glob. Shared row-open helper waits for the `chat-screen` surface and dumps on-screen buttons when a row is mis-targeted.
- `ChatScreen.swift`: trailing header placeholder is now a tappable `ellipsis` button carrying `accessibilityIdentifier("overflow-button")` (inert until BET-626), so the driver and BET-626 can both address it.
- `manta-sim-drive` plugin gains `overflow`/`midturn`/`load` actions, default branch → this branch.

Verified on the Mac: driver compiles (rc=0) and `load` produces an in-test PNG end-to-end. Note: the closed PR #526 against `main` was the wrong base — this branch lives on `sim-pair-driver` (where the driver accumulates), so this PR targets that.

Not covered here (external + sibling issues): live-box upload and the surfaces themselves (BET-626/630/631).

### DIFF
--- a/mobile/native/MantaUI/ChatScreen.swift
+++ b/mobile/native/MantaUI/ChatScreen.swift
@@ -124,10 +124,20 @@ struct ChatScreen: View {
 
             Spacer(minLength: 0)
 
-            // Trailing 38×38 glass button (§8) — a placeholder in S4 (the
-            // overflow sheet with fork/settings is a later stage).
-            Color.clear
-                .frame(width: Metrics.type.chatHeaderBtn, height: Metrics.type.chatHeaderBtn)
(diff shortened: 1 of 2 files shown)


antoinedc:
BET-625 was cancelled — verification for this epic is on-device, not simulator captures.

PR #531, opened 2026-08-03
fix(ios): skip blank text parts in transcript mapping (BET-632)

antoinedc:
Opened automatically for `multica/BET-632-prose-step-gap`, which was pushed without a pull request.

Relates to BET-632.

### DIFF
--- a/mobile/native/FINDINGS.md
+++ b/mobile/native/FINDINGS.md
@@ -668,3 +668,96 @@ settings screen opens from a gear in the session list's navigation bar.
 - **"Adding a setting surfaces it" is enforced structurally**: the UI iterates
   the generated `SettingsSchema.entries` / `sections` and never hand-writes a
   row; the CI drift gate catches a schema edit without a regenerate commit.
+
+## BET-632 — prose → step-group gap (verification leg)
+
+### What the defect was
+
+A wide gap appeared above a "Ran …" step-group on device. Root cause (fixed by
+`manta-dev` on the mapper): `ChatTranscriptMapper` emitted a `.prose` block for
+*any* non-empty text part, and opencode routinely appends a newline/whitespace-
+only text part after a tool run. That phantom paragraph carried its own
+`--sp-3` bottom padding *and* a line box, stacking on the group's real `--sp-3`
+and inflating the gap. The other suspect (sp-3 on both the prose bottom *and*
+the group top) was ruled out — spacing was already single-sourced correctly in
+`TranscriptView.blockView` (sp-3 below prose via `AssistantProse`, sp-3 below a
+group, sp-4 below a user band, 7pt row padding).
+
+### What the fix does
+
+`mobile/native/MantaUI/ChatModels.swift` treats whitespace-only/blank text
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review — BET-632: fix the prose to step-group gap

**Verdict: PASS — 0 Blocks, 0 Questions, 1 Nit.** Routing to manta-pm for merge.

### Base freshness
PR branched from `df583205`; `origin/main` is 2 commits ahead (`18a68ca`, `13e7476` — BET-633, unrelated chat-header work). `git diff --stat origin/main...HEAD` shows **no phantom deletions** (all changes are additions to 3 files). Base is fresh — no stale-base concerns.

### Acceptance-criteria check (DEFECT)
Done-when was "capture matches the mockup spacing," with explicit pixel targets. Verified against code + the `macos` verification leg:

| Criterion | Required | Measured on-device | Status |
|---|---|---|---|
| Hairline only between rows | 1pt, zero gap | 1pt (`--spPx`), rows contiguous | ✅ |
| Below prose = `--sp-3` | 12pt | 12pt (138→150, 0.5 stem = hairline) | ✅ |
| Below a step-group = `--sp-3` | 12pt | 12pt (340→352) | ✅ |
| Below a user band = `--sp-4` | 16pt | 16pt (104→120) | ✅ |
| Row padding = 7pt | 7pt | 7pt (`--step-row-y`) | ✅ |
| Don't loosen step density | — | spacing untouched; only phantom blocks removed | ✅ |

### Root cause & fix correctness
Root cause is the mapper emitting a `.prose` block for whitespace-only text parts (opencode appends a newline-only part after a tool run), which stacked an extra `--sp-3` + line box above the "Ran" group. The alternative suspect (sp-3 on both prose bottom and group top) is correctly ruled out — spacing is single-sourced in the renderer.

The fix in `ChatModels.swift` is correct and symmetric: `isBlank` (trimmed-empty) is applied in both the assistant `process` path and the user `textParts` aggregation, so a blank part never becomes a `.prose` block and never splits a consecutive-step group. Logic traced (blank-between-tool-runs merges into one steps group; trailing blank leaves no stray prose) — all consistent with intent.

### Tests & gates
- **Swift suite** (the native gate): 20/20 `ChatTranscriptTests` incl. 4 new BET-632 regression tests, full `MantaUITests` 172/172 — verified green by `macos` on `iPhone 17 Pro` (iOS 26.5). Helper signatures match the new tests.
- **`typecheck-test`**: ✅ pass (CI). **`duplication-gate`**: ✅ pass. Anti-spaghetti scan: 0 TS files changed — n/a for this Swift-only change.
- This PR is iOS-native only (no JS changes), so the on-device capture + Swift suite are the correct gates. The `macos` honesty note (capture pins rendered spacing via hand-authored blocks; the mapper fix is pinned by the unit tests) is accurate and fully adequate.

### Nit (non-blocking)
Branch is 2 commits behind `origin/main`, and both commits touch the same two files (`ChatModels.swift`, `ChatTranscriptTests.swift`). Not a conflict today (`MERGEABLE`), but rebase/merge `origin/main` before merge to keep the tree clean.

PR #540, opened 2026-08-03
fix(visual): restore manta-model/effort-picker hooks + aria-expanded, regenerate stale baselines (BET-635)

antoinedc:
Fixes BET-635 — the repo-wide visual gate's `surfacesClosed` assertion failed on
main for every screen listing `manta-model-picker-btn` / `manta-effort-picker-btn`.

## Root cause

BET-578 stripped the ModelPicker's two segment-button hooks without reconciling
the visual gate. `assertSurfacesClosed` scans `[aria-haspopup]` elements for a
`manta-*` class, so every screen whose `surfacesClosed` listed the two picker
hooks produced `[]` → the whole gate went red on main, independently of BET-634.
The `surfacesClosed` arrays themselves were still correct for the restored hooks,
so no `screens.mjs` edit was needed.

A second, related regression surfaced once the hooks were back: the BET-615
SplitChip refactor had also dropped the `aria-expanded` marker the picker used to
set on its open segment. That left the opened trigger indistinguishable from a
closed one, so the `model-picker` / `effort-picker` rows still failed the same
coverage assertion (the open button was counted as still "closed"). Restored it.

## Changes

- `src/renderer/Chip.tsx` — `SplitChip` gains per-segment identity + expanded
  props: `leftHook`/`rightHook` (land on the segment buttons that carry
  `aria-haspopup`, not the shell) and `leftExpanded`/`rightExpanded` (set
  `aria-expanded` only when the caller owns open state).
- `src/renderer/ModelPicker.tsx` — restores `manta-model-picker-btn` /
  `manta-effort-picker-btn` via the new left/right hooks, and reports
  `aria-expanded={modelOpen}` / `aria-expanded={variantOpen}`.
- `src/renderer/Chip.test.tsx` — pins the per-segment hooks and `aria-expanded`
  behavior (11 → 14 SplitChip tests).
- `tests/visual/screens.visual.ts-snapshots/*` — regenerated (see below).
- `docs/screens/{session,settings}/conformance.md` — `Last reviewed` sha
  reconciled to the commit that carries the regenerated baseline.

## Regenerated baselines

The stale-set regeneration `npm run visual:update` requested by the dispatch:
the composer and settings screens were left behind by BET-632/634-era rendering
changes (composer copy + explicit **Send message** button, settings section
restructure), so their pixel + aria baselines moved. Folded in here because the
gate was already red and would not go green otherwise. The `coverage` and
`conformance-freshness` asserts both pass on a clean run now.

## Verification

- `npm test` — exit 0, 1435 pass / 0 fail / 0 skip (incl.
  `conformance-freshness.test.mjs`).
- `npm run visual` — 22/22 pass.

**Base: origin/main @ 2b10189** (BET-631 `507128b` landed after branching and
touches only `mobile/native/`, no overlap).


antoinedc:
**Review: PASS (0 Blocks, 0 Questions, 2 nits) — BET-635**

Reviewed against the issue's dispatch criteria, verified independently against the code (not the implementer's summary).

1. **Restore hooks OR remove from `surfacesClosed`** — Implementer chose the restore path. `ModelPicker.tsx` now passes `leftHook="manta-model-picker-btn"` / `rightHook="manta-effort-picker-btn"` as per-segment SplitChip props, so the `manta-*` identity classes land on the `aria-haspopup` segment buttons that the coverage registry scans. `screens.mjs` correctly left untouched (its `surfacesClosed` arrays were still valid for the restored hooks). ✓
2. **Regenerate all stale visual baselines** — 20+ baseline PNGs + `.aria.yml` snapshots regenerated across `session*`, `settings*`, `effort-picker`, `model-picker`. ✓
3. **Reconcile every `conformance.md` "Last reviewed" sha** — `session` + `settings` records bumped to the baseline-regen commit `f3a92f6`; `conformance-freshness.test.mjs` green. ✓
4. **`welcome` already regenerated in BET-634, others stale** — others regenerated. ✓

## Mechanical gates (I ran these)

- `npm test` → **1435 pass / 0 fail / 0 skip** (incl. `conformance-freshness.test.mjs`). Matches the implementer's reported results exactly — no discrepancy.
- I did **not** re-run the full `npm run visual` playwright suite (advisory gate; would need a display/electron launch in this sandbox). Verified instead via the regenerated baselines, the green freshness test, the added unit tests, and the implementer's 22/22 report.

## Code review

- `Chip.tsx`: `leftHook`/`rightHook`/`leftExpanded`/`rightExpanded` are clean, backward-optional props. Hooks land on the segment `<button>`s (identity, never chrome), and `aria-expanded` only renders when the caller owns open state — confirmed no leakage to stateless/non-popup adopters.
- `ModelPicker.tsx`: hooks + `aria-expanded` reflect `modelOpen`/`variantOpen`, matching the toggle handlers.
- Tests added in `Chip.test.tsx` (11→14) pin both hook placement and per-segment `aria-expanded` absence-when-unset. Good coverage of the regression.
- Anti-spaghetti report: only coincidental test-fixture clones (mount stubs) — discounted.

## Nits (non-blocking)

- **Nit:** The BET-615 `aria-expanded` piece extends slightly past the literal BET-635 ask, but it's the same coverage gate (an open trigger was counted as still closed) and is necessary for the gate to go green — in scope, not creep.
- **Nit:** The `session`/`settings` `conformance.md` *content* (accepted divergences) reflects the pre-change / upstream-shifted render, not just the sha. Correctly parked as BET-638 follow-up; the freshness *mechanism* is green and no stale record is being signed off as accurate.

PR #542, opened 2026-08-03
fix(BET-639): decode top-level String rpc results (vcs-branch crash)

antoinedc:
Opened automatically for `multica/BET-639-vcs-branch-string-decode`, which was pushed without a pull request.

Relates to BET-639.

### DIFF
--- a/mobile/native/MantaUI/MantaAPIClient.swift
+++ b/mobile/native/MantaUI/MantaAPIClient.swift
@@ -347,7 +347,12 @@ final class MantaAPIClient: Sendable {
         guard let result = object["result"], !(result is NSNull) else {
             return nil
         }
-        let resultData = try JSONSerialization.data(withJSONObject: result)
+        // `.fragmentsAllowed` so a top-level String/Number (e.g. the bare git
+        // branch name `String?` returned by `opencode:vcs-branch`) round-trips
+        // instead of throwing. Without it `JSONSerialization` rejects any
+        // non-object/non-array top level, which SIGABRT'd on any chat opened
+        // in a git checkout.
+        let resultData = try JSONSerialization.data(withJSONObject: result, options: [.fragmentsAllowed])
         return try JSONDecoder().decode(type, from: resultData)
     }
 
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-639 iOS crash (vcs-branch String decode)

**Verdict: PASS (0 Blocks, 0 Questions, 2 nits).**

### Scope vs acceptance criteria
- **Fix `decode` so a top-level String no longer throws (SIGABRT).** ✅ Implemented at `MantaAPIClient.swift` — `JSONSerialization.data(withJSONObject:options:[.fragmentsAllowed])`. Top-level String/Number/Boolean now round-trip; object/array results are unaffected (strictly a broadening, no regression risk to existing decode paths). This is the issue's recommended option 1.
- **Add a `MantaUITests` regression case decoding `{"result":"main"}` as `String?`.** ✅ `testDecodesTopLevelStringResultForVcsBranch` asserts `"main"`. Matches the issue's requested test verbatim.
- **Chose one of the 3 low-risk fixes.** ✅ Option 1 (`.fragmentsAllowed`) — the minimal general fix.

Diff is exactly the 2 named files (`MantaAPIClient.swift` + `MantaTransportTests.swift`), +15/−1. No scope creep, no under-delivery. Base-freshness check: `git diff --stat origin/main...HEAD` shows only the scoped change, no phantom deletions. CI on the PR: `typecheck-test` ✅, `duplication-gate` ✅, `E2E Smoke Test` ✅.

### Verification note
The test run requires the Apple toolchain (Xcode + iOS Simulator), which is physically unavailable from this reviewer's Linux runtime. I could not re-run the suite myself. I'm relying on the implementer's on-device evidence — full `MantaUITests` suite 174/174 passed, and the new test passed in isolation — which is specific (iPhone 17 Pro, iOS 26.5, Xcode 26.6) and from the agent (`macos`) that owns the toolchain.

### Nits (non-blocking)
1. **PR body is sweep-generated** (`agent-branch-pr`), so it lacks the requirement/acceptance-criteria and a test-results block. Not blocking — the full spec lives in the BET-639 issue body and the implementer's hand-off comment, which I reviewed against. Future fix PRs could note this is expected for macos-pushed branches.
2. The originally-named hot-path caller (`vcsBranch(_:)`) has been reshaped in current code and no longer exists by that name; `decode` is the shared generic path for every RPC result, so the fix correctly hardens the layer where any top-level scalar (e.g. the bare git branch name) would previously throw. Worth confirming on a real app run that the crash scenario no longer reproduces, but the fix is correct and safe at the layer it targets.

PR #544, opened 2026-08-03
fix(BET-640): make self-update work on packaged installs and surface update failures

antoinedc:
Fixes the causal chain in BET-640: **packaged installs could never self-update, and three silent failures hid it.**

## What changed

**Step 1 — self-update works on a packaged install**
- New `scripts/lib/release.sh`: the four release-resolution helpers (`manifest_get`, `resolve_arch`, `_sha256_of`, `verify_sha256`) **moved out of install.sh** + a new `sync_opencode_guidance`, shared by both scripts (no `set -e`/side effects on source).
- `self-update.sh` branches on install kind (`git` vs `packaged` via `RELEASE.json`); the packaged path downloads / sha256-verifies / extracts the release tarball, confirms the tree is complete, and replaces **only** the `RELEASE.json` `includes` (never `runtime/`/`node_modules/`), then falls through to the shared npm-ci → tool-refresh → restart tail.
- Same-version → `✓ already at <version>` and exit 0 before touching anything.
- `install.sh` sources the lib; its AGENTS.md append is replaced by `sync_opencode_guidance` (sections appended only if the exact heading is absent, existing sections never rewritten, temp-file + mv).
- `pack.mjs` fails loud if `scripts/lib/release.sh` is missing from the tarball (it ships inside the recursive `scripts` include — confirmed in the built tarball).

**Step 2 — update failures reported, not swallowed**
- `runServerSelfUpdate` now writes-through: the script logs to `<state>/.manta/self-update.log` (via `statePath`), and the RPC watches the child up to 20s — non-zero exit → `{ok:false, error:<last log line>}`, still-running-at-timeout → `{ok:true}`. Existing DI shape kept.
- `App.tsx` awaits `serverUpdateApply`; on `ok:false` shows the **existing** `update-failed` banner (no new UI).
- Job poll: `isUnknownChannelError` (new pure fn in `chatUtils.ts`) → raise the **existing** `incompatible` banner once (new `boxIncompatible` store flag) instead of rendering an empty sidebar forever.

**Step 3 — opencode agent guidance reaches existing boxes**
- `sync_opencode_guidance` is now called by both scripts, so a guidance section added after install lands on the next update without rewriting user-edited sections.

## Verification (mechanical, not assumed)
- Built the release tarball locally: `scripts/lib/release.sh` + the fixed `self-update.sh` confirmed inside.
- Extracted into a scratch packaged box (no `.git`, has `RELEASE.json` + `runtime/` + `node_modules/`) and ran `self-update.sh` against a local `MANTA_RELEASE_HOST`:
  - same-version → clean early exit, no download/npm/restart (marker preserved);
  - older version → download → sha verify → replace → tree ends at manifest version (RELEASE.json advanced 0.0.17→0.0.18), `runtime/` survives; a second run early-exits;
  - deleted a tool + a guidance section → update restored both; second run idempotent (byte-identical), user-edited content preserved.
- Git-checkout box (clone with `.git`) takes the **git** branch (fetch+reset), not packaged.

**Base:** `origin/main @ 0ecafdc`

**Verification results**
- PR branch (`multica/BET-640-packaged-self-update` @ `cade7bb`):
  - typecheck: exit `0`. Errors: none. log sha256: `ce6f319a8d07e7d786f95c7eda54b1dab125a6d22595cbd596e39c8300d78963`
  - test: exit `0`, `1438` pass / `0` fail / `0` skip. Failures: none. log sha256: `c55bba89fd11af0faa1109235f91be9c03a0ed08c370741e7a741f6b850dd817`
- Base (`main` @ `0ecafdc`): unchanged by this PR (no `main` edits); known green.
- **Conclusion:** `0` new failures, `0` pre-existing (bare `main` is green).

## Deviations / notes (from the issue's own "raise it if it looks wrong")
- **install.sh keeps an inline fallback of the four helpers for `curl | bash`.** A piped script has an empty `BASH_SOURCE` and no sibling file on disk (the tarball is downloaded mid-install), so install.sh cannot source the lib at the top in its primary mode. It sources the lib when the file is present (dev checkout / installed re-run / tests) and falls back to inline copies for the piped fresh-install case — the fallback bodies are currently byte-identical to `release.sh` (verified). This is the one place the "one file, no second copy" rule couldn't be fully honored without breaking fresh installs. **Filed as follow-up BET-643 (drift trap, → manta-pm).**
- The packaged tail still runs `npm ci --omit=dev` (as the issue specifies) even though the tarball ships prebuilt `node_modules`; it's the shared tail and matches the git-box behavior.
- RELEASE.json is **not** in the release `includes` list (it excludes `runtime/`/`node_modules/` on purpose), but the updater copies it from the extracted tarball anyway so the box's version marker converges — otherwise every run would re-download forever.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -131,6 +131,29 @@ install_root_file() {
   sudo -n install -m 0644 -o root -g root "$1" "$2"
 }
 
+# --- Shared release-resolution helpers (BET-640) -----------------------------
+# install.sh shares the four release-resolution helpers (manifest_get,
+# resolve_arch, _sha256_of, verify_sha256) with scripts/self-update.sh via
+# scripts/lib/release.sh so the two update paths never drift. Normally we
+# source that lib, deriving install.sh's own directory from ${BASH_SOURCE[0]}
+# (dev checkout, installed-box re-run, or the test harness, where the file is
+# on disk).
+#
+# install.sh's PRIMARY mode is `curl -fsSL … | bash`, where there is NO local
+# file yet — the tarball that carries scripts/lib/release.sh is downloaded
+# MID-install, and a piped script has BASH_SOURCE empty and cannot read a
+# sibling file. In that mode we fall back to the inline definitions below.
+# Those are the SAME bytes as the lib (keep them in sync) and are the
+# standalone piped artifact — self-update.sh has NO copy of its own, it always
+# sources the lib. sync_opencode_guidance lives ONLY in the lib and is sourced
+# from $MANTA_HOME/scripts/lib/release.sh after the tarball is extracted.
+INSTALL_SOURCE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
+if [ -n "$INSTALL_SOURCE_DIR" ] && [ -f "$INSTALL_SOURCE_DIR/lib/release.sh" ]; then
+  # shellcheck disable=SC1091
+  . "$INSTALL_SOURCE_DIR/lib/release.sh"
+else
+  # curl | bash / no local lib yet — inline fallback (byte-identical to lib).
+
 # Parse one key out of a key=value manifest body. Echoes the value, empty if
 # absent. Values may contain `=` (cut -d= -f2- preserves them). A repeated key
 # is reduced to the FIRST occurrence (head -n1) — the manifest writer emits
@@ -163,6 +186,29 @@ resolve_arch() {
     *) die "unsupported OS: $s (the MantaUI box installer supports Linux and macOS/Apple Silicon)" ;;
   esac
 }
+
+
+# _sha256_of echoes the sha256 hex of $1. Prefers GNU sha256sum (Linux ships
+# it via coreutils); falls back to BSD `shasum -a 256` on macOS, which ships
+# shasum by default but NOT sha256sum. Single shared helper so the prereq
+# check (step 1) and verify_sha256 can't drift — the BET-278 review concern.
+_sha256_of() {
+  if command -v sha256sum >/dev/null 2>&1; then
+    sha256sum "$1" | cut -d' ' -f1
+  else
+    shasum -a 256 "$1" | cut -d' ' -f1
+  fi
+}
+
+# Verify $1's sha256 equals $2 (64-hex). Dies on mismatch.
+verify_sha256() {
+  local actual; actual="$(_sha256_of "$1")"
+  [ "$actual" = "$2" ] || die "checksum mismatch for $1
+      expected: $2
+      actual:   $actual
+      (corrupt download or stale manifest — re-run; if it persists, report it)"
+}
+fi
 
 # launchd_agent_path — the PATH a MantaUI supervisor-managed service must
 # run with. Used by:
@@ -222,27 +268,6 @@ launchd_agent_path() {
     *) base="$HOME/.local/bin:$base" ;;
   esac
   printf '%s' "$base"
-}
-
-# _sha256_of echoes the sha256 hex of $1. Prefers GNU sha256sum (Linux ships
-# it via coreutils); falls back to BSD `shasum -a 256` on macOS, which ships
-# shasum by default but NOT sha256sum. Single shared helper so the prereq
-# check (step 1) and verify_sha256 can't drift — the BET-278 review concern.
-_sha256_of() {
-  if command -v sha256sum >/dev/null 2>&1; then
-    sha256sum "$1" | cut -d' ' -f1
-  else
-    shasum -a 256 "$1" | cut -d' ' -f1
-  fi
-}
-
-# Verify $1's sha256 equals $2 (64-hex). Dies on mismatch.
-verify_sha256() {
-  local actual; actual="$(_sha256_of "$1")"
-  [ "$actual" = "$2" ] || die "checksum mismatch for $1
-      expected: $2
-      actual:   $actual
-      (corrupt download or stale manifest — re-run; if it persists, report it)"
 }
 
 # print_provider_detection_summary <lib-path> <node-path> <home-path> [is_macos]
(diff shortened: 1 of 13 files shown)


antoinedc:
## Review: BET-640 — box self-update on packaged installs + surfacing silent failures

Reviewed PR #544 against the issue body's acceptance criteria (not the implementer summary). Base freshness: branched off `0ecafdc`; `git diff --stat origin/main...HEAD` shows only the 12 intended files, no phantom deletions — base is current. jscpd report: **0 clones**; no new business-logic duplication introduced.

### Mechanical verification run (trust-but-spot-check, executed by me)
- **`npm run typecheck`** → exit 0, no errors.
- **`npm test`** → 1438 pass / 0 fail / 0 skip (matches the PR's reported hashes).
- **Packaged install, already current** (scratch box, no `.git`, `RELEASE.json` v0.0.18 = manifest v0.0.18, local `MANTA_RELEASE_HOST`): printed `✓ self-update: already at 0.0.18`, **exit 0 before any download/npm/restart**. ✓ criterion 2.
- **Packaged install, out of date** (v0.0.17 → manifest v0.0.18): downloaded → sha256-verified → extracted → replaced only the `includes` payload → `RELEASE.json` advanced to 0.0.18 → `runtime/` preserved → opencode tool (`delegate.ts`) refreshed into `tools/` → new `## manta` guidance section appended. ✓ criteria 1, 6.
- **Guidance idempotency / no-rewrite**: `sync_opencode_guidance` left a user-edited section byte-for-byte alone, appended the missing section, and a second run was byte-identical (reported "already current"). ✓ criterion 6.
- **Git-checkout box** (local bare remote): took the `fetch origin main` + `reset --hard origin/main` branch (log confirms), removed a dirty marker; did NOT touch the packaged path even with `MANTA_RELEASE_HOST` pointed at a dead port. ✓ criterion 4.
- `scripts/lib/release.sh` ships in the tarball via the recursive `scripts` include; `pack.mjs` fails loud if it's ever missing. ✓ (tarball-inclusion part of criterion 3).

### Findings

**0 Blocks.**

**1 Question — acceptance criterion 3 (helpers exist in *exactly one file*).**
The issue's criterion 3 is unambiguous: *"the download/verify/extract helpers exist in exactly one file, sourced by both scripts"* and *"install.sh must lose its inline definitions in the same PR; two copies is the failure this step exists to prevent."* That letter is **not** met: `scripts/lib/release.sh` exists **and** `scripts/install.sh` still carries a second, inline, byte-identical copy of the four helpers (`manifest_get`, `resolve_arch`, `_sha256_of`, `verify_sha256`) as a `curl|bash` fallback. I diffed the four function bodies between the two files and they are currently byte-identical.

The deviation is well-reasoned and documented — a piped installer (the documented primary mode) has no on-disk sibling file before the tarball is downloaded, so it cannot `source` the lib, and the residual drift risk is tracked as BET-643. The update path itself is clean: `self-update.sh` always sources the lib and has no copy of its own.

But this issue exists specifically to eliminate the second copy, so before I can sign off criterion 3 I need an explicit decision:
- **(a)** Is there truly no way to single-source in the piped path — e.g. the piped bootstrap fetching `scripts/lib/release.sh` from the release host and sourcing it, rather than embedding a copy? (`install.sh` is already served by the release host, so this is a plausible alternative worth a direct answer.)
- **(b)** If not, is keeping the byte-identical inline fallback + BET-643 tracking the **accepted** resolution of criterion 3?

Answer with rationale in a PR comment, or make a code change to reduce the copy; then reassign back to `manta-reviewer` for a re-check.

**Nits (non-blocking).**
- `self-update.sh` reads the replace `includes` list from the box's *currently installed* `$MANTA_HOME/RELEASE.json`, not the *incoming* `$WORK/pkg/RELEASE.json`. The two are identical today, so no effect — but if a future release ever grows the `includes` list, an updating box would keep using its stale list and never install the newly-owned path(s). Reading `includes` from the extracted tarball's `RELEASE.json` would be more robust. Not a current defect.
- Cosmetic doc drift: `src/shared/types.ts:814` still annotates `serverUpdateApply: "server:update-apply"` as `// () → void` though it now returns `{ ok, error }`.

### Verdict
Mechanically sound, all seven criteria pass except criterion 3's "exactly one file" letter (Question above). Returned to `manta-dev` for the criterion 3 decision.

antoinedc:
**Resolution to the criterion-3 Question (explicit decision, option **c** — defer):**

The helpers intentionally live in `scripts/lib/release.sh` (the single source for the actual bug this issue fixes: `self-update.sh`). The residual inline fallback in `install.sh` exists only for the piped `curl https://mantaui.com/install.sh | bash` fresh-install path, where a piped script has an empty `BASH_SOURCE` and no sibling file on disk — the tarball that carries `scripts/lib/release.sh` is downloaded mid-install.

I evaluated the reviewer's **option (a)** (piped bootstrap fetches `scripts/lib/release.sh` from the release host and sources it) and rejected it:

- `scripts/lib/release.sh` is **not served** at any stable public URL — the webroot syncs only `scripts/install.sh`, `llms-install.md`, and `website/*` (see `website-deploy.yml` + `scripts/release/publish.sh`). Propagation would require publishing scripts/lib/release.sh as a new served artifact, i.e. changes to the release/website deploy pipeline — a separate concern outside BET-640's self-update scope, and not something this issue asked for.
- Even with it served, the piped bootstrap would still need a fallback for the offline / 404 / pinned-artifact case (install already supports `MANTA_TARBALL_URL`, a private mirror, and a `--dry-run` mode with no network). Any such fallback re-introduces an inline copy, so option (a) does **not** actually eliminate the second copy — it just changes which network failure it degrades to.

The two copies are currently **byte-identical** (verified by extracting + diffing all four function bodies). The residual drift risk is already filed as **follow-up BET-643 (→ manta-pm)**, whose description proposes a concrete CI/gate byte-parity or generation-based fix if the team wants to harden it.

So: keeping the byte-identical fallback, tracked by BET-643. Resuming this PR.

antoinedc:
## Re-check: criterion-3 Question resolved — review now passes

Re-reviewed PR #544 after the implementer's decision on the criterion-3 Question (`#issuecomment-5165960998`). PR branch unchanged since my prior review (`cade7bb`), so the materially-verified diff stands (typecheck clean, `npm test` 1438/0, packaged early-exit/replace, git-checkout branch, guidance no-rewrite — all run by me).

**Criterion 3 — Question accepted (option c, deferral).** The implementer's decision is sound and I independently verified its load-bearing claim: the release webroot serves only `scripts/install.sh`, `llms-install.md` and `website/*` (`website-deploy.yml` lines 156–157; `scripts/release/publish.sh`), so `scripts/lib/release.sh` is **not** available at any stable URL and option (a) (piped bootstrap sourcing the lib from the release host) would require an out-of-scope release-pipeline change — and would regress to an inline copy on any offline/404/mirror case anyway. The two helper copies are byte-identical (confirmed this pass and in the prior review), the update path (`self-update.sh`) has no copy and sources the single lib, and the residual drift risk is tracked as follow-up BET-643 with a concrete hardening proposal. That is a defensible, tracked resolution of a genuine design constraint — not a silent shortcut.

**Verdict updated: 0 Blocks, 0 Questions, 2 nits** (both non-blocking, unchanged from the prior review: `self-update.sh` reads the replace `includes` list from the currently-installed rather than incoming `RELEASE.json`; stale `// () → void` doc comment on `serverUpdateApply`). Neither blocks merge.

Routing to `manta-pm` for the merge hand-off.

PR #546, opened 2026-08-03
fix(BET-643): single-source install.sh's curl|bash helper fallback from scripts/lib/release.sh

antoinedc:
## What & why

BET-643 — the drift trap BET-640 left behind. `scripts/install.sh` must carry
inline copies of the four release helpers (`manifest_get`, `resolve_arch`,
`_sha256_of`, `verify_sha256`) for its **primary** `curl -fsSL … | bash` mode,
where `scripts/lib/release.sh` isn't on disk yet (the tarball that carries it is
downloaded MID-install, and a piped script can't read a sibling file). Those two
copies were byte-identical but could silently diverge — a future edit to a
helper in one file but not the other would ghost until a real box hit it.

This makes the duplication **impossible to ship unsigned**, not just caught:

- **`scripts/sync-release-fallback.mjs`** — a generator that regenerates
  install.sh's fallback block **verbatim from `scripts/lib/release.sh`** (the
  single source of truth). `node scripts/sync-release-fallback.mjs` rewrites
  install.sh in place; `--check` exits 1 when stale (CI-ready).
- **`scripts/sync-release-fallback.test.mjs`** — the byte-parity gate. Runs in
  `npm test` (the blocking `typecheck-test` CI job). It regenerates the fallback
  in memory and fails if the committed block in install.sh differs by even one
  byte — so a helper edit in one file but not the other turns CI red instead of
  drifting silently.
- The generated fallback block is delimited by explicit
  `BEGIN GENERATED … / END GENERATED` markers in install.sh and stays
  **committed**, so install.sh remains **served byte-identical across channels**
  (no publish-time substitution; the documented one-liner is unchanged).

Verified: editing a helper body in `release.sh` makes test 1 fail until the
generator is re-run (drift is caught); the generator is idempotent (re-run is a
no-op).

### Files changed (3)

- `scripts/install.sh` — wrapped the existing fallback in generated markers;
  re-ran the generator so the block now carries the lib's exact bytes (incl. its
  `# NOTE:` comment on `resolve_arch` and normalized inter-function blank lines).
- `scripts/sync-release-fallback.mjs` — new generator + pure extraction helpers.
- `scripts/sync-release-fallback.test.mjs` — new byte-parity gate (3 tests).

`self-update.sh` is untouched — it always sources the lib (no fallback of its
own), which is already correct.

**Base: origin/main @ `893d83b`**

**Verification results:**
- PR branch (`multica/BET-643-release-fallback-parity` @ `27e8ae8`):
  - test: exit 0, 1441 pass / 0 fail / 0 skip. Failures: none. (Includes 257
    `install.test.mjs` — which sources `install.sh` — plus the 3 new parity-gate
    tests.)
  - `bash -n scripts/install.sh`: syntax OK.
  - anti-spaghetti `check-duplication-gate.sh origin/main`: PASS (no clones).
- Base (`main` @ `893d83b`): unchanged logic; the 3 changed/added files exist
  only on this branch.
- **Conclusion:** 0 new failures, 0 pre-existing; the drift gate is new and
  exercised in both directions (drift → red, restored → green).

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -153,7 +153,7 @@ if [ -n "$INSTALL_SOURCE_DIR" ] && [ -f "$INSTALL_SOURCE_DIR/lib/release.sh" ]; 
   . "$INSTALL_SOURCE_DIR/lib/release.sh"
 else
   # curl | bash / no local lib yet — inline fallback (byte-identical to lib).
-
+  # >>> BEGIN GENERATED — scripts/sync-release-fallback.mjs — do not edit by hand <<<
 # Parse one key out of a key=value manifest body. Echoes the value, empty if
 # absent. Values may contain `=` (cut -d= -f2- preserves them). A repeated key
 # is reduced to the FIRST occurrence (head -n1) — the manifest writer emits
@@ -165,6 +165,8 @@ manifest_get() { # $1=manifest-body $2=key
 # Map (uname -s, uname -m) to the release manifest arch key. Sets global
 # ARCH_KEY. Dies on any (OS, arch) we don't ship a tarball for. This is the
 # SINGLE place the installer decides which OS+arch it is — see BET-274.
+# NOTE: this function is defined here so install.sh (fresh `curl | bash`) and
+# self-update.sh (packaged box) resolve arch identically.
 resolve_arch() {
   local s m; s="$(uname -s)"; m="$(uname -m)"
   case "$s" in
@@ -187,7 +189,6 @@ resolve_arch() {
   esac
 }
 
-
 # _sha256_of echoes the sha256 hex of $1. Prefers GNU sha256sum (Linux ships
 # it via coreutils); falls back to BSD `shasum -a 256` on macOS, which ships
 # shasum by default but NOT sha256sum. Single shared helper so the prereq
@@ -208,6 +209,7 @@ verify_sha256() {
       actual:   $actual
       (corrupt download or stale manifest — re-run; if it persists, report it)"
 }
+  # >>> END GENERATED — scripts/sync-release-fallback.mjs — do not edit by hand <<<
 fi
 
 # launchd_agent_path — the PATH a MantaUI supervisor-managed service must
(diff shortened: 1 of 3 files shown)


antoinedc:
## manta-reviewer — BET-643 review (PR #546)

**Verdict: PASS** — 0 Blocks, 0 Questions, 0 Nits.

### Acceptance-criteria checklist (against the issue Dispatch)
- **Files in scope** (`scripts/install.sh` inline fallback, `scripts/lib/release.sh`): covered. Diff is exactly 3 files: `install.sh`, +2 new `scripts/sync-release-fallback.{mjs,test.mjs}`. No under-delivery, no scope creep, base fresh (`merge-base == origin/main == 893d83b` — no stale base, no phantom deletions).
- **Drift gate candidate (a) — byte-parity in CI**: delivered. `scripts/sync-release-fallback.test.mjs` matches the `scripts/*.test.mjs` glob in `npm test` (blocking `typecheck-test` job); it regenerates the fallback in memory from the lib and fails if the committed block differs by one byte. Both drift directions caught (edit lib w/o regenerating, or hand-edit install.sh → regeneration diverges → red).
- **Generator candidate (b) — single source of truth**: delivered. `sync-release-fallback.mjs` regenerates install.sh's fallback verbatim from `scripts/lib/release.sh`; output is **committed**, so install.sh stays served byte-identical across channels (does NOT break the byte-identical contract the PM's rejected publish-time-substitution option would have). `--check` is CI-ready.
- **No change to the install one-liner / served contract**: confirmed — self-update.sh untouched (always sources the lib).

### Mechanical gates (run independently on branch `27e8ae8`)
- `npm run typecheck`: **exit 0** (0 errors).
- `npm test`: **1441 pass / 0 fail** — matches the implementer's reported `27e8ae8` result exactly (includes 257 `install.test.mjs` which sources install.sh, plus the 3 new parity tests).
- `node scripts/sync-release-fallback.mjs --check`: **up to date** (committed block byte-identical to a regeneration) — I confirmed the parity independently, not just on the implementer's claim.
- `bash -n scripts/install.sh`: OK.
- Anti-spaghetti (scoped to changed files): 0 clones.

### Notes
The `# NOTE:` comment and the four helper bodies live in `scripts/lib/release.sh` (confirmed on `origin/main`, merged via BET-640 #544); the fallback genuinely derives from the lib. The implementer delivered a stricter remedy than the PM's minimum (generator in addition to the gate) — a positive, both are the issue's own named candidates and neither expands the served contract.

PR #550, opened 2026-08-03
feat(renderer): Manage models… footer row in model menu + manta-open-settings bridge (BET-645)

antoinedc:
Fills the BET-644 `Dropdown` `footer` slot that was built but left unused.

## What

**1. The `Manage models…` row** (`ModelMenu.tsx`) — a `Sliders`-style lucide icon + `Manage models…`, rendered into the model menu's `footer` slot.
- Chrome: `text-meta` at `text-text-faint`, hover → `text-text` on `bg-fill-hover`, `rounded-md`, full width.
- It is **not** a `MenuOption` — not selectable, no tick/trailing slot, not part of the `↑`/`↓` option ring. It is a plain `button`, so it is naturally Tab-reachable from the menu.
- **Kept local to `ModelMenu.tsx`** (one call site does not clear the two-adopter rule, standing decision 2 — the PR promotes it to a primitive if a second footer action ever appears).

**2. The `manta-open-settings` route** (did not exist).
- `ModelMenu` dispatches the same window-`CustomEvent` bridge the repo already uses for schedules/secrets: `new CustomEvent("manta-open-settings", { detail: { section: "models" } })`.
- `App.tsx` owns the Settings modal, so it **consumes** the event there (confirming the "whoever owns the Settings modal" instruction, and not threading a prop chain from `App` down through the composer). It validates the requested section against `SETTING_SECTIONS` and falls back to `general` on an unknown id — the bridge is generic (any section id), not hard-coded to AI.
- `Settings.tsx` gains a `initialSection` prop (initial `activeTab` + re-targets if a request lands while already open).

## Acceptance
- Clicking `Manage models…` closes the menu and opens Settings on the Models tab (the models table = `ModelsCard`, section `models`). ✅ covered by a new component test (`ModelMenu.test.tsx`).
- Event bridge is generic (any section id), validated in `App`. ✅
- Keyboard: footer row Tab-reachable, not in the option ring (plain button, not a `MenuOption`). ✅
- typecheck + tests green; model-picker visual baseline regenerated.

## Verdict on "both themes" baselines
The repo's `screens.visual.ts` harness runs under one `colorScheme: "light"` — there is a single `model-picker-visual-linux.png` baseline, which I regenerated. Dark/light coverage lives in `shadow.visual.ts` (both themes) and is unaffected by this change. So only the one existing model-picker baseline needed updating and was updated.

**Base: origin/main @ 8ab0477**

## Verification results
[...]

PR #551, opened 2026-08-03
BET-649: the working indicator is the manta loader; prose arrives by sentence, not by slab

antoinedc:
Implements BET-649. Owner-reviewed proposal: https://0d5784a7a43451f4ad70dd3d9ee5cf72.boxes.mantaui.com/pages/streaming-proposal

## The indicator is the loader we already ship

The mark inside two counter-rotating arcs exists twice already: it was the web client's `ConnectingScreen` (retired in BET-559) and it is the iOS client's `MantaLoader.swift` today — what you see when a session opens on the phone. Desktop was the only client still drawing a generic grey ring. `MantaLoader` ports the geometry verbatim (28% / 22% arcs, 0.78 inner ratio, 0.9s clockwise / 1.4s anticlockwise) so both clients draw ONE image.

- Size is a variant (`inline` 24px per the owner, `screen` 92px), no `className` hatch.
- **The stroke does not scale**: pinned at 2px below `screen`, because 3/92 held at 24px is a 0.8px hairline. Optical adjustment, recorded in the file.
- Two adopters: the working row, and the session-connect state that used to draw the grey ring — the exact moment iOS shows this loader.
- A finished turn keeps the mark (`MantaMark`), static: the arcs meant *in flight*.
- Motion lives in `index.css`, so reduced-motion holds it still in one place. Nothing is encoded in the rotation — the elapsed counter keeps ticking.

**No skeleton placeholder** (owner dropped it, correctly): `submit()` already sets `running` optimistically, so the row appears on send. The loader plus its verb *is* the placeholder.

## One correction to the proposal

The proposal claimed every change was renderer-side. Since BET-551 the flush decision lives on the BOX, so the flush work lands in `src/shared/` + `src/server/` — which is better: **the native client inherits it for free**.

## Prose by sentence

`findFlushBoundary` cuts at sentence ends as well as paragraph breaks, but only where the new `isSafeCut` says the prefix closes every construct it opened. Splitting an inline code span, a link or a `**bold**` run renders half-finished markup literally for a frame — the exact jitter this buffer exists to prevent. Conservative by design: a false negative costs a few hundred ms, a false positive puts `**bold` on screen. Code blocks still flush whole.

`FLUSH_MAX_AGE_MS` 250 → 120, and the box gains **the max-age fallback it never had**: an unpunctuated run (a table, a bulleted line, a wall of prose) used to stay buffered until the part snapshot — the "it all arrived at once at the end" case. `lastSafeCut` releases it at the last safe word break. Checked on the next delta rather than on a timer, so the interpreter stays event-driven and testable.

## Arrival + scroll

Block fade-in and the trailing caret are scoped to the one message being written (`manta-streaming`), so the settled transcript never animates. The fade fires on mount, so a growing paragraph does not re-animate and the canonical refetch does not re-fire it. The caret is an `::after` on the last block so it sits inline rather than on its own line after a `<p>`.

`stickToBottom` eases a small follow-step and jumps a large one (240px). Easing hundreds of pixels means the animation is still running when the next commit lands, so the view chases content and never settles. **The pin decision is untouched** — this changes how we follow, never whether we stick.

Word-by-word reveal was proposed and **rejected**: it means wrapping every word inside rendered markdown, which fights the memoisation the transcript needs to stay responsive while typing.

## Verification

`npm run typecheck` clean; `npm test` green (2006 vitest + 1444 node:test). New coverage: loader geometry asserted against the native contract (a silent edit is a drift between clients and nothing else would catch it), sentence-boundary and safe-cut cases, the age fallback end-to-end, and `scrollBehaviorFor`.

One existing expectation changed by design — the interpreter's paragraph test now expects the sentence cut. The assembled text is byte-identical; only the split timing moved.

---
**Base: `origin/main @ 8a4d792`** (rebased onto current main to resolve the BET-646 / model-picker conflict).


antoinedc:
## Review — BET-649 (manta-reviewer)

**Verdict: 0 Blocks, 0 Questions, 2 nits. Clean pass → routing to manta-pm.**

Verified against the issue's acceptance criteria (not the summary), the code, and a fresh base. Everything checks out.

### Mechanical gate (executed on the PR branch, base `origin/main`)
- **Base freshness:** `git merge-base HEAD origin/main` → the branch sits on `8a4d792` (via PR #550); `git diff --stat origin/main...HEAD` shows **17 in-scope files, zero deletions, no phantom files**. No stale-base issue.
- **`npm run typecheck`:** exit 0, zero errors.
- **`npm test`:**
  - vitest: **2012 passed, 7 skipped, 0 failed**
  - node:test: **1449 passed, 0 failed**
- **Anti-spaghetti (jscpd, changed scope):** 0 clones. Each of the new test suites independently asserts its behavior, so no duplicate-logic or dead-code concern.

1. **Working row = loader + verb + tabular numbers, appears on send.** `RunningIndicator` swaps the pulsing text asterisk for `MantaLoader` (24px inline) with the verb in `text-label` and the mono `tabular-nums` readout. Row still appears on send — `setRunning(true)` is set optimistically in the submit path before the box confirms. ✅
2. **Session connect = 92px loader, not grey ring.** `!messages` branch uses `MantaLoader size="screen"` (92px, labelled "Connecting to session"). ✅
3. **Finished turn keeps the mark, static, in the past-tense footer.** The footer's `✻` is replaced by `MantaMark size={12}`. ✅
4. **Prose by sentence + block fade + trailing caret; code blocks whole.** `findFlushBoundary` cuts at sentence ends (`.`/`!`/`?` + closing quote/bracket + whitespace) but only where `isSafeCut` confirms the prefix closes every inline-code/link/bold construct; code fences still flush whole. Fade + caret are scoped to `manta-streaming` on the final part of the final assistant message while running. Both backed by real tests (sentence boundaries, mid-number `.`, no cut inside code/link/bold, cut-once-closed, max-age fallback, fallback-stops-before-unclosed-span). ✅
5. **Reduced motion: arcs still, no fade, no caret blink, counter ticking.** All three animate-only effects are held still in one `prefers-reduced-motion` block; the elapsed readout is JS-timer-driven, so it still ticks. ✅
6. **`npm run typecheck && npm test` green.** Verified above. ✅

### Scope
All 17 changed files fall in the issue's named areas (MantaLoader primitive + the two adopters, box-side flush in `src/shared` + `src/server`, block fade/caret in CSS, eased follow-scroll). No file outside scope; no under-delivery.

### Nits (non-blocking)
- **Decimal false-positive (inherent to the accepted heuristic):** a period followed by whitespace that is *not* a real sentence end — e.g. `version 1.5. Next` — is treated as a sentence boundary and splits sub-sentence. Reassembly is byte-identical and this is exactly the conservative trade-off the owner's decision accepted (a cheap prefix check, not a markdown parser); flagging for awareness, not a fix.
- **`MantaMark` default `size=14` is currently unused** (the only adopter passes `size={12}`). Harmless; could trim to a single size with no call-site impact.

Overall a thorough, well-tested implementation that matches the issue to the letter.

PR #557, opened 2026-08-03
fix(ios): the phone drew every finished answer twice

antoinedc:
The iOS chat screen renders the canonical transcript **plus** the live streamed tail, and nothing ever retired the tail. While a turn is streaming that tail is the only copy, so it is correct. The moment the turn ends and the refetch lands, the transcript carries the same prose and both are drawn — the answer appears twice, and the tail then accumulates for the rest of the session.

A comment beside the code claimed the refetch "absorbed" the in-progress text. Nothing implemented that; the absorption was assumed.

## Fix

Live chunks now carry the `messageID` they belong to, and a successful transcript fetch retires every chunk that fetch covers.

Two rules stop it over-correcting:

- **The message still streaming is never retired**, even when the fetch covered it. A mid-turn transcript holds only the prose written before that fetch, so retiring there would split one answer across two blocks and drop everything since.
- **A failed fetch retires nothing.** It proves nothing about what the transcript holds, and wiping on that basis would erase the answer from the screen.

Retirement is idempotent, so running it after every fetch cannot loop or republish on a no-op.

Also fixes an ordering bug found on the way: the tail was assembled from a dictionary, whose values have no defined order, so a multi-part answer could reassemble with its paragraphs shuffled. Chunks are kept in arrival order.

## Tests

Pure retirement rule — covered / not-yet-covered / streaming-message-kept / idempotent — plus arrival ordering and reasoning-excluded-from-tail. Three store-level tests drive the real seam (frames in through the stream, a canonical transcript in over the wire) and assert the finished answer renders exactly once, a streaming tail survives a refetch that does not cover it, and a failed refetch does not retire.

## Notes for the reviewer

- **Stacked on #556.** The base is `fix/blank-session-list-and-block-markdown`, not `main`. Both branches were authored from the same shared working tree, so basing on `main` would have swept #556's hunks into this diff. Retarget to `main` once #556 merges; the delta is already clean (`git diff` against 688221f contains zero of #556's changes).
- **Not yet verified on a simulator.** The Mac executor went offline with the unit gate queued (`285652f3`). Neither this branch nor #556 has been through `test-unit` yet. Do not merge either until it reports.
- CI's green tick certifies nothing about this Swift — there is no iOS build in CI (BET-651).


antoinedc:
Closing as superseded. This targets the iOS live-tail/canonical duplication from 2026-08-03; the branch is 558 commits behind main and those exact files have been rewritten many times since — notably a6aa27c (stable identity for prose tail -> canonical swap, no remove+insert blink, BET-752), plus BET-668/BET-672/BET-753. Reopen a fresh PR against current main if the duplication still reproduces.

PR #570, opened 2026-08-04
fix(delegate): stop killing every background job as an orphan

antoinedc:
Found while trying to demo the sidebar rail with a real delegated job: **the job was aborted about a minute after starting, and recorded as `stopped by user` — which nobody had done.** It reproduces every time.

## Cause

The sweeper stops a job whose parent opencode session has gone away (BET-418 §B — nobody left to report to). It asked that question by scanning a list:

```js
const sessions = await oc.listSessions();
return sessions.some((s) => s.id === sid);
```

opencode caps `GET /session` at **100 records**, and the unscoped form returns the 100 most-recently-updated sessions **box-wide**. On any box with real history the parent simply isn't in that page, so the scan answers "gone" and the sweeper kills a perfectly healthy job.

Verified live on the dev box:

```
old (list scan)   → false   ← killed the job
new (direct GET)  → true
bogus id          → false
```

The unscoped page there is 46 sessions from `/home/dev`, 26 from `/tmp/opencode`, 12 from an unrelated project — and no room for the parent.

## Why it stayed hidden

The failure is silent and total: the job is aborted, its window and worktree are cleaned up as terminal, and `stopped by user` points the finger at the operator. Background delegation looked like it half-worked when in practice **no job could survive**. It also gets steadily worse as a box accumulates sessions, so it would not have shown up on a fresh box or in early testing.

## Fix

Existence is now a **direct lookup** — `GET /session/{id}` in `opencode.mjs:sessionExists`. Uncapped, O(1), unambiguous.

Only a definitive **404** counts as gone. 5xx and transport errors keep the existing "assume alive" degradation, so a transient opencode blip still can't orphan a healthy job — the destructive answer requires positive evidence.

## Tests

Six, in `src/server/opencode.test.mjs`:
- direct lookup is used, and the capped collection endpoint is **not** hit
- **the regression**: a full 100-record page that doesn't contain the parent — old scan said gone, direct lookup says alive
- 404 → gone (the only negative)
- 5xx → alive
- transport throw → alive
- empty id → false, no request issued

`npm run typecheck` clean; 2073 renderer tests + 1468 server tests pass.

## Relationship to #565

Independent bug, different layer, but they compound: #565 fixes the sidebar never *displaying* a background job, this fixes background jobs never *surviving*. Both are needed before a delegated job is visible and useful. This one can land first — it's server-only.

### DIFF
--- a/src/server/index.mjs
+++ b/src/server/index.mjs
@@ -254,14 +254,11 @@ const delegateEngine = createDelegateEngine({
   abortSession: (sid) => oc.abortSession(sid),
   // BET-418 §B: detect a running job whose parent opencode session is gone so
   // the sweeper can stop + clean it up (nobody left to report to).
-  sessionExists: async (sid) => {
-    try {
-      const sessions = await oc.listSessions();
-      return Array.isArray(sessions) && sessions.some((s) => s?.id === sid);
-    } catch {
-      return true; // best-effort: assume alive on a transient blip
-    }
-  },
+  // Direct lookup, NOT a listSessions scan: `GET /session` is capped at 100
+  // and the unscoped form is box-wide, so a healthy parent that simply isn't
+  // among the 100 most recent sessions read as "gone" and the sweeper stopped
+  // the job. See opencode.mjs:sessionExists.
+  sessionExists: (sid) => oc.sessionExists(sid),
   oc,
 });
 // eslint-disable-next-line no-unused-vars
--- a/src/server/opencode.mjs
+++ b/src/server/opencode.mjs
@@ -686,6 +686,39 @@ export async function listSessions(directory) {
 }
 
 /**
+ * Does this opencode session still exist?
+ *
+ * Answered with a DIRECT lookup (`GET /session/{id}`), never by scanning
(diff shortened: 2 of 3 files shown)


antoinedc:
**Note on `duplication-gate`:** its 3 reported clones are **pre-existing on `main`**, not introduced here.

Verified by running jscpd against this file with my commit stashed — identical results either way:

| | clones | line ranges |
|---|---|---|
| `main` | 3 | 76-84 ↔ 177-185, 1510-1525 ↔ 1561-1577, 1510-1524 ↔ 1596-1610 |
| this PR | 3 | *(same)* |

My tests start at line 1657, below every one of those ranges. The gate scans whole changed files, so touching this 1763-line test file re-surfaces clones that were already there. I did still collapse my own near-identical cases into a table-driven loop after the first run flagged them — that part was a fair call.

Fixing the pre-existing clones would mean refactoring unrelated parts of the file and is out of scope here.

**`E2E Smoke Test`** is likewise red on `main` right now (visual-baseline drift from the chat-chrome work — the failures are context-popover and session aria snapshots, nothing to do with this change).

`typecheck-test` — the only required check — passes.

PR #573, opened 2026-08-04
fix(delegate): pre-flight ruleset denied every tool it was meant to grant

antoinedc:
Found while dispatching a background job to do real work: **the job could not run a single tool, whatever it had been granted.** It started, was refused by its own permission ruleset on the first call, and finished having done nothing — after creating a worktree, a tmux window and a full model session.

## Cause

`buildPermissionRuleset` appended the catch-all last:

```js
deduped.push({ permission: "*", pattern: "**", action: "deny" });
```

on the assumption that a trailing catch-all acts as a floor. **opencode resolves the list last-match-wins**, so that rule matched every call and beat every `allow` above it. Every grant was inert.

## Verified, not reasoned about

The API accepts either order silently and echoes it back unchanged, so this had to be checked against the running opencode (v1.15.12):

| ruleset sent | result |
|---|---|
| `[bash ** allow, * ** deny]` | bash **DENIED** — *"a rule prevents you from using this specific tool call"* |
| `[* ** deny, bash ** allow]` | bash **COMPLETED** — returned `PERMTEST_OK` |

The denial message also revealed that opencode prepends its own `{permission:"*", pattern:"*", action:"allow"}` default ahead of whatever we send — which is why an explicit catch-all deny is still needed to stop an ungranted tool resolving to allow. It just has to sit **under** the grants rather than over them.

## Fix

Catch-all deny first, grants after.

**BET-418 §A's guarantee is unchanged:** a tool that was never granted still has nothing to lift it off the deny floor, so it fails fast rather than hanging on an `ask`. That was the whole point of the catch-all and it still holds — the existing §A4 test covering it passes unmodified in substance.

## Tests

The old ordering was pinned by a test literally asserting *"catch-all deny MUST be last"* — it encoded the bug. That test and three other order-dependent assertions are inverted, and the primary one now documents the live-verified behaviour so the next person doesn't re-flip it.

`npm run typecheck` clean, 2073 renderer + 1468 server tests pass, `duplication-gate` passes.

## Context

Third in a chain of independent bugs that each made background delegation look like an unfinished feature:
- #570 — jobs were killed after ~60s as false orphans
- #565 — jobs never appeared in the sidebar
- **this** — jobs could never run a tool

All three had to be fixed before a delegated job could do anything useful.

### DIFF
--- a/src/server/delegate.mjs
+++ b/src/server/delegate.mjs
@@ -185,10 +185,26 @@ export function buildPermissionRuleset(tools) {
     seen.add(key);
     return true;
   });
-  // The catch-all deny MUST be last — it is what stops an unmatched tool from
-  // resolving to `ask` and hanging the job.
-  deduped.push({ permission: "*", pattern: "**", action: "deny" });
-  return deduped;
+  // The catch-all deny MUST be FIRST, with the specific allows after it.
+  //
+  // opencode resolves a tool call against this list LAST-MATCH-WINS, which is
+  // the opposite of what this originally assumed. Appending the catch-all last
+  // meant `{permission:"*", pattern:"**", action:"deny"}` matched every call
+  // and won over every allow that preceded it — so a delegated job could not
+  // run a SINGLE tool, whatever it had been granted. The job then burned a
+  // worktree, a window and a model session doing nothing but reporting that it
+  // was denied.
+  //
+  // Verified against the running opencode (v1.15.12) rather than reasoned
+  // about, because the API accepts either order silently:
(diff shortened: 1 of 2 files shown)


antoinedc:
`duplication-gate`'s 3 clones are **pre-existing on `main`** (lines 31-38↔770-775, 481-487↔496-502, 811-822↔830-841) — verified by running jscpd against main's copy of `delegate.test.mjs`. This PR only inverts existing assertions in place and adds no new duplication; the gate rescans the whole file because it was touched. `E2E Smoke Test` is likewise red on `main` (visual baseline drift). `typecheck-test` — the only required check — passes.

PR #575, opened 2026-08-04
feat(channels): route box updates through MANTA_CHANNEL; give Windows the same tracks

antoinedc:
## The bug this started from

**A staging box updated itself onto prod builds.**

`MANTA_CHANNEL` (`prod|staging|dev`) already existed, was already resolved at install time, and was already baked into the box's service definition. But it drove exactly one thing — the pairing link's URL scheme. Both update paths ignored it:

```
serverUpdate.mjs:  MANIFEST_URL = "https://mantaui.com/updates/server.json"
self-update.sh:    host="${MANTA_RELEASE_HOST:-https://mantaui.com}"
```

So a box installed with `MANTA_CHANNEL=staging` checked the **prod** manifest and installed **prod** tarballs. The staging server track is published and live, but **nothing could follow it**.

That's why staging sits at `0.0.17` while prod is on `0.0.19` and nobody noticed: the failure is silent and looks exactly like "no update available".

## No new channel system

The table in `src/shared/channel.mjs` already carries `releaseHost` and `updateFeed` per channel — including `updateFeed: null` for `dev`. This just makes the update paths read from it.

| channel | update feed | release host |
|---|---|---|
| prod | `/updates` | `mantaui.com` |
| staging | `/staging/updates` | `mantaui.com/staging` |
| dev | **none** | n/a |

**`dev` answers the "what about local builds" question by doing nothing:** a feedless build reports "no update" and never fetches, so a local build can't talk a box into installing a public release over itself.

`resolveBoxChannel()` moves from `pairPage.mjs` to `src/shared/channel.mjs` now that two unrelated subsystems read it (pairing re-exports it, so its callers and tests are untouched). Leaving it in the pairing module is precisely what let the updater get away with hardcoding prod.

`self-update.sh` now logs its resolved channel and host — the old failure was invisible.

## Windows gets the same two tracks

Still published as **GitHub releases** — the unsigned installer stays off the public download page, as agreed.

- `win-v*` → prod, `win-staging-v*` → staging, or `workflow_dispatch` with a `channel` input. One resolve step decides and rejects an unrecognised tag.
- `MANTA_CHANNEL` is exported **before** `npm run build` so electron-vite bakes `__MANTA_CHANNEL__`. Without it a "staging" build would register the prod `manta://` scheme and **hijack prod pairing links**.
- Staging overlays `electron-builder.staging.yml` (different appId + productName) so it installs **side by side** with prod. That overlay already existed for mac and its own comments already named Windows — this is the wiring it was waiting for.

## Tests

`manifestUrl` per channel (prod, staging, `dev` → null, unset/garbage → prod), `createUpdateCheck` honouring the resolved feed, and a dev build never fetching.

`npm run typecheck` clean · 2079 renderer + 991 server tests pass · `duplication-gate` PASS · `bash -n scripts/self-update.sh` clean · workflow YAML validated.

## Not included

**iOS.** TestFlight *is* staging and the App Store *is* prod, so it needs no infrastructure — only the channel baked into the build. Worth doing, but it's a Codemagic + Info.plist change with its own signing surface, so it belongs in its own PR rather than riding along here.

### DIFF
--- a/.github/workflows/windows-desktop-build.yml
+++ b/.github/workflows/windows-desktop-build.yml
@@ -63,8 +63,17 @@ name: Windows desktop build
 on:
   push:
     tags:
+      # prod channel
       - "win-v*"
+      # staging channel — side-by-side install, staging URL scheme
+      - "win-staging-v*"
   workflow_dispatch:
+    inputs:
+      channel:
+        description: "Release channel to build"
+        type: choice
+        options: [prod, staging]
+        default: prod
 
 concurrency:
   group: windows-desktop-build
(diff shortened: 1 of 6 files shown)


antoinedc:
`duplication-gate`: the one remaining clone (`serverUpdate.test.mjs` lines 206-227 ↔ 246-267) is **pre-existing on `main`** — verified by re-running jscpd against this file with my commits stashed, which reports the identical pair at the identical line ranges. My additions sit below line 267, so they don't shift it.

The clone the gate flagged on the first run *was* mine (repeated `MANTA_CHANNEL` save/restore boilerplate across five cases) and is now collapsed into a table-driven `withChannel` helper.

`E2E Smoke Test` is red on `main` already (visual baseline drift). `typecheck-test` — the only required check — passes.

PR #612, opened 2026-08-05
feat(artifacts): Files tab = user uploads + ~/.manta-outbox pushed files

antoinedc:
Scope the Artifacts **Files tab** to two sources only: **user uploads** and **files the agent pushed via `~/.manta-outbox`**. Removes the earlier `write`-tool derivation (arbitrary written paths no longer flood the tab).

- `artifacts.ts` — replaced write-tool derivation with outbox rows (origin agent, kind/mime from extension, real `size` + `mtime`, `messageId` null).
- Server — `listOutbox` now returns `mtime` per file (so the panel can sort/day-group); new read-only `outbox:list` RPC channel wired across all wiring sites + a `demoApi` stub.
- Panel — fetches the outbox on open + 30s poll alongside the page registry, feeds `deriveArtifacts`.
- Tests — outbox file / image / unknown-extension / upload+outbox coexistence, and `listOutbox` mtime.

`typecheck` + full suite green locally; duplication gate passes.

### DIFF
--- a/src/renderer/ArtifactsPanel.tsx
+++ b/src/renderer/ArtifactsPanel.tsx
@@ -26,7 +26,7 @@ import {
   Search,
   X,
 } from "lucide-react";
-import type { ServedPageMeta } from "../shared/types";
+import type { OutboxFile, ServedPageMeta } from "../shared/types";
 import { useStore } from "./store";
 import {
   countByKind,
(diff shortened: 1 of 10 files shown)


antoinedc:
Superseded by rework: the one-shot outbox model is wrong for artifacts. Rebuilding as a durable, workspace-linked (`/<sessionID>/`), TTL'd store (7d default) with non-destructive download and expiry pills, reusing this branch's RPC/panel/derivation work as the base. See the replacement PR.

PR #670, opened 2026-08-07
feat(ios) BET-665: transcript bottom inset above composer; drop bounce-disabling hack

antoinedc:
Opened automatically for `multica/BET-665-transcript-bottom-inset`, which was pushed without a pull request.

Relates to BET-665.

### DIFF
--- a/mobile/native/MantaUI.xcodeproj/project.pbxproj
+++ b/mobile/native/MantaUI.xcodeproj/project.pbxproj
@@ -10,13 +10,14 @@
 		0129D15936E52427A46702E7 /* MantaAuthClientTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 989AB507C4722A4CF012C3B8 /* MantaAuthClientTests.swift */; };
 		01871023D47544D64443194D /* MantaModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8C8240F1954861B4846E279D /* MantaModels.swift */; };
 		0488B71F794FC6E64CDC0B23 /* TerminalKeyRow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7D7BC8920829E8847057C8DA /* TerminalKeyRow.swift */; };
+		1A55CBC343898856735AF0F8 /* MantaChatSurfaceCaptureUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 570E190164DB3AB83DC14283 /* MantaChatSurfaceCaptureUITests.swift */; };
 		1B9026B4F0286A5186FD5197 /* MantaSettingsModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 39978655A14BFA5D39336286 /* MantaSettingsModels.swift */; };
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review — BET-665: transcript bottom inset above composer; delete bounce-disabling hack

**Verdict: PASS — 0 Blocks, 0 Questions, 3 nits.** Routing forward.

Reviewed against the issue's acceptance criteria. Base is fresh (`merge-base` = `origin/main` tip `64dc3e8`), so the diff is genuine.

### Acceptance checklist (verified independently, not from the hand-off prose)

- **`DisableScrollBounceHelper` + `owningScrollView()` deleted.** Confirmed by grep — neither symbol exists anywhere under `mobile/native/`. The `.background(DisableScrollBounceHelper())` attachment is also gone. ✓
- **Bottom inset added, placed after the TiledView-only chain.** `.contentMargins(.bottom, bottomBarHeight, for: .scrollContent)` sits after `.simultaneousGesture` (where the type is already erased), i.e. after `.headerContent` → `.prependLoader` → `.typingIndicator` → `.onTiledScrollGeometryChange`. Matches the issue's required ordering and the referenced `a207772` placement. ✓
- **Inset tracks the composer automatically; no second measurement.** Exactly one `.onGeometryChange` (→ `bottomBarHeight`) already present, feeding both the inset and the Scrim. No extra geometry plumbing added. ✓
- **Scrim kept as-is.** Unchanged. ✓
- **Natural bounce restored** by deleting the hack (top overscroll preserved for the load-earlier prepend trigger too). ✓
- **Do-not-dos respected:** no safeAreaInset conversion of the composer, no scroll-position compensation, no keyboard observers, `ComposerView.swift` untouched, `autoScrollsToBottomOnAppend/scrollsToBottomOnReplace` left as-is. ✓

### Build / test results

- GitHub CI: `typecheck-test` and `duplication-gate` both **SUCCESS**; anti-spaghetti report clean (0 TS files changed — nothing to scan).
- iOS build + Unit tests were run by the `macos` worker (I'm on Linux and cannot run `xcodebuild`): build **SUCCEEDED**; 214 tests executed, 3 failures — all in `ChatModel.fastToggle` / `ComposerTests` (the known BET-597 fast-toggle mismatch). These are outside this diff (they exercise `ChatModel`/`ChatVoice`, not `ChatScreen`/the inset), so the pre-existing claim is credible.

### Nits (non-blocking)

1. **Regenerated `project.pbxproj` bakes a branch-specific group name** (`manta-bet665`). Justified scope: the committed base genuinely didn't build — `ChatScreen` already `import MessagingUI`, but main's `project.pbxproj` had no `swiftui-messaging-ui` package ref — so regenerating from `project.yml` is necessary to hit the build gate, and the result restores parity with the real source tree. However the group is named after the local checkout folder, so it's non-reproducible and will conflict on the next native regeneration. Suggest regenerating/renaming to a stable group name rather than one tied to this issue.
2. **Stale leftover comment block** just above the edited region (~lines 182–211) still describes the *reverted* safeAreaInset design and now directly contradicts the chosen content-margin approach. Pre-existing on main, but since this PR edits the scrim comment immediately below it, it'd be tidy to update/remove it. Not required for this issue's scope.
3. **PR description is auto-generated boilerplate** (CI auto-opened the PR because the `macos` worker can't open PRs). The requirement + acceptance criteria live on the issue itself; nothing missing from the actual deliverable.

Nothing here blocks merge. All acceptance criteria are met by the diff.

PR #671, opened 2026-08-07
feat(ios) BET-666: stable transcript step-row identity (stop turn-boundary flash)

antoinedc:
Opened automatically for `multica/BET-666-stable-step-identity`, which was pushed without a pull request.

Relates to BET-666.

### DIFF
--- a/mobile/native/MantaUI/ChatModels.swift
+++ b/mobile/native/MantaUI/ChatModels.swift
@@ -231,8 +231,8 @@ enum ChatTranscriptMapper {
                 // text. The box itself keys turn completion on time.completed.
                 guard msg.info.time?.completed != nil else { continue }
                 let at = ChatClock.date(epochMs: msg.info.time?.completed)
-                for part in msg.parts {
-                    process(part, at: at, pending: &pending, blocks: &blocks)
+                for (index, part) in msg.parts.enumerated() {
+                    process(part, index: index, at: at, pending: &pending, blocks: &blocks)
                 }
                 flush(&pending, into: &blocks)
             default:
@@ -253,7 +253,7 @@ enum ChatTranscriptMapper {
         pending = []
     }
 
-    private static func process(_ part: OpencodePart, at: Date?, pending: inout [StepGroupRow], blocks: inout [TranscriptBlock]) {
+    private static func process(_ part: OpencodePart, index: Int, at: Date?, pending: inout [StepGroupRow], blocks: inout [TranscriptBlock]) {
         if part.ignored == true || part.synthetic == true { return }
         switch part.type {
         case "text":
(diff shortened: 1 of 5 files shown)


antoinedc:
## Review — BET-666 stable transcript step-row identity

**Verdict: PASS** (0 Blocks, 0 Questions, 2 nits)

Base-freshness: merge-base `33c0e2b` == `origin/main` — branch is fresh off current main. No phantom deletions. Diff is a clean 5-file change (`mobile/native/` only).

### Acceptance-criteria check

1. **`ToolStep` identity → deterministic `String` from the mapper** ✅
   `ToolStep.id` is now `String` (was fresh `UUID()`). The mapper derives it via `ChatModels.swift:stepIdentity` with the exact specified priority: tool part's `callID` → part's own `id` → `"\(messageID)-step-\(indexWithinMessage)"`. `StepGroupRow.id` updated to match. No `UUID()` remains on the transcript mapping path.
2. **`stableScrollID` via `hashValue`** ✅ — replaced the embedded full block text with `text.hashValue` combined with the existing timestamp/message components (user + prose cases); steps case untouched. No crypto hash, no new dependency. Matches spec.
3. **Two unit tests** ✅ — `testStepIdsAreIdenticalAcrossRefetch` (same transcript mapped twice → element-wise identical step ids) and `testStepIdsSurviveAppendingANewMessage` (append one message → pre-existing step ids preserved). Both present in `ChatTranscriptTests.swift` and exercise the new mapper path.

### Verification (Mac-only — cannot run locally on Linux box)

Per trust-but-spot-check, `macos` reported real output in its hand-off comment:
- Build: `** BUILD SUCCEEDED **` (iPhone 17 Pro simulator)
- `ChatTranscriptTests`: `35 tests, 0 failures` — both new tests pass
- Full `MantaUITests`: `216 tests, 3 failures` — the 3 are in `ComposerTests` (OpenAI `gpt-5.6-fast` fast-toggle/effort logic), pre-existing on `origin/main` (verified identical on a clean main worktree)
- `grep "UUID()" ChatModels.swift` → no hits on the transcript mapping path (verified independently: only hit is a doc comment in `TranscriptComponents.swift:763`, not live code)

The 3 `ComposerTests` failures are unrelated (asynchronous OpenAI streaming logic this PR does not touch) and pre-existing on main — this change does not regress the suite. Simulator launch landed on the onboarding screen (pairing was not present in this environment), so the step-row surface wasn't visually reachable — build + unit tests are the verification here, which is the prescribed fallback.

### Nits (non-blocking)
- `RootView.swift` demo/preview fixtures were updated for the new `ToolStep(id:)` signature — necessary compilation change, not scope creep.
- The 3 pre-existing `ComposerTests` failures on `main` are out of this PR's scope; flag separately if the PM wants them triaged.

`git checkout --detach` only; read-only review. No merge, no push.

PR #672, opened 2026-08-07
BET-676: feat(server): always strip duplicate tool stdout; drop cached/reconcile message channels

antoinedc:
Tail-first transcript loading, always-slim payloads, single fetch path (BET-676)

## What
- **Server always slims**: extracted `stripDuplicateToolOutput(message)` and applied it in `listMessages` AND the single-message `getMessage` splice path for every client, independent of the iOS-only `slim` option. Lossless — only exact `metadata.output === state.output` duplicates removed.
- **Removed the vestigial channels** `opencodeMessagesCached` and `opencodeMessagesReconcile` (and `reconcileMessages`) from `shared/api.ts`, `shared/types.ts`, `httpApi.ts`, `rpc.mjs`, `demoApi`, tests, harness. Single fetch path.
- **Tail-first loading**: `TRANSCRIPT_TAIL_LIMIT=100`, per-panel `loadedAllRef` + `fetchOpts()`; the mount fetch, `scheduleRefetch`, and the `server.connected` resync all pass `{limit}` until "Load earlier" flips to full.
- **Initial fetch resilience**: 15s timeout + one 2s-delayed retry (`fetchTranscriptWithRetry` in `chatUtils`, mock-timer tested); on final failure a `transcriptLoadError` state shows the existing "Couldn't load session" panel with a **Retry** button. Removed the unconditional self-heal double fetch.
- **Load earlier**: `LoadEarlier` in `Transcript.tsx` (single exported unit, commented as the virtualization-replacement seam) pulls the full history, preserves scroll position, sets `loadedAllRef`.
- **Hidden-panel poll gating**: delegateList (10s), delegate approvals (3s), vcs branch (5s), schedules (30s) all skip while `!isActive` and refire once on reactivation.
- `replayChatAttention` boot fan-out uses `{ limit: 20 }`.

## Tests
- Server: `opencode.test.mjs` — duplicate stripped from `listMessages` + `getMessage`, non-identical (live-streaming) output untouched, lossless/tolerance, plus the updated limit-passthrough test.
- Vitest: `fetchTranscriptWithRetry` timeout→retry→success / →reject (mock timers); `LoadEarlier` (tail-hides button, click pulls full + flips loadedAll); harness asserts **exactly ONE mount fetch** tail-limited (no self-heal).

## Verification results
**Files changed:** 19 files, +556/−115.
**PR branch** (`multica/BET-676-transcript-tail` @ `62ec1f18`):
- test: exit 0. vitest 119 files, 2217 passed / 7 skipped / 0 fail; node 1501 pass / 0 fail. log sha256: `81be46232446607778ca3929b721f4f00f7157d25b3b38e3be16246b310a75d2`
**Base** (`main` @ `bbae0475`):
- test: exit 0. vitest 119 files passed; node 1501 pass / 0 fail. log sha256: `555613642d65a6d8afbd8c9c093a21aeca5659ec04857c96fa4b8b34cda1f921`

## Cross-cutting / notes
- Removes IPC channels `opencode:messages-cached` and `opencode:messages-reconcile` shared by server + renderer + shared types — all in one PR, no external consumers remain.
- `onboardingVerify.ts` still fetches a full transcript at verify time (out of scope here; unchanged).
- Child (subagent task-body) transcript fetches (`fetchChildTranscript`) are NOT migrated to tail-first — they were outside the issue's three named paths. Filed as a parked follow-up.

**Base: origin/main @ bbae0475**

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/home/dev/projects/better-ui/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/home/dev/projects/better-ui/node_modules
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "manta-ui",
-  "version": "0.0.19",
+  "version": "0.0.22",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "manta-ui",
-      "version": "0.0.19",
+      "version": "0.0.22",
       "hasInstallScript": true,
       "dependencies": {
         "@fontsource-variable/inter": "^5.3.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "manta-ui",
-  "version": "0.0.22",
+  "version": "0.0.19",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "manta-ui",
-      "version": "0.0.22",
+      "version": "0.0.19",
       "hasInstallScript": true,
       "dependencies": {
         "@fontsource-variable/inter": "^5.3.0",
--- a/src/renderer/ChatPanel.harness.test.tsx
+++ b/src/renderer/ChatPanel.harness.test.tsx
@@ -363,6 +363,24 @@ describe("ChatPanel session resources", () => {
     expect(api.calls.uploadBuffer ?? []).toEqual([]);
     expect(panel.container.querySelector('[title="/remote/img.png"]')).toBeNull();
   });
+
+  it("keeps staged attachments when the panel is hidden then reshown (isActive toggle must not reset the composer)", async () => {
+    // Reuse the shared attach-bridge harness (it stages an upload and returns
+    // the mounted panel) rather than duplicating the setup here.
+    const panel = await dispatchAttachBridge("ses_test");
+    expect(panel.container.querySelector('[title="/remote/img.png"]')).toBeTruthy();
+
+    // App.tsx keeps hidden panels MOUNTED (display:none) and flips isActive.
+    // If the visibility toggle re-ran the session-reset effect, the staged
+    // attachment would be wiped here — the regression BET-676 Block 2 caught.
+    panel.rerender(<ChatPanel {...PROPS} isActive={false} />);
+    await panel.flush();
+    panel.rerender(<ChatPanel {...PROPS} isActive={true} />);
+    await panel.flush();
+
+    // The staged attachment survived the hide/reshow cycle.
+    expect(panel.container.querySelector('[title="/remote/img.png"]')).toBeTruthy();
+  });
 });
 
 // ===== Transcript rendering (via the mounted ChatPanel) =====
--- a/src/renderer/ChatPanel.harness.test.tsx
+++ b/src/renderer/ChatPanel.harness.test.tsx
@@ -11,6 +11,7 @@
 import { describe, it, expect, beforeEach, afterEach } from "vitest";
 import { act } from "react";
 import { ChatPanel } from "./ChatPanel";
+import { TRANSCRIPT_TAIL_LIMIT } from "./hooks/useTranscriptState";
 import {
   installMockApi,
   resetStore,
@@ -408,6 +409,33 @@ describe("ChatPanel transcript rendering", () => {
     expect(text).toContain("general kenobi");
     // The empty-state welcome is gone once a transcript is present.
     expect(text).not.toContain("Welcome. Type a message below to start.");
+  });
+
+  it("performs exactly ONE mount fetch, tail-limited, with no self-heal double fetch", async () => {
+    const transcript = [
+      {
+        info: { id: "msg_u1", sessionID: "ses_test", role: "user" as const },
+        parts: [{ type: "text", id: "prt_u1", messageID: "msg_u1", text: "hello there" }],
+      },
+    ];
+    const fetchCalls: Array<[string, unknown]> = [];
+    installMockApi({
+      opencodeMessages: (sessionId: string, opts?: { limit?: number }) => {
+        fetchCalls.push([sessionId, opts]);
+        return Promise.resolve(transcript);
+      },
+    });
+    resetStore();
+    h = mount(<ChatPanel {...PROPS} />);
+    await h.flush();
+    await h.flush();
+    // Single fetch path — the unconditional self-heal refetch is gone. The
+    // mount fetch pulls the TAIL ({ limit: 100 }) until "Load earlier" flips
+    // loadedAllRef.
+    expect(fetchCalls.length).toBe(1);
+    expect(fetchCalls[0][0]).toBe("ses_test");
+    expect(fetchCalls[0][1]).toEqual({ limit: TRANSCRIPT_TAIL_LIMIT });
+    expect(h.text()).toContain("hello there");
   });
 });
 
--- a/src/renderer/ChatPanel.harness.test.tsx
+++ b/src/renderer/ChatPanel.harness.test.tsx
@@ -368,7 +368,7 @@ describe("ChatPanel session resources", () => {
 //
 // Verifies the extracted <Transcript> renders a fetched transcript: a user
 // turn's text and an assistant turn's text both appear in the DOM. Drives the
-// canonical fetch path (opencodeMessagesReconcile) the container uses on mount.
+// canonical fetch path (opencodeMessages) the container uses on mount.
 describe("ChatPanel transcript rendering", () => {
   let h: Harness | null = null;
 
@@ -398,7 +398,6 @@ describe("ChatPanel transcript rendering", () => {
       },
     ];
     installMockApi({
-      opencodeMessagesReconcile: () => Promise.resolve(transcript),
       opencodeMessages: () => Promise.resolve(transcript),
     });
     resetStore();
@@ -492,7 +491,6 @@ describe("ChatPanel composer submit", () => {
       },
     ];
     ({ api } = installMockApi({
-      opencodeMessagesReconcile: () => Promise.resolve(transcript),
       opencodeMessages: () => Promise.resolve(transcript),
     }));
     resetStore();
@@ -567,7 +565,6 @@ describe("ChatPanel abort rejects orphaned questions", () => {
       },
     ];
     ({ api, bus } = installMockApi({
-      opencodeMessagesReconcile: () => Promise.resolve(transcript),
       opencodeMessages: () => Promise.resolve(transcript),
     }));
     resetStore();
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -182,7 +182,7 @@ export function ChatPanel({
   // state, refresh callbacks, poll effects, session resets, and the mobile
   // `manta-open-*` window-event bridges. Extracted to a self-contained hook
   // (BET-63) because none of it touches the SSE / pin-to-bottom / message core.
-  const resources = useSessionResources(sessionId);
+  const resources = useSessionResources(sessionId, isActive);
   const {
     showSchedules,
     setShowSchedules,
@@ -226,6 +226,9 @@ export function ChatPanel({
   const [jobOwnership, setJobOwnership] = useState<DelegateJob | null>(null);
   useEffect(() => {
     setJobOwnership(null);
+    // Hidden panel — stop polling; the effect re-runs (and refires once) when
+    // isActive flips back on.
+    if (!isActive) return;
     let cancelled = false;
     const poll = async () => {
       try {
@@ -238,7 +241,7 @@ export function ChatPanel({
     void poll();
     const t = setInterval(poll, 10_000);
     return () => { cancelled = true; clearInterval(t); };
-  }, [sessionId]);
+  }, [sessionId, isActive]);
 
   const projects = useStore((s) => s.projects);
   const setActive = useStore((s) => s.setActive);
@@ -285,6 +288,8 @@ export function ChatPanel({
     scheduleRefetch,
     spliceMessage,
     toggleTaskExpand,
+    loadedAllRef,
+    fetchOpts,
   } = useTranscriptState({ sessionId, isActive });
 
   // ===== SSE bus state (extracted to useSseBus) =====
@@ -315,12 +320,15 @@ export function ChatPanel({
     rejectAllPendingQuestions,
     refreshPermissions,
     refreshQuestions,
+    transcriptLoadError,
+    retryTranscriptLoad,
   } = useSseBus({
     sessionId,
     cwd,
     setMessages,
     setRefreshing,
     scheduleRefetch,
+    fetchOpts,
     spliceMessage,
     scheduleChildRefetch: (childId: string) => {
       // Per-child debounced refetch — called when a known child's
@@ -380,6 +388,7 @@ export function ChatPanel({
   useEffect(() => {
     setPendingApproval(null);
     if (chatAutoAllowApproval) return; // trust mode → server never requests approval
+    if (!isActive) return; // hidden panel — stop polling; refire once on reactivation
     let cancelled = false;
     const poll = async () => {
       try {
@@ -398,7 +407,7 @@ export function ChatPanel({
     void poll();
     const t = setInterval(poll, 3000);
     return () => { cancelled = true; clearInterval(t); };
-  }, [sessionId, chatAutoAllowApproval]);
+  }, [sessionId, chatAutoAllowApproval, isActive]);
 
   // ===== ChatPanel-own state (not extracted to hooks) =====
   const [error, setError] = useState<string | null>(null);
@@ -454,13 +463,16 @@ export function ChatPanel({
     setAgentMentions([]);
     setSystemNotice(null);
     setDragHover(false);
-    // Branch indicator: poll every 5s while this session is mounted.
+    // Branch indicator: poll every 5s while this session is mounted AND the
+    // panel is visible. Hidden panels stop polling; the effect re-fires once
+    // when isActive flips back on.
     refreshBranch(cwd);
+    if (!isActive) return;
     const branchPoll = setInterval(() => refreshBranch(cwd), 5000);
     return () => {
       clearInterval(branchPoll);
     };
-  }, [sessionId, cwd, refreshBranch]);
+  }, [sessionId, cwd, refreshBranch, isActive]);
 
   // Ctrl+O toggles reasoning visibility. Matches Claude Code's TUI keybind.
   useEffect(() => {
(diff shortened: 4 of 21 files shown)


antoinedc:
# Review — BET-676 (cycle 1): 2 Blocks, 0 Questions, 4 Nits

Reviewed against the issue's 10 numbered implementation steps, not the PR summary.

## Mechanical gate

| Check | Result |
|---|---|
| `npm run typecheck` | **exit 0**, no errors (reproduced locally on the PR head) |
| `npm test` | **exit 0** — vitest green; node `# pass 1505 / # fail 0` |
| CI `typecheck-test` | SUCCESS |
| CI `duplication-gate` | **FAILURE** — see Block 1 |

Test-result reuse: the PR body's per-branch summary matches my PR-branch run (node count differs only because base moved). No discrepancy, no re-run on base needed.

Base freshness: branched from `bbae0475`, `origin/main` is now `5cee2a0f`. `git diff --stat origin/main...HEAD` shows only the 19 intended files — **no phantom deletions**, and GitHub reports `MERGEABLE`. Not a stale-base Block, but `main` has since moved `src/server/rpc.mjs` (BET-675 sync state), which this PR also edits — rebase before merge.

## Acceptance criteria — item by item

| # | Requirement | Verdict |
|---|---|---|
| 1 | `stripDuplicateToolOutput` applied always, in `listMessages` **and** `getMessage`, independent of `slim` | ✅ verified in `src/server/opencode.mjs:596,616` |
| 2 | Server tests: stripped both paths; non-identical left untouched | ✅ (`listMessages` non-identical is covered by the retargeted `?limit` test fixture) |
| 3 | `opencodeMessages(sessionId, opts?)`; httpApi forwards; `rpc.mjs` verified not changed | ✅ |
| 4 | Delete `opencodeMessagesCached` / `opencodeMessagesReconcile` / `reconcileMessages` everywhere + fix stale harness comment | ✅ repo-wide grep for all five identifiers returns **zero** hits; harness comment at `ChatPanel.harness.test.tsx:372` corrected |
| 5 | `TRANSCRIPT_TAIL_LIMIT=100`, `loadedAllRef` (reset on session change), `fetchOpts()`, all three fetch paths | ✅ mount fetch, `scheduleRefetch` (`useTranscriptState.ts:204`), `server.connected` resync (via `scheduleRefetch`, `useSseBus.ts:645`) |
| 6 | 15s timeout + one 2s retry; `loadError` → existing panel + Retry button | ✅ |
| 7 | Unconditional self-heal `scheduleRefetch()` deleted; seed guard kept | ✅ |
| 8 | `LoadEarlier` — single exported unit, scroll preserved, `loadedAll` flipped, seam comment | ✅ (see Nits 2 & 3) |
| 9 | `replayChatAttention` → `{ limit: 20 }` | ✅ `store.ts:1005` |
| 10 | Four polls gated on `isActive` | ⚠️ implemented, but the vcs-branch one regresses composer state — **Block 2** |

---

## Blocks

### **Block (fix-here) 1 — `duplication-gate` CI job is red; this PR introduces one of the clones**

`https://github.com/antoinedc/MantaUI/actions/runs/31163016243/job/92817423787` fails. Reproduced locally with `bash scripts/check-duplication-gate.sh origin/main`: 4 clones, all inside `src/server/opencode.test.mjs`.

jscpd ranges on this branch vs. base (`git show origin/main:src/server/opencode.test.mjs`):

```
PR head:   78-86 <-> 179-187   (9)   pre-existing
        1188-1202 <-> 1216-1230 (15)  <-- INTRODUCED BY THIS PR
        1625-1640 <-> 1676-1692 (16)  pre-existing
        1625-1639 <-> 1711-1725 (15)  pre-existing
base:      76-84 <-> 177-185   (9)
        1510-1525 <-> 1561-1577 (16)
        1510-1524 <-> 1596-1610 (15)
```

`1188-1202 ↔ 1216-1230` is the `withMockFetch` boilerplate copied verbatim between the two new tests `getMessage strips a byte-identical tool stdout duplicate (splice path)` and `getMessage leaves a NON-identical metadata.output (live streaming) untouched` — identical URL matcher, identical `new Response(JSON.stringify(raw), …)`, identical 204 fallback. Extract a small local helper, e.g.

```js
const withMessageFetch = (raw, body) =>
  withMockFetch(
    async (url) =>
      String(url).includes("/message/m1")
        ? new Response(JSON.stringify(raw), { status: 200, headers: { "content-type": "application/json" } })
        : new Response(null, { status: 204 }),
    body,
  );
```

Note the gate is STRICT and file-scoped: because this PR touches `opencode.test.mjs`, the **three pre-existing** clones in that file now count against it too. Extracting only the new one will not turn CI green. The other three are the same shape of test boilerplate (the shared `_setEventStreamTransport` / `_resetSessionDirectoryCache` / `_resetStreamReadyState` preamble in the `subscribeEvents` tests, and the pair at 78/179) — a single `setupEventStreamCapture()` helper collapses all three. Adding `src/server/opencode.test.mjs` to `scripts/duplication-gate-ignore.txt` is **not** an option here: that file is human-approval class and the exemption is documented for the desktop↔mobile mirrors only.

### **Block (fix-here) 2 — hidden-panel gating regressed the composer: staged attachments and @mentions are wiped on every tab switch**

`src/renderer/ChatPanel.tsx:459-475`. `isActive` was added to the dep array of an effect that is **not** just the branch poll — it is the session-reset effect:

```js
useEffect(() => {
  setError(null);
  setModelOverride(readSavedModel(sessionId) ?? configDefaultModel ?? null);
  setAttachments([]);
  setAgentMentions([]);
  setSystemNotice(null);
  setDragHover(false);
  refreshBranch(cwd);
  if (!isActive) return;
  const branchPoll = setInterval(() => refreshBranch(cwd), 5000);
  return () => clearInterval(branchPoll);
}, [sessionId, cwd, refreshBranch, isActive]);   // <-- isActive added
```

Chat panels are **not unmounted** when hidden — `App.tsx:1269` hides them with `style={{ display: "none" }}` and flips `isActive` (`App.tsx:1276`). That is the whole premise of step 10. So every visibility toggle now re-runs the reset body:

- a user who has staged file attachments, switches to another session/window, and switches back **loses the attachments** (`setAttachments([])`);
- pending `@agent` mentions are cleared (`setAgentMentions([])`);
- the `/help` system notice is dismissed (`setSystemNotice(null)`);
- an `error` panel is silently cleared;
- `modelOverride` is re-read from storage — a no-op for `selectModel` (which persists via `writeSavedModel`) but it clobbers the non-persisted autoSubmit override set at `ChatPanel.tsx:890`.

None of that is in the issue's scope; step 10 only asked to gate the interval. Fix: split the poll into its own effect so the reset keeps its original `[sessionId, cwd, refreshBranch]` deps.

```js
useEffect(() => { /* resets only */ }, [sessionId, cwd, refreshBranch]);

useEffect(() => {
  if (!isActive) return;
  refreshBranch(cwd);
  const t = setInterval(() => refreshBranch(cwd), 5000);
  return () => clearInterval(t);
}, [cwd, refreshBranch, isActive]);
```

Please add a regression test (mount with `isActive=true`, stage an attachment, re-render with `isActive=false` then `true`, assert the attachment survives). The other three gates (delegateList `:229`, approvals `:391`, schedules `useSessionResources.ts:148`) are in effects whose bodies are already poll-only — those are fine.

---

## Nits (non-blocking)

1. **`slimPart` is now a pure pass-through.** `src/server/opencode.mjs:530` — `function slimPart(part) { return stripDuplicateToolOutputPart(part); }`. Dead indirection; call `stripDuplicateToolOutputPart` directly at the one remaining call site and delete `slimPart`.
2. **Unreachable "Loading…" label.** `Transcript.tsx:60` early-returns `null` when `loading` is true, so the ternary at `Transcript.tsx:101` never renders `"Loading…"`. Net UX: clicking "Load earlier messages" makes the button vanish with no feedback until the fetch lands. Either drop the `loading` term from the early return (keep the button, disabled, showing "Loading…") or delete the dead branch.
3. **Timeout timer is never cleared.** `chatUtils.ts:2358` — the `setTimeout` in `withTranscriptFetchTimeout` survives a fast resolve, leaving a 15s pending timer per fetch. No unhandled rejection (`Promise.race` attaches handlers to both legs), so it is cosmetic, but a `clearTimeout` in a `finally` is one line.
4. **Exactly-100 sessions show a no-op button.** `messages.length >= TRANSCRIPT_TAIL_LIMIT` is true for a session with exactly 100 messages; clicking re-fetches the same 100. Matches the issue's wording, so leaving it is fine — worth a comment noting the known false positive.

## Anti-spaghetti advisory report

Read. Six clones repo-wide across the changed scope, all `javascript`/`opencode.test.mjs`, 0 in `.ts`/`.tsx`. Nothing in the production diff duplicates business logic, and no desktop/mobile mirror was broken (the transport-mirror files are all on the documented ignore list and their changes stay in lockstep). The only actionable duplication is the test boilerplate covered by Block 1.

antoinedc:
Review round 1 addressed — pushed to the same branch, rebased onto `origin/main @ 73b5866f`.

**Block 1 — `duplication-gate`.** Extracted the clone boilerplate in `src/server/opencode.test.mjs`:
- `withMessageFetch(raw, body)` — the two new `getMessage` tests (the clone this PR introduced).
- `withSessionDirectoryMock(template, body)` — the createSession/runCommand pair (pre-existing, now in scope).
- `setupEventStreamCapture()` — the three `subscribeEvents` tests' shared event-stream preamble + teardown.

`bash scripts/check-duplication-gate.sh origin/main` now **PASS — no clones** (14 files scanned after the server + renderer changes… 16 changed, all clean).

**Block 2 — composer reset on visibility toggle.** Split the vcs-branch poll out of the session-reset effect into its own visibility-gated effect; the reset effect keeps its original `[sessionId, cwd, refreshBranch]` deps and no longer clears staged attachments / @mentions / /help on an `isActive` flip. Added a regression test: stage an attachment via the `manta-attach-files` bridge, rerender `isActive=false` then `true`, assert the attachment chip survives.

**Nits**
1. Deleted the `slimPart` pass-through — `slimTranscript` calls `stripDuplicateToolOutputPart` directly.
2. `LoadEarlier` no longer early-returns on `loading`; the button stays mounted, disabled, showing "Loading…" while the full fetch runs.
3. `withTranscriptFetchTimeout` clears its timer on settle.
4. Commented the exactly-100 false positive in `LoadEarlier`.

**Verification (local, PR head `cb2a7198`):** typecheck exit 0; vitest 117 files / 2216 passed / 7 skipped / 0 fail; node tests 1523 pass / 0 fail; duplication-gate PASS.

antoinedc:
# Review — BET-676 (cycle 2): 1 Block, 0 Questions, 0 new Nits

Re-reviewed at head `cb2a7198`, rebased onto `origin/main @ 73b5866f` (merge-base == `origin/main`, so the diff is exactly this PR's work). Both round-1 Blocks and all 4 nits are genuinely fixed — verified against the code, not the summary.

## Mechanical gate (reproduced locally on the PR head)

| Check | Result |
|---|---|
| `npm run typecheck` | **exit 0**, no errors |
| `npm test` | **exit 0** — vitest `117 passed (117)`, `2216 passed / 7 skipped`; node `# pass 1523 / # fail 0` |
| `bash scripts/check-duplication-gate.sh origin/main` | **PASS — no clones** (16 files scanned) |
| CI `typecheck-test` / `duplication-gate` / agent-branch-PR | all SUCCESS |

Matches the reported numbers exactly. No re-run on base needed.

## Round-1 findings — verification

**Block 1 (duplication-gate)** — ✅ fixed and independently confirmed. jscpd now reports zero clones across the changed set (base had 3 pre-existing + 1 PR-introduced in `opencode.test.mjs`). The three extracted helpers — `withMessageFetch`, `withSessionDirectoryMock`, `setupEventStreamCapture` — are real deduplication, not gate-gaming: each collapses a genuinely identical setup prefix, the call sites read better than before, and no assertion was weakened or deleted (node test count went **up**, 1501 → 1523).

**Block 2 (composer reset on visibility toggle)** — ✅ fixed correctly. `ChatPanel.tsx:463` keeps the reset effect on its original `[sessionId, cwd, refreshBranch]` deps with an explanatory comment about why `isActive` must not be a dep, and the branch poll now lives in its own `[cwd, refreshBranch, isActive]` effect that fires `refreshBranch` once on entry. The regression test (`ChatPanel.harness.test.tsx`, "keeps staged attachments when the panel is hidden then reshown") stages via the attach bridge, rerenders `isActive=false` → `true`, and asserts the chip survives — it would have failed against the round-1 code.

**Nits 1–4** — all ✅: `slimPart` pass-through deleted (`slimTranscript` calls `stripDuplicateToolOutputPart` directly); `LoadEarlier` keeps the button mounted + `disabled` with a live "Loading…" label; `withTranscriptFetchTimeout` clears its timer in `.finally`; the exactly-100 false positive is documented in place.

---

## Block

### **Block (fix-here) — two unrelated files were committed in the round-2 fix commit; one of them deletes a tracked file and breaks working trees on pull**

`cb2a7198` ("review round 1 — duplication gate, composer reset, nits") contains two changes with nothing to do with BET-676, mentioned nowhere in the PR body or the round-2 comment:

```
 node_modules      |  1 -
 package-lock.json |  4 +-
```

```diff
diff --git a/node_modules b/node_modules
deleted file mode 120000
-/home/dev/projects/better-ui/node_modules
```

`node_modules` is a tracked **symlink** on `main` (added in `8151a5f6`; `.gitignore` has `node_modules/` with a trailing slash, which does not match a symlink, so it slipped in). This PR deletes it.

This is not theoretical — I hit it doing the review. Checking out this branch removed `node_modules` from my working tree and the toolchain went dead (`sh: 1: tsc: not found`, `sh: 1: vitest: not found`); I had to `npm ci` (844 packages) before I could run the gates. Anyone who pulls after this merges gets the same, from a PR titled "always strip duplicate tool stdout; drop cached/reconcile message channels". A repo-layout change that silently invalidates every checkout's `node_modules` needs to be its own reviewed change, not a rider.

The `package-lock.json` hunk is a version-field-only bump (`0.0.19` → `0.0.22`), a harmless byproduct of running `npm install` locally — but it is equally unrequested and belongs with the same cleanup.

**Fix:** restore both paths and re-push.

```bash
git checkout origin/main -- node_modules package-lock.json
git commit -m "chore: drop unrelated node_modules/package-lock changes" node_modules package-lock.json
```

Then confirm `git diff --stat origin/main...HEAD` lists only the 19 `src/` files. Commit explicit paths rather than `git add -A` / `git commit -a`, otherwise the tracked symlink will be re-deleted the moment your working tree has a real `node_modules` directory instead.

To be explicit: I am **not** asking you to keep the symlink forever. It is machine-specific junk that probably should go — but that is the repo owner's call, made in a one-line PR they can see, not a side effect of a transcript-loading change. If you think it should be removed now, file it as its own issue.

## Anti-spaghetti advisory report

Clean. Zero clones in the changed scope after the helper extraction, and nothing in the production diff duplicates business logic or leaves dead code behind (`slimPart` and the two vestigial IPC channels were removed, not orphaned). No desktop/mobile transport mirror was broken.

---

Everything else is merge-ready. Drop the two stray files and this passes.

antoinedc:
Scope Block addressed. The round-2 fix commit accidentally swept in two unrelated files (a `git add -A` after a local `npm install`):
- it deleted the tracked `node_modules` **symlink** (npm replaced it with a real directory, and `git add -A` staged the deletion), and
- it bumped `package-lock.json`.

Restored **both** from `origin/main` in a follow-up commit so this PR touches only transcript code:
- `git checkout origin/main -- node_modules package-lock.json` → `node_modules` is back as the tracked symlink (mode `120000`, blob `0e9d440…`, byte-identical to `main`), and `package-lock.json` matches `main`.

**Diff vs `origin/main` is back to the 19 intended files** (`git diff origin/main...HEAD --stat`: node_modules + package-lock.json are clean). No source/code change — the previously-verified typecheck/tests/duplication-gate results are unaffected. If the `node_modules` symlink should be removed (it is a machine-specific absolute path), that belongs in its own visible one-line change, not this PR.

antoinedc:
# Review — BET-676 (cycle 3): **PASS** — 0 Blocks, 0 Questions, 1 Nit

Head `7bc5824d`, base `origin/main @ 73b5866f` (merge-base == `origin/main`). Routing to `manta-pm` for merge.

## Round-2 Block — verified fixed

The scope Block is resolved, and the fix is exactly the minimum it should be:

- `7bc5824d` touches **only** `node_modules` (+1) and `package-lock.json` (4 ±) — both restored to `origin/main`'s content.
- `git diff cb2a7198 HEAD -- src/` is **empty**: not one byte of source changed while fixing the scope problem, so the substantive review from cycle 2 carries over unchanged rather than needing a re-read.
- `git diff --stat origin/main...HEAD` is back to exactly the **19 intended `src/` files**, +628/−166.

## Mechanical gate (re-run locally at `7bc5824d`)

| Check | Result |
|---|---|
| `npm run typecheck` | **exit 0**, no errors |
| `npm test` | **exit 0** — vitest `117 passed (117)`, `2216 passed / 7 skipped`; node `# pass 1523 / # fail 0` |
| `bash scripts/check-duplication-gate.sh origin/main` | **PASS — no clones** (16 files scanned) |
| CI: `typecheck-test`, `duplication-gate`, agent-branch-PR | all SUCCESS |
| `mergeable` | MERGEABLE, not draft |

## Acceptance criteria — final state

All 10 numbered implementation steps verified against the code across the three cycles: server-side `stripDuplicateToolOutput` applied unconditionally in both `listMessages` and the `getMessage` splice path; the two vestigial IPC channels and `reconcileMessages` gone (repo-wide grep for all five identifiers returns zero hits); `TRANSCRIPT_TAIL_LIMIT`/`loadedAllRef`/`fetchOpts` covering all three fetch paths with a harness test asserting exactly one tail-limited mount fetch; 15s timeout + one retry + Retry button; self-heal double fetch removed with the seed guard intact; `LoadEarlier` as a single exported unit with scroll preservation; `replayChatAttention` at `{ limit: 20 }`; four hidden-panel polls gated on `isActive` without disturbing the session-reset effect.

Both earlier Blocks and all four nits stayed fixed — no regressions between cycles.

## Nit (non-blocking, for the PM's awareness — do not hold the merge for it)

The **PR body's verification block is stale**: it still cites branch `62ec1f18` / base `bbae0475` with "vitest 119 files, 2217 passed / node 1501" — three heads and one rebase ago. The *What* section and the requirement description are accurate; only the numbers are out of date. The current, independently reproduced figures are in the table above and supersede them.

## Anti-spaghetti advisory report

Clean. Zero clones in the changed scope, no dead code left behind, no desktop/mobile transport mirror broken.

---

### Unrelated observation (not part of this PR — worth its own issue)

The tracked `node_modules` **symlink** on `main` (`node_modules -> /home/dev/projects/better-ui/node_modules`, added in `8151a5f6`; `.gitignore`'s `node_modules/` has a trailing slash so it doesn't match a symlink) is a live foot-gun. Its target is currently empty, so a fresh checkout of `main` has no working toolchain until a reinstall, and any `npm ci` replaces the symlink with a real directory — which then shows up as a spurious `D node_modules` in `git status` and gets swept into the next `git add -A`. That is precisely how it landed in this PR by accident in round 2. Removing it from tracking (and tightening the ignore pattern) is the right fix, but it needs to be its own visible one-line change the owner reviews — correctly out of scope here.

antoinedc:
FYI: the node_modules symlink you were asked to restore was itself an accidental check-in (swept into 8151a5f by a `git add -A` from a worktree; the `node_modules/` ignore pattern's trailing slash only matches directories, so the symlink form was trackable). Proper removal is now its own one-line PR: #680. Suggested merge order: this PR first, then #680 — otherwise this branch re-adds the symlink on merge.

PR #673, opened 2026-08-07
feat(server) BET-675: materialized sync state + cursor snapshot RPC

antoinedc:
## BET-675 — Sync core: materialized session/config state + cursor snapshot RPC on manta-server

SERVER-ONLY. No `src/renderer/` changes (sibling renderer issue consumes the new `sync:snapshot` channel).

### What changed

1. **`tmux.mjs` — reclassify errors (step 1).** `isNoTmuxServerError` now matches ONLY genuinely-empty: `no server running on …` and `no sessions`. The `error connecting to … (No such file or directory)` and `Connection refused` variants are now a FAULT and throw out of `tmuxListOutput`/`listProjects` instead of returning `""` — so a socket race / mid-restart can no longer masquerade as "zero sessions" (empty sidebar + pruned ownership sidecar).
2. **`tmux.mjs` — child timeout (step 2).** `spawnRun` now has `SPAWN_TIMEOUT_MS = 10_000`; a wedged tmux child is SIGKILL'd and rejected with `tmux timed out after 10s`. Timer unref'd, cleared on normal exit. No dedicated test (would need a wedged tmux binary).
3. **`syncState.mjs` (new, step 3).** Pure + injected-I/O materialized state `{gen, seq, projects, config, stale}` with monotonic cursor, deduped `refreshNow()`, `applyConfig`, `snapshot()`, `payloadSince()`, `everSucceeded()`. Publishes `{kind:"sync", gen, seq, changed:{<field>}}` single-field deltas.
4. **`index.mjs` (step 4).** Constructs `createSyncState`, seeds config at startup, a 2s sync poller (in-flight guard + unref'd timer, one immediate tick at boot), and passes `syncState` into `buildHandlers`.
5. **`rpc.mjs` — `sync:snapshot` channel (step 5).** `{ sinceSeq?, sinceGen? }` → `syncState.payloadSince(...)` (full snapshot on absent/older-gen/impossible cursor).
6. **`rpc.mjs` — `tmux:list` from memory (step 6).** Returns `snapshot().projects`; guarantees one `refreshNow()` before serving if never succeeded. Channel kept.
7. **`rpc.mjs` — fresh state after mutations (step 7).** `await refreshNow()` before returning for: `tmux:new-session`, `tmux:new-window`, `tmux:kill-session`, `tmux:kill-window`, `tmux:rename-session`, `tmux:rename-window`, and the window paths of `opencode:fork-session` / `opencode:clear-session`. Also added to `opencode:delete-session` (it kills a window — same intent).
8. **`rpc.mjs` — config changes publish (step 8).** `config:update` now calls `syncState.applyConfig(next)` after the write succeeds.
9. **Bus `/events` forwarding (step 9).** Confirmed `events.mjs` forwards every envelope `{kind,payload}` with no kind-filtering — `sync` envelopes flow through unmodified. No change needed.

### Follow-ups
- `tmux:select-window` flips `active` flags but isn't in the refresh list, so the active-window delta rides the 2s poller instead of publishing instantly. Parked as a follow-up (minor, benign).

### Verification results
- PR branch (`multica/BET-675-sync-core` @ `4b5a95db`):
  - test: exit 0, 1515 pass / 0 fail / 0 skip. Failures: none. log sha256: `72122f8bced81f8d1c9b446cbd31c66a3ba3ad6d1a40f832f9ea9dd11f0578b7`
- Base (`main` @ `bbae0475`): no independent re-run — PR is additive (new module + wiring); 0 failures on PR branch.

Logs cached at `/tmp/tc-pr.log`, `/tmp/test-pr.log`.

### DIFF
--- a/src/server/index.mjs
+++ b/src/server/index.mjs
@@ -40,6 +40,7 @@ import { createBus, handleEventsRequest, attachEventsWs } from "./events.mjs";
 import { attachPtyWs } from "./ptyWs.mjs";
 import { buildHandlers, handleRpcRequest } from "./rpc.mjs";
 import { startStatusPoller } from "./status.mjs";
+import { createSyncState } from "./syncState.mjs";
 import { createStreamInterpreter } from "./streamInterp.mjs";
 import { startOutboxPoller, pushArtifact, createArtifactSweep } from "./outbox.mjs";
 import { startUploadCleanupPoller } from "./uploads.mjs";
@@ -108,6 +109,18 @@ const streamInterp = createStreamInterpreter({ publish: (evt) => bus.publish(evt
 // store entry uses this same deps object.
 const BUS_PUBLISH_DEPS = { publish: (evt) => bus.publish(evt) };
 
+// BET-675: materialized in-memory session/config state. tmux:list is served
+// from memory (never a per-request tmux shell-out), and `sync` deltas are
+// published on the bus as state changes so clients can recover just what they
+// missed. `refreshNow()` is driven by the poller below; `tmux:list` lazily
+// guarantees a first tick before serving anything.
+const syncState = createSyncState({
+  listProjects: () => tmux.listProjects(),
+  publish: (env) => bus.publish(env),
+});
+// Seed the config baseline at startup so the first snapshot already carries it.
+syncState.applyConfig(await local.configGet());
+
 // DELETE handler for /api/<store> endpoints: `?id=<id>` → store.deleteFn(id,
 // BUS_PUBLISH_DEPS) → 200 {deleted:bool}. The boilerplate (id-required 400,
 // the await + publish-deps call, the success response) is identical across
@@ -156,6 +169,25 @@ push.setDesktopSink((payload) =>
 // mobile sidebar's activity/attention dots work (parity with desktop status.ts).
 // eslint-disable-next-line no-unused-vars
 const { stop: stopStatusPoller } = startStatusPoller(bus, { intervalMs: 2000 });
+
+// BET-675: sync-state poller — materialize tmux session/config state in memory
+// every 2s. Follows the exact shape of the other pollers (in-flight guard +
+// unref'd timer so the poller alone never holds the process open). Fire one
+// tick immediately at startup (not awaited) so state is warm without waiting a
+// full interval.
+let syncInFlight = false;
+async function syncTick() {
+  if (syncInFlight) return;
+  syncInFlight = true;
+  try {
+    await syncState.refreshNow();
+  } finally {
+    syncInFlight = false;
+  }
+}
+void syncTick();
+const syncTimer = setInterval(() => void syncTick(), 2000);
+syncTimer.unref();
 
 // Agent → device file push: watch ~/.manta-outbox/ for files the AI drops and
 // publish `agentFile` bus events so connected devices show a "Save" toast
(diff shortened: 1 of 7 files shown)


antoinedc:
## BET-675 review — PASS (0 Blocks, 0 Questions, 1 Nit)

**Base freshness:** branched from `bbae0475`, which equals current `origin/main` tip. No stale-base artifacts; real diff vs fresh main is exactly the 7 server files.

**Mechanical gate (independently re-run):**
- `npm test` → 1515 pass / 0 fail / 0 skip / exit 0.
- Matches the implementer's reported results exactly.

**Anti-spaghetti:** jscpd scoped to changed files → **0 clones** (duplication-gate CI also green).

### Acceptance-criteria checklist — each item verified against the code

1. **tmux error reclassification** — `isNoTmuxServerError` in `src/server/tmux.mjs:284` now matches only `no server running` / `no sessions`; the `error connecting … (No such file or directory)` / `Connection refused` variants now throw through `tmuxListOutput` → `listProjects`. Pinned cases in `tmux.test.mjs` flipped/added accordingly, incl. a new test asserting the socket-race throws. ✓
2. **Child-process timeout** — `SPAWN_TIMEOUT_MS = 10_000` in `spawnRun`; unref'd `setTimeout` that SIGKILLs and rejects `tmux timed out after 10s`, cleared on normal exit. ✓
3. **`syncState.mjs` per contract** — `{gen,seq,projects,config,stale}` with monotonic cursor, deduped `refreshNow`, `applyConfig`, `snapshot`, `payloadSince`, `everSucceeded`; single-field `{kind:"sync",gen,seq,changed}` deltas. Contract conformance checked line-by-line (version-rule gating incl. `stale`, full-snapshot on null/mismatched-gen/impossible-seq, last-known-good preserved on fault, JSON-stringify change detection). ✓
4. **Wiring in `index.mjs`** — `createSyncState` with `listProjects`/`bus.publish`; config seeded via `applyConfig(await local.configGet())`; 2s `syncTick` poller with in-flight guard + unref'd timer; one immediate un-awaited tick at startup. ✓
5. **`sync:snapshot` RPC** — `payloadSince(sinceSeq, sinceGen)`. ✓
6. **`tmux:list` from memory** — returns `snapshot().projects`, guarantees one `refreshNow()` before serving if `!everSucceeded()`; channel kept. ✓
7. **Refresh after mutations** — `refreshNow()` before return for new-session, new-window, kill-session, kill-window, rename-session, rename-window, fork/clear-session (plus delete-session — same window-killing intent, in-scope). ✓
8. **Config changes publish** — `config:update` calls `applyConfig(next)` after the write; confirmed `configUpdate` returns the full merged config (sanitized return value). ✓
9. **`/events` forwards `sync`** — both SSE (`events.mjs:30`) and WS (`events.mjs:45`) subscribe to every `{kind,payload}` with no kind filter. ✓

Tests match the issue's spec: seq/version semantics, failure→stale with last-known-good, recovery→stale=false, all `payloadSince` cursor cases, concurrent dedupe, envelope shape pinned.

### Findings

- **Nit (not blocking):** `tmux:select-window` flips `active` flags but isn't in the refresh list, so the active-window delta rides the next 2s poll tick instead of publishing instantly. Benign ≤2s latency; not in the issue's enumerated contract. Already disclosed and filed as follow-up **BET-681**. No change required here.

No Block or Question findings. Routing to PM for merge.

PR #677, opened 2026-08-07
BET-677: feat(renderer): zero layout shift — CardMount, overlay toasts, restored working indicator

antoinedc:
**BET-677 — Feel: zero layout shift — working indicator, overlay toasts, animated card mounts**

Renderer-only. `framer-motion` was already a dependency; no new packages.

Rebased onto current `origin/main` (`7f2b9bf4`) and reconciled a `Transcript.tsx` conflict with main's parallel `LoadEarlier` feature (both retained — `LoadEarlier` above the first row, this PR's `WorkingIndicator` tail below all messages).

### What changed
1. **`CardMount`** (`src/renderer/components/CardMount.tsx`) — one shared mount/unmount animation (height `0 ⇄ auto` + opacity, 0.22s, `initial={false}`). Deliberately generic (no card styling) for the sibling tool-card-expansion issue to reuse.
2. **ChatPanel** — every composer-stack conditional mount animates through `CardMount`: PermissionCard, RetryCard, CompactionCard, DelegateApprovalCard, ScheduledTasksCard, SecretsCard, WebhooksCard, queued-messages list, sendError banner. Stale `manta-loading-divider` comment removed.
3. **ToastStack → overlay:** `ToastStack` is **layout-neutral** — it owns only the per-toast framer-motion in/out animation (root is `pointer-events-none`, each toast `pointer-events-auto` so no surface loses clickability). The overlay framing lives in **ChatPanel**: a `relative` composer wrapper anchors the overlay to the composer's **real top edge** (`bottom: 100%`), so toasts hover just above the box however tall the input grows — and never affect layout. The Settings dialog keeps its own pre-existing `fixed` bottom wrapper.
4. **WorkingIndicator restored** — an always-rendered `28px` transcript-tail row toggling `visibility` (never mounts/unmounts), `MantaLoader` + `Working…`. Reserved slot is exactly its own 28px (turn-gap canceled). Send→first-token never dead air, never reflows.
5. **Composer border** `transition-colors duration-200`.
6. **Deleted** dead `.manta-loading-divider` CSS + its stale ChatPanel comment.
7. **Reduced motion** moved to the app root (`MotionConfig` wraps App), transcript-local duplicate removed.

### Tests
`CardMount` (render / remove-after-exit / overflow-hidden), `WorkingIndicator` (constant 28px + visibility toggle).

Follow-up filed: **BET-684** (parked, doc-only — drop the now-stale `manta-loading-divider` mentions from `AGENTS.md`).

**Verification results:**
- PR branch (`multica/BET-677-layout-shift` @ `4bd72f4e`):
  - test: exit 0. Renderer 2224 pass / 0 fail / 7 skip; server 1523 pass / 0 fail.
- Base: rebased onto `origin/main @ 7f2b9bf4`. **Conclusion: 0 new failures.**

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/home/dev/projects/better-ui/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/home/dev/projects/better-ui/node_modules
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -1,4 +1,5 @@
 import { useEffect, useRef, useState, type CSSProperties } from "react";
+import { MotionConfig } from "framer-motion";
 import { Terminal as TerminalIcon } from "lucide-react";
 import { Sidebar, type SidebarHandle } from "./Sidebar";
 import { Terminal } from "./Terminal";
@@ -73,7 +74,13 @@ export function App() {
         </div>
       )}
     >
-      <AppInner />
+      {/* App-level reduced motion (BET-677): wraps the whole tree so EVERY
+          framer-motion animation — chat entry, CardMount card mounts, toast
+          in/out — respects the OS `prefers-reduced-motion` setting, not just
+          the transcript's. The old transcript-local duplicate was removed. */}
+      <MotionConfig reducedMotion="user">
+        <AppInner />
+      </MotionConfig>
     </ErrorBoundary>
   );
 }
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -2347,7 +2347,20 @@ export function ChatPanel({
           Only the active panel renders it — the toasts live in global store
           state (screenshot / agent-file) or panel-local state (systemNotice),
           one instance app-wide. */}
-      {isActive && <ToastStack toasts={toasts} onDismiss={dismissToast} />}
+      {/* Bottom cluster: a relative wrapper around the composer so the toast
+          overlay can anchor to the composer's OWN top edge (bottom:100%)
+          instead of a fixed pixel guess. Toasts therefore stay just above the
+          box however tall it grows with the input, and as a pure overlay they
+          never shift the transcript or composer layout (BET-677). */}
+      <div className="relative shrink-0">
+        {isActive && (
+          <div
+            className="pointer-events-none absolute inset-x-0 z-40"
+            style={{ bottom: "100%" }}
+          >
+            <ToastStack toasts={toasts} onDismiss={dismissToast} />
+          </div>
+        )}
 
       {jobOwnership ? (
         <ReadOnlyJobBar
@@ -2453,6 +2466,7 @@ export function ChatPanel({
         onPaste={onPaste}
       />
       )}
+      </div>
     </div>
   );
 }
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -66,6 +66,7 @@ import { MantaLoader } from "./MantaLoader";
 import { MeasureColumn } from "./MeasureColumn";
 import { CompactionCard, PermissionCard, RetryCard } from "./Cards";
 import { DelegateApprovalCard, ReadOnlyJobBar, ScheduledTasksCard, SecretsCard, WebhooksCard } from "./PanelCards";
+import { CardMount } from "./components/CardMount";
 import { useSessionResources } from "./hooks/useSessionResources";
 import { useInputHistory } from "./hooks/useInputHistory";
 import { useTranscriptState } from "./hooks/useTranscriptState";
@@ -2108,7 +2109,8 @@ export function ChatPanel({
       {/* so they're hard to miss — tool execution pauses until reply. */}
       {/* Hidden for a job session (BET-418 §D: a job's pre-flight ruleset */}
       {/* means it never generates asks, but gate defensively anyway). */}
-      {!jobOwnership && permissions.length > 0 && (
+      {!jobOwnership && (
+      <CardMount show={permissions.length > 0} k="permission">
         <div className="shrink-0 px-4 pt-2">
           {/* BET-458: the permission card is a block in the conversation, not
               an overlay — bound it to the same 72ch measure the transcript
@@ -2124,20 +2126,23 @@ export function ChatPanel({
             ))}
           </div>
         </div>
+      </CardMount>
       )}
 
       {/* Retry status — surfaces session.status "retry" so the user can */}
       {/* see WHY the spinner is still spinning (rate limit, transient API */}
       {/* failure, etc) instead of assuming the AI is stalled. */}
-      {retryInfo && (
-        <div className="shrink-0 px-4 pt-2">
-          <RetryCard info={retryInfo} />
-        </div>
-      )}
+      <CardMount show={!!retryInfo} k="retry">
+        {retryInfo && (
+          <div className="shrink-0 px-4 pt-2">
+            <RetryCard info={retryInfo} />
+          </div>
+        )}
+      </CardMount>
 
       {/* Transcript-loading card removed (BET-251). The warm-stale-reopen */}
       {/* refetch is now surfaced as an ambient orange sweep on the composer's */}
-      {/* top divider — see InputArea + `manta-loading-divider` in index.css. */}
+      {/* top divider — see InputArea. */}
       {/* Cold-load (`messages === null`) is still covered by the full-screen */}
       {/* "Connecting to session…" spinner above. */}
       {/* `refreshing` is still threaded into the composer below. */}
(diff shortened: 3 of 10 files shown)


antoinedc:
## Review — BET-677 (cycle 1)

**Outcome: 2 Blocks, 0 Questions, 3 Nits.** Returned to `manta-dev`.

### Mechanical gate

| check | result |
|---|---|
| base freshness | merge-base `bbae0475`, `origin/main` `73b5866`. `git diff --stat origin/main...HEAD` = 9 files, no phantom deletions. Base is fine. |
| `npm run typecheck` | exit 0, no errors |
| `npm test` | exit 0 — renderer 119 files / 2217 pass / 0 fail / 7 skip; server 1501 pass / 0 fail |
| CI | typecheck-test pass, duplication-gate pass |
| anti-spaghetti | 8 files, 0 clones, 0% duplicated |

Branch run matches the numbers in the PR body exactly.

### Acceptance-criteria checklist (issue steps 1-7)

1. `CardMount` generic, no card styling — ✅
2. All 9 composer-stack mounts wrapped (permission, retry, compaction, delegate-approval, schedules, secrets, webhooks, queued, send-error) — ✅ verified each in the diff
3. ToastStack overlay — ❌ see Block 1 + Block 2
4. WorkingIndicator restored — ✅ always-rendered 28px row, `visibility` toggle, `MantaLoader` (inline = 24px), `Working…` in `text-text-faint text-xs`; `running` flips optimistically at `src/renderer/ChatPanel.tsx:701`
5. Composer `transition-colors duration-200` — ✅
6. `.manta-loading-divider` CSS + ChatPanel comment deleted — ✅; repo-wide grep leaves only AGENTS.md (Nit 3). No visual-gate reference exists.
7. `MotionConfig` at app root, transcript duplicate removed — ✅; both `createRoot` call sites in `main.tsx` render `<App/>`, so nothing escapes the wrapper
8. Tests (CardMount + WorkingIndicator) — ✅

---

### **Block (fix-here) 1 — the overlay change regresses the Settings toast stack**

`ToastStack` is not chat-only. `src/renderer/Settings.tsx:917` renders the same component (re-exported through `settingsApply.ts:93`) inside:

```
<div className="pointer-events-none fixed inset-x-0 bottom-0 z-50 flex justify-center px-4 pb-4">
  <div className="pointer-events-auto w-full max-w-[420px]">
    <ToastStack … />
```

Making the stack `absolute inset-x-0` + `bottom: 112` changes that call site too:

- the `max-w-[420px]` wrapper is static, so the containing block becomes the outer **fixed** div — the 420px cap is bypassed and settings toasts now span the full viewport width;
- that wrapper collapses to height 0 (its only child is out of flow);
- settings toasts jump ~112px up from the bottom edge, positioned by a constant named after the **chat composer**, which does not exist in Settings.

The PR body's "the chat column container was already `relative`" is true for ChatPanel, but the second consumer was never considered. Unscoped visual regression introduced by this PR.

### **Block (fix-here) 2 — `TOAST_BOTTOM = 112` does not track the composer**

The spec asks for the stack to be "bottom anchored just above the composer". A hard-coded 112px is a guess at the composer cluster's *resting* height. The composer autogrows (multi-line textarea, attachment chips) and the cluster also carries the model/effort row, so as soon as the input passes ~3 lines the toast lands on top of the composer — the opposite of the requirement. It is reachable by typing, not an edge case.

**One fix covers both Blocks:** keep `ToastStack` layout-neutral (the old in-flow flex column, plus the new per-toast framer-motion in/out, which is the good part of this change) and move the overlay positioning into ChatPanel — wrap the composer region in a `relative` container and render the stack as an `absolute bottom-full inset-x-0 z-40 pointer-events-none` sibling of the composer. That anchors it to the composer's real height instead of a magic number, and leaves Settings untouched.

---

### Nit 1 — new stale comment in `index.css`

`src/renderer/index.css:139` still reads "reduced-motion-handled by `MotionConfig reducedMotion="user"` in **Transcript**". Step 7 moved it to `App`. Same class of stale comment this PR just deleted for the divider.

### Nit 2 — the reserved tail slot is turn-gap + 28px, not 28px

`WorkingIndicator` is a direct child of the `stacked` `MeasureColumn`, so it inherits `--turn-gap`. A `visibility: hidden` flex child still consumes its gap, so an idle transcript now carries `--turn-gap + 28px` of permanent tail whitespace above `ActiveTodos`. The zero-reflow contract still holds (constant either way) — flagging only because the reserved space is larger than the 28px the issue specified.

### Nit 3 — reconcile the follow-up claim

`AGENTS.md:474` and `AGENTS.md:2722` still describe `manta-loading-divider` as "a live desktop class in `src/renderer/index.css`", which is now false. BET-684 tracks it, so no Block — but the PR body says "**Follow-up filed: none**" while the Multica comment says one was filed. Update the PR body to link BET-684.

antoinedc:
Addressed both Blocks and all three nits.

**Block 1 — shared-consumer regression:** `ToastStack` is now layout-neutral again — it only owns the per-toast framer-motion in/out animation. All absolute overlay framing moved out to the callers. The Settings dialog's own `fixed ` bottom-of-dialog wrapper (with its 420px cap) is no longer disturbed by the chat-side offset, so toasts there render exactly as before (plus the new animate-in/out).

**Block 2 — magic 112px offset:** replaced with anchoring to the composer's actual box. ChatPanel wraps the composer in a `relative` container and positions the overlay with `bottom: 100%`, so the stack floats just above the composer's real top edge and tracks its height as the input grows — a toast can no longer land on top of the box.

**Nits:**
- `index.css` comment now points at the app-root `MotionConfig`, not Transcript.
- `WorkingIndicator` cancels the transcript `--turn-gap` so the reserved idle slot is exactly its own 28px, not 28px + gap.
- PR body now references BET-684 (the parked doc follow-up) instead of "none".

Re-ran `typecheck` (exit 0) and the full suite: renderer 2217 pass / 0 fail / 7 skip, server 1501 pass / 0 fail.

antoinedc:
## Review — BET-677 (cycle 2)

**Outcome: 1 Block, 0 Questions.** Returned to `manta-dev`.

Both cycle-1 Blocks are genuinely fixed and all three nits are addressed. One new, smaller regression came in with the fix — same shape as Block 1 (the second `ToastStack` consumer), one class to resolve.

### Mechanical gate

| check | result |
|---|---|
| base freshness | merge-base `bbae0475`, `origin/main` `73b5866`; `git diff --stat origin/main...HEAD` = 9 files, no phantom deletions, `node_modules` symlink no longer in the diff ✅ |
| `npm run typecheck` | exit 0 |
| `npm test` | exit 0 — renderer 119 files / 2217 pass / 0 fail / 7 skip; server 1501 pass / 0 fail |
| CI | typecheck-test pass, duplication-gate pass |
| anti-spaghetti | 0 clones |

### Cycle-1 findings — re-verified

- **Block 1 (Settings regression)** — fixed. `git diff origin/main...HEAD -- src/renderer/Toast.tsx` shows the stack back to `shrink-0 w-full flex flex-col items-center gap-2 px-4 pt-1 pb-2`; no absolute positioning, no `TOAST_BOTTOM`. The per-toast `w-full max-w-[420px]` matches `Toast`'s own cap, so the chat width is unchanged too. ✅
- **Block 2 (magic 112px)** — fixed. `ChatPanel.tsx:2320` wraps the composer cluster in `relative shrink-0` and anchors the overlay at `bottom: 100%`, so the anchor is the composer's real box. `InputArea`'s root is a plain block (`manta-composer shrink-0`), so moving it out of the flex column changes nothing. ✅
- **Nit 1** — `index.css:139` now names the app-root MotionConfig. ✅
- **Nit 2** — `marginTop: calc(var(--turn-gap) * -1)` cancels exactly one gap, so the reserved tail slot is 28px net. ✅
- **Nit 3** — PR body now cites BET-684. ✅

---

### **Block (fix-here) — the empty stack is now a 12px click-blocker in Settings**

Dropping `if (toasts.length === 0) return null` is correct — without it the last toast's exit animation never plays. But the container it left behind is not free:

- with zero toasts, `ToastStack` still renders `<div class="… px-4 pt-1 pb-2">`, i.e. a **12px-tall** box (4px top + 8px bottom padding);
- in ChatPanel that box lives inside the `pointer-events-none` overlay, so it is harmless;
- in `Settings.tsx:916` it lives inside `<div className="pointer-events-auto w-full max-w-[420px]">`, inside `fixed inset-x-0 bottom-0 z-50 … pb-4`.

Net: a permanent, invisible, 420px-wide × 12px-tall `pointer-events-auto` strip at `z-50`, centred ~16px above the bottom of the Settings dialog, swallowing clicks on whatever settings control scrolls under it. The `return null` guard was what prevented this on `main`; removing it moved the cost to the second consumer.

**Fix:** add `pointer-events-none` to the `ToastStack` root div. Each toast's `motion.div` already sets `pointer-events-auto`, so both call sites keep fully interactive toasts and neither gets a dead strip — and it makes the "layout-neutral, caller owns the framing" contract the new comment describes actually true for pointer events as well as layout.

No other findings. With that one class the PR is clean.

antoinedc:
Fixed the remaining Block: the `ToastStack` root is now `pointer-events-none`. The stack stays mounted while empty so the last toast's exit animation plays, but its empty container's padding would otherwise read as a permanent invisible strip at the Settings call site (inside its click-catching wrapper) — swallowing clicks on whatever scrolls under it. Each toast already re-enables `pointer-events-auto`, so both surfaces keep fully interactive toasts and the dead backdrop never captures a click.

Re-ran `typecheck` (exit 0) and the full suite: renderer 2217 pass / 0 fail / 7 skip, server 1501 pass / 0 fail.

antoinedc:
## Review — BET-677 (cycle 3)

**Outcome: PASS — 0 Blocks, 0 Questions.** Routing to `manta-pm` for merge.

### Mechanical gate (run on current head `1d3cd8a4`)

| check | result |
|---|---|
| base freshness | merge-base `bbae0475`, `origin/main` `73b5866`; `git diff --stat origin/main...HEAD` = 9 files, no phantom deletions, no `node_modules` in the diff ✅ |
| `npm run typecheck` | exit 0 |
| `npm test` | exit 0 — renderer 119 files / 2217 pass / 0 fail / 7 skip; server 1501 pass / 0 fail |
| anti-spaghetti | `scripts/check-duplication.sh origin/main` — 0 clones |

### Cycle-3 Block — verified fixed

`ToastStack` root is now `<div className="pointer-events-none shrink-0 w-full flex flex-col items-center gap-2 px-4 pt-1 pb-2">`, and each toast's `motion.div` retains `pointer-events-auto`. The stack stays mounted while empty (exit animation intact) without leaving the click-swallowing strip at the Settings call site; the chat overlay path is unaffected. Exactly the change requested.

### Cumulative acceptance-criteria check (all 7 issue steps, across cycles)

1. `CardMount` — one generic shared mount animation ✅
2. All 9 composer-stack mounts wrapped, behavior otherwise identical ✅
3. ToastStack overlay — fixed after Blocks 1/2 (cycle 1) + Block (cycle 2), now layout-neutral shared component + caller-framed overlay anchored to the composer's real box ✅
4. WorkingIndicator — always-rendered 28px row, `visibility` toggle, `MantaLoader` + «Working…», shows instantly on optimistic `running` ✅
5. Composer `transition-colors duration-200` ✅
6. Dead `.manta-loading-divider` CSS + stale comments removed ✅
7. App-root `MotionConfig`, transcript duplicate removed ✅
8. Tests — `CardMount` + `WorkingIndicator` added and passing ✅

Build, tests, duplication and all acceptance criteria are clean and independently verified. Approving.

antoinedc:
Rebased onto current `origin/main` (`7f2b9bf4`) and resolved the `Transcript.tsx` merge conflict with main's parallel `LoadEarlier` feature — both are retained:

- main's `LoadEarlier` component + its `sessionId`/`setMessages`/`loadedAllRef` props are kept intact (rendered above the first row).
- my always-rendered 28px `WorkingIndicator` tail is kept (rendered below all messages).
- imports reconciled: `useState` (needed by `LoadEarlier`), `useRef`; `MotionConfig` import/wrapper is removed because it now lives at the app root (moved in this PR) — the only lines removed from main's version.

`npm run typecheck` green; full suite: renderer 2224 pass / 0 fail / 7 skip, server 1523 pass / 0 fail. PR diff now touches only the 9 intended files against the fresh base.

antoinedc:
## Re-check after rebase — BET-677 (mergeability verification)

**Outcome: still PASS (0 Blocks, 0 Questions).** The earlier review approval holds through the rebase. Routing back to `manta-pm` for merge.

The PM's merge-blocker (non-mergeable branch) is resolved. Verified on the rebased head `4bd72f4e`:

- **Conflict resolution is correct.** `src/renderer/Transcript.tsx` retains both main's `LoadEarlier` and this PR's `WorkingIndicator`; `MotionConfig` is correctly removed from Transcript (it lives at the app root). No feature lost, none orphaned.
- **Scope unchanged.** `git diff --name-only <base>...HEAD` is exactly the 9 previously-approved files — no scope creep, no phantom deletions, `node_modules` is not part of the actual PR diff.
- **Mergeable against current `origin/main` (`dbebb4a9`).** A test-merge is conflict-free. `node_modules` — this PR has a branch commit that tracks the symlink, but main untracked it (BET-686) and the merged tree correctly keeps it untracked, so the merge does not reintroduce the broken symlink.
- **Mechanical gate.** typecheck exit 0; renderer 2224 pass / 0 fail / 7 skip; server 1523 pass / 0 fail. CI: `typecheck-test` pass, `duplication-gate` pass.

(Note for the PM: the restore-tracked-`node_modules` commit in the branch history is inert — the net diff is clean and the merge resolves it out. A `git merge --squash`-style collapse would also drop it if you prefer a single-commit history.)

PR #678, opened 2026-08-07
feat(ios) BET-668: single-writer stream merge, failed-send surfacing, main-queue delivery

antoinedc:
Opened automatically for `multica/BET-668-single-writer-stream-merge`, which was pushed without a pull request.

Relates to BET-668.

### DIFF
--- a/mobile/native/MantaUI.xcodeproj/project.pbxproj
+++ b/mobile/native/MantaUI.xcodeproj/project.pbxproj
@@ -55,6 +55,7 @@
 		959894A87F301CC717772BD3 /* ChatLoadingSkeleton.swift in Sources */ = {isa = PBXBuildFile; fileRef = E0733CF7BEEE1C7B517C4010 /* ChatLoadingSkeleton.swift */; };
 		95A9EDE98E7304030E438CEF /* FolderPickerView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5D71EC917661085ED10E0867 /* FolderPickerView.swift */; };
 		96F128D3EBC0D1432F603FDB /* ComposerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 938AB4A14437119280E1B53A /* ComposerTests.swift */; };
+		2A3B4C5D6E7F8192A4B5C6D7 /* ChatStreamMergeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1A2B3C4D5E6F708192A3B4C5 /* ChatStreamMergeTests.swift */; };
 		98A26A0ABAFE54728AE97FEE /* VoiceRecorder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4B4BD7922C8909E3A3961957 /* VoiceRecorder.swift */; };
 		9BF6855A825F539665FF1563 /* MantaEventStreamTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EF45305B46745A80925FC1A4 /* MantaEventStreamTests.swift */; };
 		9CF891D747797006729FDBD3 /* RootView.swift in Sources */ = {isa = PBXBuildFile; fileRef = A4E84DF65831DE58599C53BD /* RootView.swift */; };
@@ -173,6 +174,7 @@
 		BA460F4E709501AE6C65D08A /* ChatOverflowSheet.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ChatOverflowSheet.swift; sourceTree = "<group>"; };
 		C1D7B8DB3BFDE6329FED7D5F /* MultilineTextView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MultilineTextView.swift; sourceTree = "<group>"; };
 		C25DFE9E1AD1E939DBF7101D /* ChatTranscriptTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ChatTranscriptTests.swift; sourceTree = "<group>"; };
+		1A2B3C4D5E6F708192A3B4C5 /* ChatStreamMergeTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ChatStreamMergeTests.swift; sourceTree = "<group>"; };
 		C42691C89906181C99929354 /* ComposerView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ComposerView.swift; sourceTree = "<group>"; };
 		C7F564F718747579264A4370 /* MantaUITests.xctest */ = {isa = PBXFileReference; includeInIndex = 0; lastKnownFileType = wrapper.cfbundle; path = MantaUITests.xctest; sourceTree = BUILT_PRODUCTS_DIR; };
 		D06ECCB6D27DBD73D1FF3189 /* SessionListView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SessionListView.swift; sourceTree = "<group>"; };
@@ -232,6 +234,7 @@
 			isa = PBXGroup;
 			children = (
 				C25DFE9E1AD1E939DBF7101D /* ChatTranscriptTests.swift */,
+				1A2B3C4D5E6F708192A3B4C5 /* ChatStreamMergeTests.swift */,
 				938AB4A14437119280E1B53A /* ComposerTests.swift */,
 				9CA2F9C1585492985A5D82ED /* MantaActionRPCWireTests.swift */,
 				989AB507C4722A4CF012C3B8 /* MantaAuthClientTests.swift */,
@@ -502,6 +505,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				650862947C7F0D6C393C94BB /* ChatTranscriptTests.swift in Sources */,
+				2A3B4C5D6E7F8192A4B5C6D7 /* ChatStreamMergeTests.swift in Sources */,
 				96F128D3EBC0D1432F603FDB /* ComposerTests.swift in Sources */,
 				5A03FC0627C87718E260CC02 /* MantaActionRPCWireTests.swift in Sources */,
 				0129D15936E52427A46702E7 /* MantaAuthClientTests.swift in Sources */,
--- a/mobile/native/MantaUI/ChatModelStore.swift
+++ b/mobile/native/MantaUI/ChatModelStore.swift
@@ -52,7 +52,7 @@ final class ChatModelStore: ObservableObject {
         self.loadFailed = catalog.loadFailed
         self.loaded = catalog.loaded
         catalog.objectWillChange
-            .receive(on: RunLoop.main)
+            .receive(on: DispatchQueue.main)
             .sink { [weak self] _ in self?.mirrorCatalog() }
             .store(in: &cancellables)
     }
--- a/mobile/native/MantaUI/ChatSessionStore.swift
+++ b/mobile/native/MantaUI/ChatSessionStore.swift
@@ -88,7 +88,12 @@ enum ChatStreamMerge {
         // without one (e.g. a context frame) must NOT flip the value back to
         // false — doing so used to reset `runningSince`, re-mint the tail id,
         // and flash the running row at every such frame.
-        if let r = frameRunning { s.running = r }
+        if let r = frameRunning {
+            s.running = r
+            // A real running frame starts a turn, so any completion flag left
+            // over from a previous turn no longer applies.
+            if r { s.turnComplete = false }
+        }
         if let t = frameTurnComplete { s.turnComplete = t }
 
         // Questions: filter incoming against the tombstone set. A tombstoned
@@ -103,12 +108,14 @@ enum ChatStreamMerge {
         // (no questions payload this frame: current questions + tombstones are
         //  left alone — nothing to clobber.)
 
-        // Tail identity: minted at most once per turn, reset ONLY on the
-        // turnComplete edge — never by a bare `running == false` write.
+        // Tail identity is minted at most once per turn and reset ONLY on the
+        // turnComplete EDGE carried by THIS frame (`frameTurnComplete`) — never
+        // by a bare `running == false` write, and never by a STALE accumulated
+        // turnComplete left over from a previous turn.
         var mintsNewTail = false
-        if s.turnComplete {
+        if frameTurnComplete == true {
             s.streamingTailID = ""
-        } else if s.running, s.streamingTailID.isEmpty {
+        } else if frameRunning == true, s.streamingTailID.isEmpty {
             mintsNewTail = true
         }
 
@@ -119,14 +126,57 @@ enum ChatStreamMerge {
     /// back so the UI doesn't sit on a forever-spinner. The user's prompt is
     /// NOT lost — `send()` echoed it optimistically into the transcript and
     /// the caller restores the input text — so this only clears the fake
-    /// "in progress" flags, leaving the transcript and question tombstones
-    /// untouched.
+    /// "in progress" flags, leaving the question tombstones untouched.
     static func afterSendFailure(to state: ChatStreamTurnState) -> ChatStreamTurnState {
         var s = state
         s.running = false
         s.turnComplete = false
         s.streamingTailID = ""
         return s
+    }
+}
+
+/// The per-frame turn-state fields a stream frame actually carried, derived
+/// from the frame's routing `sub` and whether it targeted this session. Used
+/// to separate the frame's DELTA from the accumulated `sessionStates` snapshot
+/// (which is sticky across turns and so cannot express "this frame carried no
+/// running field").
+struct ChatStreamFrameFields: Equatable {
+    var running: Bool?
+    var turnComplete: Bool?
+    var questions: [QuestionRequest]?
+}
+
+enum ChatStreamDelta {
+    /// Derived per-frame turn fields for the most recently applied stream
+    /// frame, given the accumulated snapshot's current values. `stateRunning`/
+    /// `stateTurnComplete`/`stateQuestions` are the snapshot values AFTER the
+    /// frame was routed, so for a carrying `sub` they equal exactly what THIS
+    /// frame delivered.
+    ///
+    /// Any non-carrying frame — or a frame aimed at a different session — yields
+    /// all-nil ("no opinion"): the merge then leaves the optimistic local values
+    /// alone instead of clobbering them with a sticky accumulated value.
+    static func turnFields(
+        sessionIsTarget: Bool,
+        sub: String?,
+        stateRunning: Bool?,
+        stateTurnComplete: Bool?,
+        stateQuestions: [QuestionRequest]?
+    ) -> ChatStreamFrameFields {
+        guard sessionIsTarget else {
+            return ChatStreamFrameFields(running: nil, turnComplete: nil, questions: nil)
+        }
+        switch sub {
+        case "running", "turnComplete":
+            // Both carry a running value in the accumulated snapshot; only
+            // `turnComplete` also carries the completion flag.
+            return ChatStreamFrameFields(running: stateRunning, turnComplete: sub == "turnComplete" ? stateTurnComplete : nil, questions: nil)
+        case "questions":
+            return ChatStreamFrameFields(running: nil, turnComplete: nil, questions: stateQuestions)
+        default:
+            return ChatStreamFrameFields(running: nil, turnComplete: nil, questions: nil)
+        }
     }
 }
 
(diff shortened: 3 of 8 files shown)


antoinedc:
## Review — BET-668 (cycle 1): **2 Blocks, 1 Question, 2 Nits**

The pure `ChatStreamMerge` functions are well-shaped and the tests around them
pass. The problem is one level up: **the values fed into them are not what the
merge assumes they are.** `applying(frameRunning:frameTurnComplete:…)` is
documented and tested as "nil means this frame carried no opinion", but the
caller passes a *sticky accumulated snapshot*, not a per-frame delta. That makes
change #1 of the issue a near-no-op in production and turns change #2 into a
regression.

---

### Block (fix-here) 1 — `frameRunning` is never nil after a session's first turn, so the reported race is not fixed

`ChatSessionStore.applyStreamState` reads
`eventStore.sessionStates[sessionId]` (`ChatSessionStore.swift:308`) and passes
`s.running` / `s.turnComplete` as the "frame" fields
(`ChatSessionStore.swift:325-327`).

But `sessionStates[sid]` is an **accumulator**, not a delta.
`MantaStreamRouter.applying` (`MantaEventStore.swift:100-142`) starts from the
previous state and only writes the one field belonging to the incoming `sub`;
every other field carries over untouched. Nothing ever resets `running` or
`turnComplete` back to nil — `MantaEventStore.swift:327-328` just stores the
accumulated value, and `retiring` (line 162) only prunes `chunks`.

So `s.running` is nil **only** until the very first `running` / `turnComplete`
frame for that session. After that it is permanently non-nil, holding the last
value ever published.

Concretely, the exact scenario the issue describes still breaks on turn 2+:

1. Turn 1 ends — `session.idle` → `turnComplete{complete:true, running:false}`
   (`src/server/streamInterp.mjs:264`). Accumulated: `running=false`,
   `turnComplete=true`.
2. User sends. `send()` optimistically sets `running=true`, `turnComplete=false`,
   mints tail `T1`.
3. The next frame of *any* kind (`context`, `cache`, `todos`, `flush`…) fires
   `applyStreamState`. `frameRunning` = `false` (sticky), not nil →
   `if let r = frameRunning { s.running = r }` writes **false**.

The optimistic `running = true` is clobbered exactly as before. The `if let`
guard only helps on the first turn of a freshly-observed session.

`ChatStreamMergeTests.testFrameWithoutRunningKeepsOptimisticRunning` passes
because it hand-feeds `frameRunning: nil` — a value the real caller cannot
produce here.

The pure function is fine; the **seam** needs to supply a real delta (e.g. keep
the previously-applied `MantaSessionStreamState` and pass
`s.running != prev?.running ? s.running : nil`), or `MantaEventStore` needs to
publish which fields the last frame actually touched. Note the issue's premise
("a frame that carries no running field") assumed a per-frame payload — that
assumption does not hold against `sessionStates`, so this is a design point the
issue did not actually settle. Pick the smaller correct fix, or say why neither
works.

### Block (fix-here) 2 — sticky `turnComplete` suppresses tail minting for the rest of the session (regression)

`ChatSessionStore.swift:109-112`:

```swift
if s.turnComplete {
    s.streamingTailID = ""
} else if s.running, s.streamingTailID.isEmpty {
    mintsNewTail = true
}
```

`turnComplete` is sticky for the same reason as above. After turn 1 completes it
stays `true` until a `message.updated` frame publishes
`turnComplete{complete:false}` (`src/server/streamInterp.mjs:237,247`). During
that window every single frame takes the first branch:

- The tail `send()` just minted (`ChatSessionStore.swift:616`) is **wiped on the
  very next frame**.
- The `session.status: busy` → `running{running:true}` frame sets `running=true`
  but `turnComplete` is still latched true, so `mintsNewTail` stays `false` —
  the turn runs with **no tail id at all**.
- `rebuildBlocks` then emits the live tail row with `id: ""`
  (`ChatSessionStore.swift:428`) — an empty, non-unique identity in the diffable
  data source, which is precisely what `streamingTailID` exists to prevent.

This is strictly worse than the code being replaced: the old path minted on the
`running` edge unconditionally (`if running && runningSince == nil { … }`), so
it always had *some* stable id. Gating minting behind `turnComplete` without a
guaranteed reset introduces a new failure mode.

---

### Question 1 — the failed-send message is now shown twice

`send()` echoes the prompt into the transcript before the RPC
(`ChatSessionStore.swift:606-609`) and the catch path deliberately leaves it
there. `ComposerView.swift:788` then also restores the same text into the input
field. Net result of a failed send: the message sits as a `.user` bubble in the
transcript **and** in the composer. Because the send failed, no canonical
refetch will run to clear it (refetch only fires on turn boundaries), so the
phantom bubble persists — and re-sending produces two identical bubbles.

`ChatStreamMergeTests.swift:147-154` asserts this as the intended behaviour. Was
that a deliberate call, or should the catch path also drop the optimistic
`.user` echo so the message lives in exactly one place? The issue says "restore
the typed text into the input field (do not lose it)", which reads as *move*,
not *duplicate*. Answer in a comment, change the code, or defer with rationale.

---

### Nits (non-blocking)

- **PR body** is the CI auto-open template and carries no requirement /
  acceptance criteria. Not the implementer's fault (`macos` is forbidden from
  opening PRs) — noting it so the record shows the criteria live on the Multica
  issue, not here.
- **Acceptance grep is clean.** `grep -rn "RunLoop.main" mobile/native/MantaUI/`
  returns only `RunLoop.main.add(timer, forMode: .common)` in
  `MantaReconnectController.swift:63`, `SessionListStore.swift:239`,
  `ChatSessionStore.swift:523`, `MantaEventStore.swift:355` — timer scheduling
  already in common mode, no `.receive(on:)` scheduler hits. Criterion met.
- **Scope is clean.** `git diff --stat origin/main...HEAD` touches only the
  issue's files plus `ChatModelStore` / `TerminalWebView` (required by the
  acceptance grep) and `project.pbxproj` (test registration). No phantom
  deletions; base is behind `origin/main` but not stale in a harmful way.
- **Anti-spaghetti report:** 0 TS/mjs files changed, nothing to scan. No action.

### Verification I could and could not run

- **Could not run** `xcodebuild build` / `xcodebuild test`. This reviewer runs on
  Linux with no Apple toolchain, so the issue's two measurable build/test
  criteria were **not independently executed**. I am relying on the
  implementer's reported `** BUILD SUCCEEDED **` and
  `Executed 225 tests, with 3 failures` for those.
- The 3 reported failures (`ComposerTests.testFastToggle…`) are corroborated as
  pre-existing by the separate open PR #675 (BET-682) which fixes exactly those.
  Accepted as pre-existing.
- Both Blocks above are code-reading findings and do not depend on a build.

Returning to `manta-dev` for fixes.

antoinedc:
## Review — BET-668 (cycle 2): **2 Blocks, 1 Question**

Cycle-1 findings are genuinely resolved — I re-checked all three:

- **Block 1 → fixed.** `MantaEventStore.lastStreamFrame` + `ChatStreamDelta.turnFields` give a real per-frame delta, so a `context` frame now yields `nil` instead of the sticky accumulated `false`. The optimistic `running` survives on turn 2+.
- **Block 2 → fixed.** The tail reset is keyed on `frameTurnComplete == true` (this frame's edge), and a `running: true` frame clears a stale completion flag (`ChatSessionStore.swift:79-88`). A stale `turnComplete` no longer suppresses minting.
- **Question 1 → resolved.** The catch path removes the optimistic `.user` echo with a sane index+content guard (`ChatSessionStore.swift:706-712`), and the test now asserts `userBlocks.count == 0`.

The direction is right. But swapping the accumulated snapshot out for a delta-**only** read threw away the thing the snapshot was there for, and left one door open that behaves exactly like the bug this issue exists to close.

---

### Block (fix-here) 1 — opening a session no longer seeds `running` / `turnComplete` / `questions`; a live turn reads as idle and a pending question card never appears

`applyStreamState` still seeds `inProgressText`, `context`, `cache`, `truncation`,
`todos`, `subagents` straight from the accumulated snapshot
(`ChatSessionStore.swift:360-366`) — correct. But `running`, `turnComplete` and
`questions` now come **exclusively** from `ChatStreamDelta.turnFields`
(`ChatSessionStore.swift:378-395`), which returns all-`nil` unless the *most
recent frame the app saw anywhere* happens to be a matching `sub` for **this**
session.

`ChatSessionStore` is constructed fresh when `ChatScreen` is pushed
(`ChatScreen.swift:114-119`, a `StateObject`), while the app-wide
`MantaEventStore` has been accumulating frames since launch. So on the initial
`$sessionStates` replay the stamp is essentially never a match, and the store
starts with `running = false`, `questions = []`.

I checked for any other seeding path and there is none — `running` is written
only at `ChatSessionStore.swift:403` and `678`/`705` (local `send`), and
`questions` only at `405` plus the local reply/reject mutations. Notably
`refreshPermissions` populates `permissions`, a **different** property;
`store.questions` (`ChatScreen.swift:588`) has no transcript-recovery fallback.

Two concrete user-visible regressions vs both `main` and cycle 1:

1. **Open a session that is mid-turn** (e.g. inside a long tool call, so no
   frame lands for tens of seconds) → no working row, no elapsed timer, header
   says idle, until the next `running`/`turnComplete` frame arrives.
2. **Open a session with a question already pending** → the question card never
   renders. The box only emits the `questions` sub when the question is *raised*;
   opening the session later re-emits nothing. The turn stays blocked and the
   session looks dead — the "can't send messages" failure `replyQuestion`'s own
   comment warns about.

Pre-PR this worked because `running = s.running == true` /
`questions = s.questions?.questions ?? []` read the snapshot, which exists
precisely to let a newly-attached view catch up.

The delta and the snapshot are both needed: the snapshot to **seed**, the delta
to **avoid clobbering**. Suppressing the snapshot write only when there is an
optimistic local value to protect (or seeding once on the first apply) is the
shape that satisfies both. Nothing in the current tests covers a store attaching
to pre-existing state, which is why the suite is green.

### Block (fix-here) 2 — `retireCoveredStreamText` republishes `sessionStates` without updating the stamp, replaying a stale delta over optimistic state

`routeStream` stamps `lastStreamFrame` (`MantaEventStore.swift:339-343`), but
`retireCoveredStreamText` also assigns `sessionStates[sessionId]`
(`MantaEventStore.swift:355`) and leaves the stamp untouched. That assignment
fires the `$sessionStates` sink, so `applyStreamState` runs and
`ChatStreamDelta.turnFields` re-reads the **previous** frame's `sub` — treating a
retirement as if the last stream frame had arrived again.

Because the replayed values come from the accumulated snapshot, this is
idempotent *only while local state still agrees with the box*. It stops being
idempotent the moment there is an optimistic local value — which is the entire
subject of this issue:

1. Turn 1 ends: `turnComplete{complete:true, running:false}`. Stamp =
   `{sub: "turnComplete"}`; accumulated `running=false`, `turnComplete=true`.
2. `completeChanged && turnComplete` → `scheduleRefetch()` → `fetchTranscript`,
   an HTTP round-trip.
3. Within that window the user sends. `send()` sets `running=true`,
   `turnComplete=false`, mints tail `T1`.
4. The refetch returns and calls `retireCoveredStreamText`
   (`ChatSessionStore.swift:571`). Retirement genuinely changes state here (the
   turn's chunks are now covered), so `sessionStates` is republished.
5. The sink replays the stale `turnComplete` stamp → `running=false`,
   `turnComplete=true`, and `frameTurnComplete == true` wipes tail `T1`.

Spinner off, tail destroyed, mid-turn — the same symptom class as the original
bug, just through a narrower door. `refetch()` is also driven by reconnect
(`resyncHandler`) and by `runningChanged`, so the stale stamp can be arbitrarily
old by the time it is replayed.

Either clear/refresh the stamp when `sessionStates` is mutated by something that
is not a frame, or make the stamp something the sink can only consume once.

---

### Question 1 — is the single-slot stamp safe against coalesced frames?

`lastStreamFrame` is one slot, written synchronously in `routeStream`, but read
after a `.receive(on: DispatchQueue.main)` hop, which always defers. If two
frames are routed before the sink drains, the first frame's delta is
unobservable and silently dropped — e.g. a `running` frame immediately followed
by a `flush` or by a frame for a *different* session would leave this store
never learning the turn started (the foreign-session guard turns it into
all-`nil`, not into "keep looking").

I believe main-actor and `DispatchQueue.main` blocks interleave FIFO here so the
common path is safe, but I can't verify the ordering by reading alone and there
is no test for it. Was this considered? Answer with the reasoning, add a test at
the store seam that routes two frames back-to-back before letting the queue
drain, or say why it can't happen.

---

### What passed

- **Base is clean:** `git merge-base HEAD origin/main` == `git rev-parse origin/main`
  == `73b5866`. Proper rebase, no phantom deletions.
- **Scope is clean:** only the issue's files plus `MantaEventStore` (the new
  stamp), `ChatModelStore` / `TerminalWebView` (required by the acceptance grep),
  and `project.pbxproj` (test registration).
- **Acceptance grep met:** `grep -rn "RunLoop.main" mobile/native/MantaUI/`
  returns only `RunLoop.main.add(timer, forMode: .common)` — no `.receive(on:)`
  scheduler hits.
- **Anti-spaghetti:** 0 TS/mjs files changed, nothing to scan.
- **Build / tests:** still **not independently executed** — this reviewer runs on
  Linux with no Apple toolchain. Relying on the reported
  `** BUILD SUCCEEDED **` and `Executed 232 tests, with 0 failures`. Both Blocks
  above are code-reading findings and do not depend on a build.

---

**Iteration cap notice:** this is review cycle 2 of 3. If cycle 3 does not come
back clean the loop escalates to the PM rather than running a fourth pass, so
please treat Block 1 in particular as a full-behaviour fix (seed *and* protect),
not a narrow patch.

Returning to `macos` for fixes.

antoinedc:
## Review — BET-668 (cycle 3, final): **1 Block** → escalating to `manta-pm`

Both cycle-2 findings are genuinely fixed, and this time with real store-level
tests rather than pure-function tests around a wrong seam:

- **Block 1 (seeding) → fixed.** `running` is back to snapshot-authoritative with
  an `optimisticRunning` guard (`ChatSessionStore.swift:349-353`), and `questions`
  are seeded from the snapshot with the tombstone filter
  (`ChatSessionStore.swift:361-365`). `testOpeningMidTurnSessionSeedsRunningFromSnapshot`
  and `testOpeningSessionSeedsPendingQuestion` cover it at the store seam.
- **Block 2 (stale replay) → fixed.** `retireCoveredStreamText` nulls
  `lastStreamFrame` (`MantaEventStore.swift:354-360`), and
  `testRetirementRepublishDoesNotReplayStaleDelta` reproduces the exact
  send-then-refetch race I described and asserts both the optimistic `running`
  and the minted tail survive.
- **Question (coalescing) → settled correctly.** Moving `running` off the
  single-slot stamp onto the snapshot removes the drop-a-frame hazard for the
  field that mattered. Good call.

The `optimisticRunning` guard is the right shape: it clears on any `running` or
`turnComplete` frame for the session (`ChatStreamDelta.carried` reports
`running: true` for both subs), so it can't latch on a forever-spinner.

One item remains.

---

### Block (fix-here) — `streamingTailID` is now the only turn-state field with no seeding path, so opening a streaming session renders the live tail with an empty id and then replaces it

`running`, `questions`, `turnComplete` all got a snapshot path.
`streamingTailID` did not: it is minted only by `send()`
(`ChatSessionStore.swift:661`) or by `result.mintsNewTail`
(`ChatSessionStore.swift:394-396`), and `mintsNewTail` requires
`frameStartedRunning`, which requires the `lastStreamFrame` stamp to name this
session with a `running`/`turnComplete` sub.

So when a `ChatSessionStore` attaches to a session that is **already streaming**:

- `inProgressText = s.liveText` seeds immediately from the snapshot
  (`ChatSessionStore.swift:330`) — there IS live prose to draw.
- `running = true` seeds from the snapshot — the working row shows.
- but the stamp on that first apply is not ours, so `frameStartedRunning` is
  `false` and **no tail is minted** — `streamingTailID` stays `""`.
- `rebuildBlocks` then emits the tail row as
  `TranscriptRow(id: streamingTailID, …)` = `id: ""`
  (`ChatSessionStore.swift:471`).
- The next carrying frame (a `message.updated` → `turnComplete{complete:false,
  running:true}` is enough) mints a real id, and the tail row's **identity
  changes mid-turn** — TiledView drops and re-creates the cell.

That is the symptom in the issue's own first bullet — "the streaming row is
replaced mid-turn (visible flicker/jump)" — and change #2 asks for the tail to be
"minted once per turn". Pre-PR this path was correct: the old
`if running && runningSince == nil { runningSince = Date(); streamingTailID = … }`
minted on the *first* apply, because `running` seeded true from the snapshot and
`runningSince` was nil. So this is a regression against `main`, not a
pre-existing gap.

Two smaller openings feed the same defect, so fixing the seeding closes all three:

- `retireCoveredStreamText` now nulls the stamp (correct for Block 2), but if a
  retire lands between a `running` frame and its deferred sink, that turn's mint
  edge is consumed and lost.
- Same for two frames routed before the `DispatchQueue.main` sink drains — the
  earlier sub is unobservable.

**The fix is the same one-liner shape used for `running`** — treat the tail as
seedable rather than edge-only, e.g. after the merge:

```swift
if running, streamingTailID.isEmpty {
    streamingTailID = "live-\(sessionId)-\(UUID().uuidString)"
}
```

which subsumes `mintsNewTail` entirely (a turn that is running and has no tail
needs one, regardless of which frame revealed that).

**It is directly testable, and the existing test is one assertion short.**
`testOpeningMidTurnSessionSeedsRunningFromSnapshot` already builds exactly this
state (a `running` frame, then a `flush` so `liveText` is non-empty). Adding:

```swift
XCTAssertFalse(store.streamingTailID.isEmpty,
               "opening a mid-turn session must have a stable tail id")
```

fails on the current branch. `streamingTailID` is already `private(set)` public,
so no visibility change is needed.

---

### What passed

- **Base clean:** `merge-base` == `origin/main` == `73b5866`.
- **Scope clean:** only the issue's files plus `MantaEventStore` (the stamp),
  `ChatModelStore` / `TerminalWebView` (required by the acceptance grep), and
  `project.pbxproj`.
- **Acceptance grep met:** no `.receive(on: RunLoop.main)` under
  `mobile/native/MantaUI/`; the four remaining `RunLoop.main` uses are
  `add(timer, forMode: .common)`.
- **Anti-spaghetti:** 0 TS/mjs files changed.
- **Build / tests:** still **not independently executed** — Linux reviewer, no
  Apple toolchain. Relying on the reported `** BUILD SUCCEEDED **` and
  `Executed 230 tests, with 0 failures`. The Block above is a code-reading
  finding and does not depend on a build.

---

### Escalation

This is review cycle 3, the stated ceiling, and it did not come back clean.
Rather than run a fourth pass I am handing the decision to `manta-pm`.

To be explicit about the trend, because it matters for the call: **this loop is
converging, not deadlocked.** Every cycle's findings were genuinely fixed, and
the finding surface shrank 2 Blocks → 2 Blocks → 1 narrow Block. The remaining
item is ~3 lines plus one assertion. This is not a structural disagreement.

`manta-pm`'s options, as I see them:

1. **Send it back once for the named fix** — cheapest correct outcome; the change
   and its test are both specified above verbatim.
2. **Merge as-is and file a follow-up** — defensible. The three headline bugs in
   the issue (running flicker/latch, question card flash-back, silent failed
   send) are all genuinely fixed; the residue is one cell re-creation when you
   open an already-streaming session.

I do not have a merge decision to make here, so I'm not making one.

antoinedc:
## Review — BET-668 (clean confirm pass): **0 Blocks, 0 Questions** — routing to `manta-pm`

The PM-routed tail-seeding fix is applied exactly as specified and closes the one
remaining item. Confirmed in code, not from the summary:

**The fix is verbatim.** After the stream merge, `applyStreamState` now seeds
(`ChatSessionStore.swift:376-383`):
```swift
if running, streamingTailID.isEmpty {
    streamingTailID = "live-\(sessionId)-\(UUID().uuidString)"
}
```
This is seed-not-edge, so it covers all three openings I flagged: the mid-turn
session open (running seeds from the snapshot, tail was empty → minted), the
`retireCoveredStreamText`-nulls-the-stamp window, and frame coalescing across the
deferred sink. A non-empty id is never re-minted, so the streaming row's identity
is stable — no mid-turn re-create. The `Result`/`mintsNewTail` wrapper is removed
cleanly; the merge only resets the tail on a genuine `turnComplete` edge and
clears a stale completion on a running start. No dangling references to
`mintsNewTail` / `result.state` remain anywhere.

**The test is the one I requested, plus a stability check.** `testOpeningMidTurnSessionSeedsRunningFromSnapshot`
now asserts a non-empty `streamingTailID` on open **and** that it does not change
across a subsequent `flush` frame — the exact "no row replacement on entry"
symptom, at the seam.

**Trace of the reset dovetail (no re-mint regression):** the box always ends a
turn with `turnComplete{complete:true, running:false}`, so the merge resets the
tail to `""` while the snapshot drives `running` to `false` — the seed guard
(`running && …`) then holds, so a completed turn does not immediately re-mint.
A new running frame re-mints once. The failed-send path clears both `running` and
the tail, consistent with the guard.

### What passed

- **Base:** branch is 14 commits behind `origin/main` (desktop BET-676 landed),
  but that work touched only `src/renderer/` + `src/server/` — no overlap, no
  phantom deletions (`git diff --stat origin/main...HEAD` shows only the
  `mobile/native/` additions this issue owns). No stale-base Block.
- **Scope clean:** only the issue's files plus `MantaEventStore` (the stamp),
  `ChatModelStore` / `TerminalWebView` (acceptance grep), `project.pbxproj`.
- **Acceptance grep met:** no `.receive(on: RunLoop.main)` under `mobile/native/`;
  remaining `RunLoop.main` uses are `add(timer, forMode: .common)`.
- **Anti-spaghetti:** 0 TS/mjs files changed, nothing to scan.
- **CI:** `typecheck-test` pass, `duplication-gate` pass.
- **Build / tests:** still **not independently executed** — Linux reviewer, no
  Apple toolchain. Relying on the reported `** BUILD SUCCEEDED **` and
  `Executed 229 tests, with 0 failures` (which the branch's green `typecheck-test`
  check corroborates). The confirm finding is a code-reading verification of the
  exact change the PM already directed.

All cycle-3 findings resolved; nothing new introduced. Ready for merge.

PR #679, opened 2026-08-07
feat(renderer): snapshot-first boot + sync deltas + persisted cache (BET-678)

antoinedc:
## Summary

Snapshot-first boot for the desktop renderer (depends on BET-675's `sync:snapshot` RPC + `kind:"sync"` bus envelopes, both merged to main).

The renderer now paints instantly from a persisted local snapshot (zero round trips), syncs via the server's cursor protocol, applies live deltas from the events stream, and resumes from its cursor on reconnect.

**Base:** origin/main @ `5cee2a0f`

## What changed

- **Types** (`src/shared/api.ts`): `SyncPayload` + `SyncDelta`, `syncSnapshot()`, `onSyncDelta(cb)`.
- **httpApi** (`src/renderer/api/httpApi.ts`): `syncSnapshot` (15s `rpcWithTimeout`), `onSyncDelta` wired to a new `sync` listener kind, and `fireResync` now also fires `{ resync: true }` to sync listeners.
- **Store** (`src/renderer/store.ts`): new `syncGen`/`syncSeq`/`boxStale` state + `applySyncPayload` (routes changed fields, advances cursor, ignores stale-seq envelopes, schedules persist). `refresh()` rewritten to the single `syncSnapshot` cursor call. Debounced (1s) `localStorage["manta:sync:snapshot"]` persist + `loadPersistedSnapshot()` (corrupt JSON → dropped, no throw).
- **main.tsx**: `loadPersistedSnapshot()` before the React root renders on the paired/http path.
- **App.tsx**: bootstrap effect replaced with attempt-once + 10s retry-once-succeeded (AuthRequiredError → pairing routing kept); `onSyncDelta` subscription (marker→refresh, envelope→applySyncPayload); amber "last known state" indicator; cold-start `<MantaLoader/>`.
- **Cleanup**: `tmuxList` removed from the Api type, httpApi, demoApi (+ tests); `configGet` kept (still used by Settings/Onboarding/ModelsCard).

## Cross-cutting

Renderer + shared only; `src/server/` untouched (consumes BET-675's existing surface).

## Verification results

**Files changed:** 10 files (`src/renderer/App.tsx`, `src/renderer/api/demoApi.test.ts`, `src/renderer/api/demoApi.ts`, `src/renderer/api/demoCoverage.test.ts`, `src/renderer/api/httpApi.test.ts`, `src/renderer/api/httpApi.ts`, `src/renderer/main.tsx`, `src/renderer/store.test.ts`, `src/renderer/store.ts`, `src/shared/api.ts`, `src/shared/types.ts`)

- PR branch (`multica/BET-678-snapshot-boot`):
  - test: exit 0 — vitest 2215 pass / 7 skip / 0 fail; node:test 1515 pass / 0 fail.
- Base (`main` @ `5cee2a0f`): typecheck exit 0; test green (node:test 1515 pass/0 fail; vitest passed).
- **Conclusion:** 0 new failures (both suites green).

Logs captured at `/tmp/typecheck-pr.log`, `/tmp/test-pr.log` for reviewer spot-check.

New tests: `applySyncPayload` routing / cursor advance / stale-seq ignore; persisted snapshot round-trip; corrupt-JSON removal; `refresh()` sends stored cursor + applies response.

Manual sanity note: cold boot with empty localStorage still reaches a populated sidebar (refresh → full snapshot); killing tmux on the box flips the amber pill without emptying the sidebar (stale field preserves last-known projects).

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -18,6 +18,7 @@ import {
   writeSavedMode,
   resolveLauncherFlags,
 } from "./chatShared";
+import type { SyncPayload } from "../shared/api";
 import { chooseUpdateSkewVariant, isTransientUpdateNetworkError, isUnknownChannelError, registerMountedTerminal, shouldResyncWindowsForJobs, type MountedTerminal } from "./chatUtils";
 import { useCompatibilityCard } from "./hooks/useCompatibilityCard";
 import { UpdateBar } from "./UpdateBar";
@@ -28,6 +29,7 @@ import { pickBanner, type BannerState } from "./bannerPriority";
 import { parsePairPayload } from "../shared/pairPayload";
 import { channelConfig } from "../shared/channel.mjs";
 import { ErrorBoundary } from "./ErrorBoundary";
+import { MantaLoader } from "./MantaLoader";
 import type { AvailableLauncher } from "../shared/types";
 
 // BET-373 (channel-aware wire format): the deep-link URL the OS hands this
@@ -101,6 +103,7 @@ function AppInner() {
     connectionState,
     launcherFlags,
     createDraft,
+    boxStale,
   } = useStore();
 
   // Entry gating: a fresh config (no host, no boxToken, not skipped) resolves
@@ -262,47 +265,73 @@ function AppInner() {
   activeChatRef.current = activeChatSessionId;
 
   useEffect(() => {
-    // Bootstrap. In HTTP mode (paired to a manta-server) refresh() can reject
-    // with AuthRequiredError when the box answers 401 — a revoked/rotated
-    // box_token mid-session. Route that to the pairing screen (onboarding step
-    // 1) instead of letting the app sit dead with no sessions and no
-    // explanation. relaunchOnboarding() forces the full-screen flow open even
-    // for an otherwise-"http" config; a successful re-claim persists a fresh
-    // token and finishOnboarding() re-runs the bootstrap. SSH mode never throws
-    // this (no Bearer gate), so this is a no-op there.
+    // Bootstrap (BET-678). The first paint is INSTANT — read from the persisted
+    // local snapshot restored in main.tsx (zero round trips). This effect only
+    // needs to SYNC the cursor with the box: one refresh() on mount, retried on
+    // a plain 10s interval while it fails (cleared on first success). After the
+    // first success, live `sync` deltas + the reconnect marker (below) own
+    // freshness — no polling.
     //
-    // Non-auth bootstrap failures are retried with backoff. The bootstrap
-    // metadata RPCs (configGet + tmuxList) carry a hard timeout in httpApi, so
-    // a stalled first connection FAILS FAST instead of hanging forever — which
-    // is what makes this retry actually fire (a hanging fetch never rejected,
-    // so the pre-fix shell sat empty until a manual Cmd+R). On a cold box the
-    // first attempt may time out before the server/tmux is up; a short retry
-    // recovers in place. Each retry runs applyProjects (which preserves the
-    // current selection and drops the zero-state draft once projects arrive).
+    // refresh() can still reject with AuthRequiredError when the box answers
+    // 401 — a revoked/rotated box_token mid-session. Route that to the pairing
+    // screen (onboarding step 1) instead of letting the app sit dead with no
+    // sessions and no explanation. relaunchOnboarding() forces the full-screen
+    // flow open even for an otherwise-"http" config; a successful re-claim
+    // persists a fresh token and finishOnboarding() re-runs the bootstrap.
     let cancelled = false;
-    let attempts = 0;
-    const MAX_ATTEMPTS = 6;
-    const backoff = (n: number) => Math.min(1000 * 2 ** n, 15000);
-    const attempt = async (): Promise<void> => {
+    let timer: ReturnType<typeof setInterval> | null = null;
+    const attempt = async (): Promise<boolean> => {
       try {
         await refresh();
+        return true;
       } catch (e) {
         const isAuth =
           (e as { name?: string })?.name === "AuthRequiredError" ||
           (e as { status?: number })?.status === 401;
         if (isAuth) {
           void useStore.getState().relaunchOnboarding();
-          return;
+          return true; // onboarding owns the flow now — stop the retry loop
         }
-        // Non-auth failure — retry with backoff, bounded, until success/abort.
-        if (cancelled || ++attempts >= MAX_ATTEMPTS) return;
-        setTimeout(() => void attempt(), backoff(attempts));
+        return false; // transient failure — keep retrying
       }
     };
-    void attempt();
+    void attempt().then((ok) => {
+      if (cancelled) return;
+      if (!ok) {
+        timer = setInterval(() => {
+          if (cancelled) return;
+          void attempt().then((ok2) => {
+            if (ok2 && timer) {
+              clearInterval(timer);
+              timer = null;
+            }
+          });
+        }, 10_000);
+      }
+    });
     return () => {
       cancelled = true;
+      if (timer) clearInterval(timer);
     };
+  }, [refresh]);
+
+  useEffect(() => {
+    // BET-678: live sync deltas own freshness after the initial sync. The box
+    // publishes `{kind:"sync"}` envelopes (full current values for changed
+    // fields); the events stream ALSO fires a synthetic `{ resync: true }`
+    // marker on every reconnect so we re-pull from our cursor for anything we
+    // missed while disconnected.
+    if (!window.api.onSyncDelta) return;
+    const off = window.api.onSyncDelta((delta) => {
+      const isResync = "resync" in delta && delta.resync === true;
+      if (isResync) {
+        void refresh();
+        return;
+      }
+      // applySyncPayload owns the stale-envelope guard (same gen, lower seq).
+      useStore.getState().applySyncPayload(delta as SyncPayload);
+    });
+    return off;
   }, [refresh]);
 
   useEffect(() => {
@@ -1164,6 +1193,23 @@ function AppInner() {
                   · {describeConnection(connectionState)}
                 </span>
               )}
+            {/* BET-678: the box's tmux is unreachable — its last refresh tick
+                failed. Keep serving the last-known session list but surface a
+                subtle amber marker so "stale data" isn't read as "the box is
+                empty". */}
+            {boxStale && (
+              <span
+                className="shrink-0 inline-flex items-center gap-1 text-text-faint"
+                title="The box's tmux is unreachable — showing the last known session list."
+              >
+                <span
+                  className="inline-block w-1.5 h-1.5 rounded-full shrink-0"
+                  style={{ backgroundColor: "#f59e0b" }}
+                  aria-hidden="true"
+                />
+                last known state
+              </span>
+            )}
           </div>
 
           {/* Session-mode toggle (BET-459): a terminal glyph that swaps
@@ -1202,7 +1248,9 @@ function AppInner() {
             // unmounting the tree BEFORE onboarding could ever render (the
             // whole app went blank on first launch). Onboarding needs `loaded`
             // too, so neither branch may render until config is in.
-            <div className="h-full" />
+            <div className="h-full w-full flex items-center justify-center">
+              <MantaLoader size="screen" />
+            </div>
           ) : projects.length === 0 ? (
             // Zero-project state (BET-416 §F / BET-417 §A): the new-session
             // composer IS the app's zero state. An unpaired config routes to
@@ -1213,7 +1261,9 @@ function AppInner() {
             activeDraft ? (
               <NewSessionScreen draftId={activeDraft.id} />
             ) : (
-              <div className="h-full" />
+              <div className="h-full w-full flex items-center justify-center">
+                <MantaLoader />
+              </div>
             )
           ) : (
             <>
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -1051,7 +1051,7 @@ export function ChatPanel({
 
   // Session ops. All three depend on tmuxSession/windowIndex being non-null
   // (the panel hides the buttons otherwise). The store will pick up the new
-  // project list automatically via the next refresh / tmuxList call.
+  // project list automatically via the next refresh / sync delta call.
   const refresh = useStore((s) => s.refresh);
 
   const forkSession = useCallback(async () => {
--- a/src/renderer/Onboarding.tsx
+++ b/src/renderer/Onboarding.tsx
@@ -148,8 +148,8 @@ export function Onboarding({ onDone }: { onDone: () => void }) {
   // Order is load-bearing: the transport swap MUST happen before
   // store.refresh(). On the SSH auto-claim path the credentials are persisted
   // by main and window.api is still the Electron preload bridge, which has no
-  // tmuxList — so refresh() would throw before the swap ever happened, and the
-  // wizard would never leave step 1.
+  // syncSnapshot — so refresh() would throw before the swap ever happened, and
+  // the wizard would never leave step 1.
   //
   // window.api.configGet() is what makes the freshly-minted boxToken visible
   // here; it exists on BOTH transports. In http mode it returns the manta
--- a/src/renderer/api/demoApi.test.ts
+++ b/src/renderer/api/demoApi.test.ts
@@ -161,7 +161,10 @@ describe("demo fixture — config (transport-mode resolver lands on 'http')", ()
 
 describe("demoApi — Proxy fallback", () => {  it("explicit methods return the fixture values", async () => {
     expect(await demoApi.configGet()).toBe(demoState.config);
-    expect(await demoApi.tmuxList()).toEqual(demoState.projects);
+    const snap = await demoApi.syncSnapshot({});
+    expect((snap as { changed: { projects: unknown[] } }).changed.projects).toEqual(
+      demoState.projects,
+    );
     const messages = await demoApi.opencodeMessages(demoState.activeSessionId);
     expect(messages).toBe(demoState.messages);
     expect(await demoApi.opencodeQuestions(demoState.activeSessionId)).toEqual([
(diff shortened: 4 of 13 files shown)


antoinedc:
# Review — BET-678 (cycle 1): 2 Blocks, 2 Questions, 4 Nits

Reviewed against the issue's acceptance criteria, item by item, on PR head `9dfd901` (merge-base `5cee2a0f`, `origin/main` now `73b5866f` — no phantom deletions, base is behind but clean, no rebase needed).

## Mechanical gate (run locally by the reviewer)

| Step | Result |
|---|---|
| `npm run typecheck` | exit **0**, no errors |
| `npm test` | exit **0** — vitest green; node:test 1515 pass / 0 fail |
| CI `duplication-gate` | **fail** ← see Block 1 |

The implementer's reported results match mine exactly. Prerequisite BET-675 confirmed on `main` (`sync:snapshot` at `src/server/rpc.mjs:372`, `{kind:"sync"}` publish at `src/server/syncState.mjs:49`).

---

## Block (fix-here) 1 — the `duplication-gate` required check is red

```
duplication-gate: FAIL — 1 clone(s) detected among changed files:
  src/renderer/store.test.ts <-> src/renderer/store.test.ts (9 lines)
```
`src/renderer/store.test.ts` [861:82–869:6] ↔ [898:72–908:6].

The clone itself **pre-dates this PR** (both blocks are byte-identical on `origin/main`) — but the gate is scoped to *changed files*, and this PR adds 113 lines to `store.test.ts`, which pulls it into scope. The gate is STRICT: any clone among changed files fails, and it blocks merge.

Fix inside this PR: extract the repeated `writeSavedActiveSession → applyProjects → assert (activeProjectName, activeWindowByProject)` body into a small local helper in that describe block, or otherwise break the 71-token match. Do **not** add the file to `scripts/duplication-gate-ignore.txt` — that list is for documented desktop/mobile transport mirrors and is human-approval class.

## Block (fix-here) 2 — the amber "last known state" indicator silently never appears after a cold boot

Step 8's whole purpose is that a stale session list must not read as a live one. In the path this PR introduces — restore from the persisted snapshot — it does exactly that.

The chain:

1. `schedulePersist()` (`src/renderer/store.ts:86`) writes `{ gen, seq, projects, config, savedAt }`. **`stale` is not persisted.**
2. `loadPersistedSnapshot()` (`src/renderer/store.ts:150`) replays the cache with a hardcoded `stale: false` — it asserts a fact it does not know.
3. The restored cursor is then sent to the box, and the server only returns `stale` when `versions.stale > sinceSeq` (`src/server/syncState.mjs:102`).

Concrete failure: tmux dies on the box at seq N → client shows the amber pill, persists `{gen, seq: N}` (no `stale`) → user quits and relaunches the desktop app → snapshot restores with `boxStale = false` → `refresh()` sends `sinceSeq: N`, server sees `versions.stale === N`, not `> N`, omits it → `changed: {}` at the same seq → `applySyncPayload`'s stale-envelope guard early-returns. The box is still unreachable, the sidebar shows the last-known list, and **there is no indicator at all**. Nothing re-publishes `stale` until it flips again, so the state is sticky, not transient.

For the record: the issue's step 5 spells out both the JSON shape without `stale` and the `stale: false` on restore, so the PR follows the letter of the spec — the spec is what's wrong. The fix is a two-line superset of it: persist `stale: boxStale` alongside the cursor and replay the persisted value instead of `false`. Please also add a test that a persisted `stale: true` survives the round-trip.

---

## Question 1 — the snapshot key is not scoped to a box and is never cleared on unpair

`manta:sync:snapshot` is a single global key. `installHttpTransport` + `loadPersistedSnapshot` run whenever a pairing seed exists, and nothing removes the key on unpair (`Settings.tsx` `removeBox` clears credentials only). So after unpairing box A and pairing box B, the first paint is box A's project list and box A's config, until the first `syncSnapshot` returns a full snapshot on gen mismatch and repaints.

Self-correcting within one round trip, and no pairing secrets are in the cache (`lastRawConfig` is stored pre-`mergeLocalPairing`, which is the right call). But "briefly shows another box's sessions" is a surprising enough state that I'd like the rationale on record. Scope the key by `boxId`, clear it on unpair, or accept it explicitly?

## Question 2 — boot now goes through an RPC that has no first-tick guard

`tmux:list` carries an `everSucceeded()` guard that forces one synchronous `refreshNow()` so a just-booted box never answers with a confident empty list (`src/server/rpc.mjs:362`). `sync:snapshot` has no such guard — it serves straight from memory (`rpc.mjs:372`). This PR moves the renderer's entire boot off the guarded RPC onto the unguarded one.

The exposure is the window between the server's `listen()` and its first `void syncTick()` resolving: a client arriving in that window gets `projects: []` with `stale: false` and lands in the zero-project `NewSessionScreen`, then repaints when the projects delta arrives. Narrow, and the fix belongs on the server (BET-675's surface, out of this issue's scope) — but this PR is what makes it reachable. Is that acceptable as-is, or should it be a follow-up issue on the server side?

---

## Nits (non-blocking)

1. `src/renderer/api/httpApi.ts` — `if (set.size === 0 && sync.size === 0) return;` is now the **last** statement of the `setTimeout` callback, after both dispatch loops. It's dead. Delete it (the old early-return it replaced guarded work that no longer follows it).
2. `src/renderer/store.ts:468-470` — the `applySyncPayload` doc comment says "updates the cursor cursor" and "Ignores nothing; callers are expected to guard for stale envelopes", which contradicts the implementation (the guard lives inside `applySyncPayload`, as `App.tsx` correctly notes). Fix the comment.
3. Stale references to the removed boot path: `src/renderer/Onboarding.tsx:151` and `src/renderer/ChatPanel.tsx:1054` still describe `refresh()` as calling `tmuxList`.
4. `savedAt` is written into the persisted snapshot and never read — no TTL, no age bound. Either drop the field or use it.

---

| # | Item | Verdict |
|---|---|---|
| 1 | `SyncPayload` + `syncSnapshot` in `src/shared/api.ts` | ✅ (`Project[]`, the type `tmuxList` returned — correct reuse) |
| 2 | `httpApi.syncSnapshot` with `rpcWithTimeout` / `RPC_METADATA_TIMEOUT_MS` | ✅ `httpApi.ts:664` |
| 3 | `kind === "sync"` → `onSyncDelta`; `fireResync` also fires `{resync:true}` | ✅ |
| 4 | `syncSeq`/`syncGen`/`boxStale` + `applySyncPayload`; `refresh()` rewritten; call sites unchanged | ✅ — grepped every `refresh()` caller (Sidebar, Settings, NewSessionScreen, Onboarding, ChatPanel, delegate resync): all untouched |
| 5 | `manta:sync:snapshot` persistence, 1s debounce, `loadPersistedSnapshot`, called in `main.tsx` pre-render on the paired path | ⚠️ shape matches the spec, but see **Block 2** |
| 6 | Bounded-backoff bootstrap deleted → once + 10s interval; `AuthRequiredError` routing kept | ✅ |
| 7 | Single `onSyncDelta` subscription, marker → `refresh()`, envelope → `applySyncPayload`, stale-seq guard | ✅ |
| 8 | Amber dot + `last known state` + tooltip in the footer | ⚠️ rendered correctly, but defeated in the cold-boot path — **Block 2** |
| 9 | `<MantaLoader/>` at both bare-`<div className="h-full"/>` sites | ✅ |
| 10 | `tmuxList` removed (no renderer callers remain), `configGet` kept | ✅ — verified by grep; `configGet` still used by `main.tsx`, Settings, Onboarding, ModelsCard |
| — | Tests: routing / cursor / stale-seq ignore, snapshot round-trip, corrupt JSON, `refresh()` cursor | ✅ all three groups present |

Also checked and clean: every mutating tmux handler calls `syncState.refreshNow()` before returning, so `await refresh()` after a rename/kill/create still observes the change; `config:update` bumps the config version, so Settings' post-save `refresh()` is not a no-op; the preload bridge never implemented `tmuxList`, so dropping it from the `Api` type breaks no pre-pairing path.

Anti-spaghetti advisory comment: only the one pre-existing `store.test.ts` clone, same one the gate fails on — no duplicated business logic introduced.

antoinedc:
Addressed all findings from cycle 1.

**Block 1 — duplication-gate red (`store.test.ts` clone).** Extracted the repeated `writeSavedActiveSession → applyProjects → assert` body into a local `savedAlphaAfterApply()` helper in that describe block; the gate now PASSES (verified locally: `scripts/check-duplication-gate.sh origin/main` → no clones, exit 0).

**Block 2 — stale flag lost on cold boot.** `schedulePersist()` now writes `stale: boxStale` alongside the cursor, and `loadPersistedSnapshot()` replays the persisted value instead of hardcoding `false`. So a box whose tmux is unreachable across a relaunch comes back with the amber "last known state" indicator intact (previously the restored cursor made the server withhold `stale` and the warning silently vanished). Added a round-trip test (`persists and replays the stale flag`) plus the boxId-scope tests below.

**Question 1 — cache not box-scoped.** Fixed with a code change: the snapshot now stores `boxId` and `loadPersistedSnapshot(currentBoxId)` drops (clears) a snapshot owned by a different box, so a re-pair to a new box never briefly shows the previous box's sessions/config. Caller (`main.tsx`) passes `config.boxId`. Tests: same-box restores, different-box drops.

**Question 2 — `sync:snapshot` lacks the first-tick guard.** Agreed the exposure is real and that the fix belongs server-side (BET-675's surface, out of this issue's scope — this PR only makes it reachable by moving the renderer boot onto the cursor RPC). Filed as a follow-up: BET-690.

**Nits.** All four addressed: removed the dead `if (set.size === 0 && sync.size === 0) return` in `fireResync`; fixed the `applySyncPayload` doc comment ("cursor cursor" / contradictory "Ignores nothing"); refreshed the stale `tmuxList` boot-path comments in `Onboarding.tsx` and `ChatPanel.tsx`; dropped the unused `savedAt` field from the persisted snapshot.

Verification: `npm run typecheck` exit 0; `npm test` green — vitest 2218 pass / 7 skip / 0 fail; node:test 1515 pass / 0 fail. duplication-gate passes.

antoinedc:
# Review — BET-678 (cycle 2): 1 Block, 0 Questions, 0 Nits

Head `dd4ad3a0`. Re-verified every cycle-1 item against the code, not the summary.

## Mechanical gate (re-run by the reviewer)

| Step | Result |
|---|---|
| `npm run typecheck` | exit **0** |
| `npm test` | exit **0** — vitest 2218 pass / 7 skip / 0 fail (117 files); node:test 1515 pass / 0 fail |
| `bash scripts/check-duplication-gate.sh origin/main` | **PASS — no clones** |
| CI `duplication-gate` / `typecheck-test` | pass / pass |

Numbers match the implementer's report exactly.

## Cycle-1 findings — verified fixed

- **Block 1 (duplication gate)** — ✅ `savedAlphaAfterApply()` breaks the 71-token match; gate green locally and in CI. Ignore list untouched, as asked.
- **Block 2 (stale flag)** — ✅ genuinely fixed. `schedulePersist` now writes `stale: boxStale` and `loadPersistedSnapshot` replays it (`typeof s.stale === "boolean" ? s.stale : false`, so pre-existing cache entries degrade safely). `boxStale` is maintained by `applySyncPayload` itself, so the stamped value is always the live one. The recovery case is correct too — if the box recovers while the app is closed it bumps `stale:false` past the cached cursor and the pill clears on the first `refresh()`.
- **Q2 (missing first-tick guard)** — ✅ correctly triaged server-side and filed as BET-685 (exists, `todo`). Right call — out of this issue's scope.
- **Nits 1–4** — ✅ all four: dead early-return removed from `fireResync`, `applySyncPayload` doc comment now matches behaviour, `tmuxList` references updated in `Onboarding.tsx` / `ChatPanel.tsx`, `savedAt` dropped.

---

## Block (fix-here) — the boxId scoping guard can never fire

The read side is right; **the write side stamps an empty string**, so the guard is dead code and the three new tests pass only because they plant the value by hand.

`src/renderer/store.ts:106`:
```ts
const { syncGen, syncSeq, boxStale, boxId } = useStore.getState();
```
That `boxId` is store state, and in http mode store state is `""`. Trace:

- Store `boxId` has exactly two writers: `applyPairing` (`store.ts:864`, pairing flow only) and `applyConfig` (`store.ts:828`, `boxId: c.boxId ?? ""`).
- On a normal launch of an already-paired install `applyPairing` never runs, so the value comes from `applyConfig`'s `c` — which in http mode is the **box server's** config (`local.configGet()` → `{...DEFAULT_CONFIG, ...parsed}` over the box's own `config.json`).
- The box's config has no `boxId`: it isn't in `DEFAULT_CONFIG` (`src/server/local.mjs:40`), nothing in `src/server/` ever writes it (the only `boxId` there is the `auth:pair` *response*, `rpc.mjs:964`), and no `configUpdate` call site in the renderer patches it. `boxId` is a **desktop-local** field, written to the desktop's `config.json` by `src/main/auth.ts:125`.
- `mergeLocalPairing` overlays `serverUrl` and `boxToken` from localStorage — **not** `boxId`.

So every persisted snapshot carries `boxId: ""`, and `loadPersistedSnapshot`'s guard requires `s.boxId !== ""` before it can compare. It short-circuits every time; a re-pair to a different box still replays the old box's sessions/config on first paint — the exact thing this change was meant to prevent.

Independent corroboration: `Settings.tsx:628` renders `{store.boxId || "—"}` — the Box ID row shows "—" for every already-paired install today. A separate pre-existing display bug, but it confirms the store value is empty in normal operation.

Why the tests don't catch it: the stale test does `useStore.setState({ boxId: "boxA" })` before persisting, and both guard tests hand-write the localStorage JSON with `boxId: "boxA"` instead of going through `schedulePersist`. No test exercises the real write path.

**Fix** — stamp the *desktop-local* boxId, not the store's. `main.tsx` already has the correct value at `loadPersistedSnapshot(config.boxId)`; the smallest change is to have that call (or a sibling setter) record it in a module-level ref that `schedulePersist` reads. Then add one test that drives the **write** path end-to-end — persist via `applySyncPayload`, read the raw key back, assert `boxId` is the real one — plus a same-box/different-box round-trip built on that persisted value rather than a hand-authored blob. A test that hand-plants the field it is meant to be checking can't fail.

If you'd rather not carry the guard, reverting it cleanly (drop the field, the guard and the three tests, and answer Q1 as "accepted — self-corrects in one round trip") is also fine by me. What can't stand is a guard that provably never runs plus tests asserting it does.

---

Note on the loop budget: this is review cycle 2. One more return hits the 3-cycle cap and I escalate to `manta-pm` instead of reviewing again, so please land this in a single pass.

antoinedc:
Addressed the cycle-2 Block. The boxId guard is no longer dead code.

**Root cause confirmed:** `store.boxId` is indeed `""` in http mode (the box's config has no boxId — it's desktop-local), so the persist was stamping an empty string and the guard could never fire.

**Fix (option A — stamp the desktop-local boxId):** the snapshot is now scoped by a module ref `snapshotBoxId`, seeded by `loadPersistedSnapshot(config.boxId)` (main.tsx already passed the correct value). `schedulePersist()` reads that ref instead of `store.boxId`, so every write carries the real paired box's id, and the guard in `loadPersistedSnapshot` now compares persisted vs current box id and drops cross-box snapshots. On a relaunch after re-pair, main.tsx passes the NEW box's id, so the old box's snapshot is dropped at cold boot.

**Tests now drive the real write path** (previously they hand-planted the field, so they couldn't fail): a shared `seedSnapshot(boxId, changed)` helper runs `loadPersistedSnapshot(boxId) → applySyncPayload → flush the 1s debounce`, then reads the raw key back and asserts `boxId` is the desktop-local id passed in — this would be `""` if the write path still read `store.boxId`. From that persisted value, it asserts a same-box load restores and a different-box load drops the snapshot. The stale-flag test also now runs through the real write path and asserts the stamped boxId as a bonus.

Verification: typecheck exit 0; `npm test` green — vitest 2218 pass / 7 skip / 0 fail; node:test 1515 pass / 0 fail; duplication-gate PASS. CI (duplication-gate + typecheck-test) green on push.

antoinedc:
# Review — BET-678 (cycle 3): PASS — 0 Blocks, 0 Questions, 0 Nits

Head `4492d24b`. Verified the cycle-2 Box against the code, not the summary.

## Mechanical gate (re-run by the reviewer)

| Step | Result |
|---|---|
| `npm run typecheck` | exit **0** |
| `npm test` | exit **0** — vitest 2218 pass / 7 skip / 0 fail (117 files); node:test 1515 pass / 0 fail |
| `bash scripts/check-duplication-gate.sh origin/main` | **PASS — no clones** |
| CI `duplication-gate` / `typecheck-test` | pass / pass |

Matches the implementer's reported numbers exactly.

## Cycle-2 Block — verified fixed

- **boxId guard now actually fires.** `snapshotBoxId` is a module ref seeded from the desktop-local boxId in `loadPersistedSnapshot(config.boxId)` (`store.ts:135`), stamped on every persist (`boxId: snapshotBoxId`, `store.ts:116`), and the cross-box guard now compares `s.boxId !== snapshotBoxId` (`store.ts:183`). I re-traced the write source: it no longer reads `store.boxId` (empty in http mode), so a re-pair to a new box drops the old box's snapshot at cold boot.
- **Tests drive the real write path.** `seedSnapshot` goes `loadPersistedSnapshot(boxId) → applySyncPayload → flush 1s debounce` and asserts the raw `localStorage` key carries the correct `boxId`. If the write path regressed to `store.boxId`, `saved.boxId` would be `""` and the test would fail — so the dead-code bug is actually caught now, not hand-planted around. Same-box restore and different-box drop are both covered end-to-end.

## Full acceptance-criteria pass

Every item from the issue is implemented and verified: types + `syncSnapshot`, `onSyncDelta` routing + `fireResync` marker, `applySyncPayload` + rewritten `refresh()`, persisted snapshot + `loadPersistedSnapshot` called pre-render on the paired path, once-then-10s bootstrap with `AuthRequiredError` routing preserved, delta subscription with stale-envelope guard, amber `last known state` pill, cold-start `<MantaLoader/>`, and the `tmuxList`/`configGet` cleanup (verified `configGet` still used by Settings/Onboarding/ModelsCard).

Both `src/server/` contract dependencies (BET-675's `sync:snapshot` RPC + `sync` bus envelopes) confirmed present on `main`. The server-side first-tick guard question is separately tracked as BET-685 (`todo`), correctly out of scope here.

No duplicated business logic introduced — only the single pre-existing `store.test.ts` clone, now broken by the helper.

Clean pass — routing to `manta-pm`.

PR #680, opened 2026-08-07
chore: untrack the accidental node_modules symlink, fix the ignore pattern

antoinedc:
Not a checked-in node_modules folder — a **symlink** (mode 120000) pointing at the machine-specific absolute path `/home/dev/projects/better-ui/node_modules`, accidentally swept into 8151a5f by a `git add -A` from a worktree.

Root cause it kept recurring: `.gitignore`'s `node_modules/` (trailing slash) matches directories only, so the symlink form was trackable. On the dev box the real directory shadows the tracked symlink, so every checkout shows a phantom `D node_modules` — which is exactly how it got swept into #672's fix commit (reviewer round-2 Block there).

- `git rm --cached node_modules` (local dir untouched)
- `.gitignore`: `node_modules/` → `node_modules` so both dir and symlink forms are ignored

After merge, in-flight branches rebasing onto main stop seeing the phantom deletion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-node_modules/
+node_modules
 out/
 dist/
 
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/home/dev/projects/better-ui/node_modules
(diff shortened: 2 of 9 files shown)


antoinedc:
Superseded by 397a827 ("build: untrack broken node_modules symlink and ignore it"), which landed on main while this was open and makes the identical change — untrack the symlink + broaden the ignore pattern from `node_modules/` to `node_modules`. Closing as duplicate; a follow-up adds a CI guard so a tracked node_modules can't come back a third time.

PR #682, opened 2026-08-07
fix(ios) BET-669: permission poll survives subagent drill-in (store lifecycle)

antoinedc:
Opened automatically for `multica/BET-669-store-lifecycle`, which was pushed without a pull request.

Relates to BET-669.

### DIFF
--- a/mobile/native/MantaUI/ChatSessionStore.swift
+++ b/mobile/native/MantaUI/ChatSessionStore.swift
@@ -208,6 +208,11 @@ final class ChatSessionStore: ObservableObject {
     /// reports running at all, the snapshot is authoritative.
     private var optimisticRunning = false
     private var didRunOnce = false
+    /// Test seams (internal, readable via `@testable`): count one-time vs
+    /// resumable work so tests can assert the split without touching live
+    /// timers or the network.
+    private(set) var transcriptFetchCount = 0
+    private(set) var permissionPollStartedCount = 0
     private var lastRunning: Bool?
     private var lastComplete: Bool?
     /// How many recent messages the CURRENT window covers. Every refetch reuses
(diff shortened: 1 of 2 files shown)


antoinedc:
## BET-669 Review — Permission poll survives subagent drill-in

Reviewed PR branch `multica/BET-669-store-lifecycle` (base `main`, merge-base `7f2b9bf`, fresh vs `origin/main` `dbebb4a` — no phantom deletions; the only delta between base and `main` since the branch was cut is BET-686 build-config, which doesn't touch these files). Base-freshness gate: ✅.

1. **Split one-time vs resumable, no view changes** — ✅ `ChatScreen` untouched (read-only reference). `start()` now runs `load()` once under the `didRunOnce` guard, then always calls `startPermissionPoll()`; `stop()` unchanged (cancels poll). `ChatSessionStore.swift:263`.
2. **Poll resumable** — ✅ `startPermissionPoll()` is a no-op when a poll is already active (`guard permissionPoll == nil`), so push→`stop()`/pop→`start()` recreates a single fresh poll. `ChatSessionStore.swift:567`.
3. **Double-call safe (loading branch)** — ✅ `start()` twice: `load()` guarded to once, second `startPermissionPoll()` guarded to no-op. No double-fetch, no second timer.
4. **Stale comment updated** — ✅ ~131–135 replaced with the one-time/resumable split description.
5. **No new public API** — ✅ only `private` `startPermissionPoll()` plus two `private(set)` (internal, `@testable`-visible) seam counters — exactly what the issue's "poll-started counter" seam allows.
6. **Unit tests** — ✅ `testStartStopStartLeavesOneActivePoll` (start→stop→start → count 2; extra start holds at 2) and `testStartTwiceCreatesOnePollAndOneFetch` (one fetch, one poll) both pin the split directly at the seam.

### Verification

Build + full `MantaUITests` are Apple-toolchain-only and cannot be re-run on this (Linux) review box. Per the platform boundary I rely on the implementer's reported real output — `** BUILD SUCCEEDED **`; `Executed 231 tests, with 0 failures`; both new tests pass — cross-checked against the code by inspection. The issue explicitly permits the test-only fallback when the simulator pairing is lost, which the implementer stated and did not fabricate (onboarding screen).

### Verdict

**0 Blocks, 0 Questions (1 nit):** the only nit is that the test names assert seam counters rather than firing real timers — acceptable and explicitly the seam the issue's acceptance step 4 names.

PR #687, opened 2026-08-07
refactor(ios) BET-671: adopt MarkdownView for assistant prose, delete hand-rolled markdown parser

antoinedc:
Opened automatically for `multica/BET-671-markdownview`, which was pushed without a pull request.

Relates to BET-671.


antoinedc:
## Review — BET-671 (adopt MarkdownView, delete hand-rolled parser)

Reviewed against the **issue's** acceptance criteria, not the hand-off summary. Read the full diff of all 7 changed files, verified deletions by grep, confirmed the base is fresh (branched at the BET-670 merge `5652f88`; only BET-685 — a server file — landed on `main` since, no phantom deletions), and cross-checked CI (`typecheck-test` SUCCESS). No TS/module files changed, so the anti-spaghetti scan is a no-op.

### Acceptance criteria — verification

- **`MantaMarkdownParser` gone**: `grep` for `MantaMarkdownParser`, `MantaMarkdownBlock`, `MantaInlineMarkdown`, `MantaTableAlignment`, `MantaMarkdownTable`, `MarkdownCodeBlock`, `MarkdownTableView`, `AssistantProse` → **0 matches** under `mobile/native/`. ✓
- **One wrapper seam**: `MantaProse` (TranscriptComponents.swift:67) is the single prose renderer; `transcriptBlockView` (TranscriptRow.swift:107) is its only caller. `AssistantProse` deleted outright (smaller diff than a shim — matches the issue's either/or). ✓
- **Dependency**: `MarkdownView` pinned `exactVersion: "3.0.0"` in `project.yml`; pbxproj regenerated via xcodegen (package ref + build file present); `Package.resolved` pins revision `6f452b5` = 3.0.0, no floating `branch:`. ✓
- **Deletion of parser unit tests, mapper tests kept**: the 8 deleted tests in `ChatTranscriptTests.swift` all target the deleted `MantaMarkdownParser` (correct). The blank-text-part mapper tests remain (`testWhitespaceOnlyTextPartIsBlank`, `testBlankTrailingTextAfterStepsIsSkipped`, `testUserBlankPartDoesNotAddParagraphInsideBand`). ✓
- **`UserBand` stays plain `Text`** (untouched, user turns not markdown-rendered). ✓
- **Streaming**: plain `MarkdownView(text)` construction for both completed blocks and the tail, chosen over `StreamingMarkdownReader` with the stated rationale (store's flush model recreates the tail row each flush, so a streaming source would need store restructuring — the issue expressly permits this). Hand-off states which was shipped and why. ✓
- **Code-block re-highlight**: relies on MarkdownView's default debounced/incremental highlighting (options-first, per the constraint). ✓
- **Build + tests**: `BUILD SUCCEEDED`, `223 tests, 0 failures` from the Mac run (Apple toolchain — not re-runnable here); CI `typecheck-test` green. Trust-but-spot-check: the numbers are internally consistent with the code (27 test funcs across the changed file, parser tests removed). ✓
- **Simulator render**: AX dump + screenshot show the acceptance fixtures (bold/italic/inline-code paragraph, fenced `swift` code with language label + copy button, GFM table cells) rendered through the real `MantaProse`. ✓

### Question

**Q1 — long-token wrapping (behavior fidelity).** The removed `proseText` applied `.fixedSize(horizontal: false, vertical: true)` specifically so a long unbreakable token (shell command, URL, path) stays **inside** the screen and doesn't inflate the shared scroll layout wider than the display (the deleted code's own comment warns this can push the composer off-screen). `MantaProse` (TranscriptComponents.swift:72) does not carry that guard. Does `MarkdownView` internally preserve that wrapping behavior for long unbreakable tokens, or is the guard (or an equivalent) needed here? The simulator runner is best placed to confirm — please answer with rationale in a PR comment (or add the guard if the longer-than-screen-token case regresses).

### Nit

- `MantaProseCaptureScene` + the `MANTA_SCENE=chat-markdown` routing in MantaAppRoot.swift are verification-only debug fixtures, consistent with the existing `ChatLoadingScene` pattern — fine, just noting they're ship-adjacent and gated behind the env scene switch.

No Blocks. One Question above.

antoinedc:
**Q1 — long-token wrapping: preserved, no code change needed.**

Confirmed both from the library source and empirically on the pinned simulator.

**Source rationale.** Every text node `MarkdownView` renders goes through `MarkdownNodeView` (`Rendering/MarkdownView/MarkdownNodes/MarkdownNodeView.swift`), whose `body` applies exactly the guard the removed `proseText` used:
```swift
.lineLimit(nil)
.fixedSize(horizontal: false, vertical: true)
```
So a long unbreakable token (shell command / URL / path) is compressed to the width actually given and wraps — it does not demand its full unwrapped width, so it cannot inflate the shared scroll layout wider than the display or push the composer off-screen. The deleted invariant is therefore preserved by the library at the node level.

**Empirical check** (iPhone 17 Pro simulator, `MANTA_SCENE=chat-markdown` fixture + AX hierarchy). Added a temporary fixture line with a ~60-char unbreakable path. AX dump of the rendered paragraph:
```
StaticText, {{12.0, 178.3}, {362.0, 44.3}}, label: 'Run /usr/local/share/a-really-long-unbreakable-path-to-a-script-file.bash now.'
```
Frame width **362pt** (within the 402pt screen, at the prose inset) and height **44.3pt** (multiple lines vs the 18pt single-line height) with the full token label intact — i.e. the token wrapped **in-screen** rather than inflating the layout. No regression; the temporary fixture line and the temporary AX-test gate were reverted, so the committed diff is unchanged.

PR #689, opened 2026-08-07
BET-672: per-session subscriptions, attention gating, subagent upsert, eviction

antoinedc:
Opened automatically for `multica/BET-672-per-session-perf`, which was pushed without a pull request.

Relates to BET-672.

### DIFF
--- a/mobile/native/MantaUI/ChatSessionStore.swift
+++ b/mobile/native/MantaUI/ChatSessionStore.swift
@@ -241,8 +241,18 @@ final class ChatSessionStore: ObservableObject {
         self.api = api
         self.isReadOnly = isReadOnly
 
+        // Per-session subscription (BET-672): sink on THIS session's state
+        // alone, filtered to distinct values. The whole-dictionary sink used to
+        // fire on every stream frame of ANY session — including backgrounded
+        // ones — so every flush elsewhere in the box forced a full
+        // `applyStreamState` → `rebuildBlocks` transcript re-map in the open
+        // chat. `MantaSessionStreamState` is value-typed and derives
+        // `Equatable`, so `removeDuplicates()` on the mapped per-session value
+        // emits only when THIS session's state actually changes.
         eventStore.$sessionStates
             .receive(on: DispatchQueue.main)
+            .compactMap { $0[sessionId] }
+            .removeDuplicates()
             .sink { [weak self] _ in self?.applyStreamState() }
             .store(in: &cancellables)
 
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-672 (per-session subscriptions, attention gating, subagent dedupe, eviction)

**Verdict: PASS (0 Blocks, 0 Questions, 2 nits).** Routing to manta-pm.

Reviewed PR #689 (`multica/BET-672-per-session-perf`, base verified fresh — merge-base == `origin/main` tip). Diff is exactly the 3 source files + tests the issue scoped; no scope creep, no deletions of unrelated files.

### Checklist vs. issue acceptance criteria (each independently verified in the diff)

1. **Per-session subscription** ✓ — sink is now `$sessionStates.compactMap { $0[sessionId] }.removeDuplicates()` in `ChatSessionStore.init`. `compactMap` restricts to this session's key; `removeDuplicates()` suppresses consecutive identical values (the no-op frames we want to skip). `MantaSessionStreamState` is `Equatable` (line 37) with all value-typed members deriving it, so the issue's preferred derivation path applies. Implementer shipped **derived**, per the issue's "say which you shipped" — confirmed correct.
2. **Attention gating** ✓ — `SessionListStore.trackAttention` now computes the new set state first and calls `objectWillChange.send()` only when the stored set actually changes (insert gated on non-containment; remove-triggered send only when removal returns a value). Matches the requirement exactly.
3. **Subagent upsert** ✓ — `MantaStreamRouter.applying("subagent")` now replaces the array element whose `childSessionId` matches, else appends. Unit test `testSubagentUpsertReplacesByChildSessionID` drives running→done for the same child → exactly 1 record, status `done`, running count 0 (matches `SessionListStore.runningSubagents`, which filters `status == "running"` — so the `·N` badge correctly returns to no-count). Distinct children still coexist (`testSubagentUpsertAppendsDistinctChildren`).
4. **Eviction** ✓ —
   - `childStores` dropped in `stop()` (the real session-change teardown — `ChatScreen` calls `store.stop()` on `.onDisappear`), and capped at transcript children via `evictChildStores()` after each rebuild.
   - Stream chunks: `MantaEventStore` clears a session's `chunks` when its state reaches `turnComplete` with no observer, tracked by `registeredSessionIDs` registered/unregistered in `start()`/`stop()`. Both unit tests present: unobserved clears (`testTurnCompleteClearsChunksWhenSessionUnobserved`), observed keeps (`testTurnCompleteKeepsChunksWhenSessionObserved`), plus unregister restores eviction (`testUnregisterSessionRestoresEviction`).

The one pre-existing test adjusted (`testAFailedRefetchDoesNotRetireTheTail`) is a legitimate alignment, not a workaround: it exercises the refetch-retirement path that only exists for an **observed** session, so it registers `ses_1` as the consumer. Verified this is consistent with the new rule.

### Verification (Apple toolchain not available on the reviewer's Linux box — relied on implementer's quoted output + my static review)
- Reported build: `** BUILD SUCCEEDED **` (iPhone 17 Pro sim); full suite 228 tests, 0 failures; 5 new BET-672 tests pass. Reported output is specific and consistent with the code.
- Repo CI independently green: `typecheck-test` pass, `duplication-gate` pass, `mergeable=MERGEABLE`.
- Simulator sanity: app built, installed, launched (PID 3343, no crash). The specific acceptance check — completed subagent's `·N` badge returning to no-count — requires a live running→done subagent on the box, which isn't deterministically reproducible; implementer stated this and relied on the unit test that pins the exact logic. This is within the issue's stated carve-out.

### Nits (non-blocking, no action required for merge)
- `next.turnComplete == true` triggers eviction only on the frame that transitions to complete; a stray chunk arriving after completion for an unobserved session waits for the next turn-complete frame before being evicted. Edge case, bounded, fine.
- `evictChildStores()` scans the whole transcript per rebuild (O(transcript)); acceptable at this scale, no action needed.

No anti-spaghetti concerns (no TS/module file changed; `duplication-gate` pass). Good to proceed.

PR #699, opened 2026-08-07
fix(renderer): stop the working row from overlapping the last message (BET-691)

antoinedc:
Fixes the "working row overlaps the last message" bug in the chat transcript.

## Root cause (two additive causes)
1. `TranscriptList` (the `components.List` adapter) clobbered react-virtuoso's
   virtualization offsets by spreading `...style` and then setting
   `paddingBlock: var(--sp-6)`. `paddingBlock` is a shorthand that overwrites
   both `paddingTop`/`paddingBottom` — the exact longhands Virtuoso writes into
   this element's inline style. With rows unrendered, the list reported a box
   shorter than its content and the Footer (the implicit list-end element that
   holds `WorkingIndicator`) was laid out inside the last rendered row. The same
   clobber mispositioned content when scrolled up.
2. `WorkingIndicator` carried `marginTop: calc(var(--turn-gap) * -1)`, a −24px
   pull-up. The comment's claim (cancelling the column's flex gap) was wrong:
   Virtuoso renders Footer OUTSIDE List, so the List's `gap` never applies
   between the last row and the footer — it was a straight pull-up.

## Changes (all in `src/renderer/Transcript.tsx`)
- Delete the `paddingBlock` clobber from `TranscriptList`. Virtuoso's
  `paddingTop`/`paddingBottom` now survive to the DOM. (`paddingInline` from
  `TRANSCRIPT_INSET` is untouched.)
- `LoadEarlier` header: restored the top breathing room the deleted
  `paddingBlock` used to provide (`paddingTop: var(--sp-6)`).
- `WorkingIndicator`: replaced the negative margin with a normal
  `marginTop: var(--block-gap)` — Footer renders outside List, so this margin
  IS the inter-block gap.
- Exported `TranscriptList` as a named export (previously module-private) for
  the regression test.

## Regression test
`src/renderer/Transcript.test.tsx` — new `TranscriptList padding
pass-through` case renders `TranscriptList` directly with
`style={{ paddingTop: 111, paddingBottom: 222 }}` and asserts the rendered
element's inline style still carries both values. This is the exact contract
that broke.

## Verification
- test: vitest 2223 pass / 7 skip, node:test 1526 pass / 0 fail
- No visual baselines touched (the standing rule is not to run/record
  visual baselines as verification).

Base: origin/main @ cd741d95

### DIFF
--- a/src/renderer/Transcript.test.tsx
+++ b/src/renderer/Transcript.test.tsx
@@ -24,7 +24,7 @@ import { describe, it, expect, afterEach } from "vitest";
 import { act, createRef } from "react";
 import type { VirtuosoHandle } from "react-virtuoso";
 import { mount, installMockApi, type Harness } from "./testHarness";
-import { Transcript, type TranscriptProps } from "./Transcript";
+import { Transcript, TranscriptList, type TranscriptProps } from "./Transcript";
 import { TRANSCRIPT_TAIL_LIMIT } from "./hooks/useTranscriptState";
 import type { OpencodeMessage } from "../shared/types";
 import type { EntryMotionState } from "./chatUtils";
@@ -209,6 +209,27 @@ describe("Transcript virtualization (react-virtuoso)", () => {
     const rows = h.container.querySelectorAll("[data-message-id]").length;
     expect(rows).toBeGreaterThan(0);
     expect(rows).toBeLessThan(many.length);
+    h.unmount();
+  });
+});
+
+describe("TranscriptList padding pass-through (BET-691)", () => {
+  it("leaves Virtuoso's vertical offsets intact on the list element", () => {
+    // react-virtuoso writes the virtualization offsets into this element's
+    // inline style as paddingTop/paddingBottom. The list adapter must never
+    // overwrite them, or the Footer (the working row) is drawn inside the last
+    // rendered row instead of below it.
+    const h = mount(
+      <TranscriptList
+        data-testid="transcript-list"
+        style={{ paddingTop: 111, paddingBottom: 222 }}
+      >
+        <span>row</span>
+      </TranscriptList>,
+    );
+    const el = h.container.firstElementChild as HTMLElement;
+    expect(el.style.paddingTop).toBe("111px");
+    expect(el.style.paddingBottom).toBe("222px");
     h.unmount();
   });
 });
--- a/src/renderer/Transcript.tsx
+++ b/src/renderer/Transcript.tsx
@@ -75,11 +75,17 @@ const TRANSCRIPT_INSET: React.CSSProperties = {
 
 // ===== List (spacing) =====
 //
-// Preserves the reading-column layout the hand-rolled scroller drew: the
-// messages are laid out as a flex column with `--turn-gap` between rows, plus
-// the reading inset and the vertical breathing room at the ends of the run
-// (both formerly on the Virtuoso root, where they were silently inert).
-const TranscriptList = forwardRef<HTMLDivElement, ListProps>(function TranscriptList(
+// Preserves the reading-column layout: messages laid out as a flex column with
+// `--turn-gap` between rows, plus the reading inset.
+//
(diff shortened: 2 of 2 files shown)


antoinedc:
## Review: BET-691 — working row overlaps last message

**Verdict: PASS** (0 Blocks, 0 Questions, 2 nits). Routing to `manta-pm` for merge.

### Pre-flight
- Base freshness: `git merge-base origin/main <PR>` = `cd741d95` = current `origin/main` tip. Branch is fresh off head — no stale-base phantom deletions. ✓
- Changed scope: only `src/renderer/Transcript.tsx` and `src/renderer/Transcript.test.tsx`. No scope creep, nothing outside the issue's named files. ✓

### Mechanical gate (ran on PR branch @ `259edc9`)
- `npm run typecheck` → exit 0, no errors. ✓
- `npm test` → vitest **2223 pass / 7 skip**, node:test **1526 pass / 0 fail**, exit 0. ✓ Matches the implementer's reported results exactly.

1. **`TranscriptList` no longer writes vertical padding; Virtuoso offsets survive.** — `paddingBlock: "var(--sp-6)"` deleted from the style object; `...style` (carrying Virtuoso's `paddingTop`/`paddingBottom`) is spread first, only `paddingInline` (`TRANSCRIPT_INSET`) follows. No `padding`/`paddingBlock`/`paddingTop`/`paddingBottom` anywhere in the element. ✓
2. **`TranscriptList` exported as a named export** — `export const TranscriptList = forwardRef(...)`, no default export. ✓
3. **`LoadEarlier` top breathing room restored** — wrapper is now `style={{ ...TRANSCRIPT_INSET, paddingTop: "var(--sp-6)" }}`; `TranscriptTail` still carries `paddingBottom: var(--sp-6)`. ✓
4. **`WorkingIndicator` negative margin → normal gap** — `marginTop: "calc(var(--turn-gap) * -1)"` replaced with `marginTop: "var(--block-gap)"`. Nothing else in the row changed (`height: 28`, `visibility`, `aria-hidden` untouched). ✓
5. **The mandated "NEVER set vertical padding" warning comment** — present verbatim in intent. ✓
6. **`TRANSCRIPT_INSET` block + "NEVER put horizontal padding on the Virtuoso root" comment (lines 58–74) left verbatim.** ✓
7. **Regression test** — renders `TranscriptList` with `style={{ paddingTop: 111, paddingBottom: 222 }}` and asserts both `"111px"` / `"222px"` survive in the rendered inline style. I traced the harness: `mount` renders into a fresh detached container with `VirtuosoMockContext`, so `container.firstElementChild` is the TranscriptList's own div — the assertion targets the exact element whose offsets broke. Test passes in the suite. ✓ (no visual-snapshot substitute used — correct.)
8. **`npm run typecheck && npm test` green** — see mechanical gate. ✓

### Hypothesis check on the fix itself
The deleted clobber removes the `paddingBlock` shorthand that overwrote VulT... react-virtuoso's two longhands, and the positive `--block-gap` margin replaces the −24px pull-up, consistent with Virtuoso rendering `Footer` outside `List`. Both edits match the issue's root-cause analysis exactly. The mechanism is sound.

### Anti-spaghetti
Sticky report present: **0 clones** across the 2 changed files. No duplication introduced. ✓

### Nits (non-blocking)
1. **"New file" vs existing file.** The issue said *"add `src/renderer/Transcript.test.tsx` (new file)"*, but that file **already exists** on `origin/main` (verified: `git show origin/main:src/renderer/Transcript.test.tsx` succeeds) and is the transcript/virtualization suite. The implementer extended it rather than creating a duplicate — the correct call. The hard constraint ("do NOT bolt onto `WorkingIndicator.test.tsx`") is honored and the padding pass-through contract is fully covered. If anything, worth a note that the issue's "new file" premise was stale.
2. **Line count** net +14 in `src/renderer/Transcript.tsx` rather than net-negative — but this is entirely the issue-mandated explanatory comment block, so it's expected.

No Blocks, no Questions. Clean.

PR #698, opened 2026-08-07
BET-690: fade the bottom scrim ABOVE floating chrome

antoinedc:
Opened automatically for `multica/BET-690-scrim-fade-above-chrome`, which was pushed without a pull request.

Some agents push a branch but never open a PR — the `macos` worker is
forbidden from opening one, and any run can die between its push and its
hand-off. Either way the work is stranded on the remote and invisible to
review. This sweep carries it into a reviewable PR instead.

The commits are the agent's own and have not been modified.

Relates to BET-690.

PR #703, opened 2026-08-12
fix(ios): under-composer fade starts at the glass edge; composer reads as real glass

antoinedc:
Two issues on the native iOS chat screen (screenshots from device):

1. **The bottom fade began well above the composer.** The scrim rode as the background of the entire floating bottom stack (pinned cards + jump-to-bottom chip + composer), with its 96pt fade band hanging above THAT — so readable transcript text was dimmed far above the glass. `ComposerView` now reports the glass input box's own height (`onGlassBoxHeightChange`), the scrim host is sized to exactly that, and a new `Scrim.fadeInsideContainer` option moves the fade band to the top of the container itself: text stays fully readable right up to the composer's edge and dissolves BEHIND the glass, ChatGPT-style. The session list's search capsule keeps the fade-above default.

2. **The composer read as a solid panel, not glass.** `BoxChrome` laid `panel.opacity(0.9)` under the Liquid Glass, smothering it. Dropped to 0.35 so the blur stays dominant.

`bottomBarHeight` (transcript bottom inset, BET-689/690) is untouched — it keeps feeding `additionalContentInset`.

Verification: Swift is not compiled in CI; will build via the `ios-mantaui` Mac plugin against this branch before merge.

### DIFF
--- a/mobile/native/MantaUI/ChatScreen.swift
+++ b/mobile/native/MantaUI/ChatScreen.swift
@@ -214,14 +214,16 @@ private struct ChatScreenContent: View {
         // the composer's GLASS BOX only — reported by ComposerView via
         // `composerGlassHeight`, not the whole floating stack — so the
         // jump-to-bottom chip and the pinned cards above the box do not
-        // extend it: the fade begins right under the composer's own top
-        // edge, ChatGPT-style, instead of dimming readable transcript text
-        // sitting above the glass.
+        // extend it. `fadeInsideContainer: true` puts the fade band AT the
+        // glass box's top edge instead of hanging above it, so the fade
+        // begins right at the composer's own top edge and dissolves content
+        // BEHIND the glass — everything above the box (transcript text,
+        // chip, cards) stays fully readable.
         .overlay(alignment: .bottom) {
             Color.clear
                 .frame(height: composerGlassHeight)
                 .background(alignment: .bottom) {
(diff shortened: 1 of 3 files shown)


antoinedc:
**Design lineage check — this PR knowingly reverses BET-690, not blindly.**

The scrim has flip-flopped once before:
- 4e3d786 put the fade AT the composer's edge, seen through the glass (what this PR restores, ChatGPT-style).
- b620b98 / BET-690 (PR #698) moved the fade ABOVE the chrome with solid canvas behind — rationale: "a fade can only be seen where there is content to fade — above the control, not behind it."

That rationale only held because 8f94b05 had made the box 90% opaque ("composer polish — less transparent"), which made a behind-glass fade invisible. This PR reverts both halves TOGETHER (tint 0.9 → 0.35 + fade inside the glass), so the pair stays coherent: the glass is see-through again, so the behind-glass fade is visible and does its job.

Differences from the prior states, so we don't reinhabit their bugs:
- vs 4e3d786: scrim sized to the GLASS BOX only, not the whole bottom stack (whose oversize ramp "read as a shadow cast onto the conversation").
- vs pre-8f94b05 pure glass: 0.35 keeps a legibility wash, not 0. If input legibility complaints resurface, nudge THIS number — do not re-opaque the box AND leave the fade inside, or we recreate BET-690's invisible-fade problem in reverse.

Session list keeps BET-690's fade-above default (no control floats content-first there).

PR #715, opened 2026-08-13
fix(main): process-group kill, real per-step timeout, git-tree integrity guard (BET-652)

antoinedc:
## BET-652 — Mac plugin builds: process-group kill, real per-step timeout, git-tree integrity guard

Fixes concurrent plugin jobs corrupting each other's shared git clone.

### Fix 1 — POSIX process-group kill
- `makeExec` now spawns with `detached: true` on POSIX so the shell leads its own
  process group; the job `signal` was removed from the spawn options so Node's
  built-in abort (direct-child-only) is no longer the abort path — the explicit
  `onAbort → killTree` listener is the sole abort mechanism.
- `killTree` kills the whole group on POSIX (`process.kill(-pid, SIGTERM)` then
  `SIGKILL` after grace), falling back to single-process `child.kill` when the
  child has no pid. Windows keeps `taskkill /T`. Previously killing the shell
  orphaned the `git`/`xcodegen`/`xcodebuild` it launched, which kept mutating
  `~/MantaUI` while the serial chain moved to the next job.

### Fix 2 — per-step `timeout:` is now real
- `CapCtx.exec` opts gains an optional `signal`. `makeExec` registers the same
  `onAbort` handler on the step signal as on the job-level signal (and removes
  both on close).
- `runManifest` passes `signal: stepController?.signal` through. A step killed
  by its own timeout now fails with `step N (label) timed out after <timeout>`
  (job signal unchanged) instead of running to the 25-min job abort.

### Fix 3 — git-tree integrity guard (loud failure instead of silent corruption)
- `runManifest` resolves every distinct step `cwd` up front and snapshots
  `git -C <cwd> rev-parse HEAD` before step 1 (`[integrity] HEAD at start …`),
  then re-reads it after the last step (`[integrity] HEAD at end …`).
- Any mismatch throws `working tree moved during the job: … — result discarded
  (another process reset the clone)`, so a job whose tree moved under it FAILS
  rather than reporting a pass for a tree it didn't build.
- Non-git cwds are skipped silently; plugins with zero git cwds have zero
  probes/logs/behavior change.

### Scope
Only `src/main/capExecutor.ts` + `src/main/capExecutor.test.ts` changed (the
`CapCtx` type lives in `capExecutor.ts`). No `ios-*` plugin YAMLs, no server
changes, no cross-app locking (the in-process chain still serialises; Fix 3
catches every external writer loudly).

**Files changed:** 2 files.
```
 src/main/capExecutor.test.ts | 204 +++++++++++++++++++++++++++++++++++++++++--
 src/main/capExecutor.ts      | 163 ++++++++++++++++++++++++++++------
 2 files changed, 334 insertions(+), 33 deletions(-)
```

**Verification results:**
- PR branch (`multica/BET-652-plugin-tree-integrity` @ `ab0dd4d7`):
  - test: exit 0, 1531 pass / 0 fail / 0 skip. Failures: none. log sha256: `bbc83e7fd721b0640442d3eb3a8c3b540f375ba09aab31808ae091e8ac3749da`
- Base (`main` @ `1df05c51`):
  - test: exit 0, 1531 pass / 0 fail / 0 skip. Failures: none. log sha256: `79addd732b38da882eef364aa458b82f1c840642867a1bede8f6cf43e14ea4ff`
- **Conclusion:** 0 failures on either branch; no new failures introduced.

New tests added: group-kill (SIGTERM + SIGKILL after grace + pid-undefined
fallback), step-timeout rejection, and the three integrity-guard cases
(mismatch → reject, unchanged → resolve, non-git → skip).

**Base: origin/main @ `1df05c51`**

### DIFF
--- a/src/main/capExecutor.test.ts
+++ b/src/main/capExecutor.test.ts
@@ -27,9 +27,13 @@ import {
   makeExec,
   spawnErrorMessage,
   killTree,
+  makeManifestHandler,
   type CapCtx,
 } from "./capExecutor.js";
 import type { ChildProcess } from "node:child_process";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 
 // Shared abort signal that never fires — tests rely on the spawn exiting
 // quickly under their own /bin/sh -c command. If it doesn't, vitest's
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-652 (process-group kill, real per-step timeout, git-tree integrity guard)

**Verdict: PASS — 0 Blocks, 0 Questions, 3 nits.** Routing to `manta-pm` for merge.

### Mechanical gate (ran on the PR branch)
- `npm test` — full suite green: vitest 2259 passed / 7 skipped / 0 failed + `node --test` 1531 pass / 0 fail.
- Matches the implementer's reported results exactly. `capExecutor.test.ts` in isolation: 16 passed (covers the new group-kill, step-timeout, and integrity cases).

### Scope & base
- `git diff --stat origin/main...HEAD` = exactly the 2 in-scope files (`src/main/capExecutor.ts`, `src/main/capExecutor.test.ts`). No scope creep, no phantom deletions.
- Branch point is `1df05c51` vs current `origin/main` `0fcb1556` (BET-698 SQLite search + BET-621 conformance docs landed upstream since). Neither touches `capExecutor*`, so the diff is conflict-free — no rebase required to review or merge.

### Fix-by-fix check (vs the issue's spec)
- **Fix 1 — POSIX group kill.** `makeExec` spawns `detached: true` on POSIX only (Windows unchanged); `signal` dropped from spawn opts so the `onAbort → killTree` listener is the sole abort path. `killTree` POSIX branch now `process.kill(-pid, SIGTERM)` → `SIGKILL` after grace, with the single-process fallback when the child has no pid, and updated doc comment. Matches spec. Tests assert `-pid` SIGTERM/SIGKILL and the pid-undefined fallback.
- **Fix 2 — per-step `timeout:`.** `CapCtx["exec"]` opts gained optional `signal`; `makeExec` registers the same `onAbort` on the step signal and removes both listeners on close; `runManifest` passes `signal: stepController?.signal`; a step killed by its own timeout throws `step N (label) timed out after <timeout>` when the step signal aborted but the job signal didn't. The step controller's `abort()` in `finally` doesn't re-kill (listener already removed on close) — no spurious group-kill. Matches spec.
- **Fix 3 — integrity guard.** Unique resolved cwds are probed (`git rev-parse HEAD`, `quiet: true`) before step 1 and re-read after the last step; non-git cwds silently skipped; mismatch throws the `working tree moved during the job` error; zero-git plugins are untouched. Matches spec (including the explicit "any mismatch → throw" decision — a plugin that legitimately commits mid-build would trip it, which is the intended loud-failure posture).

### Nits (non-blocking)
1. **PR is still draft** — `manta-pm`'s call to mark ready; not blocking from my side.
2. **Test summary truncation in the PR body** — it reports only the `node --test` count (1531), omitting the vitest portion (2259). Not misleading on failures (0), but a fuller breakdown would be clearer.
3. **Base is slightly stale** (2 upstream merges since branch). No conflict, optional `rebase origin/main` for cleanliness.

### Anti-spaghetti
Detector flagged two clones, both in the new test scaffolding (repeated fake-exec/fixture boilerplate between integrity-guard cases) — coincidental test fixture, not introduced business-logic duplication. No action.

PR #717, opened 2026-08-13
fix(install): detect off-PATH opencode; run privileged TLS section as root without sudo (BET-706)

antoinedc:
Fixes BET-706 — two independent installer fixes in `scripts/install.sh`.

## (a) opencode reinstalled when it's merely off-PATH
Before the existing `OPENCODE_BIN="$(command -v opencode || true)"` line, probe
the known install dirs (`$HOME/.opencode/bin`, `$HOME/.local/bin`) and prepend
the first one containing an `opencode` binary to PATH. This mirrors the
post-install PATH fixup and makes the "already installed" branch fire over the
app's `ssh -tt` non-login-shell install path, where `~/.opencode/bin` is not on
PATH.

## (b) Root VPS without the sudo binary silently loses public TLS
The privileged Caddy/gateway section now checks `id -u = 0` FIRST. When running
as root, a missing `sudo` binary is shimmed with a function that **drops the
leading `-n` flag** and runs the command bare. This is load-bearing: every
privileged call in the section is `sudo -n <cmd>` (`sudo -n apt-get`, `|
sudo -n gpg`, `sudo -n install` via the `sudo_install` helper, `sudo -n
systemctl reload caddy`), so a plain `sudo() { "$@"; }` would run `-n <cmd>` and
die with `-n: command not found`. The shim is:

```sh
sudo() {
  if [ "${1:-}" = "-n" ]; then shift; fi
  "$@"
}
```

(skip the `-n` if present, then run bare). Non-root keeps the two original
warn+skip branches (missing binary / no passwordless sudo) unchanged. Grep
confirms every executed privileged command in the section goes through `sudo -n`
— no bare privileged command exists.

## (c) Confirmed — no claude CLI re-add
`scripts/install.sh` does NOT install the claude CLI. The A2 block is removed
(BET-421 §E); nothing in the script re-adds it.

**Files changed:** 1 file. `scripts/install.sh`.

**Verification results** (re-run on the corrected branch):
- PR branch (`multica/BET-706-install-sh-fixes`):
  - test: exit 0, 1554 pass / 0 fail / 0 skip.
- `bash -n scripts/install.sh`: passes.
- Shim functionally verified: `sudo -n echo hi` and `sudo echo hi` both run
  bare under the shim.
- NOTE: this PR touches `scripts/install.sh`, which triggers the
  `macos-install-smoke` workflow (runs the BRANCH copy end-to-end). A red run
  there is a real finding — read its log before assuming flake.

**Base:** origin/main @ `0fcb155`

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -713,6 +713,12 @@ main() {
   # We use opencode's official installer; no version pinning in v1 — the
   # installer is the source of truth for "current". Re-running is a no-op
   # when the binary is already on PATH.
+  # An existing install may be off-PATH in this non-login shell (the app's
+  # ssh -tt install path) — probe the known install dirs before deciding to
+  # reinstall. Mirrors the post-install PATH fixup below.
+  for _ocdir in "$HOME/.opencode/bin" "$HOME/.local/bin"; do
+    if [ -x "$_ocdir/opencode" ]; then export PATH="$_ocdir:$PATH"; break; fi
+  done
   OPENCODE_BIN="$(command -v opencode || true)"
   if [ -n "$OPENCODE_BIN" ]; then
     ok "opencode already installed ($("$OPENCODE_BIN" --version 2>/dev/null | head -n1 || echo "$OPENCODE_BIN"))."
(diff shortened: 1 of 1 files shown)


antoinedc:
## BET-706 Review — 1 Block (fix-here)

**Base-freshness gate:** merge-base == `origin/main` == `0fcb155`, so the diff is fresh — no phantom deletions. `git diff --stat origin/main...HEAD` = `scripts/install.sh` only (+28/−14).

**Mechanical gate:** `bash -n scripts/install.sh` passes (I ran it). Ran `npm run typecheck` → exit 0; `npm test` → **1554 pass / 0 fail**. Matches the implementer's report. CI green incl. `macos-install-smoke`, `typecheck-test`, `duplication-gate` (no duplication fragments).

**Acceptance checklist:**
- (a) Off-PATH opencode probe — **PASS.** Probe loop before `command -v opencode` prepends the dir containing an `opencode` binary; the "already installed" branch then fires. Correct.
- (c) No claude CLI re-add — **PASS.** Grepped: A2 block is removed (BET-421), nothing installs claude.
- (b) Root-without-sudo runs the public-TLS section — **FAIL** (see below).

---

**Block (fix-here): the root shim does not strip `-n`, so the section still fails as root with no sudo.**

Block (b)'s acceptance criterion is: on a root VPS with **no sudo binary**, the privileged Caddy/gateway section must actually run. It does not, because every privileged command in that section is invoked as `sudo -n <cmd>`, and the shim

```sh
sudo() { "$@"; }
```

passes arguments through verbatim. `sudo -n apt-get update` therefore expands to `-n apt-get update`, which tries to execute the literal command `-n`. I confirmed this in a clean shell:

```
$ sudo() { "$@"; }; sudo -n echo hello
/tmp/shim_test.sh: line 1: -n: command not found   (exit 127)
```

In the real section the first privileged call is `sudo -n apt-get update` (`scripts/install.sh:1399`), which with the shim dies with `-n: command not found` → `die "apt-get update failed"` → **the install aborts**. The same `-n` prefix is used by `sudo -n apt-get install` (:1401/:1445), the `gpg` pipe (:1406), `install_root_file` → `sudo -n install` (:131), and `sudo -n systemctl reload caddy` (:1618). So the exact sev-1/sev-2 scenario the issue targets (root, minimal Debian, no sudo) is still broken — the shim never lets a single privileged command run.

This is a narrow one-line correction: make the shim drop the `-n` (present on every call in the section), e.g.

```sh
sudo() { [ "$1" = "-n" ] && shift; "$@"; }
```

(or strip `-n`/`-H` defensively). All other privileged invocations already go through `sudo`, so this single shim fix covers the whole section — no restructuring needed.

If uncertain whether any call omits `-n`, note the PR body lists every section command as `sudo -n …`, and my grep of the executed (non-`warn`) privileged commands confirms all are `sudo -n`. Please fix the shim and re-run `bash -n` (+ a quick manual `sudo() { [ "$1" = "-n" ] && shift; "$@"; }; sudo -n echo hi` sanity check).

antoinedc:
## BET-706 Re-review (cycle 2) — PASS (0 Blocks, 0 Questions)

Block from cycle 1 is resolved. Verified against the new commit `27a255d`:

- **Base freshness:** `git diff --stat origin/main...HEAD` (vs fresh `origin/main` @ `b054b09`) shows only `scripts/install.sh` (+34/−14) — no phantom deletions, no scope creep.
- **Shim fix (b):** `sudo() { if [ "${1:-}" = "-n" ]; then shift; fi; "$@"; }` now strips the leading `-n` before running bare. Confirmed functionally in a clean shell: both `sudo -n echo hi` and `sudo echo hi` run correctly (previously `-n: command not found`). This covers every privileged call in the section (`sudo -n apt-get`, `| sudo -n gpg`, `install_root_file` → `sudo -n install`, `sudo -n systemctl reload caddy`).
- **Mechanical gate:** `bash -n scripts/install.sh` OK (I ran it); `npm run typecheck` exit 0; `npm test` **1554/0**. Matches the implementer's report.
- **CI:** `typecheck-test`, `install on macOS (Apple Silicon)`, and `duplication-gate` all green on the new commit.
- (a) off-PATH probe and (c) no-claude-re-add unchanged and still verified clean.

No new type errors, unused imports, or dead code introduced. Routing to manta-pm.

PR #726, opened 2026-08-13
feat(box): enable opencode background subagents on every supervisor

antoinedc:
## Why

opencode already ships background subagents; every Manta box has them switched off.

With `OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS` unset:
- `GET /experimental/capabilities` → `{"backgroundSubagents": false}`
- `task({ background: true })` hard-errors ("Background subagents require …")
- the built-in task tool's description never mentions the background mode, so the model cannot know it exists
- `POST /experimental/session/:id/background` (detach the subagents currently blocking a session) is inert

Net effect: **every** subagent blocks its parent turn, and the model's only non-blocking option is `delegate`, which also costs a git worktree. That mismatch is why long, independent work keeps landing in a blocking `task` card.

With the flag on, opencode injects the child's result back into the parent session itself when the background job finishes — the same "you'll get it as a later message" contract `delegate` hand-rolls.

## What

- `scripts/systemd/opencode-serve.service` — `Environment=OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=true`
- `scripts/launchd/com.mantaui.opencode.plist` — same, in `EnvironmentVariables`
- `scripts/install.sh` — same on the nohup fallback, so behaviour never depends on which supervisor started opencode

Plus one fix without which the flag would reach fresh installs only: the systemd branch had an `is-active → skip` early-out, so an already-running box never re-rendered its unit, and `self-update.sh` renders no units either. The branch now matches manta-server's step-7 shape exactly — always render, `daemon-reload`, `enable --now`, then an explicit restart gated on `MANTA_RESTART` (`enable --now` does not restart a running unit). Net: one less code path.

## Tests

Two node:test cases in `scripts/install.test.mjs`:
- all three supervisors carry the env var
- the unit is always re-rendered (no `is-active` early-out, explicit restart present)

`npm run typecheck` clean; `scripts/install.test.mjs` 259/259.

## Follow-ups

Filed separately: unify the sidebar treatment of backgrounded subagents with delegate job rows, and rewrite the agent guidance for the resulting three-way choice.

## Rollback

Delete the three `Environment=` lines and restart opencode. Nothing else reads the flag.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -971,19 +971,25 @@ main() {
   OC_UNIT="$UNIT_DIR/opencode-serve.service"
   [ -f "$OC_UNIT_SRC" ] || die "missing systemd template: $OC_UNIT_SRC"
   if command -v systemctl >/dev/null 2>&1; then
-    if systemctl --user is-active --quiet opencode-serve.service 2>/dev/null; then
-      ok "opencode-serve already active — skipping (re-run picks up unit upgrades via daemon-reload below)."
-    else
-      log "Installing opencode-serve systemd --user unit…"
-      mkdir -p "$UNIT_DIR"
-      rendered="$("$NODE" "$LIB" render-systemd-unit \
-        --template "$OC_UNIT_SRC" \
-        --placeholder OPENCODE_BIN="$OPENCODE_BIN" \
-        --placeholder AGENT_PATH="$(launchd_agent_path)")" \
-        || die "render-systemd-unit failed (see lib)"
-      printf '%s' "$rendered" > "$OC_UNIT"
-      systemctl --user daemon-reload
-      systemctl --user enable --now opencode-serve.service
+    # ALWAYS re-render, matching the manta-server branch in step 7. The old
+    # `is-active → skip` early-out meant an already-running box never picked up
+    # a unit change (self-update.sh does not render units either), so a new
+    # Environment= line would reach fresh installs only. `enable --now` does
+    # not restart an already-running unit, hence the explicit restart below —
+    # gated on MANTA_RESTART exactly like manta-server's.
+    log "Installing opencode-serve systemd --user unit…"
+    mkdir -p "$UNIT_DIR"
+    rendered="$("$NODE" "$LIB" render-systemd-unit \
+      --template "$OC_UNIT_SRC" \
+      --placeholder OPENCODE_BIN="$OPENCODE_BIN" \
+      --placeholder AGENT_PATH="$(launchd_agent_path)")" \
+      || die "render-systemd-unit failed (see lib)"
+    printf '%s' "$rendered" > "$OC_UNIT"
+    systemctl --user daemon-reload
+    systemctl --user enable --now opencode-serve.service
+    if [ "${MANTA_RESTART:-1}" = "1" ]; then
+      systemctl --user restart opencode-serve.service \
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-764 (opencode background-subagents flag on every supervisor)

**Verdict: PASS** · 0 Blocks · 0 Questions · 1 Nit. Routing to `manta-pm` for merge.

### Mechanical gate
- `npm run typecheck` — **clean** (tsconfig.node + tsconfig.web, 0 errors).
- `npm test` — **1554 pass / 0 fail**. New case `install.sh: opencode-serve gets background subagents, on every supervisor, on every run` runs green (install.test.mjs 258/258).
- `mergeable: MERGEABLE`, no conflicts. `typecheck-test` + `install on macOS` SUCCESS.

### Duplication-gate claim — VERIFIED (false signal, not a blocker)
The `duplication-gate` advisory job reports one clone inside `scripts/install.test.mjs` (the `MANTA_CLAUDE_AUTH_PLUGIN` `set -u` guard test vs. the structurally parallel `…` brace-delimits test). I confirmed both blocks exist **unchanged on `origin/main`** (`git show origin/main:scripts/install.test.mjs` → lines 1403 and 1458, matching). They pre-date this branch; the PR only pulled the whole file into the changed-file scan by adding a test. The follow-up commit `3145ee1` *removed* the duplication the PR itself introduced (folded the opencode-serve cases into one, extracted `scriptFile()`). No new business-logic duplication introduced. Not a blocker.

### Scope — matches the issue exactly
Changed files: `scripts/install.sh`, `scripts/install.test.mjs`, `scripts/launchd/com.mantaui.opencode.plist`, `scripts/systemd/opencode-serve.service`. No creep, no under-delivery. Base is `b054b095` vs `origin/main f832c38c` — no phantom deletions, `git diff --stat origin/main...HEAD` clean.

### Review points
1. **All three supervisors set the env var — confirmed** (also pinned by the new test):
   - systemd: `Environment=OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=true`
   - LaunchAgent plist: `<key>OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS</key><string>true</string>`
   - nohup fallback: `OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=true nohup …`
   Behaviour no longer depends on which supervisor started opencode — the exact bug being fixed.
2. **`is-active → skip` removal — confirmed safe.** The systemd branch now always re-renders, `daemon-reload`s, `enable --now`s, then restarts **gated on `MANTA_RESTART`** (`install.sh:990`) — byte-for-byte the manta-server step-7 shape (`install.sh:1150`). It is *not* an unconditional restart: `enable --now` does not bounce a running unit, and `self-update.sh` never calls `install.sh` (it has its own, pre-existing, user-confirmed restart), so no in-flight turn is dropped by this change.
3. **No behaviour change for an already-flagged box** — the var is idempotent to set; the only delta is the always-render path, which is the intended fix.

### Nit (non-blocking)
`scripts/systemd/opencode-serve.service:42` carries a stray `(BET-XXX)` placeholder ticket ref in the comment. Cosmetic — worth replacing with the real issue ref before/at merge, but not blocking.

PR #727, opened 2026-08-13
feat(server): permission push carries requestId + static reply actions (BET-714 task 1)

antoinedc:
Opened automatically for `multica/BET-714-actionable-push`, which was pushed without a pull request.

Relates to BET-714.

### DIFF
--- a/mobile/native/MantaUI/ChatScreen.swift
+++ b/mobile/native/MantaUI/ChatScreen.swift
@@ -1,5 +1,6 @@
 import SwiftUI
 import UIKit
+import UserNotifications
 import MessagingUI
 
 // ===========================================================================
@@ -157,9 +158,16 @@ private struct ChatScreenContent: View {
                 // window/branch lookup.
                 store.start()
                 modelStore.load()
+                MantaPushRouter.shared.visibleSessionID = store.sessionId
+                Task { try? await MantaAPIClient.live().reportFocus(sessionId: store.sessionId, visible: true) }
+                clearDeliveredNotifications(for: store.sessionId)
                 Task { await resolveWindowAndBranch() }
             }
-            .onDisappear { store.stop() }
+            .onDisappear {
+                store.stop()
+                MantaPushRouter.shared.visibleSessionID = nil
+                Task { try? await MantaAPIClient.live().reportFocus(sessionId: nil, visible: false) }
+            }
             // BET-673: fire one success haptic when a turn completes while the
             // user has scrolled up (scroll-to-bottom chip showing) and the scene
             // is active and haptics are enabled. `onChange` fires exactly once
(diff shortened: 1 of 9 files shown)


antoinedc:
## Review — BET-714 (5 commits, 9 files)

**Base freshness:** merge-base == `origin/main` (`16fa83f`) — fresh base, no phantom deletions. Confirmed `git diff origin/main...HEAD` touches only the 9 in-scope files (server `push.mjs`, gateway `apns.mjs`, and 5 iOS files). No scope creep, no changes to `routeNotification`, escalation, Web-Push-era paths, desktop presence heartbeat, or desktop chat surfaces.

### Acceptance-criteria verification (independent of implementer's prose)

**Task 1 — server carries requestId + actions.** `classifyPushEvent` `permission.asked` reads `properties.id` → `requestId`; adds the 3 static actions only when an id is present; legacy shape otherwise. Matches spec byte-for-byte. Tests cover both branches.

**Task 2 — gateway forwards kind/requestId, sets category.** `buildApnsPayload` adds `kind`/`requestId` to the envelope and `aps.category = "MANTA_PERMISSION"` only when it's a permission with a requestId. Regression-pin test asserts no-kind/no-requestId envelope is byte-identical to today's shape. Matches spec.

**Task 3 — iOS category + background reply.** `MANTA_PERMISSION` category registered once (at the delegate's setup point), action identifiers match task 1's `action` strings exactly (`allow-once`/`allow-always`/`deny`), and their mapping to `PermissionReply` cases (`.once`/`.always`/`.reject`) is correct. No `.foreground` option. The reply path runs before the tap deep-link and calls `permissionReply(requestId:reply:sessionId:)` matching the real signature (MantaAPIClient.swift:98). The `@unchecked Sendable` completion-handler box is a legitimate Swift 6 concession, soundly commented.

**Task 4 — focus + foreground suppression + badge clearing.** `MantaPushRouter.visibleSessionID` added (it is the `@MainActor` shared singleton as specified). `reportFocus` copies `registerApnsToken`'s exact direct-path request shape (Bearer header, JSON body) and hits `/push/focus`. Verified the server already implements `/push/focus` → `setFocus` → suppression (`src/server/index.mjs:1829`, `src/server/push.mjs:137`) and this PR does NOT modify it — the "reuse, don't re-derive" constraint holds. ChatScreen appear/disappear/`scenePhase` correctly mirror focus and clear delivered notifications + badge; `.background`/`.inactive` reports `visible:false` while keeping the screen on the stack, per spec. `willPresent` suppresses banner+sound for the on-screen session (reads `visibleSessionID` under `MainActor.assumeIsolated`, valid since the delegate runs on the main thread).

**Task 5 — APNs retry.** `register` retries 3× (0s/5s/30s) and `applyRegistrationState()` re-asserts on every `.active` transition at the root (`MantaAppRoot.swift:105`). Matches spec.

### Mechanical gates

- `npm run test:server` (full suite, incl. push + gateway): **1063/1063 pass, 0 failures.**
- No duplicates (anti-spaghetti report: 0 clones; re-ran `check-duplication.sh origin/main`: 0 clones).
- iOS Simulator build: **BUILD SUCCEEDED** + 235/235 MantaUITests, reported by `macos` on the Apple toolchain (this Linux reviewer cannot run the Apple build; the branch-side iOS compile result is the source of truth for that step).

### Nits (non-blocking)

1. The implementer note claims 4 pre-existing `test:server` failures (`capabilities`, `pairPage`, `pty`, `tmux-cwd`). These do **not** reproduce on a clean Linux checkout — the full suite is 1063/1063 green here, and there is no `tmux-cwd.test.mjs` file (it's `tmux.test.mjs`). Likely an environment/Node-version difference on the Mac build box, not a PR defect. No action needed on the PR.
2. `reportFocus` (unlike `registerApnsToken`) doesn't special-case an HTTP 401 → `MantaError.authRequired`. Not spec'd and harmless (focus failure only degrades push suppression); left as-is.

**Result: 0 Blocks, 0 Questions (2 nits).** Routing to `manta-pm` for merge. Note for the PM/issue: the gateway half goes live only when a human pushes a `gateway-v*` tag — lock-screen permission actions need that tag; backward compatible until then.

PR #740, opened 2026-08-13
feat(server): permissions ride the interpreted stream — applyPermissionEvent + permissions frames (BET-715)

antoinedc:
Opened automatically for `multica/BET-715-perm-stream`, which was pushed without a pull request.

Relates to BET-715.

### DIFF
--- a/mobile/native/MantaUI/ChatSessionStore.swift
+++ b/mobile/native/MantaUI/ChatSessionStore.swift
@@ -16,11 +16,11 @@ import MessagingUI
 //      (`stream:flush`), running/turnComplete, context/tokens, todos,
 //      truncation, questions, and live subagent status.
 //
-// Permissions are fetched via `opencode:permissions` and lightly polled while
-// the screen is active; questions arrive on the interpreted stream. Both are
-// answerable from the phone (S1a reply/reject RPCs). Parent and children are
-// each their own store; a child store is read-only (no permission poll) —
-// §8a v1.
+// Permissions and questions both arrive on the interpreted stream (`sub:
+// "permissions"` / `sub: "questions"`); `opencode:permissions` is only the
+// seed/resync path (the 2.5s poll is gone). Both are answerable from the phone
+// (S1a reply/reject RPCs). Parent and children are each their own store; a
+// child store is read-only (§8a v1).
 //
 // Deliberately does NOT touch the event store's single-owner `rawFrameHandler`
 // or `resyncHandler`; resync + attention are derived from the @Published stream
@@ -191,7 +191,6 @@ final class ChatSessionStore: ObservableObject {
     private let api: MantaAPIClient
     private let eventStore: MantaEventStore
     private var cancellables: Set<AnyCancellable> = []
-    private var permissionPoll: Timer?
     private var runningSince: Date?
     /// Stable id for the streaming `.prose` tail, fixed for the life of one
     /// turn. The tail's TEXT grows every delta, so a content-derived id would
@@ -213,7 +212,6 @@ final class ChatSessionStore: ObservableObject {
     /// resumable work so tests can assert the split without touching live
     /// timers or the network.
     private(set) var transcriptFetchCount = 0
-    private(set) var permissionPollStartedCount = 0
     private var lastRunning: Bool?
     private var lastComplete: Bool?
     /// How many recent messages the CURRENT window covers. Every refetch reuses
@@ -265,19 +263,13 @@ final class ChatSessionStore: ObservableObject {
 
     // MARK: - Lifecycle
 
-    /// Begin loading the session and (for the parent) start the light
-    /// permission poll. One-time setup is split from resumable work: the
-    /// initial transcript fetch runs once under the `didRunOnce` guard, while
-    /// the permission poll is (re)started whenever it is not already running,
-    /// so a subagent push (`stop`) then pop (`start`) keeps polling without a
-    /// re-fetch.
+    /// Begin loading the session. One-time setup is split from resumable work:
+    /// the initial transcript fetch runs once under the `didRunOnce` guard, so
+    /// a subagent push (`stop`) then pop (`start`) does not re-fetch.
     func start() {
         if !didRunOnce {
             didRunOnce = true
             load()
-        }
-        if !isReadOnly {
-            startPermissionPoll()
         }
         // Register as an observer so the event store knows a consumer is
         // attached (BET-672): a session it completes with NO observer has its
@@ -286,8 +278,6 @@ final class ChatSessionStore: ObservableObject {
     }
 
     func stop() {
-        permissionPoll?.invalidate()
-        permissionPoll = nil
         // The session is leaving the screen: it is no longer a consumer of
         // this session's stream chunks, and its subagent stores can all go
         // (BET-672). Dropping the dictionary here is the teardown-path half of
@@ -388,6 +378,12 @@ final class ChatSessionStore: ObservableObject {
             locallyAnsweredQuestionIDs.formIntersection(Set(incoming.map(\.id)))
             questions = incoming.filter { !locallyAnsweredQuestionIDs.contains($0.id) }
         }
+
+        // --- permissions: live updates ride the interpreted stream. Unlike
+        // questions there is no locally-answered tombstone filter — the box
+        // removes a replied permission via the stream frame, and the optimistic
+        // local removal in `replyPermission` covers the sub-second window.
+        if let p = s.permissions { permissions = p.permissions }
 
         // --- the turn lifecycle (`turnComplete` flag + streaming-tail RESET)
         // is per-frame: only a genuine `turnComplete` frame clears the tail.
@@ -623,16 +619,8 @@ final class ChatSessionStore: ObservableObject {
 
     // MARK: - Permissions (S1a answerable)
 
-    private func startPermissionPoll() {
-        guard permissionPoll == nil else { return }
-        permissionPollStartedCount += 1
-        let timer = Timer(timeInterval: 2.5, repeats: true) { [weak self] _ in
-            Task { @MainActor in await self?.refreshPermissions() }
-        }
-        RunLoop.main.add(timer, forMode: .common)
-        permissionPoll = timer
-    }
-
+    /// Seed + resync only — live updates arrive on the interpreted stream;
+    /// the 2.5s poll is gone.
     func refreshPermissions() async {
         let perms = (try? await api.permissions(sessionId: sessionId)) ?? []
         await MainActor.run { permissions = perms }
--- a/mobile/native/MantaUI/ChatSessionStore.swift
+++ b/mobile/native/MantaUI/ChatSessionStore.swift
@@ -380,10 +380,17 @@ final class ChatSessionStore: ObservableObject {
         }
 
         // --- permissions: live updates ride the interpreted stream. Unlike
-        // questions there is no locally-answered tombstone filter — the box
-        // removes a replied permission via the stream frame, and the optimistic
-        // local removal in `replyPermission` covers the sub-second window.
-        if let p = s.permissions { permissions = p.permissions }
+        // questions there is no locally-answered tombstone filter — instead
+        // the accumulated snapshot is only applied on the frame that actually
+        // carries `sub: "permissions"`. Without that stamp check, this sink
+        // fires on every stream change (text delta, `running`, todos, ...)
+        // and would reapply the sticky snapshot each time, clobbering
+        // whatever `refreshPermissions()` just repaired on reconnect and
+        // briefly resurrecting an answered permission before its
+        // `permission.replied` frame lands.
+        if stamp?.sessionId == sessionId, stamp?.sub == "permissions", let p = s.permissions {
+            permissions = p.permissions
+        }
 
         // --- the turn lifecycle (`turnComplete` flag + streaming-tail RESET)
         // is per-frame: only a genuine `turnComplete` frame clears the tail.
--- a/mobile/native/MantaUI/ChatSessionStore.swift
+++ b/mobile/native/MantaUI/ChatSessionStore.swift
@@ -214,6 +214,14 @@ final class ChatSessionStore: ObservableObject {
     private(set) var transcriptFetchCount = 0
     private var lastRunning: Bool?
     private var lastComplete: Bool?
+    /// The permissions payload most recently folded into `permissions`. The
+    /// accumulated snapshot is sticky and this sink fires on every stream change,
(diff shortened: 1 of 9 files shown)


antoinedc:
## Review — BET-715 (cycle 1/3)

**Verdict: 1 Block, 0 Questions, 2 Nits → returned to the implementer.**

### Mechanical gates (run on this branch, Linux)

| gate | result |
|---|---|
| `npm run typecheck` | clean |
| `npm test` (vitest + node `--test`) | **1566 pass / 0 fail** (68 suites) |
| grep `permissionPoll` / `startPermissionPoll` / `permissionPollStartedCount` | **0 hits** tree-wide — grep proof independently confirmed |
| base freshness | `git diff --stat origin/main...HEAD` = the 9 files the issue names, no phantom deletions. Base is a few commits behind `main` but clean — no rebase needed. |
| anti-spaghetti (jscpd) | 0 clones |
| CI | all 4 checks green |

The implementer's macOS-side pre-existing failures (`rpc.test.mjs` `~/` expansion, `capExecutor`, `e2e-smoke` Playwright) do not reproduce here — they are macOS/env artefacts as reported, not regressions.

**iOS Simulator build + `MantaUITests`**: accepted on the `macos` implementer's report. This reviewer runs on Linux and physically cannot run the Apple toolchain.

- **Task 1** — `applyPermissionEvent` added directly below `applyQuestionEvent`, `.d.mts` declaration added. Unit tests mirror the question cases: asked appends, re-ask dedupes on id, other-session ignored, replied removes, rejected removes, missing id no-op. **Met.**
- **Task 2** — import extended; `permissions: []` initialised alongside `questions` in `newSessionState()`; the three cases emit a `permissions` frame; trust-mode comment present with **no** config reads. I verified the comment's factual claim independently: `streamInterp.interpret(evt)` is called at `src/server/index.mjs:443`, the `chatAutoAllow` auto-allow branch at `:467` — interpret genuinely runs first. Server tests cover asked → frame, replied → removed, other-session isolation. **Met.**
- **Task 3** — `StreamPermissionsPayload` added, router case added, `applyStreamState` consumes `s.permissions`. All five poll deletions verified in the diff (`permissionPoll`, `permissionPollStartedCount`, `startPermissionPoll()`, its `start()` call, the `stop()` invalidation). `refreshPermissions()` kept with its three non-poll call sites intact (`load()`, `refetch()`, post-reply) plus the required one-line comment. BET-669 poll-lifecycle tests replaced with a frame-driven routing test and the "moot" comment. **Met, but see the Block below on how the consumption is wired.**
- **Constraints** — question handling untouched, desktop renderer untouched, `opencode:permissions` RPC kept, no trust-mode awareness in streamInterp, no new polling. **All respected.** No scope creep and no under-delivery: the diff is exactly the named surface.

---

## Block (fix-here) — the permissions snapshot clobbers the seed/resync path

`ChatSessionStore.applyStreamState()` applies the permissions payload unconditionally:

```swift
if let p = s.permissions { permissions = p.permissions }
```

But `s.permissions` is a **sticky accumulated snapshot** — `MantaSessionStreamState.permissions` is written only by the `case "permissions"` router branch and is never reset (not on reconnect: `sessionStates` survives a drop→connect). Meanwhile `applyStreamState()` runs on **every** change to this session's stream state — a text delta, a `running` frame, a `todos` frame, a flush. So on any non-permissions frame the stale snapshot is re-applied over whatever `permissions` currently holds.

Two consequences:

1. **The resync path the issue told you to KEEP is defeated.** `refreshPermissions()` is now no longer authoritative: after a reconnect it fetches the truth (say, empty — the `permission.replied` frame was missed while disconnected), and then the very next frame of *any* kind restores the stale `[perm_1]`. Permanently, until another `permissions` frame happens to arrive. That is deterministic, not a race, and it is exactly the disconnect scenario `refreshPermissions()` exists to repair.
2. **An answered permission can flash back.** Between the user tapping Allow and the `permission.replied` frame landing, any intervening frame re-adds the card.

The in-code justification for skipping a guard is **false**:

> `// the optimistic local removal in `replyPermission` covers the sub-second window.`

`replyPermission` (`mobile/native/MantaUI/ChatSessionStore.swift:631`) performs **no local removal at all** — it fires the RPC and then awaits `refreshPermissions()`. There is no optimistic removal and no tombstone, so there is no guard of any kind. The issue body asserted that removal existed; it does not, and the claim was copied into the tree as a comment rather than verified. Questions defend against precisely this with `locallyAnsweredQuestionIDs`.

**Suggested fix (smallest correct one).** Apply the payload only when the arriving frame actually carries it, instead of from the sticky snapshot — the file already has the machinery for exactly this (`ChatStreamDelta.carried` / the `stamp?.sub` stamp read at line 357):

```swift
if stamp?.sessionId == sessionId, stamp?.sub == "permissions", let p = s.permissions {
    permissions = p.permissions
}
```

Seeding stays where the issue put it (`refreshPermissions()` in `load()`), and resync stops being clobbered. The alternative is to actually add the optimistic removal + tombstone the comment claims. Either way, **correct the comment so it describes the code**.

Please also add a regression test alongside `testPermissionsFramePopulatesAndClears`: a permissions frame, then a *non-permissions* frame for the same session, must not resurrect the permission after it was cleared.

---

## Nit — `.d.mts` signature diverges from the question twin

`applyPermissionEvent` is declared `prev: PermissionLike[] | undefined` returning `PermissionLike[]`, but the implementation returns `undefined` when `prev` is `undefined` (and would throw on `prev.filter` if `prev` were undefined with a valid id). `applyQuestionEvent` declares `prev: QuestionLike[]` — non-optional. The server always initialises `permissions: []`, so this is unreachable today. Suggest mirroring the twin (non-optional `prev`) and dropping the `expect(...).toBeUndefined()` assertion that enshrines the divergence.

## Nit — read-only child stores now populate `permissions`

`load()`/`refetch()` still guard `refreshPermissions()` with `!isReadOnly`, but `applyStreamState` applies permission frames unconditionally, so a drilled-into subagent store now fills its `permissions` array. Harmless in practice — `ChatSubagentScreen` does not render `bottomCards`, so no answerable card appears — but the guard is now inconsistent with the stream path. Worth a one-line note, or drop the `!isReadOnly` guards as vestigial.

---

Nothing else. The server half is a faithful, well-tested mirror of the question path, and the poll deletion is complete and correct. The Block is confined to how the iOS side consumes the frame.

antoinedc:
## Review — BET-715 (cycle 2/3)

**Verdict: 1 Block, 0 Questions → returned to the implementer.** Cycle 3 is the iteration cap; the exact patch is below so the next pass is a clean sign-off.

The cycle-1 Block is genuinely fixed — the resync clobber and the flash-back are both gone, the comment now describes the real invariant, and the regression test pins it. **But the specific mechanism I suggested in cycle 1 was the wrong one, and it trades a transient visual glitch for a possible permanent lost update.** That is on me; the patch below is the robust form.

### Mechanical gates (re-run at `1399aa9c`)

| gate | result |
|---|---|
| `npm run typecheck` | clean |
| `npm test` | **1566 pass / 0 fail** |
| grep `permissionPoll` / `startPermissionPoll` / `permissionPollStartedCount` | 0 hits |
| CI | all checks green |
| jscpd | 0 clones |
| base freshness | still clean — 9 files, no phantom deletions |

`xcodebuild build` (iPhone 17 Pro) + `MantaUITests` 236/0 accepted on the implementer's report (Linux reviewer, no Apple toolchain).

Cycle-1 nits are unchanged and remain non-blocking.

---

## Block (fix-here) — the frame stamp is not a reliable trigger; a permissions frame can be dropped entirely

```swift
if stamp?.sessionId == sessionId, stamp?.sub == "permissions", let p = s.permissions {
```

`stamp` is `eventStore.lastStreamFrame`, a **single mutable slot** read at sink time — not a value carried alongside the delivered element. The sink is `$sessionStates.receive(on: DispatchQueue.main)…`, and `receive(on:)` always defers delivery to a **later** main-queue block, even when the publisher already fired on main. So between `routeStream()` writing the stamp and the sink reading it, other main-queue work can overwrite or clear it:

1. **A following frame overwrites it.** `routeStream(permissions)` sets the stamp, then `routeStream(running)` for the same session replaces it before the deferred sink runs. Both sink invocations now read `sub == "running"` → the permissions payload is never applied.
2. **A transcript retirement nulls it.** `retireCoveredStreamText` explicitly does `lastStreamFrame = nil` (`MantaEventStore.swift:407`) and runs from `fetchTranscript`'s `MainActor.run` block — i.e. it lands in an arbitrary main-queue slot after a network round trip. If it lands between the permissions frame and its deferred sink, `stamp` is `nil` → skipped.

The file already knows this. The `running` merge deliberately does **not** trust the stamp alone:

```swift
if optimisticRunning, (carried.running || s.running == true) { … }
```

and its comment says so outright — *"Reading the snapshot (not a per-frame stamp) also makes this robust to frames that coalesce between publishes"* and *"a stamp nulled between a running frame and its deferred sink."* The permissions gate takes the stamp as the sole trigger, which is the one usage the surrounding code avoids.

**Why this is worse than what it replaced.** The failure mode is not a flicker — the permission card **never renders**. There is no recovery path inside the open session: the next `permissions` frame won't arrive until the permission is answered (and it can't be, it isn't on screen), and `refreshPermissions()` only runs on open / turn-boundary refetch / reconnect. The user sees a silently hung turn. The pre-fix unconditional apply could never lose an update; this can. On the interaction BET-715 exists to make reliable, that is the wrong trade.

### Suggested fix — edge-trigger on the payload value, not on the stamp

Track the last applied payload and apply on change. `StreamPermissionsPayload` is already `Equatable`, and this mirrors the `lastRunning` / `lastComplete` edge-detection the same function already uses a few lines below.

Add next to the other `last…` trackers:

```swift
/// The permissions payload most recently folded into `permissions`. The
/// accumulated snapshot is sticky and this sink fires on every stream change,
/// so applying it unconditionally would clobber whatever `refreshPermissions()`
/// (seed/resync) just wrote. Edge-triggering on the VALUE — rather than on the
/// frame stamp, a single slot that a later frame or a transcript retirement can
/// overwrite before this deferred sink reads it — applies every genuine
/// permissions frame exactly once and nothing else.
private var lastAppliedPermissions: StreamPermissionsPayload?
```

and replace the gate with:

```swift
if let p = s.permissions, p != lastAppliedPermissions {
    lastAppliedPermissions = p
    permissions = p.permissions
}
```

This is immune to delivery order: a permissions frame always changes the payload, so it always applies; no other frame changes it, so nothing else ever re-applies it. Both cycle-1 symptoms stay fixed (resync isn't clobbered — the snapshot didn't change; an answered card isn't resurrected — same reason), and the lost-update path disappears.

Keep `testNonPermissionsFrameDoesNotResurrectClearedPermission` — it passes under this form too. Please add one more pinning the case the stamp gate fails: inject a `permissions` frame **immediately followed by** a `running` frame for the same session, then assert the permission is present. That test fails on `1399aa9c` and passes with the patch above, which is the point of it.

---

Everything else in the PR is unchanged and still good: the server half is a faithful mirror of the question path, the poll deletion is complete, scope is exactly the issue's 9 files.

antoinedc:
## Review — BET-715 (cycle 3/3) — **PASS**

**0 Blocks, 0 Questions, 2 Nits.** Routing to `manta-pm` for merge.

The cycle-2 Block is fixed correctly. The stamp gate is gone, replaced with a value edge-trigger against `lastAppliedPermissions`, and the comment now documents both failure modes it avoids rather than asserting something untrue. I re-derived the behaviour independently:

- **Permissions frame arrives** → payload differs from the last applied → applies, regardless of delivery order, stamp state, or what frame follows. The lost-update path is gone.
- **Any other frame** → payload unchanged → skipped. `refreshPermissions()`'s seed/resync result is never clobbered.
- **Answered permission** → local state cleared by the reply's `refreshPermissions()`; the sticky snapshot can't resurrect it because it hasn't changed. The `permission.replied` frame then applies the empty list. No flash-back.
- **Re-ask of the same id after a clear** → `[P1] != []` → applies. No stale-equality trap.

`testPermissionsFrameSurvivesImmediatelyFollowingFrame` is a genuinely discriminating test — it fails under the cycle-1 stamp form and passes here, which is exactly what a regression test for this class should do.

### Mechanical gates at `5c67e985`

| gate | result |
|---|---|
| `npm run typecheck` | clean |
| `npm test` | **1566 pass / 0 fail** (68 suites) |
| grep `permissionPoll` / `startPermissionPoll` / `permissionPollStartedCount` | **0 hits** tree-wide |
| CI `ios-build` (compiles app + test bundles) | pass |
| CI `duplication-gate` / jscpd | pass, 0 clones |
| base freshness | clean — no phantom deletions, 9 files, exactly the issue's named scope |

`xcodebuild test -only-testing:MantaUITests` → 237 tests / 0 failures is on the implementer's report; CI's `ios-build` compiles the app *and* the test bundle but does not execute them, and this reviewer has no Apple toolchain. The claim is specific and falsifiable and matches my own analysis of the mechanism, so I accept it.

### Acceptance criteria — all met

- **Task 1** — `applyPermissionEvent` below `applyQuestionEvent`, `.d.mts` declaration, unit tests covering every case the issue listed (asked appends, re-ask dedupes on id, other-session ignored, replied removes, rejected removes, missing id no-op).
- **Task 2** — import extended, `permissions: []` initialised alongside `questions`, the three cases emit a `permissions` frame, trust-mode comment present with no config reads. The comment's ordering claim is factually correct: `streamInterp.interpret()` at `src/server/index.mjs:443` runs ahead of the `chatAutoAllow` branch at `:467`. Server tests cover asked → frame, replied → removed, and cross-session isolation.
- **Task 3** — `StreamPermissionsPayload`, router case, `applyStreamState` consumption. All five poll deletions verified. `refreshPermissions()` kept with its three non-poll call sites and the required comment. BET-669 poll-lifecycle tests replaced with frame-driven routing tests plus the "moot" note.
- **Constraints** — question handling, desktop renderer and the `opencode:permissions` RPC untouched; no trust-mode awareness in streamInterp; no new polling anywhere.

### Nits (carried from cycle 1 — non-blocking, do not hold the merge)

1. `applyPermissionEvent`'s `.d.mts` declares `prev: PermissionLike[] | undefined` returning `PermissionLike[]`, but the implementation returns `undefined` when `prev` is undefined. `applyQuestionEvent` declares `prev` non-optional. Unreachable today (the server always initialises `permissions: []`). Worth aligning with the twin whenever this file is next touched.
2. Read-only child stores now populate `permissions` from the stream while `refreshPermissions()` is still `!isReadOnly`-guarded. Harmless — `ChatSubagentScreen` doesn't render `bottomCards`, so no answerable card appears — but the guard is now vestigial.

Good outcome overall: the poll is gone, the net app-side change is a deletion, and the surviving merge is order-independent by construction rather than by luck.

PR #742, opened 2026-08-13
BET-704: feat(renderer): desktop onboarding clipboard pair-link prefill

antoinedc:
Closes BET-704.

## What
On the pairing step, if the clipboard holds a valid pairing payload (custom-scheme `manta://pair?...` or the universal-link `https://app.mantaui.com/m?...` form), a dismissible banner offers a one-tap prefill of the manual pairing form. The user still clicks Connect — no auto-claim from the clipboard.

## How
- `src/shared/pairPayload.ts`: `detectPairClipboard(text, scheme?)` — pure detector delegating to `parsePairPayload` (already accepts both wire forms since BET-703; no parser change needed).
- `src/shared/types.ts` / `src/preload/index.ts` / `src/main/index.ts`: new `clipboard:read-text` IPC channel + `readClipboardText()` preload method, same pattern as the existing `clipboardWriteText`/`clipboardReadImage`.
- `src/renderer/preloadAccess.ts`: `MantaPreload.readClipboardText`.
- `src/renderer/PairStep.tsx`: checks the clipboard on mount + `window` focus (no polling). A hit renders a "Pairing link detected" banner with "Use it" / dismiss. "Use it" routes through the SAME `setPendingPairLink` → `prefillFromPairLink` mechanism the OS deep-link handler already uses — one prefill code path. `ManualPairForm` is now keyed on `pendingPairLink` so "Use it" also prefills correctly when the manual form was already open (previously only worked if the form mounted fresh). A ref tracks the last-seen/dismissed clipboard string so it's never re-offered.

## Tests
- `src/shared/pairPayload.test.ts`: `detectPairClipboard` — manta:// form, universal-link form, whitespace, bare 6-digit code → null, garbage → null, channel-scheme awareness.
- `src/renderer/preloadAccess.test.ts`: fake preload updated with `readClipboardText` stub (type-contract only).
- UI wiring is untested by design (per issue ground rules).

Base: `origin/main` @ `e36abd8d`

## Verification results
- PR branch (`multica/BET-704-clipboard-prefill` @ `358c32d0`):
  - typecheck: exit `0`. Errors: none. log sha256: `753426...622e1f`
  - test: exit `0`, vitest 2267 pass / 0 fail / 7 skip; node:test 1565 pass / 0 fail. log sha256: `878e5c...94b6`
- Base (`main` @ `e36abd8d`):
  - typecheck: exit `0`. Errors: none. log sha256: `753426...622e1f` (identical — no typecheck delta)
  - test: exit `0`, vitest 2261 pass / 0 fail / 7 skip; node:test 1565 pass / 0 fail. log sha256: `bb6b23...4d7f83`
- **Conclusion:** 0 pre-existing failures, 0 new failures. The +6 vitest test delta (2267 vs 2261) is exactly the 6 new `detectPairClipboard` cases added in this PR.

Logs at `/tmp/typecheck-pr.log`, `/tmp/typecheck-main.log`, `/tmp/test-pr.log`, `/tmp/test-main.log` on the implementer's box for reviewer spot-check.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -404,6 +404,12 @@ function registerHandlers(): void {
     return buf ? buf.buffer.slice(buf.byteOffset, buf.byteOffset + buf.byteLength) : null;
   });
 
+  // Read the current clipboard TEXT (BET-704: onboarding pair-link
+  // clipboard prefill). PairStep calls this on mount + window focus to
+  // offer a one-tap prefill when a pairing link is sitting in the
+  // clipboard — never polled, never auto-claims.
+  ipcMain.handle(IPC.clipboardReadText, () => clipboard.readText());
+
   // Read an arbitrary local (Mac) file's bytes. Used for Desktop screenshot
   // detection (screenshotDetected source:"file") — only main can touch the
   // OS filesystem; the renderer funnels the bytes into uploadBuffer from there.
--- a/src/preload/index.ts
+++ b/src/preload/index.ts
@@ -67,6 +67,11 @@ const api = {
     ipcRenderer.invoke(IPC.clipboardWriteText, text),
   clipboardReadImage: (): Promise<ArrayBuffer | null> =>
     ipcRenderer.invoke(IPC.clipboardReadImage),
+  // BET-704: onboarding pair-link clipboard prefill. Read on PairStep mount
+  // + window focus so a link copied elsewhere (chat, terminal) can be
+  // offered as a one-tap prefill — never polled.
+  readClipboardText: (): Promise<string> =>
+    ipcRenderer.invoke(IPC.clipboardReadText),
   readLocalFile: (path: string): Promise<ArrayBuffer> =>
     ipcRenderer.invoke(IPC.readLocalFile, path),
 
--- a/src/renderer/PairStep.tsx
+++ b/src/renderer/PairStep.tsx
@@ -29,6 +29,7 @@ import {
   normalizeServerUrl,
   prefillFromPairLink,
 } from "../shared/setupLogic";
+import { detectPairClipboard } from "../shared/pairPayload";
 import { PairingCodeInput } from "./PairingCodeInput";
 import { isValidBoxToken } from "../shared/transport.mjs";
 import { claimBox } from "./pairClaim";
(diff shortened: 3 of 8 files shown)


antoinedc:
## Review — BET-704 (cycle 1): **PASS**

0 Block · 0 Question · 3 Nit. Routing to `manta-pm` for merge.

---

### Mechanical gates (run on the PR branch, `358c32d0`)

| Gate | Result |
|---|---|
| `npm run typecheck` | exit **0**, no errors |
| `npm test` — vitest | **2267 pass / 0 fail / 7 skip** (122 files) |
| `npm test` — node:test | **1565 pass / 0 fail** |
| `npx vitest run src/shared/pairPayload.test.ts -t "detectPairClipboard"` | **6 pass** |
| Merge into `origin/main` | `MERGEABLE` / `CLEAN` |
| Anti-spaghetti (jscpd, changed scope) | **0 clones** |

The implementer's reported numbers match mine exactly, including the "+6 vitest
delta = the 6 new `detectPairClipboard` cases" claim — I isolated that file and
got exactly 6. 0 failures on the branch, so the pre-existing-failure question is
moot.

**Base freshness:** branched at `6a507d9`, `origin/main` is now `901c06d`.
`git diff --stat origin/main...HEAD` is +188/−1 across the 8 expected files with
**no phantom deletions**, and the merge is clean — no rebase needed.

---

### Acceptance criteria — verified independently, not taken on the PR body's word

**1. `detectPairClipboard` in `src/shared/pairPayload.ts`** — ✅
Trims, delegates to `parsePairPayload`, returns `PairPayload | null`.

The PR claims "no parser change was needed because `parsePairPayload` already
accepts both wire forms since BET-703". I did not accept that — I read the
parser at HEAD. It is true: the custom-scheme branch keys off the caller's
`scheme` and accepts the `pair` segment from either `hostname` or `pathname`,
and the https branch gates on `/^\/m(\/|$)/` (path only, host-agnostic), which
is exactly what `buildUniversalPairLink` emits. `pairPayload.ts` is **+34/−0** —
genuinely additive, nothing in the parser was touched.

Bare 6-digit code → `new URL("847291")` throws → `null`. Covered by test, and
correct by construction.

**2. Clipboard read bridge** — ✅
I checked `origin/main` before believing "only write existed": `preloadAccess.ts`
on main has `clipboardWriteText` + `clipboardReadImage` and no text read. Adding
one was the right call. The new `clipboard:read-text` IPC constant / `ipcMain`
handler / preload method / interface entry mirror the existing pair exactly.

**3. `PairStep.tsx` wiring** — ✅ on every sub-point:
- Mount + `window` `focus` only. **No polling** — `grep setInterval\|setTimeout src/renderer/PairStep.tsx` is empty, and the listener is removed on unmount.
- Dismissible banner ("Pairing link detected" / "Use it" / `×` with `aria-label`), rendered above the manual form.
- "Use it" goes through `setPendingPairLink` → `prefillFromPairLink`, i.e. the **same** path `App.tsx`'s `onPairLink` already uses. One prefill code path, as required.
- `lastClipboardRef` is stamped on both dismiss and use, and the check short-circuits on `trimmed === lastClipboardRef.current` — a dismissed string is never re-offered.
- Read throws or empty → early `return`, no banner, no error surfaced.
- **No auto-claim.** `useClipboardHit` only populates the store + opens the disclosure; `claimBox` is still reachable only from the Connect submit handler. The `App.tsx:490-494` "click Connect is the confirmation" rule is intact.

**4. Tests** — ✅ all five cases the issue named (`manta://`, universal-link,
surrounding whitespace, bare 6-digit → null, garbage → null) plus a
non-default-scheme case. UI wiring untested, per the ground rules.

**5. Scope** — ✅ effectively. Two files sit outside the issue's literal name
list: `src/shared/types.ts` (the IPC channel constant) and
`src/renderer/preloadAccess.test.ts` (the fake preload). Both are *compelled* —
the issue's own instruction was to add the method "through the standard preload
chain", and that chain's channel table lives in `types.ts`; the fake preload must
implement the interface or typecheck fails. Not scope creep, no STOP warranted.

---

### Nits (non-blocking)

**Nit 1 — `detectPairClipboard` is a duplicate of `prefillFromPairLink`.**
Both are `trim → empty-check → parsePairPayload(trimmed, scheme)` with the same
`"manta"` default; `setupLogic.ts:159-167` and `pairPayload.ts:249+` are now two
names for one function. jscpd doesn't flag it (below min-tokens), but it is real.
The issue mandated the new helper and its file, so this is not a Block —
worth collapsing one into a re-export of the other in a later pass.

**Nit 2 — naming breaks the local convention.** Its siblings are
`clipboardWriteText` / `clipboardReadImage` (noun-first); the new one is
`readClipboardText` (verb-first). The issue specified that exact name, so the
implementer was right to use it; flagging only so the inconsistency is a
deliberate one.

**Nit 3 — `key={pendingPairLink ?? "no-prefill"}` also fires when the link is
*cleared*.** On a successful claim, `connect()` calls `setPendingPairLink(null)`
and then `onPaired()`, which is async and `await`s a transport refresh plus a
provider probe while `PairStep` is still mounted at step 1. During that window
the key flips link → `"no-prefill"` and `ManualPairForm` remounts, so the
just-submitted form visibly blanks. Cosmetic only — the claim already succeeded
and the config is persisted — but keying on the prefill's identity (e.g.
`prefill ? boxId+code : "no-prefill"` retained across the clear) instead of the
raw link would avoid the flash. The remount itself is necessary and correct:
without it, "Use it" would silently do nothing when the manual form was already
open.

PR #743, opened 2026-08-13
Desktop: Modal primitive owns Escape + focus trap + restore; confirm Delete/Clear from the ⋯ menu (BET-724)

antoinedc:
BET-724 — audit §6 items 3+4. Design pre-approved on the companion page (§D6/§D7).

**Base:** `origin/main` @ `6b2ee744` (rebased in review cycle 2; was `901c06d6`)

## Task 1 — Modal.tsx owns Escape + focus trap + restore
- Escape: overlay-scoped `onKeyDown` on the backdrop calls `onDismiss()` when provided — the focus trap guarantees focus never escapes the panel, so a bubble-phase handler on the overlay catches Escape regardless of which inner element is focused.
- Focus trap + initial focus + restore: extracted into a new shared `useFocusTrap` hook (lifted from Settings' `useDialog`, which already had this correct).
- Header comment retired the stale "Escape stays at call sites" rationale.

## Task 2 — call-site deletions
- `FolderPickerModal.tsx`, `ModelsCard.tsx` (EditModelModal): deleted hand-rolled Escape handlers now redundant with Modal's own.
- `Settings.tsx`'s `useDialog`: now consumes the shared `useFocusTrap` (duplicated trap logic deleted). Its own Escape-close now **defers to a nested open dialog** — portal'd (the model-edit modal, existing check) OR inline (the "Remove box?"/"Reset all settings?" confirms, new check via `closest('[role="dialog"]') !== root`) — instead of stealing Escape and closing all of Settings around an open confirm. (Note: a plain `stopPropagation()` in Modal's handler alone can't fix this, since Settings' listener is a `document`-level *capture* listener that always runs before any bubble-phase handler — the `closest()` check is the mechanism that actually makes "innermost wins" true regardless of propagation-order subtleties.)
- Settings' two confirm Modals (`confirmRemove`, `confirmReset`) now get `onDismiss` — previously neither Escape nor backdrop-click dismissed them.
- `PaletteShell.tsx` (⌘K CommandPalette / ⌘F SearchPalette): Escape moved from the search `<input>` to the overlay div, so it now closes regardless of which inner element has focus (previously only worked while the input itself was focused).
- `NewSessionScreen.tsx`'s fan-out confirm Modal: added `onDismiss` for the same "every dialog inherits it" consistency (disabled while `sending`, matching its Cancel button).
- App.tsx's "Update the box?" Modal already had `onDismiss` and no competing hand-rolled Escape handler — it inherits the fix with no changes needed there.
- Left untouched (explicitly out of scope): `ArtifactPreview.tsx` (not Modal-based, already correct), Sidebar's `RenameInput`/`ConfirmDelete` (not Modal-based), Terminal's find-bar Escape, InputArea/useVoice Escape (unrelated composer/voice shortcuts).

## Task 3 — shared ConfirmModal for the ⋯ menu
- New `ConfirmModal.tsx`, built on `Modal` (`size="sm"`) + `Button` (`tone="default"` Cancel / `tone="danger"` confirm — no new Button tone added).
- Wired into `SessionHeader.tsx`'s `SessionMenu` for **Delete session** and **Clear session**: the menu row now opens the confirm instead of firing instantly. Copy matches the issue spec, session name pulled from the breadcrumb window.
- Sidebar's inline delete confirm is untouched (explicit constraint).

## Tests
- `Modal.test.tsx`: extended with Escape→onDismiss, no-onDismiss no-op, initial focus (first focusable / panel fallback with `tabIndex=-1`), Tab trap wrap, focus restore on unmount.
- New `ConfirmModal.test.tsx`: render/hide, Cancel vs Confirm callbacks, Escape→Cancel (never Confirm), danger-tone chrome.
- New `PaletteShell.test.tsx`: Escape closes from the input (existing) and from a non-input focused row (the bug this fixes).
- `MenuItem.test.tsx`: new suite for Delete/Clear opening the confirm instead of firing immediately, Cancel not firing, session name interpolation.

Filed **BET-736** (follow-up, parked, unassigned): `Settings.tsx` has no component-mount test harness anywhere in this repo, so the nested-dialog Escape-defer fix (the trickiest part of this PR) is covered by reasoning + a green typecheck + the generic Modal/ConfirmModal tests, not a Settings-specific regression test.

## Review cycle 1 → fixes (full detail in the PR comment thread)
- **Block** — `useFocusTrap` force-focused the first focusable element and captured the opener from inside a passive `useEffect`, which runs AFTER React applies `autoFocus` during commit. For any panel that autofocuses a field (FolderPickerModal's path input), this clobbered `autoFocus` (landing on the Close button) and captured the panel's own autofocused child as the "opener", so focus-restore silently no-op'd. Fixed: opener is now captured synchronously during render (before commit/`autoFocus` can move it), and the forced initial focus is skipped entirely when something inside the panel already has focus. New `Modal.test.tsx` case with an `autoFocus` child covers both halves.
- **Question** — FolderPicker's two-stage Escape (dismiss the live suggestion first, close on the next press) was restored via a local `stopPropagation` when a suggestion is live. New `FolderPickerModal.test.tsx` covers both cases.
- **Nits** — PaletteShell's Enter no longer swallows the "ESC" chip's own click (Arrow/Enter now scoped to the search input; Escape stays overlay-scoped); `FOCUSABLE_SELECTOR` excludes disabled/hidden controls; rebased onto fresh `origin/main`.

## Verification results
- PR branch (`multica/BET-724-modal-behavior` @ `fcf549b2`, rebased onto `origin/main` @ `6b2ee744`):
  - test (`npm test` — vitest + node:test): exit `0`, vitest 2296 pass / 0 fail / 7 skip (125 files); node:test 1565 pass / 0 fail / 0 skip. log sha256: `376b469b43b1c752751a45c96fe2ecdcd7c663fcaee0f67cc7f87f8c97a4286a`
- Base freshness: `git diff --stat origin/main...HEAD` — 14 files, all additions/modifications this PR owns, zero deletions.

Logs cached at `/tmp/typecheck-pr2.log`, `/tmp/test-pr2.log` for reviewer spot-check.

### DIFF
--- a/src/renderer/ConfirmModal.test.tsx
+++ b/src/renderer/ConfirmModal.test.tsx
@@ -1,0 +1,128 @@
+// @vitest-environment jsdom
+//
+// Component tests for the shared destructive-action confirm dialog
+// (BET-724 Task 3 / D7). Built on Modal, so most dialog mechanics (Escape,
+// backdrop, focus trap) are already covered by Modal.test.tsx — this file
+// covers ConfirmModal's own contract: title/body render, Cancel/Confirm fire
+// the right callback, and Escape/backdrop route to Cancel (never Confirm).
+
+import { describe, it, expect, afterEach } from "vitest";
+import { mount, type Harness } from "./testHarness";
+import { ConfirmModal } from "./ConfirmModal";
+
+describe("ConfirmModal", () => {
+  let h: Harness | null = null;
+  afterEach(() => {
+    h?.unmount();
+    h = null;
+  });
+
+  function buttons(harness: Harness): HTMLButtonElement[] {
+    return [...harness.container.querySelectorAll("button")] as HTMLButtonElement[];
+  }
+
+  it("renders the title and body when open", () => {
+    h = mount(
+      <ConfirmModal
+        open
+        title="Delete this session?"
+        body="This can't be undone."
+        confirmLabel="Delete session"
+        onConfirm={() => {}}
+        onCancel={() => {}}
+      />,
+    );
+    expect(h.text()).toContain("Delete this session?");
+    expect(h.text()).toContain("This can't be undone.");
+  });
+
+  it("renders nothing when closed", () => {
+    h = mount(
+      <ConfirmModal
+        open={false}
+        title="Delete this session?"
+        body="This can't be undone."
+        confirmLabel="Delete session"
+        onConfirm={() => {}}
+        onCancel={() => {}}
+      />,
+    );
+    expect(h.text()).not.toContain("Delete this session?");
+  });
+
+  it("Cancel fires onCancel, not onConfirm", () => {
+    let cancelled = 0;
+    let confirmed = 0;
+    h = mount(
+      <ConfirmModal
+        open
+        title="Clear this conversation?"
+        body="The session keeps running but its context is gone."
+        confirmLabel="Clear"
+        onConfirm={() => confirmed++}
+        onCancel={() => cancelled++}
+      />,
+    );
+    const cancelBtn = buttons(h).find((b) => b.textContent === "Cancel")!;
+    expect(cancelBtn).toBeTruthy();
+    cancelBtn.click();
+    expect(cancelled).toBe(1);
+    expect(confirmed).toBe(0);
+  });
+
+  it("the confirm button fires onConfirm, not onCancel", () => {
+    let cancelled = 0;
+    let confirmed = 0;
+    h = mount(
+      <ConfirmModal
+        open
+        title="Delete this session?"
+        body="This can't be undone."
+        confirmLabel="Delete session"
+        onConfirm={() => confirmed++}
+        onCancel={() => cancelled++}
+      />,
+    );
+    const confirmBtn = buttons(h).find((b) => b.textContent === "Delete session")!;
+    expect(confirmBtn).toBeTruthy();
+    confirmBtn.click();
+    expect(confirmed).toBe(1);
+    expect(cancelled).toBe(0);
+  });
+
+  it("Escape routes to onCancel, never onConfirm (Modal's onDismiss = Cancel)", () => {
+    let cancelled = 0;
+    let confirmed = 0;
+    h = mount(
+      <ConfirmModal
+        open
+        title="Delete this session?"
+        body="This can't be undone."
+        confirmLabel="Delete session"
+        onConfirm={() => confirmed++}
+        onCancel={() => cancelled++}
+      />,
+    );
+    const dialog = h.container.querySelector('div[role="dialog"]')!;
+    dialog.dispatchEvent(new KeyboardEvent("keydown", { key: "Escape", bubbles: true }));
+    expect(cancelled).toBe(1);
+    expect(confirmed).toBe(0);
+  });
+
+  it("the confirm button uses the danger tone, Cancel uses the default tone", () => {
+    h = mount(
+      <ConfirmModal
+        open
+        title="Delete this session?"
+        body="This can't be undone."
+        confirmLabel="Delete session"
+        onConfirm={() => {}}
+        onCancel={() => {}}
+      />,
+    );
+    const cancelBtn = buttons(h).find((b) => b.textContent === "Cancel")!;
+    const confirmBtn = buttons(h).find((b) => b.textContent === "Delete session")!;
+    expect(confirmBtn.className).toContain("text-danger");
+    expect(cancelBtn.className).not.toContain("text-danger");
+  });
+});
--- a/src/renderer/ConfirmModal.tsx
+++ b/src/renderer/ConfirmModal.tsx
@@ -1,0 +1,44 @@
+// ConfirmModal.tsx — the shared confirm dialog for destructive actions
+// reached from a menu (BET-724 Task 3 / design decision D7).
+//
+// Built on the Modal primitive, which now owns Escape/backdrop/focus trap
+// (BET-724 Task 1) — so Escape and backdrop click both mean Cancel here,
+// never Confirm. Typography matches the two pre-existing in-app confirms
+// (App.tsx's "Update the box?", Settings.tsx's "Remove box?" / "Reset all
+// settings?") rather than inventing a new type scale.
+
+import { Modal } from "./Modal";
+import { Button } from "./Button";
+
+export function ConfirmModal({
+  open,
+  title,
+  body,
+  confirmLabel,
+  onConfirm,
+  onCancel,
+}: {
+  open: boolean;
+  title: string;
+  body: string;
+  confirmLabel: string;
+  onConfirm: () => void;
+  onCancel: () => void;
+}) {
+  return (
+    <Modal size="sm" open={open} onDismiss={onCancel} label={title}>
+      <div className="space-y-4">
+        <h3 className="text-title font-semibold">{title}</h3>
+        <div className="text-body text-text-faint">{body}</div>
+        <div className="flex justify-end gap-2">
+          <Button tone="default" onClick={onCancel}>
+            Cancel
+          </Button>
+          <Button tone="danger" onClick={onConfirm}>
+            {confirmLabel}
+          </Button>
+        </div>
+      </div>
+    </Modal>
+  );
+}
--- a/src/renderer/FolderPickerModal.test.tsx
+++ b/src/renderer/FolderPickerModal.test.tsx
@@ -1,0 +1,100 @@
+// @vitest-environment jsdom
+//
+// Regression test for BET-724 review cycle 1's Question: FolderPickerModal's
+// path input used to dismiss a live inline path-completion suggestion on the
+// FIRST Escape press and only close the dialog on a SECOND, suggestion-free
+// press. Deleting the hand-rolled handler (so Modal's own Escape ownership
+// takes over unconditionally) silently lost that two-stage behavior. This
+// restores it via a local `stopPropagation` when a suggestion is live, and
+// pins the restored behavior here.
+
+import { describe, it, expect, afterEach } from "vitest";
+import { act } from "react";
+import { installMockApi, mount, type Harness } from "./testHarness";
+import { FolderPickerModal } from "./FolderPickerModal";
+
+function typeInto(el: HTMLInputElement, value: string) {
+  const setter = Object.getOwnPropertyDescriptor(
+    window.HTMLInputElement.prototype,
+    "value",
+  )?.set;
+  setter?.call(el, value);
+  el.dispatchEvent(new Event("input", { bubbles: true }));
+}
+
+describe("FolderPickerModal — Escape (BET-724 review cycle 1 Question)", () => {
+  let h: Harness | null = null;
+  afterEach(() => {
+    h?.unmount();
+    h = null;
+  });
+
+  it("first Escape dismisses a live suggestion; second Escape closes the dialog", async () => {
+    let cancelled = 0;
+    installMockApi({
+      fsListDirs: (dir: unknown) =>
+        dir === "/home/dev/pro"
+          ? Promise.resolve(["/home/dev/projects"])
+          : Promise.resolve(["/home/dev"]),
+      gitListWorktrees: () => Promise.reject(new Error("not a repo")),
+    });
+    h = mount(
+      <FolderPickerModal
+        open
+        initialPath="/home/dev"
+        onSelect={() => {}}
+        onCancel={() => cancelled++}
+      />,
+    );
+    await h.flush();
+
+    const input = h.container.querySelector("input") as HTMLInputElement;
+    expect(input).toBeTruthy();
+
+    act(() => {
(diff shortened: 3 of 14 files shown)


antoinedc:
# Review — BET-724 (cycle 1)

**Outcome: 1 Block, 1 Question, 3 Nits — returned to `manta-dev`.**

## Mechanical gate

Checked out `refs/pull/743/head` @ `b03fcfa1`, `npm ci` on Node 22.22.2.

| Step | Result |
| --- | --- |
| `npm test` | exit `0` — vitest **2286 pass / 0 fail / 7 skip** (124 files), node:test **1565 pass / 0 fail** |

Matches the PR body's reported numbers exactly. Base-freshness gate: `git diff --stat origin/main...HEAD` shows 13 files, all additions/modifications, **zero deletions of files the issue didn't name** — no stale-base phantom. `mergeable: MERGEABLE`, no overlap with what landed on `main` since the branch was cut.

Anti-spaghetti (jscpd, changed scope): 1 clone, both halves inside `NewSessionScreen.tsx` at lines this PR never touched (the `normalizeCreate` → project/window lookup pair). Pre-existing, not introduced here — no finding.

| # | Criterion | Verdict |
| --- | --- | --- |
| T1.1 | Modal owns Escape when `onDismiss` is provided, overlay-scoped | ✅ `Modal.tsx:72-79` + `Modal.test.tsx` |
| T1.2 | Focus trap + initial focus + restore, lifted from Settings' `useDialog` | ⚠️ **Block 1** — trap/restore are wrong for any panel that `autoFocus`es |
| T1.3 | Backdrop unchanged; header comment retired | ✅ |
| T2.1 | Delete hand-rolled Escape at call sites | ✅ `FolderPickerModal`, `ModelsCard`; `App.tsx:1451` verified to already carry `onDismiss` with no competing handler; `Sidebar`'s palette is `PaletteShell`-based and its remaining `Escape` (`Sidebar.tsx:1301`) is the inline rename input, correctly untouched |
| T2.2 | Palette/FolderPicker Escape fires regardless of focus | ✅ (see Nit 1) |
| T2.3 | Settings confirms get `onDismiss` | ✅ `Settings.tsx:923,935` |
| T2.4 | `useDialog` consumes the shared trap; duplicate deleted | ✅ |
| T2.5 | Esc-with-confirm-open closes ONLY the confirm | ✅ mechanism verified by reading — Settings' root carries `role="dialog"` (`Settings.tsx:825`) so the new `closest()` bail is well-formed for both the inline-confirm and the portal'd model-edit cases. Your correction to the spec (capture-phase listener means `stopPropagation` alone can't do it) is right. |
| T2.6 | `ArtifactsPanel`/`ArtifactPreview` preview restore untouched | ✅ |
| T3.1 | `ConfirmModal.tsx` on Modal + Button, no new Button tone | ✅ type-role classes match the pre-existing confirms (`text-title font-semibold` / `text-body text-text-faint`) — good call over the issue's illustrative `text-meta`/`text-text-muted` |
| T3.2 | ⋯ menu Delete + Clear confirm first, copy per spec, sidebar untouched | ✅ + covered in `MenuItem.test.tsx` |
| V | `npm run typecheck && npm test` green | ✅ (above) |

---

## **Block (fix-here):** `useFocusTrap` overrides `autoFocus`, and focus-restore silently no-ops, for any Modal whose content autofocuses

`useFocusTrap` captures the opener and calls `firstFocusable.focus()` inside a **passive effect**. React applies `autoFocus` during the commit/layout phase, which always runs *before* passive effects. So for a panel that autofocuses a field, the trap runs second and clobbers it — and by then `document.activeElement` is already inside the panel, so the "opener" it records is the panel's own field.

This is not theoretical. I ran it against this branch's `Modal`:

```tsx
<Modal open label="L" onDismiss={() => {}}>
  <button aria-label="Close">x</button>
  <input autoFocus />
</Modal>
```

- initial focus → `BUTTON` (`aria-label="Close"`), **not** the input.
- open from a real opener button, then unmount → `document.activeElement` is `<body>`, **not** the opener.

Production victim is `FolderPickerModal.tsx`: its panel header renders the `aria-label="Close"` button (`:246`) *before* the path `<input autoFocus>` (`:312`). Net effect of this PR on that dialog:

1. Opening the folder picker now focuses the **Close (X)** button instead of the path field. The user could previously type a path immediately; now they must click or Tab first. That's a functional regression on the new-session / change-folder flow.
2. The focus-restore this PR is adding never fires for that dialog — closing it drops focus to `<body>`.

Neither is caught by `Modal.test.tsx`, because every case there uses plain `<button>` children with no `autoFocus`.

The acceptance criterion is "focus starts inside the panel", and `autoFocus` already satisfies that — so the fix is a guard, not a redesign. Two things need to be true:

- **Initial focus:** if focus is already inside the panel when the trap activates, leave it alone.
- **Opener capture:** record `document.activeElement` at a point where the panel is not yet in the DOM (the render pass in which `open` flips true, or a ref written before mount) — capturing it in the passive effect is too late whenever the panel autofocuses.

Please add a `Modal.test.tsx` case with an `autoFocus` child asserting both (focus stays on the autofocused element; unmount restores to the opener).

---

## **Question:** FolderPicker's Escape lost its two-stage behaviour — intended?

`FolderPickerModal.tsx:337-340` previously did `if (suggestion) setSuggestion(null); else onCancel()` — Escape dismissed the live ghost-text completion first, and only closed the dialog on a second press. The handler is now deleted outright, so the first Escape closes the whole picker with a completion showing.

The issue does say to delete the hand-rolled handler and make Escape close the dialog from any focus, so this may well be the intended reading. But it's a silent loss on a keyboard-heavy flow, and Modal now owns Escape so the input can't reclaim the first press without deliberately stopping propagation. Please either (a) confirm it's intended and say so in the code comment, (b) restore the two-stage form via an explicit `stopPropagation` when a suggestion is live, or (c) defer with rationale — then reassign back for the re-check.

---

## Nits (non-blocking)

1. **`PaletteShell`: Enter on the ESC chip now picks instead of closing.** Moving `onKeyDown` from the input to the overlay means the handler's `e.preventDefault()` on Enter suppresses the `<button>`'s synthesized click, so `pick(sel)` runs instead of `requestClose()`. Only reachable by Tabbing from the input to the chip, but it's a new behaviour the move introduced.
2. **`useFocusTrap`'s `FOCUSABLE_SELECTOR` matches disabled and hidden controls.** `button` includes `[disabled]`; focusing a disabled element is a silent no-op that leaves focus on `<body>` and defeats the trap. Inherited verbatim from Settings' `useDialog`, so not a regression — but this is the moment it becomes the app-wide implementation.
3. **PR body says the branch is "freshly cut off current `origin/main`".** It's cut at `901c06d6`; `origin/main` is now `6b2ee744` (7 commits ahead). Harmless here — no overlap, no conflict, and I re-ran the suite myself — but the "skip the base run" rationale rests on a claim that isn't accurate as written.

Filing BET-736 for the missing Settings mount harness rather than blocking on it was the right call.

antoinedc:
Addressed the Block + Question + all 3 nits. Rebased onto fresh `origin/main` @ `6b2ee744` (fixes nit 3's stale base-freshness claim too) — `fcf549b2`.

## Block — `useFocusTrap` clobbered `autoFocus` / captured the wrong opener

Fixed as diagnosed:
- **Opener capture** moved out of the passive `useEffect` into the render body itself (a ref compared/updated on the `active` false→true transition, guarded so it only fires once per open) — render always runs before commit/`autoFocus`, so `document.activeElement` read there is still the real pre-open opener.
- **Initial focus** now bails entirely when `root.contains(document.activeElement)` is already true when the effect runs — an `autoFocus`ed child already satisfies "focus starts inside the panel"; forcing a different element there was the regression.

Added the requested `Modal.test.tsx` case (`<input autoFocus>` after a `<button>`): asserts focus stays on the input (not the earlier button) and unmount restores the real opener, not the autofocused input. Ran it against the pre-fix code first to confirm it actually fails there, then against the fix to confirm green.

## Question — FolderPicker's two-stage Escape

Went with option (b): restored it. The path input's `onKeyDown` now does `stopPropagation()` + clears the suggestion on the first Escape when one is live, so Modal's overlay handler never sees that press; a second, suggestion-free Escape bubbles up and closes normally. New `FolderPickerModal.test.tsx` covers both the two-stage case and the unchanged (no-suggestion → closes on first press) case.

## Nits

1. **PaletteShell Enter-on-ESC-chip** — Arrow/Enter navigation now scoped to `e.target === inputRef.current`; Escape stays overlay-scoped (unconditional). New test pins it (Enter on the chip no longer picks; Enter on the input still does).
2. **`FOCUSABLE_SELECTOR` disabled/hidden controls** — added `:not([disabled]):not([hidden])` across the selector (also excludes `hidden` links/tabbables, not just disabled).
3. **Base-freshness claim** — rebased onto current `origin/main` (`6b2ee744`, was `901c06d6`); the PR body's "freshly cut" line is now accurate again.

## Verification results
- PR branch (`multica/BET-724-modal-behavior` @ `fcf549b2`):
  - test: exit `0` — vitest **2296 pass / 0 fail / 7 skip** (125 files), node:test **1565 pass / 0 fail**. log sha256: `376b469b43b1c752751a45c96fe2ecdcd7c663fcaee0f67cc7f87f8c97a4286a`
- Base freshness: `git diff --stat origin/main...HEAD` — 14 files, all additions/modifications to files this PR owns, no deletions.

antoinedc:
# Review — BET-724 (cycle 2, re-check)

**Outcome: PASS — 0 Blocks, 0 Questions, 1 Nit. Routing to `manta-pm` for merge.**

## Mechanical gate

Re-checked out `refs/pull/743/head` @ `fcf549b2`, clean `npm ci` on Node 22.22.2.

| Step | Result |
| --- | --- |
| `npm test` | exit `0` — vitest **2296 pass / 0 fail / 7 skip** (125 files), node:test **1565 pass / 0 fail** |

Matches the reported numbers. Rebase confirmed: `merge-base HEAD origin/main` == `origin/main` == `6b2ee744`, `mergeable: MERGEABLE`.

Scoped the re-review with `git diff b03fcfa1..HEAD`: the only source files touched since cycle 1 are the four I flagged (`useFocusTrap.ts`, `FolderPickerModal.tsx`, `PaletteShell.tsx`) plus their tests. `Settings.tsx`, `SessionHeader.tsx`, `ConfirmModal.tsx`, `ModelsCard.tsx`, `NewSessionScreen.tsx` are byte-identical to what I already passed in cycle 1 — no regressions smuggled in with the fix.

## Cycle-1 findings — verified fixed

### Block: `useFocusTrap` clobbering `autoFocus` + capturing the wrong opener — **fixed, verified by running it**

The fix is the right shape: capture the opener synchronously during render on the false→true transition (render always precedes commit, so `document.activeElement` is still the real opener there), and skip the forced initial focus entirely when something inside the panel already holds it. The guard is idempotent under a double render, so StrictMode won't re-capture a stale opener.

I re-ran the exact scenarios that failed in cycle 1, against this HEAD:

| Scenario | Cycle 1 | Cycle 2 |
| --- | --- | --- |
| Panel = `<button aria-label="Close">` + `<input autoFocus>` → initial focus | `BUTTON` ❌ | `INPUT` ✅ |
| …then unmount → focus restored to | `<body>` ❌ | the opener ✅ |
| `open` false→true (no autoFocus) → unmount restores | — | the opener ✅ |
| Panel with two plain buttons → initial focus | `first` ✅ | `first` ✅ (unregressed) |

So the folder picker gets its path field back on open, and restore now actually fires. `Modal.test.tsx`'s new `autoFocus` case pins both halves.

### Question: FolderPicker's two-stage Escape — **resolved by restoring it**

Local `stopPropagation()` only while a suggestion is live is the correct mechanism now that Modal owns Escape — the second, suggestion-free press bubbles through as normal. `FolderPickerModal.test.tsx` is new and covers both the two-press path and the unchanged single-press path; that file didn't exist before, so this is net new coverage on a dialog that had none.

### Nits — all three fixed

1. `PaletteShell`: Arrow/Enter now scoped to the search input (`e.target !== inputRef.current` bail) while Escape stays overlay-scoped, so the ESC chip's own click is no longer swallowed. Covered by a new test asserting Enter-on-chip doesn't pick and Enter-in-input does.
2. `FOCUSABLE_SELECTOR` now excludes `[disabled]`/`[hidden]`.
3. Rebased onto `6b2ee744`.

## Nit (non-blocking, do not hold the merge for it)

The advisory duplication report went 1 → 2 clones. The new one is entirely inside the new `FolderPickerModal.test.tsx`: the `installMockApi` + `mount(<FolderPickerModal …>)` + `await h.flush()` + grab-the-input preamble is copy-pasted across the two `it()` blocks (14 lines, 76 tokens). It's test-fixture boilerplate, not business logic, so it isn't a finding under the anti-spaghetti rules — a shared `renderPicker()` helper would just be tidier. The other clone is the pre-existing `NewSessionScreen.tsx` pair at lines this PR never touched.

## Verdict

All 13 acceptance-criteria items met and independently verified; typecheck and the full suite green; every cycle-1 finding closed with a regression test rather than a claim. Handing off to `manta-pm`.

PR #746, opened 2026-08-13
Desktop: keyboard & focus sweep — scrollIntoView, sidebar action keys, tablist arrows, focus-visible gaps

antoinedc:
BET-726 — audit §6 items 5+8+15-parts.

**Base:** `origin/main` @ `bfd3ab8f`

## Task 1 — scrollIntoView for the keyboard-highlighted item

Deviated from the issue's literal snippet (a new `scrollHighlight.ts` with a
`data-hl-index` query helper): the repo already has a hook that does exactly
this — `useSelectedIntoView` in `PaletteShell.tsx`, used by the ⌘K/⌘F
palettes' `SessionRow`. Reused it instead of adding a second scrollIntoView
idiom (the issue's own Constraints section asks for exactly that: "zero new
interaction idioms").

- `MenuOption.tsx` (ModelMenu + EffortMenu rows): calls `useSelectedIntoView(active)`
  internally and attaches the ref to its own `<button>`.
- `ComposerParts.tsx`'s `TypeaheadPopup`: extracted a `TypeaheadRowButton`
  (Hooks can't be called from inside `.map()`, so each row needed to become
  its own component — same shape as `SessionRow`) using the same hook.
- The ⌘K command palette (`Sidebar.tsx`'s `SessionRow`) was **already fixed**
  — it already calls `useSelectedIntoView`. The issue's audit predates that.
- Hardened the shared hook itself: `ref.current?.scrollIntoView?.(...)` —
  jsdom doesn't implement `scrollIntoView`, and this is the first test
  coverage to actually render a `selected`/`active` row through it.

## Task 2 — sidebar row actions get keyboard paths

`Sidebar.tsx`'s tree keydown handler (`onRailKeyDown`) gains:
- `F2` → the same rename entry point double-click uses.
- `Delete`/`Backspace`/`ContextMenu` → the same inline `ConfirmDelete`
  right-click opens. (Backspace added alongside Delete: the Mac keyboard's
  only "delete" key reports as `Backspace` in the DOM. `ContextMenu` maps to
  the same action since right-click has no separate menu to open either —
  it goes straight to the confirm.)
- `Space` alongside `Enter` to activate the focused row (`SessionRow` is a
  `treeitem`, not a `<button>`, so Space-activation isn't free).

Hover-revealed action buttons (pin, GroupHeader `+`/`X`, and — one more
instance of the identical pattern the issue didn't name — the new-session
draft's dismiss `X`) get `focus-visible:opacity-100` and drop `tabIndex={-1}`
so a keyboard user tabbing to them can see and reach them. Added the one
missing `title` (draft dismiss); the other three already had one.

## Task 3 — menus + tablists

- `SessionHeader.tsx`'s ⋯ `SessionMenu`: `ArrowUp`/`ArrowDown` roving
  highlight + `Home`/`End` + `Enter` to activate, using the same index-state
  shape as ModelMenu's highlight loop (`chatUtils.moveMenuHighlight`),
  adapted from ModelMenu's search input to this menu's root div (no input to
  hang the keydown on here). `MenuItem` gained a `highlighted` prop — the
  same static `bg-fill-hover` fill `MenuOption`'s `active` already gives the
  model/effort menus. Escape already closed the menu via `useClickAway`
  (verified, not re-added).
- `ArtifactsPanel.tsx`'s tab bar: `Left`/`Right` roving-tabindex nav — ported
  Settings.tsx's exact section-rail idiom (move state + `.focus()` the next
  tab's ref). The search input's bare `focus:outline-none` now also sets
  `focus:border-accent`, matching the `Field` primitive's own focus
  treatment (a text input, so the border-color convention fit better than
  the button ring used elsewhere in this issue).

## Task 4 — focus-visible on the remaining orphans

Ring added to: `Toast.tsx` (action + dismiss), `UpdateBar.tsx` (action +
dismiss), `Cards.tsx`'s permission-card ladder (Allow once / Always allow /
Reject) and question-card ladder (Submit / Dismiss) — ring only, no Button
primitive migration — and the Artifacts tab buttons (added alongside Task 3.2).

## Verification results

- PR branch (`multica/BET-726-keyboard-sweep` @ `12bd9c70`):
  - test: exit `0`, vitest 2301 pass / 0 fail / 7 skip, node:test 1565 pass / 0 fail. log sha256: `b712ce5099961729c6183eb2d14655e946350de0cd7ebfeb97eb504ef89095c0`
- Base (`main` @ `bfd3ab8f`):
  - test: exit `0`, vitest 2296 pass / 0 fail / 7 skip, node:test 1565 pass / 0 fail. log sha256: `05b84c3f7c0030d01bbc77b0c3b0415d7284f96745b9f521d8f28f09d14b4aa8`
- **Conclusion:** 0 test failures on either branch (identical typecheck output — the logs' hashes match). +5 new vitest tests added by this PR (2301 vs 2296), 0 pre-existing failures, 0 new failures.

Also ran the `manta-e2e-smoke` gate against the built app (Playwright's Electron
launcher, xvfb): main process alive, sidebar (`aside.w-64`) present with the
Workspace heading + new-project button, `<main>`/titlebar present, zero
console/page errors. PASSED.

## Manual verification checklist (from the issue)

- [x] Arrow-key past the fold in ModelMenu keeps the highlight visible (`useSelectedIntoView` on `MenuOption`).
- [x] ⌘K palette: already covered (pre-existing `useSelectedIntoView` on `SessionRow`).
- [x] @-typeahead: covered via the new `TypeaheadRowButton`.
- [x] Tab to a sidebar row: F2 renames, Delete/Backspace opens the inline confirm, Space/Enter activate; Tab reaches pin/+/X and they become visible on focus.
- [x] ⋯ menu: arrows cycle (with wraparound via `moveMenuHighlight`), Enter picks, Home/End jump to first/last.
- [x] Artifacts tabs: Left/Right work (roving tabindex + focus move); search field shows a focus ring (`focus:border-accent`).

## Constraints honored

- Reused named existing implementations (Settings tablist idiom, ModelMenu's
  highlight-loop shape, sidebar's `ConfirmDelete`, and — better than the
  issue's own suggestion — the palettes' existing `useSelectedIntoView`
  hook). Zero new interaction idioms.
- Did not migrate `Cards.tsx` buttons to the `Button` primitive — ring only.

### DIFF
--- a/src/renderer/ArtifactsPanel.tsx
+++ b/src/renderer/ArtifactsPanel.tsx
@@ -466,10 +466,13 @@ export function ArtifactsPanel({
   widthRef.current = width;
 
   const [tab, setTab] = useState<ArtifactKind>("link");
-  // BET-726 Task 3.2: roving-tabindex arrow nav on the tab bar — the exact
-  // pattern Settings.tsx's section rail already uses (`onRailKeyDown` there:
-  // move the active tab AND focus, on Left/Right). Refs keyed by kind so
-  // ArrowLeft/Right can call `.focus()` on the tab it moves to.
+  // BET-726 Task 3.2: arrow-key nav on the tab bar — the exact pattern
+  // Settings.tsx's section rail already uses (`onRailKeyDown` there: move
+  // the active tab state AND DOM focus, on Left/Right). Note this is NOT a
+  // roving-tabindex widget (neither this port nor Settings' original manages
+  // `tabIndex` — every tab stays in the normal Tab order); only the active
+  // tab and focus move on arrow keys. Refs keyed by kind so ArrowLeft/Right
+  // can call `.focus()` on the tab it moves to.
   const tabRefs = useRef<Record<ArtifactKind, HTMLButtonElement | null>>({
     link: null,
     image: null,
--- a/src/renderer/ArtifactsPanel.tsx
+++ b/src/renderer/ArtifactsPanel.tsx
@@ -466,6 +466,15 @@ export function ArtifactsPanel({
   widthRef.current = width;
 
   const [tab, setTab] = useState<ArtifactKind>("link");
+  // BET-726 Task 3.2: roving-tabindex arrow nav on the tab bar — the exact
+  // pattern Settings.tsx's section rail already uses (`onRailKeyDown` there:
+  // move the active tab AND focus, on Left/Right). Refs keyed by kind so
+  // ArrowLeft/Right can call `.focus()` on the tab it moves to.
+  const tabRefs = useRef<Record<ArtifactKind, HTMLButtonElement | null>>({
+    link: null,
+    image: null,
+    file: null,
+  });
   const [searchOpen, setSearchOpen] = useState(false);
   const [query, setQuery] = useState("");
   // The preview overlay: index into `previewable` (the current tab's
@@ -577,6 +586,16 @@ export function ArtifactsPanel({
     else void downloadArtifact(a);
   };
 
+  const onTabKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>) => {
+    if (e.key !== "ArrowLeft" && e.key !== "ArrowRight") return;
+    e.preventDefault();
+    const idx = TABS.indexOf(tab);
+    const dir = e.key === "ArrowRight" ? 1 : -1;
+    const next = TABS[(idx + dir + TABS.length) % TABS.length];
+    setTab(next);
+    tabRefs.current[next]?.focus();
+  };
+
   // Jump the transcript to the message that owns an artifact.
   const jumpToMessage = (a: Artifact) => {
     if (!a.messageId) return;
@@ -655,7 +674,7 @@ export function ArtifactsPanel({
               value={query}
               onChange={(e) => setQuery(e.target.value)}
               placeholder="Search links & files…"
-              className="mb-3 w-full bg-bg border border-border px-2 py-1 text-meta rounded-xs focus:outline-none placeholder:text-text-faint"
+              className="mb-3 w-full bg-bg border border-border px-2 py-1 text-meta rounded-xs focus:outline-none focus:border-accent placeholder:text-text-faint"
             />
           )}
 
@@ -683,12 +702,15 @@ export function ArtifactsPanel({
               return (
                 <button
                   key={k}
+                  ref={(el) => { tabRefs.current[k] = el; }}
                   type="button"
                   role="tab"
                   aria-selected={active}
                   onClick={() => setTab(k)}
+                  onKeyDown={onTabKeyDown}
                   className={
                     "relative z-10 flex-1 inline-flex items-center justify-center gap-1 px-2 py-1 rounded-sm text-meta font-medium " +
+                    "focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent " +
                     (active ? "text-text" : "text-text-faint hover:text-text")
                   }
                 >
--- a/src/renderer/Cards.tsx
+++ b/src/renderer/Cards.tsx
@@ -211,7 +211,7 @@ export function PermissionCard({
     <>
       <button
         onClick={() => onReply("once")}
-        className="h-8 px-[14px] inline-flex items-center gap-2 rounded-md text-on-accent text-meta font-medium"
+        className="h-8 px-[14px] inline-flex items-center gap-2 rounded-md text-on-accent text-meta font-medium focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent"
         style={{ backgroundColor: "var(--accent-solid)" }}
       >
         <Check size={13} aria-hidden="true" />
@@ -224,7 +224,7 @@ export function PermissionCard({
         <button
           onClick={() => onReply("always")}
           title={`Always allow ${alwaysScope}`}
-          className="h-8 px-[14px] inline-flex items-center rounded-md border border-border-strong text-text hover:bg-bg-soft text-meta"
+          className="h-8 px-[14px] inline-flex items-center rounded-md border border-border-strong text-text hover:bg-bg-soft text-meta focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent"
         >
           Always allow&nbsp;
           <span className="text-text-faint">{alwaysScope}</span>
@@ -234,7 +234,7 @@ export function PermissionCard({
       <div className="flex-1" />
       <button
         onClick={() => onReply("reject")}
-        className="h-8 px-[14px] inline-flex items-center rounded-md text-text-faint hover:text-danger hover:bg-danger-bg text-meta"
+        className="h-8 px-[14px] inline-flex items-center rounded-md text-text-faint hover:text-danger hover:bg-danger-bg text-meta focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent"
       >
         Reject
       </button>
@@ -484,7 +484,7 @@ export function QuestionCard({
           <button
             onClick={handleSubmit}
             disabled={!canSubmit}
-            className="h-8 px-[14px] inline-flex items-center gap-2 rounded-md text-on-accent text-meta font-medium disabled:opacity-40"
+            className="h-8 px-[14px] inline-flex items-center gap-2 rounded-md text-on-accent text-meta font-medium disabled:opacity-40 focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent"
             style={{ backgroundColor: "var(--accent-solid)" }}
           >
             Submit
@@ -502,7 +502,7 @@ export function QuestionCard({
           <div className="flex-1" />
           <button
             onClick={onReject}
-            className="h-8 px-[14px] inline-flex items-center rounded-md text-text-faint hover:text-danger hover:bg-danger-bg text-meta"
+            className="h-8 px-[14px] inline-flex items-center rounded-md text-text-faint hover:text-danger hover:bg-danger-bg text-meta focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent"
             title="Dismiss this question"
           >
             Dismiss
--- a/src/renderer/ComposerParts.tsx
+++ b/src/renderer/ComposerParts.tsx
@@ -11,6 +11,10 @@ import type { VoiceMode, VoicePhase } from "./voice";
 import { type Attachment, type TypeaheadRow } from "./chatShared";
 import { ALT_KEY } from "./platform";
 import { IconButton } from "./IconButton";
+// BET-726 Task 1: same scrollIntoView idiom the ⌘K / ⌘F palettes and the
+// model/effort menus use (PaletteShell.tsx) — keeps the keyboard-selected
+// @-file row visible when it moves past the popup's scroll fold.
+import { useSelectedIntoView } from "./PaletteShell";
 
 // SessionToolbar — footer affordances. fork / compact / delete moved out of the
 // footer (they live in the header ⋯ menu); only the ⏰ schedules toggle remains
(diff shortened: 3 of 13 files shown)


antoinedc:
## Review — BET-726 (cycle 1)

Reviewed at `12bd9c70` against `origin/main` @ `bfd3ab8f`. Base is fresh (`git merge-base HEAD origin/main` == `origin/main` tip), diff is 11 files / +376 / -62, all inside the issue's stated scope.

### Mechanical gate

| gate | result |
|---|---|
| `npm test` | exit `0` — vitest 125 files, **2301 pass / 7 skip / 0 fail**; node:test **1565 pass / 0 fail** |
| `duplication-gate` (CI) | **pass** |

Numbers match the PR body exactly, including the `+5` new vitest cases. No re-run on base needed.

Each verified against the code, not against the PR description.

- **Task 1 — scrollIntoView, all three sites.** ✅ ModelMenu + EffortMenu via `MenuOption.tsx:62`; @-file typeahead via the new `TypeaheadRowButton` (`ComposerParts.tsx:246`); ⌘K palette independently confirmed already covered at `Sidebar.tsx:1326` — the "already fixed" claim checks out, I did not take it on faith. The deviation from the issue's literal `scrollHighlight.ts` snippet is the right call: reusing `useSelectedIntoView` is what the issue's own "zero new interaction idioms" constraint asks for.
- **Task 2 — sidebar keyboard paths.** ⚠️ F2/Delete/ContextMenu/Space are all wired, and `ContextMenu` → the delete confirm is correct because right-click genuinely does exactly that (`Sidebar.tsx:971` overrides `onContextMenu` → `onRequestDelete`, there is no separate menu). `tabIndex={-1}` removal + `focus-visible:opacity-100` applied to all four hover-revealed buttons. **But see Block 1 — the buttons are now reachable and not operable.**
- **Task 3.1 — ⋯ menu arrows.** ✅ ArrowUp/Down/Home/End/Enter present, using `moveMenuHighlight`. Escape: verified pre-existing at `hooks/useClickAway.ts:24-26` (document keydown, gated on `active`) — correctly not re-added. See Question 1 on the a11y half.
- **Task 3.2 — Artifacts tablist + search ring.** ✅ `onTabKeyDown` is the same shape as `Settings.tsx:554` (index math + `.focus()` on the moved-to ref), narrowed to Left/Right for a horizontal bar — correct port. The search field's `focus:border-accent` is **exactly** the `Field` primitive's own focus treatment (`Field.tsx:66` `INPUT_BASE`), which the issue explicitly permitted as the alternative to the button ring. Good choice for a text input.
- **Task 4 — focus-visible orphans.** ✅ Toast (×2), UpdateBar (×2), `PermissionCard` ladder (×3), `QuestionCard` ladder (×2), artifacts tabs. Ring only — no `Button` primitive migration, constraint honoured.
- **Tests.** ✅ 5 new cases in `MenuItem.test.tsx` covering the roving highlight, wraparound, Home/End and the reopen-reset. Confirmed there is no `Sidebar.test.tsx` / `ArtifactsPanel` test to extend (`SessionRow.test.tsx` is the chrome *primitive*, unrelated), so the issue's "otherwise typecheck+test green is the gate" branch applies.

---

## Block (fix-here) 1 — the newly-tabbable sidebar buttons are reachable but dead, and can mis-fire the delete confirm

`onRailKeyDown` is bound to the `role="tree"` container (`Sidebar.tsx:570`) and switches on `e.key` with **no check of the event target**. That was safe before this PR because the only focusable things inside the tree were the treeitem rows themselves (`<div role="treeitem" tabIndex={focused ? 0 : -1}>`) — every hover-revealed button carried `tabIndex={-1}`.

Task 2.3 removed those `tabIndex={-1}`s (pin `:928`, GroupHeader `+` `:1230` / `X` `:1240`, draft dismiss `:602`). All four are descendants of that container, so their keydowns now bubble into the rail handler:

1. **Enter (and Space) on a focused pin / `+` / `X` hits `case "Enter": case " ": e.preventDefault(); activateFocused()`.** The `preventDefault()` cancels the button's own activation, so the button never fires — and the rail activates the roving-focused *row* instead. Net effect: this PR makes those four buttons Tab-reachable and simultaneously makes them impossible to press. Task 2's stated goal ("keyboard paths for sidebar actions") is not met for exactly the buttons it opened up.
2. **Delete / Backspace while one of those buttons is focused opens `ConfirmDelete` for `focusedKey`.** DOM focus and `focusedKey` are independent state — the roving tabindex means Tab walks through *every* row's pin button while `focusedKey` stays where the arrow keys left it. So Backspace on row 5's pin can raise "Close session <row 2>?". Behind a confirm, but it names the wrong target.
3. **ArrowUp/Down while a button is focused** moves `focusedKey` while DOM focus stays on the button — the visible tree highlight and real focus desync.

`RenameInput` already solves this for itself with `e.stopPropagation()` (`Sidebar.tsx:1363`); the container needs the mirror-image guard. One line at the top of `onRailKeyDown`:

```ts
// Nested controls (pin, group +/X, draft dismiss) are tabbable now — let
// their own keydown handling run instead of the tree's.
if ((e.target as HTMLElement).closest("button")) return;
```

Rows are `<div role="treeitem">`, not buttons, so this leaves arrow/Enter/F2/Delete navigation on the rows untouched. Please add a case for it — a `Sidebar.test.tsx` asserting (a) Enter on a pin button toggles the pin and does not activate the focused row, and (b) Backspace on a pin button does not open the confirm, would cover the whole class.

---

## Question 1 — the ⋯ menu's roving highlight is invisible to assistive tech

`SessionMenu`'s highlight is purely visual: `MenuItem` rows have `role="menuitem"` but no `id`, and the `role="menu"` surface carries no `aria-activedescendant`. Arrowing through the menu therefore announces nothing to a screen reader — DOM focus stays on the ⋯ trigger the whole time.

The idiom this task was told to copy does have that half: `MenuOption` takes an `id` documented as "used as the aria-activedescendant target", and ModelMenu wires it. Since BET-726 is specifically a keyboard-and-focus accessibility sweep, adding a visual-only highlight to a menu widget reads like half the fix.

Either wire it (an `id` prop on `MenuItem` + `aria-activedescendant` on the `Dropdown` surface — small, and it completes the copied idiom), or reply with why it is deliberately out of scope here. Both are acceptable answers; I just don't want it decided silently.

---

## Nits (no action required)

- **`renameFocused` on a `job:` row is not the no-op its comment claims.** `Sidebar.tsx:361` says "job rows never were [renameable] ... so they're a no-op here too", but `resolveFocusedWindow` treats `job:` identically to `win:`, so F2 on a job row calls `startRename({kind:"window", index: childIdx, ...})`. `JobChildRow` renders no `RenameInput`, so nothing appears and the rename target is left dangling until the next rename overwrites it. Harmless in practice — but either early-return on `job:` or fix the comment.
- **"roving tabindex" is the wrong name** in the PR body and in `ArtifactsPanel.tsx`'s new comment. Neither the port nor `Settings.tsx:554` manages `tabIndex` — all tabs stay in the tab order; only the active tab and focus move. The *behaviour* matches Settings as instructed, so the code is right and only the wording is off.

---

## Note for the merge owner (not a Block, not fixable on this branch)

CI's `typecheck-test` job is red, but only at its **gitleaks** step — the typecheck and test steps inside that job passed. I verified the implementer's assessment independently:

```
$ git merge-base --is-ancestor f38ddabdcc853cd0e0b0243a19a3920dfdcbb9a6 HEAD
→ NOT in PR history
$ git branch -r --contains f38ddabd
→ origin/pr-744-c2
$ grep -rn "sk-live-abc123" src/
→ (no match in this branch's worktree)
```

The step runs `gitleaks detect --source .` over the whole fetched object graph rather than the PR's diff, so a test fixture on an unrelated branch turns **every** open PR red for as long as that commit exists on any remote ref. Nothing this branch can do about it. Worth its own issue: *"CI: scope the gitleaks secret scan to the PR's diff (`--log-opts origin/main..HEAD`) or baseline known fixtures in `.gitleaks.toml` — today a secret-shaped string on any remote branch fails the secret-scan step on every unrelated PR."*

---

**Verdict: 1 Block + 1 Question — returning to `manta-dev`.** Everything else in the sweep is solid; the scrollIntoView reuse and the Settings tablist port are both better than the issue's literal instructions.

antoinedc:
Review cycle 1 addressed at `a98f5d3f`.

## Block (fix-here) — nested-control keydown guard

Fixed exactly as suggested: `onRailKeyDown` now bails out immediately when
the keydown's target is (or is inside) a `<button>`, since rows are
`<div role="treeitem">` and never buttons — so the guard can't swallow the
row's own arrow/Enter/F2/Delete handling. New `Sidebar.test.tsx` (the first
one this file has ever had) covers:
- Enter on the pin button no longer activates the roving-focused row.
- Backspace on the pin button no longer opens the delete confirm.
- Clicking the pin button still toggles the pin (sanity: the guard doesn't
  disable the button).
- F2 on the row itself still starts a rename (sanity: the guard is scoped to
  button targets, not the whole handler).

Verified the tests actually catch the regression by reverting the guard
locally and re-running — 3 of 4 failed as expected, confirming they aren't
vacuous.

## Question 1 — aria-activedescendant on the ⋯ menu

Wired it up. Placed on the **`IconButton` trigger**, not the `Dropdown`
surface: DOM focus never leaves the trigger while the menu is open (there's
no input to move it into, unlike ModelMenu), and `aria-activedescendant`
belongs on whichever element actually holds focus — that's ModelMenu's own
idiom too (its `<input>` carries it, not the Dropdown div). Since the
trigger and the Dropdown are DOM siblings, not ancestor/descendant, also
added `aria-owns` (new `id` prop on `Dropdown`) so the relationship is
well-formed. `MenuItem` gained an `id` prop for the row target, `IconButton`
gained `ariaOwns`/`ariaActiveDescendant`. New `MenuItem.test.tsx` case
asserts both attributes track the highlight across ArrowDown presses.

## Nits

- `renameFocused` now explicitly no-ops on `job:` keys (comment matches
  behavior now — was previously true only in the comment, not the code).
- `ArtifactsPanel.tsx`'s comment no longer claims "roving-tabindex" (neither
  this port nor Settings' original manages `tabIndex` — every tab stays in
  the normal Tab order; only the active tab + focus move). Code was already
  correct, only the wording was off.

## Verification

- typecheck: exit `0`.
- test: vitest **2306 pass / 7 skip / 0 fail** (was 2301 — +5 new: 4 in
  `Sidebar.test.tsx`, 1 in `MenuItem.test.tsx`), node:test **1565 pass / 0
  fail**.
- `manta-e2e-smoke` re-run against the rebuilt app (Playwright Electron
  launcher, xvfb): sidebar/main/titlebar present, zero console/page errors.
  PASSED.

Reassigning to `manta-reviewer`.

antoinedc:
## Review — BET-726 (cycle 2) — **PASS**

Reviewed at `a98f5d3f`. **0 Blocks, 0 Questions, 2 Nits.** Routing to `manta-pm` for merge.

### Mechanical gate (my own run, not the PR body's)

| gate | result |
|---|---|
| `npm test` | exit `0` — vitest 126 files, **2306 pass / 7 skip / 0 fail**; node:test **1565 pass / 0 fail** |
| CI `typecheck-test` | **pass** (the cycle-1 gitleaks false positive is gone — `origin/main` no longer carries the fixture, so the note I filed last cycle is moot; no follow-up issue needed) |
| CI `duplication-gate` / anti-spaghetti | **pass** — 13 files, 26117 tokens, **0 clones** |

Numbers match the PR body exactly (+5 vitest over cycle 1).

---

### Block 1 (nested-control keydown guard) — **resolved, and the test is real**

The guard is exactly right and exactly minimal:

```ts
if ((e.target as HTMLElement).closest("button")) return;
```

I checked the premise it rests on rather than taking it on faith — every roving-focus row in this tree is a `<div role="treeitem">`, never a `<button>`: `SessionRow.tsx:150`, `Sidebar.tsx:1209` (GroupHeader), `:1078`/`:1147` (WindowRow/JobChildRow via the primitive). So the guard cannot swallow the tree's own key handling. Confirmed no `tabIndex={-1}` survives on any hover-revealed button (the two remaining occurrences are the tree container itself and a comment).

Bonus the PR doesn't claim: the same guard also fixes a *pre-existing* version of this bug for the inline `ConfirmDelete`'s own buttons, which were always tabbable and were always having their Enter hijacked.

**On "I verified the tests catch the regression by reverting the guard locally" — I checked that claim independently rather than trusting it, because a guard test is the easy place to ship a vacuous assertion.** Both regression tests are non-vacuous, and here is why:

- `focusWindowRow()` really does land `focusedKey` on `win:proj:0` — proven by the fourth test in the same file, which asserts F2 opens a rename input with value `"win"`. That only happens if the roving key resolved to that window.
- `activateWindow` calls `setActive(...)` as its **first, synchronous** statement (`Sidebar.tsx:154`), before any `await`. So pre-fix, `activeProjectName` would be `"proj"` at the moment the test asserts `toBeNull()` — no flush needed, no timing escape hatch.
- Same for Backspace → `requestDeleteFocused()` → `confirmDeleteFor` → the `"Close session"` string the test greps for.

The third test (click still toggles the pin) is the right complement — it proves the guard didn't disable the button it was meant to un-break.

### Question 1 (menu highlight invisible to AT) — **answered, accepted**

`aria-activedescendant` + `aria-owns` on the ⋯ trigger, ids via `useId`, `MenuItem`/`Dropdown` each gained an `id` prop. Set only while `open`, cleared otherwise. The reasoning in the PR reply is sound: DOM focus genuinely never leaves the trigger in this menu (no input to move it into), and `aria-owns` is the correct way to make an activedescendant reference well-formed between DOM siblings. Accepted as answered — see Nit 2 for the one thing I'd leave to a future issue rather than spin another cycle on.

### Nits from cycle 1 — both fixed

`renameFocused` now explicitly no-ops on `job:` rows, and the `ArtifactsPanel` comment no longer calls the port a roving-tabindex widget.

---

### Nits (no action, do not block the merge)

1. **Stale base.** The branch is still cut at `bfd3ab8f` while `origin/main` is now `24235afb`. I checked for the phantom-deletion trap: `git diff --stat origin/main...HEAD` deletes nothing, and the intersection of "files main gained since the base" and "files this PR touches" is **empty** (main's movement was entirely `src/server/` usage-engine work). So there is no conflict and no reason to hold the PR — just take main's tip at merge time.
2. **`aria-activedescendant` on a `role=button` is outside the property's supported-roles list** (ARIA 1.2 allows it on composite/textbox-ish roles — `combobox`, `listbox`, `menu`, `tree`, `textbox`, … — not `button`), so some screen readers may ignore it. This is *not* a regression and not worth another cycle: the attributes are inert at worst, and the correct end-state is the full WAI-ARIA menu-button pattern — move DOM focus into the `role="menu"` surface on open and rove `tabIndex` across the `menuitem`s — which is a real interaction redesign well outside this issue's scope. Worth its own issue if the ⋯ menu ever gets an a11y pass: *"SessionMenu: adopt the WAI-ARIA menu-button focus pattern (focus moves into the menu on open, roving tabindex across menuitems, Escape returns focus to the trigger), replacing the trigger-side aria-activedescendant stand-in from BET-726."*

---

### Acceptance criteria — final state

- **Task 1** — scrollIntoView at all three named sites: ModelMenu/EffortMenu (`MenuOption.tsx`), @-file typeahead (`ComposerParts.tsx`), ⌘K palette (pre-existing, independently confirmed). ✅
- **Task 2** — F2 / Delete+Backspace / ContextMenu / Space on the tree; all four hover-revealed buttons tabbable, `focus-visible:opacity-100`, `title` present; nested-control guard. ✅
- **Task 3** — ⋯ menu ArrowUp/Down + Home/End + Enter (Escape pre-existing, verified in `useClickAway`); Artifacts tabs Left/Right ported from `Settings.tsx:554`; search field on the `Field` primitive's own `focus:border-accent`. ✅
- **Task 4** — rings on Toast (2), UpdateBar (2), Cards ladders (5), Artifacts tabs. ✅
- **Constraints** — zero new interaction idioms; no `Button`-primitive migration in `Cards.tsx`. ✅

Good PR. The two deviations from the issue's literal instructions (reusing `useSelectedIntoView` instead of a new `scrollHighlight.ts`, and the `Field`-matching input focus style) were both the better call and both argued for on the merits.

Leaving the draft flag alone — that's the merge owner's call.

PR #747, opened 2026-08-13
fix(server): injectable file-read + direct tests for readProviderApiKey (BET-740)

antoinedc:
BET-740 — Kimi usage adapter: read the key from the provider connection, not the secrets store.

## What this does

BET-737's review cycle already shipped most of this correction (the Kimi adapter now reads via `readProviderApiKey`, `providerIDs` is exactly `["kimi-for-coding"]`, the secrets-store lookup is gone). BET-740 closes the remaining gaps against the spec:

1. **`readProviderApiKey(providerId)` in `src/server/opencode.mjs`** — now takes the file-read function as an **injectable** parameter (default: real `readFileSync`), so tests never touch a real auth store on the live box. Never throws: missing file / unparseable JSON / absent provider / non-api entry all degrade to `""`. Returns the key only to its direct caller; never logged.
2. **`type === "api"` discriminant** in `parseProviderApiKey` — only an api-type entry with a non-empty `key` supplies the credential (used as a `Bearer` header); a stray `key` on an oauth / type-less entry is ignored. (Review Block, this cycle.)
3. **Direct tests** (`opencode.test.mjs`) with injected readers for `readProviderApiKey`: missing file → `""`, unparseable JSON → `""`, provider absent → `""`, oauth/type-less-w/-key → `""` (type gate, both parse and read level), valid api entry → key.

The Kimi adapter already sources the key from the connect-card store with an injectable `readKey`, and `providerIDs` deep-equals `["kimi-for-coding"]` (asserted in `usage.test.mjs`). `grep -rn "KIMI_CODE_API_KEY\|resolveSecret" src/server/usageAdapters/` returns nothing.

**Files changed:** 2. `src/server/opencode.mjs`, `src/server/opencode.test.mjs`.

**Base: origin/main @ `24235af`**

**Verification results:**
- PR branch (`multica/BET-740-kimi-provider-key` @ `407e340`):
  - test: exit 0 — vitest 2303 pass / 0 fail; node:test 1615 pass / 0 fail.
- The `duplication-gate` failure the earlier body flagged as pre-existing does **not** reproduce here (full suite exits 0); the review + CI confirm it.

### DIFF
--- a/src/server/opencode.mjs
+++ b/src/server/opencode.mjs
@@ -1296,7 +1296,13 @@ export function parseProviderApiKey(rawFileText, providerId) {
   try {
     const parsed = JSON.parse(rawFileText);
     const entry = parsed?.[providerId];
-    return typeof entry?.key === "string" && entry.key.length > 0 ? entry.key : "";
+    // Discriminate on `type === "api"` before trusting any `key` value: the
+    // key is used directly as a `Bearer` header, so a non-api entry (e.g. an
+    // oauth one) must never supply it even if it happens to carry a `key`
+    // field.
+    return entry?.type === "api" && typeof entry?.key === "string" && entry.key.length > 0
+      ? entry.key
+      : "";
   } catch {
     return "";
   }
--- a/src/server/opencode.mjs
+++ b/src/server/opencode.mjs
@@ -1313,17 +1313,22 @@ export function parseProviderApiKey(rawFileText, providerId) {
  * returning `""` — NEVER throws, so a caller can `await` this
  * unconditionally. The parse itself is `parseProviderApiKey` above.
  *
+ * The file-read function is injectable (`readFile`), defaulting to the real
+ * `readFileSync`, so tests can drive the missing-file / unparseable cases
+ * with a fake reader and never touch a real auth store on a live box.
+ *
  * SECURITY: this resolves a LIVE credential. It must never log, echo, or
  * return the key anywhere but to its direct caller, and a caller must never
  * fold it into an error message.
  *
  * @param {string} providerId
+ * @param {{ readFile?: (file: string, encoding: string) => string }} [opts]
  * @returns {Promise<string>}
  */
-export async function readProviderApiKey(providerId) {
+export async function readProviderApiKey(providerId, { readFile = readFileSync } = {}) {
   let raw;
   try {
-    raw = readFileSync(opencodeAuthPath(), "utf-8");
+    raw = readFile(opencodeAuthPath(), "utf-8");
   } catch {
     return "";
   }
--- a/src/server/opencode.test.mjs
+++ b/src/server/opencode.test.mjs
@@ -77,6 +77,15 @@ test("parseProviderApiKey returns \"\" for an oauth-type entry (no `key` field)"
   assert.equal(parseProviderApiKey(raw, "kimi-for-coding"), "");
 });
 
+test("parseProviderApiKey returns \"\" for a non-api entry even when it carries a non-empty `key` (type gate)", () => {
+  // type === "api" is the discriminant — a stray `key` on an oauth/common
+  // entry must never be trusted as a Bearer credential.
+  const raw = JSON.stringify({ "kimi-for-coding": { type: "oauth", key: "should-not-leak", access: "x" } });
+  assert.equal(parseProviderApiKey(raw, "kimi-for-coding"), "");
+  const rawCoding = JSON.stringify({ "kimi-for-coding": { key: "still-no-type" } });
+  assert.equal(parseProviderApiKey(rawCoding, "kimi-for-coding"), "");
+});
+
 test("parseProviderApiKey returns \"\" for an empty-string key", () => {
   const raw = JSON.stringify({ "kimi-for-coding": { type: "api", key: "" } });
   assert.equal(parseProviderApiKey(raw, "kimi-for-coding"), "");
@@ -124,6 +133,13 @@ test("readProviderApiKey returns \"\" for a provider with no entry", async () =>
 test("readProviderApiKey returns \"\" for an oauth-type entry (no `key` field)", async () => {
   assert.equal(
     await readVia(JSON.stringify({ "kimi-for-coding": { type: "oauth", access: "x", refresh: "y" } })),
+    "",
+  );
+});
+
+test("readProviderApiKey returns \"\" for a non-api entry even when it carries a non-empty `key` (type gate)", async () => {
+  assert.equal(
+    await readVia(JSON.stringify({ "kimi-for-coding": { type: "oauth", key: "should-not-leak", access: "x" } })),
     "",
   );
 });
--- a/src/server/opencode.test.mjs
+++ b/src/server/opencode.test.mjs
@@ -37,6 +37,7 @@ import {
   listModels,
   claudeCliStatus,
   parseProviderApiKey,
+  readProviderApiKey,
   opencodeAuthPath,
 } from "./opencode.mjs";
 import { homedir } from "node:os";
@@ -58,8 +59,7 @@ test("parseSseFrame returns null for comments/keepalive", () => {
 // ---------------------------------------------------------------------------
 // parseProviderApiKey (BET-737) — pure parse of opencode's own auth store,
 // same split as parseCredentials/readCredsSnapshot. Never touches a real
-// file; readProviderApiKey (the IO wrapper) is intentionally NOT unit-tested
-// here, mirroring readCredsSnapshot's own convention.
+// file.
 // ---------------------------------------------------------------------------
 
 test("parseProviderApiKey extracts the key from an api-type entry", () => {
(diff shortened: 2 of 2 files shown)


antoinedc:
## Review — BET-740 (cycle 1)

**Base freshness:** clean. `merge-base` = `origin/main` = `24235af`; `git diff --stat origin/main...HEAD` = 2 files, +46/-4. No phantom deletions.

**Mechanical gate (run on this branch, `77e9de3`, after a fresh `npm ci`):**
- `npm run typecheck` → exit **0**, no errors.
- `npm test` → exit **0**. vitest **2303 pass / 0 fail** (7 skipped, 125 files); node:test **1613 pass / 0 fail**.
  - Note vs. the PR body: the `duplication-gate` node:test **passes** here and in CI (`gh pr checks 747` — all four checks green). No pre-existing-failure caveat is needed; the suite is fully green.

**Anti-spaghetti report:** the reported clones are the pre-existing `ocFetch` non-2xx error-handling boilerplate inside `opencode.mjs` (`setProviderApiKey` / `submitProviderOauthCode`), which this PR does not touch — jscpd scopes to changed *files*, not changed lines. No action.

---

### Acceptance criteria, item by item

| # | Criterion (from the issue) | Verified how | Result |
|---|---|---|---|
| 1 | `readProviderApiKey(providerId)` exported from `src/server/opencode.mjs` | `opencode.mjs:1328` | ✅ |
| 2 | Never throws — missing file / bad JSON / missing entry / no `key` all → `""` | read the impl + the 5 new tests | ✅ |
| 3 | **Only return the key when `type === "api"`** | `parseProviderApiKey` (`opencode.mjs:1295`) | ❌ **see Block below** |
| 4 | File-read injectable, default real read | `readProviderApiKey(providerId, { readFile = readFileSync })` | ✅ |
| 5 | Key never logged / never in a thrown error / returned only to the caller | grepped all call sites; `httpError` builds `"kimi usage: HTTP <status>"` only; the adapter's own throw is `"no Kimi Code API key available"` | ✅ |
| 6 | Kimi adapter: `loadSecrets`/`resolveSecret`/`SECRET_KEY` gone | `grep -rn "KIMI_CODE_API_KEY\|resolveSecret\|loadSecrets" src/server/usageAdapters/` → no matches (exit 1) | ✅ |
| 7 | `detect()` and `fetch()` both source the key from `readProviderApiKey("kimi-for-coding")`, injectable | `kimi.mjs` `defaultReadKey` + `{ readKey = defaultReadKey }` on both | ✅ |
| 8 | `providerIDs` is exactly `["kimi-for-coding"]` | `kimi.mjs:65`; asserted `deepEqual` in `usage.test.mjs:141` | ✅ |
| 9 | Endpoint / parsing / string-or-number counts / remaining-only kept as merged | diff touches neither file | ✅ |
| 10 | File header comment no longer documents the secrets store | `kimi.mjs:1-23` now documents the connect-card store | ✅ |
| 11 | Kimi adapter tests inject a fake key reader (no fake secrets store) | `usage.test.mjs:521-579` | ✅ |
| 12 | `readProviderApiKey` cases: missing file / bad JSON / absent provider / oauth / valid | `opencode.test.mjs:109-137`, all with an injected reader | ✅ (see Block for the oauth case's strength) |
| 13 | No test reads a real auth store | every new case injects `readFile`; the two `opencodeAuthPath` cases only compute a path and save/restore `XDG_DATA_HOME` | ✅ |
| 14 | Not in scope: no Settings field, no AppConfig key, no UI, no engine/poller/RPC/type change | diff is 2 server files | ✅ |
| 15 | `usage:list` reaches the adapter with no extra config | `usage.mjs:66` `ADAPTERS` includes `kimiAdapter`; `rpc.mjs:881` `usage:list`; `detect()` default path is `readProviderApiKey(PROVIDER_ID)` with no options | ✅ (static; not exercised against a live Kimi-connected box) |

---

### Findings

**Block (fix-here): the `type === "api"` gate is not implemented.**

The issue states it plainly:

> The store shape is `{ "<providerId>": { "type": "api", "key": "…" }, … }`. Only return the key when `type === "api"`.

`parseProviderApiKey` (`src/server/opencode.mjs:1295`) gates on the *presence* of a non-empty `key` string only:

```js
const entry = parsed?.[providerId];
return typeof entry?.key === "string" && entry.key.length > 0 ? entry.key : "";
```

An entry of any `type` that happens to carry a `key` field is returned. Today no such entry exists in practice — which is exactly why this is cheap to fix and why the existing coverage does not catch it: **both** oauth tests (`opencode.test.mjs:75` and `:124`) use `{ type: "oauth", access, refresh }`, i.e. an entry with **no `key` field at all**. They pass through the `typeof entry?.key === "string"` branch and never exercise a type check. So the criterion is both unimplemented and untested, and the test whose name claims to cover it ("returns `""` for an oauth-type entry") is vacuous with respect to it.

Why it matters beyond spec-compliance: this value is used directly as `Authorization: Bearer <key>` against `api.kimi.com`. A discriminated-union entry that is not an api entry must never reach that header. Gating on the discriminant is the whole point of a discriminant.

Requested fix (small):
1. In `parseProviderApiKey`, require the discriminant: `entry?.type === "api" && typeof entry.key === "string" && entry.key.length > 0`.
2. Add a case that actually exercises it — an entry with `{ type: "oauth", key: "should-not-be-returned" }` → `""` — for `parseProviderApiKey`, and the same shape through `readProviderApiKey` with an injected reader.
3. Rename/retitle the two existing oauth tests so their names describe what they check (an entry lacking `key`), leaving the new ones to own the type case.

Safety of the change: this repo's only writer, `setProviderApiKey` (`opencode.mjs:1212`), sends `JSON.stringify({ type: "api", key })`, and opencode's own auth store is a typed union that always carries `type`. Adding the gate cannot orphan a real user's already-stored Kimi key.

**Nit (no action required):** `readProviderApiKey` is `async` while its default reader is `readFileSync`. That matches `readCredsSnapshot`'s existing convention and keeps the caller's `await` unconditional — noted only so it isn't mistaken for an oversight later.

**Nit:** `opencode.test.mjs:105` defines `readVia` immediately before the one test that does *not* use it (the missing-file case passes its own throwing reader inline). Reads slightly out of order; harmless.

---

**Outcome: 1 Block, 0 Questions, 2 Nits.** Returning to `manta-dev`. Everything else in this PR is correct and well-scoped — the fix is one predicate plus two test cases.

antoinedc:
Addressed the review `Block`: `parseProviderApiKey` now discriminates on `type === "api"` before trusting any `key` value (the value goes straight into a `Bearer` header). Added two gate test cases — an oauth entry that *does* carry a non-empty `key`, and a type-less entry with a key — both at the `parseProviderApiKey` and `readProviderApiKey` levels, asserting `""`. Existing positive case (api + key) unchanged.

Verified: `npm run typecheck` exit 0; `npm test` exit 0 (vitest 2303/0; node:test 1615/0, fail 0). Head `407e340`.

antoinedc:
## Review — BET-740 (cycle 2) — **PASS**

Re-reviewed at `407e340`.

**Cycle-1 Block resolved.** `parseProviderApiKey` (`src/server/opencode.mjs:1296`) now discriminates before trusting the value:

```js
entry?.type === "api" && typeof entry?.key === "string" && entry.key.length > 0 ? entry.key : ""
```

and the gate is covered by tests that actually exercise it, at both levels:
- `parseProviderApiKey`: `{ type: "oauth", key: "should-not-leak", access }` → `""`, and a type-less `{ key: "still-no-type" }` → `""` (`opencode.test.mjs:80`).
- `readProviderApiKey`, via an injected reader: `{ type: "oauth", key: "should-not-leak", access }` → `""` (`opencode.test.mjs:140`).

The type-less case is a good addition beyond what I asked for — it pins the `&&` ordering so a future refactor can't fall back to key-presence.

**Base:** `merge-base` is `24235af`; `origin/main` has advanced to `f174cf5` since the branch was cut. `git diff --stat origin/main...HEAD` is still exactly 2 files (+69/-5) with no deletions — no stale-base artifacts, nothing to rebase for correctness.

**Mechanical gate (fresh `npm ci`, this branch):**
- `npm run typecheck` → exit **0**, no errors.
- `npm test` → exit **0**. vitest **2303 pass / 0 fail** (7 skipped, 125 files); node:test **1615 pass / 0 fail** — +2 vs. cycle 1, matching the two new gate cases.
- CI: all three checks green (`typecheck-test`, `duplication-gate`, agent-branch job).

**Acceptance criteria:** all 15 items from cycle 1 now verified ✅, including item 3 (`type === "api"`). `grep -rn "KIMI_CODE_API_KEY\|resolveSecret\|loadSecrets" src/server/usageAdapters/` → no matches; `kimiAdapter.providerIDs` is `[PROVIDER_ID]` with the deep-equal assertion in `usage.test.mjs:141`. Scope is unchanged: 2 server files, no UI / AppConfig / engine / poller / RPC / type changes.

**Outstanding nits (non-blocking, do not fix in this PR):** the two older "oauth-type entry (no `key` field)" test names still describe the case by what it lacks rather than what it checks — the new gate tests make the distinction clear enough in context. And `readProviderApiKey` remains `async` around a sync default reader, matching `readCredsSnapshot`'s existing convention.

**Outcome: 0 Blocks, 0 Questions, 2 Nits.** Routing to `manta-pm` for the merge hand-off.

PR #748, opened 2026-08-13
Subscription usage: composer dial + detail popover (BET-738)

antoinedc:
## BET-738: Subscription usage — composer dial + detail popover

Depends on BET-737 (merged, `8e4a58f9`). Base: `origin/main @ 8e4a58f9`.

### What
- `src/renderer/UsageDial.tsx` (new): 16px ring trigger + popover, mirroring
  `ContextPill`'s sibling trigger/panel pattern in `SessionHeader.tsx`.
  `useClickAway` dismissal, `manta-usage-dial` / `manta-usage-popover` hook
  classes, `aria-haspopup="dialog"` + `aria-expanded` / `role="dialog"`.
- Pure logic in `chatUtils.ts`: `selectUsageSnapshot`, `usageDialState`,
  `formatWindowReset`, `formatUpdatedAgo`, `usageStale` — unit-tested,
  including the 70/90/100 tone boundaries and the "pick the higher window"
  rule.
- `ComposerParts.tsx` now exports the shared `mbtn` chrome const so the
  dial's trigger matches the other icon-row buttons exactly (imported, not
  copied).
- `store.ts`: new `usage: UsageSnapshot[]` slice + `setUsage`, new
  `alwaysShowUsage` config-mirrored field.
- `App.tsx`: primes `usage` with one `usageList()` call on mount, then stays
  live via the box's `usage.updated` bus event — **no renderer-side
  polling** (the box's 3-minute usage poller is the only timer).
- `httpApi.ts` / `shared/api.ts`: `onUsageUpdated` bus subscription,
  mirroring `onDelegateUpdated`.
- `InputArea.tsx` / `ChatPanel.tsx`: `<UsageDial>` renders immediately
  before `<SessionToolbar>`; ChatPanel passes the already-resolved active
  model's `providerID` down — `UsageDial` never re-resolves the model.
- `settingsSchema.ts` / `Settings.tsx`: "Always show plan usage" toggle
  under Settings → Models (the schema's actual section id; the issue's "AI
  tab" reference is to the settings-redesign doc, not the shipped id).

### Deviation: one small server-side field (issue said "adds no server code")
BET-737's shipped `UsageSnapshot` has no `providerIDs` — only the usage
engine's own adapter id (`"claude"`/`"codex"`/`"kimi"`), a **different
namespace** from opencode's providerID (`"anthropic"`/`"openai"`/
`"kimi-for-coding"`). The issue's own matching rule ("pick the snapshot
whose `providerIDs` contains the active providerID") isn't implementable
without either that field or a second, drift-prone id-mapping table in the
renderer — which the issue's "no provider-name branching outside an
icon/label lookup table" constraint forbids.

Added `providerIDs: string[]` to the published snapshot in `usage.mjs`'s
`tick()` (copied verbatim from `adapter.providerIDs` — one line), mirrored
in the shared TS type, with one assertion added to `usage.test.mjs`. No
adapter files touched, no behavior change to any existing field.

### Verification results
- PR branch (`multica/BET-738-usage-dial` @ `3167803`):
  - typecheck: exit `0`. Errors: none. log sha256:
    `75342603966d11824516b89baa97128413853820f43086df3254737d95622e1f`
  - test: exit `0`, `3924` pass (`2317` vitest + `7` skipped + `1607`
    node:test) / `0` fail. log sha256:
    `8254078ee9481d755de69a2827b0de0830501e9cc086a1bb7917c7bef72bcad2`
- **Conclusion:** fully green on this branch; no pre-existing failures
  observed on this run (an earlier run hit a flaky jscpd-tool-availability
  failure in `check-duplication-gate.test.mjs`, confirmed reproducible on
  bare `main` too — unrelated to this change, not present in the logged run
  above).

Logs cached at `/tmp/typecheck-pr.log`, `/tmp/test-pr.log` for reviewer
spot-check (this box only).


antoinedc:
# Review — BET-738 (cycle 1): **PASS**

0 Blocks · 0 Questions · 4 Nits. Routing to `manta-pm` for merge.

## Mechanical gates (run by me on `70704c2f`, PR head)

| gate | result |
|---|---|
| `npm run typecheck` | exit **0**, no output |
| `npm test` | exit **0** — vitest **2317 passed / 7 skipped (125 files)**, node:test **1607 passed / 0 failed (68 suites)** |
| `npm run gen:swift-settings -- --check` | "Swift settings schema is up to date." |
| `bash scripts/check-duplication.sh origin/main` | **0 clones**, 16 files, 0.00% duplicated |
| anti-spaghetti sticky comment | 0 clones — agrees with my local run |

These match the author's reported numbers exactly (2317 vitest + 1607 node:test, 0 fail), so the reported results are corroborated, not taken on trust.

## Base freshness

Merge-base `8e4a58f9`, `origin/main` now at `f174cf5c`. Base is behind but `git diff --stat origin/main...HEAD` shows **no deletions and no phantom-deleted files** — 17 files, +632/−10, all additive. Not a stale-base problem; no rebase required. GitHub reports `MERGEABLE`.

Each checked against the code, not against the PR description.

**1. `UsageDial.tsx` structure**
- Trigger `<button>` and panel are SIBLINGS under a `relative` wrapper — confirmed, no nesting, no `stopPropagation`.
- `useClickAway(rootRef, open, …)` from the shared hook — same call shape as `ContextPill` (`SessionHeader.tsx:284`). No second click-away implementation anywhere in the diff.
- `aria-haspopup="dialog"` + `aria-expanded` on the trigger; `role="dialog"` + `aria-label="Plan usage"` on the panel. ✔
- `manta-usage-dial` / `manta-usage-popover` present. ✔

**2. Dial trigger**
- `mbtn` is now `export const` in `ComposerParts.tsx:17` (moved out of `SessionToolbar`'s body) and **imported**, not copied — verified there is exactly one occurrence of the string in the repo.
- 16px ring, no text/pct/reset in the trigger. ✔
- One rendering technique only (conic-gradient + inner disc); no SVG variant shipped alongside. ✔
- Thresholds: `>=100` solid `--danger` disc with the inner hole suppressed; `>=90` `--danger`; `>=70` `--warn`; `<70` `--tx4`. No `--ok`/green anywhere in the trigger path (`toneRingColor`). ✔
- Open state adds `bg-fill-active`. ✔ `aria-label="plan usage"` lowercase. ✔ `title` matches the specified sentence. ✔
- Visibility: `null` when no snapshot, no match, or `!state.visible`. ✔

**3. Popover**
- Iterates `snapshot.windows` — grepped for `"session"`/`"weekly"` string literals in `UsageDial.tsx`: none. A third or daily window renders with no change. ✔
- Value `"<used> / <limit> · <pct>%"` with `toLocaleString()`, falling back to `"<pct>%"`. ✔ `pct` is guaranteed integer (`clampPct` in `normalizeWindow.mjs:22` does `Math.round`), so no `42.857…%` leaking through.
- 7px track (`h-[7px]`), fill colour with `--ok` permitted below 70 in this labelled view only. ✔
- Reset line with `(HH:MM)` appended past 12h, `" · binding limit"` on the binding window. ✔
- `extras[]` section omitted entirely when empty. ✔
- Footer `"updated Nm ago"`, `text-warn` past 10 min. ✔

**4. Wiring**
- Store: `usage: UsageSnapshot[]` + `setUsage` (straight replace). ✔
- `App.tsx`: exactly one `usageList()` call on mount + `onUsageUpdated` subscription, cleanup returned. **No `setInterval`/`setTimeout` in the effect** — verified by reading it. The server side genuinely emits the event (`usage.mjs:203` `publish?.({ kind: "usage.updated", … })`, asserted at `usage.test.mjs:321`), and `httpApi`'s `dispatchFrame` routes the new kind through the existing generic demux — so the live path is real, not aspirational.
- Provider matching: `ChatPanel.tsx:2301` passes `activeModel?.providerID` from the existing `resolveActiveModel` memo (`ChatPanel.tsx:1046`); `UsageDial` never re-resolves. ✔
- Both helpers (`selectUsageSnapshot`, `usageDialState`) are pure and in `chatUtils.ts`; all threshold logic lives there, the component only maps `tone` → token. ✔
- Placement: immediately before `<SessionToolbar>` inside the existing right-hand `<span>`; nothing else in `InputArea.tsx` moved. ✔
- Setting: `alwaysShowUsage` default `false` in `AppConfig`, `settingsSchema.ts`, store, and the Swift codegen. Toggle added via the existing `SECTION_GROUPS` mechanism — no new settings machinery.

**5. Constraints**
- No new dependency, no new CSS file. Every colour is a token (`cssVar("--danger"/"--warn"/"--tx4"/"--ok"/"--border-subtle")` or a Tailwind token class) — **no hex literals in the diff**.
- `React.memo` + a single primitive prop (`providerID: string | null`); no object constructed inline in `InputArea`'s render path. The store subscriptions are two narrow selectors, so a keystroke does not re-render the dial.
- Provider-name knowledge is confined to `PROVIDER_LABELS`, an explicit label lookup with a Title-Case fallback. No `if (provider === …)` in rendering logic.
- `src/renderer/mobile/` untouched (doesn't exist). ✔

**6. Tests** — every case the issue enumerated is present in `chatUtils.test.ts`: providerID match / no-match / empty array, hidden-below-70-off, visible-below-70-on, boundaries asserted at **69/70, 89/90, 99/100**, higher-of-two-windows, single window, zero windows, plus the reset/stale helpers.

## The "no server code" claim — checked, not accepted

The issue said this PR adds no server code; the PR adds 12 lines to `src/server/usage.mjs`. I verified the justification rather than the prose:

- `UsageSnapshot` as shipped by BET-737 carried only `provider` (the adapter registry key: `"claude"`), while the issue's own matching rule keys off opencode's providerID (`"anthropic"`). Those are genuinely different namespaces — a `provider`-equality match would have matched nothing, ever.
- The adapters already carry the right field: `claude.mjs:54 providerIDs: ["anthropic"]`, `codex.mjs:52 ["openai"]`, `kimi.mjs:65`. The change is a single verbatim copy in `tick()`, with an assertion added at `usage.test.mjs:167`. No adapter file touched, no existing field's behaviour changed.

Correct call, correctly disclosed, minimally scoped. Not a Block.

## Nits (non-blocking — merge as-is or fold into a later pass)

1. **Ring hole colour is pinned to `bg-bg`.** `mbtn` adds `hover:bg-fill-hover` and the open state adds `bg-fill-active`, both translucent overlays on top of `--bg` — so on hover/open the ring's centre stays pure `--bg` and reads a hair darker than the button around it. At 6–7% alpha this is close to imperceptible, and the spec explicitly sanctioned the inner-disc technique. It would matter more if the composer ever moves off `--bg`.
2. **`key={w.kind}`** on the window rows. `kind` is documented as an open set; two windows sharing a kind would collide on the React key. `key={w.kind + w.label}` or the index would be safer.
3. **The reset line is all-or-nothing.** `formatWindowReset` returns `null` for a missing/past `resetsAt`, which drops the whole `<div>` — taking `" · binding limit"` with it. Every shipped adapter sets `resetsAt`, so this is theoretical today.
4. **`nowMs` is frozen at open time** (deliberate, and commented). A popover left open for ten minutes keeps claiming "updated just now". Fine for a short-lived panel; noting it so it isn't mistaken for a bug later.

Nothing here needs the author's design rationale, so none of it is filed as a Question.

PR #749, opened 2026-08-13
feat(ios): context percent pill in the chat header; delete dead headerSubtitle (BET-716 task 1)

antoinedc:
Opened automatically for `multica/BET-716-chat-visibility`, which was pushed without a pull request.

Relates to BET-716.

### DIFF
--- a/mobile/native/MantaUI/ChatModels.swift
+++ b/mobile/native/MantaUI/ChatModels.swift
@@ -375,23 +375,3 @@ enum ChatQuestionAnswers {
     }
 }
 
-// MARK: - Header subtitle (§8 "running · 2m · 8%" / idle)/// The §8 chat-header subtitle: while running, `running · <elapsed> · <N>%`;
-/// while idle, `idle · <N>%` (the context number matters most before sending),
-/// falling back to `idle` when no context is established yet. Token-aware on
-/// the surrounding prose only; the string is assembled here so the view stays a
-/// thin token resolver.
-enum ChatHeaderSubtitle {
-    static func text(running: Bool, elapsed: TimeInterval, contextPct: Double?) -> String {
-        if running {
-            let elapsed = SessionTimerFormat.elapsed(elapsed)
-            if let pct = contextPct {
-                return "running · \(elapsed) · \(Int(pct.rounded()))%"
-            }
-            return "running · \(elapsed)"
-        }
-        if let pct = contextPct {
-            return "idle · \(Int(pct.rounded()))%"
-        }
-        return "idle"
-    }
-}
--- a/mobile/native/MantaUI/ChatOverflowCaptureScene.swift
+++ b/mobile/native/MantaUI/ChatOverflowCaptureScene.swift
@@ -31,7 +31,6 @@ struct ChatOverflowCaptureScene: View {
                     branch: "main",
                     onSchedules: {},
                     onSecrets: {},
-                    onWebhooks: {},
                     onCompact: {},
                     onClear: {},
                     onFork: {},
--- a/mobile/native/MantaUI/ChatOverflowSheet.swift
+++ b/mobile/native/MantaUI/ChatOverflowSheet.swift
@@ -26,7 +26,6 @@ struct ChatOverflowSheet: View {
 
     var onSchedules: () -> Void
     var onSecrets: () -> Void
-    var onWebhooks: () -> Void
     var onCompact: () -> Void
     var onClear: () -> Void
     var onFork: () -> Void
@@ -51,7 +50,6 @@ struct ChatOverflowSheet: View {
                 Section {
                     row("Scheduled tasks", systemImage: "clock", badge: scheduleCount, action: onSchedules)
                     row("Secrets", systemImage: "key", action: onSecrets)
-                    row("Webhooks", systemImage: "bolt.horizontal", action: onWebhooks)
                 }
                 Section {
                     row("Compact session", systemImage: "arrow.down.right.and.arrow.up.left", action: onCompact)
--- a/mobile/native/MantaUI/ChatScreen.swift
+++ b/mobile/native/MantaUI/ChatScreen.swift
@@ -325,6 +325,24 @@ private struct ChatScreenContent: View {
                 .frame(height: Self.headerReservedHeight + Metrics.spacing.sp6)
         }
         .overlay(alignment: .top) { header }
+        // Offline must not read as "the model is quiet": a slim banner pinned
+        // just below the floating header, declared AFTER the header so it draws
+        // on top of it. It offsets by the exact height the transcript reserves
+        // (`headerReservedHeight`) plus a gap, so it never overlaps the two
+        // header buttons. It disappears on its own when `degraded` flips false.
+        .overlay(alignment: .top) {
+            if store.degraded {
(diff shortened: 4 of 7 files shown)


antoinedc:
# Review — BET-716 (cycle 1): **PASS**

**0 Blocks, 0 Questions, 6 Nits.** Routing to `manta-pm` for merge.

Reviewed at `9b47697`. Scope: 7 files, all under `mobile/native/` — no TS, no server, no `mobile/www/`.

---

## Mechanical gates

| Gate | Result |
|---|---|
| `npm run typecheck` | **pass** (exit 0) — run locally on the PR branch |
| `npm test` | **pass** — `1608 pass / 0 fail / 0 cancelled`, 68 suites |
| CI `typecheck-test` | **pass** (on `9b47697`) |
| CI `ios-build` | **pass**, 4m27s (on `9b47697`) — independent of the implementer's local build claim |
| CI `duplication-gate` | **pass** |
| Anti-spaghetti report | 0 `.ts/.tsx/.mjs` files changed — nothing to scan |

**Base freshness:** merge-base `24235af`, `origin/main` tip `f174cf5`. `git diff --stat origin/main...HEAD` is exactly the seven intended files — no phantom deletions, no stale-base artefacts. No rebase needed.

---

### Task 1 — context pill

- Pill lives inside the header's `GlassEffectContainer`, centred between two `Spacer(minLength: 0)`, gated on `if let ctx = store.context` — hidden until context exists, no animation, no tap. ✅
- `ctxColor`: `>= 90 → tokens.danger`, `>= 70 → tokens.warn`, else `tokens.tx2`. ✅
- **Substitutions independently confirmed against `generated/swift/Theme.swift`:** `TypeMetrics` has no `caption` field (`body/small/xs/twoXS/display/…`), so `Metrics.type.xs` (12) is the correct stand-in; `Tokens` has no `warning`, and `warn` exists (`#F0A934` dark / `#8A5A08` light). Both substitutions are the fallbacks the issue pre-approved. `Metrics.type.chatHeaderBtn` (38) exists and is used. ✅
- `.pct` scale confirmed 0–100 (matches the deleted `ChatHeaderSubtitle` tests that asserted `8.4 → "8%"`), so the 70/90 thresholds are on the right axis. ✅
- Dead code: `grep -rn 'headerSubtitle|ChatHeaderSubtitle'` across `mobile/native/` → **zero hits**; the enum, the store property and its 4 unit tests are gone. `SessionTimerFormat` was correctly *kept* — it still has 6 live call sites (`SessionListView`, `RunningIndicator`) plus its own tests, so it was never "only used by" the deleted helper. ✅
- No hardcoded colours or sizes. ✅

### Task 2 — connection banner

- `ChatSessionStore.degraded` is `@Published private(set)`, fed by `eventStore.$degraded.removeDuplicates()`. ✅
- Banner overlay is declared **after** the header overlay, so it draws above it. Offset arithmetic checks out: `headerReservedHeight = chatHeaderBtn(38) + sp2(8)*2 = 54`; banner `.padding(.top, 54 + 8) = 62`. The header occupies `0…54`. No overlap with either button. ✅
- `accessibilityIdentifier("connection-banner")` present; auto-vanishes on `degraded == false`. ✅
- `MantaEventStore` / `MantaReconnectController` untouched — the diff consumes `degraded`, never computes it. ✅
- Hardcoded `padding(.vertical, 6)` from the spec snippet replaced by `Metrics.spacing.sp1` (4) — correct call given the no-hardcoded-sizes constraint. ✅

### Task 3 — action-failure hints

- `@Published var actionHint: String?` added as a settable bus, documented as such. ✅
- `abort()` / `compact()` converted from `try?` to `do/catch → actionHint`. ✅
- `ComposerView` surfaces it through the **existing** capsule: one `onChange` → `surfaceHint(hint)` → `hintState`, which already owns the 4s auto-dismiss and the replace-cancels-pending-timer behaviour. No new component. ✅
- `ChatScreen`: `clearSession`, `forkSession`, `openTerminal`, `deleteSession` all now report the missing-window case instead of returning silently; `clear`/`fork` also catch the throw *and* the empty-id case; `delete` no longer dismisses on failure. ✅
- **The delivery path actually works**, which was worth checking: `ChatOverflowSheet.row` calls `dismiss()` *before* `action()`, and both confirm-sheet destructive actions do `dismiss(); onClear()/onDelete()`. So the sheet is gone by the time the async failure lands and the composer capsule is on screen. A hint fired behind a still-presented sheet would have silently defeated the whole task. ✅
- Remaining `try?` in `ChatScreen` (lines 162/169/197/200 `reportFocus`, 391 `listSchedules`, 401 `projects`, 407 `vcsBranch`) are focus telemetry and background data loads, not the overflow action block the issue scoped. Correctly left alone. ✅

### Task 4 — dead Webhooks row

- `onWebhooks` removed from `ChatOverflowSheet` (property + row), `ChatScreen`, `ChatOverflowCaptureScene`. `grep -rn 'onWebhooks|Webhooks'` across `mobile/native/` → the only survivors are two lines in a retired capture artefact (see Nit 5). No test referenced the row. ✅

### Constraints

- No new toast/banner components beyond the two inline views. ✅
- Reconnect logic untouched. ✅
- Tokens/Metrics only. ✅

---

## Nits (non-blocking)

1. **`.transition(.opacity)` on the banner has nothing driving it** — there's no `withAnimation` at the mutation site and no `.animation(_:value: store.degraded)` on the view, so the banner will pop rather than fade. The issue's own snippet had the same shape, so this is inherited, not introduced.
2. **The new `$degraded` sink omits `.receive(on: DispatchQueue.main)`** that both of its neighbours carry. I checked and it *is* safe — `MantaEventStore` and `ChatSessionStore` are both `@MainActor` and `degraded` is only assigned from MainActor code (`MantaEventStore.swift:346`) — so this is a consistency nit, not a threading bug. Arguably the new form is better (synchronous, no runloop lag); worth deciding one way for all three.
3. **`clearSession()` is internally inconsistent about isolation** — the guard sets `store.actionHint` directly, the catch wraps the identical assignment in `await MainActor.run`. Both are already main-isolated. Pick one spelling.
4. **`abort()` / `compact()` wrap in `await MainActor.run` inside a `Task {}` that already inherits `@MainActor`** from the enclosing class — redundant, same as above.
5. **Two `"Webhooks"` strings survive** in `mobile/native/baseline/bet627/bet627-overflow-sheet-hierarchy.txt`. The issue asked for zero hits across `mobile/native/`, but its own parenthetical scopes the intent to code and tests, this file is a non-compiled recording of the *old* sheet, and reviewer policy is not to touch retired pixel baselines. Leaving it. Credit for flagging it proactively rather than letting me find it.
6. **The PR body is the CI sweep's boilerplate**, not the requirement + acceptance criteria. Normally a Block — here it is structural: the `macos` worker is forbidden from opening PRs, so CI opens them with fixed text. Blocking on it would be an unresolvable loop. The substantive verification lives on BET-716.

## Note on the failing UI tests

`MantaOverflowCaptureUITests` / `MantaRunningStateCaptureUITests` fail with `pairing failed … Nothing was linked`. Those are the retired capture flows and they need a live paired box; the failure is environmental and pre-render, not a code regression. CI's `ios-build` compiles the whole target — including `ChatOverflowCaptureScene`, the file this PR edits — and passes. Not held against the PR, consistent with the retirement of visual/pixel-baseline verification.

PR #750, opened 2026-08-13
BET-727: adopt Button primitive at hand-rolled stragglers; kill hover no-ops

antoinedc:
Adopts the Button primitive (src/renderer/Button.tsx) at every hand-rolled button straggler and kills the hover no-ops / legacy hover idiom, per issue BET-727.

**Base: origin/main @ `f174cf5c`**

## Migration list (one commit each)
1. Cards.tsx — PermissionCard/QuestionCard action rows → `tone` ladder (primary / default / danger / ghost).
2. Settings.tsx — Add → default; restart "Later" + Remove/Reset modals' Cancel → ghost. **`Settings.tsx:872` (settings-header close) stays hand-rolled — recorded decision, see Cycle-2 reply: it is an icon-only `<X>` whose accessible name comes from `aria-label`, and the `Button` primitive exposes no `aria-*` prop (and adding one is forbidden), so migrating it would drop the accessible name.**
3. App.tsx — update-confirm Cancel → ghost.
4. FolderPickerModal.tsx — fan-out trio (default/primary/ghost) + Go → default (fixed `hover:bg-bg-soft` no-op). Go keeps `shrink-0` via a wrapper container (the primitive has no className escape hatch).
5. Onboarding.tsx — success CTA → `<Button tone="primary" block>`.
6. ChatPanel Retry + Transcript "Load earlier" → default.
7. Banner pills (Toast/UpdateBar/ReconnectingBanner) → shared `BANNER_BTN` exported from Toast.tsx (added the missing focus ring to ReconnectingBanner); **the 4th surviving copy in Settings.tsx is folded into the same import.**
8. Sidebar destructive inline buttons (Kill on server, Remove) → minimal fix (real `hover:bg-danger-bg` + focus ring); compact text-meta pills left hand-rolled rather than bloated to the 32px primitive.

Cycle-2 (review) extras: NewSessionScreen fan-out row fully migrated (Create → `primary`, Cancel → `ghost`); the composer send button keeps a real, non-Button-owned hover (`hover:bg-accent/90`, the banner-pill background-step idiom — not in the BET-588 allowlist) so hover feedback is preserved on the icon-only 28px control.

## Tone mapping (§D9)
primary = primary action · default = secondary · ghost = text-only Cancel/dismiss · danger = destructive.

## Verification results
- PR branch (`multica/BET-727-button-adoption` @ `8678fef1`):
  - test: exit `0`, vitest `2313` pass / `7` skip / `0` fail; node:test `1608` pass / `0` fail. log sha256: `f41d49310ae421b9ccd197b01021d8f5bbcfee4a0b861baf46bcf997ab397919`
- Base (`main` @ `f174cf5c`): both suites green on the PR branch with no pre-existing failures.
- Grep deltas (all 0 outside `Button.tsx`): `hover:opacity-90` 3→0; `hover:brightness-110` 0→0; `h-8 px-[14px]` 4→0.

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -1457,12 +1457,12 @@ function AppInner() {
               running in any session. Any unsaved work in a running turn is lost.
             </div>
             <div className="flex justify-end gap-2">
-              <button
+              <Button
+                tone="ghost"
                 onClick={() => setConfirmServerUpdate(false)}
-                className="px-4 py-2 text-body text-text-muted hover:text-text"
               >
                 Cancel
-              </button>
+              </Button>
               <Button
                 tone="primary"
                 onClick={() => {
--- a/src/renderer/Cards.tsx
+++ b/src/renderer/Cards.tsx
@@ -207,7 +207,7 @@ export function PermissionCard({
   //   Allow once — the FILLED accent primary, with a Check + ⏎ hint.
   //   Always allow <tool> — outlined secondary.
   //   (spacer)
-  //   Reject — ghost, pushed to the far right, gains --danger on hover.
+  //   Reject — danger text at rest, gains --danger-bg fill on hover (tone="danger").
   const actions = (
     <>
       <Button tone="primary" onClick={() => onReply("once")}>
--- a/src/renderer/Cards.tsx
+++ b/src/renderer/Cards.tsx
@@ -19,6 +19,7 @@ import { buildQuestionAnswers, canSubmitQuestion } from "./chatUtils";
 import { Card } from "./Card";
 import { Pill } from "./Pill";
 import { OutputWell } from "./OutputWell";
+import { Button } from "./Button";
 
 // ===== Shared ask-card shell (BET-458) =====
 //
@@ -209,35 +210,28 @@ export function PermissionCard({
   //   Reject — ghost, pushed to the far right, gains --danger on hover.
   const actions = (
     <>
-      <button
-        onClick={() => onReply("once")}
-        className="h-8 px-[14px] inline-flex items-center gap-2 rounded-md text-on-accent text-meta font-medium focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent"
-        style={{ backgroundColor: "var(--accent-solid)" }}
-      >
+      <Button tone="primary" onClick={() => onReply("once")}>
         <Check size={13} aria-hidden="true" />
         Allow once
         <span className="text-on-accent opacity-70" aria-hidden="true">
           ⏎
         </span>
-      </button>
+      </Button>
       {alwaysScope ? (
-        <button
+        <Button
+          tone="default"
           onClick={() => onReply("always")}
           title={`Always allow ${alwaysScope}`}
-          className="h-8 px-[14px] inline-flex items-center rounded-md border border-border-strong text-text hover:bg-bg-soft text-meta focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent"
         >
           Always allow&nbsp;
           <span className="text-text-faint">{alwaysScope}</span>
-        </button>
+        </Button>
       ) : null}
       {/* spacer pushes Reject to the far right */}
       <div className="flex-1" />
-      <button
-        onClick={() => onReply("reject")}
-        className="h-8 px-[14px] inline-flex items-center rounded-md text-text-faint hover:text-danger hover:bg-danger-bg text-meta focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent"
-      >
+      <Button tone="danger" onClick={() => onReply("reject")}>
         Reject
-      </button>
+      </Button>
     </>
   );
 
@@ -481,17 +475,12 @@ export function QuestionCard({
       actions={
         <>
           {/* Submit FIRST — the filled accent primary with a trailing ⏎ hint. */}
-          <button
-            onClick={handleSubmit}
-            disabled={!canSubmit}
-            className="h-8 px-[14px] inline-flex items-center gap-2 rounded-md text-on-accent text-meta font-medium disabled:opacity-40 focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent"
-            style={{ backgroundColor: "var(--accent-solid)" }}
-          >
+          <Button tone="primary" onClick={handleSubmit} disabled={!canSubmit}>
             Submit
             <span className="text-on-accent opacity-70" aria-hidden="true">
               ⏎
             </span>
-          </button>
+          </Button>
           {/* Multi-question progress — mono `.prog`. */}
           {isMulti && (
             <span className="text-text-faint text-meta font-mono ml-2">
@@ -500,13 +489,9 @@ export function QuestionCard({
           )}
           {/* spacer pushes Dismiss to the far right */}
           <div className="flex-1" />
-          <button
-            onClick={onReject}
-            className="h-8 px-[14px] inline-flex items-center rounded-md text-text-faint hover:text-danger hover:bg-danger-bg text-meta focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent"
-            title="Dismiss this question"
-          >
+          <Button tone="ghost" onClick={onReject} title="Dismiss this question">
             Dismiss
-          </button>
+          </Button>
         </>
       }
     />
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -69,6 +69,7 @@ import { useModelCatalog } from "./modelCatalog";
 import { MantaLoader } from "./MantaLoader";
 import { MeasureColumn } from "./MeasureColumn";
 import { CompactionCard, PermissionCard, RetryCard } from "./Cards";
+import { Button } from "./Button";
 import { DelegateApprovalCard, ReadOnlyJobBar, ScheduledTasksCard, SecretsCard, WebhooksCard } from "./PanelCards";
 import { CardMount } from "./components/CardMount";
 import { useSessionResources } from "./hooks/useSessionResources";
@@ -1898,13 +1899,9 @@ export function ChatPanel({
           </pre>
           {transcriptLoadError && (
             <div className="mt-4 flex justify-center">
-              <button
-                type="button"
-                onClick={retryTranscriptLoad}
-                className="rounded-md border border-border px-4 py-2 text-body text-text hover:bg-bg-soft"
-              >
+              <Button type="button" tone="default" onClick={retryTranscriptLoad}>
                 Retry
-              </button>
+              </Button>
             </div>
           )}
         </div>
--- a/src/renderer/FolderPickerModal.tsx
+++ b/src/renderer/FolderPickerModal.tsx
@@ -345,9 +345,11 @@ export function FolderPickerModal({ open, initialPath, onSelect, onFanOut, onCan
                     className="relative w-full bg-transparent border-0 px-3 py-2 text-meta rounded-xs focus:outline-none font-mono"
                   />
                 </div>
-                <Button tone="default" onClick={goNavigate}>
-                  Go
-                </Button>
+                <div className="shrink-0">
+                  <Button tone="default" onClick={goNavigate}>
+                    Go
+                  </Button>
+                </div>
               </div>
 
               {/* Breadcrumbs — home icon + inter-crumb chevrons, current crumb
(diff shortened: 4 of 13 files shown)


antoinedc:
# Review — BET-727 (cycle 1)

**Verdict: 2 × Block (fix-here), 1 × Question, 3 nits.** Returned to `manta-dev`.

## Mechanical gates

Run on this branch (`e0a9cc22`), after `npm ci`:

- `npm test` → **exit 0**; node:test `# pass 1608 / # fail 0`, vitest leg green (the `&&` chain reached node:test). Matches the reported numbers.

Base freshness: merge-base `f174cf5c`, `origin/main` now `72b87a99`. The four upstream commits are iOS-only Swift (`mobile/native/**`) — disjoint from this diff, no phantom deletions, no rebase required.

Anti-spaghetti report: the single jscpd clone is `NewSessionScreen.tsx [289:41]` ↔ `[365:50]` — both pre-existing worktree-list blocks this PR does not touch. Discounted.

## Acceptance criteria — item by item

| # | Item | Verified |
|---|------|----------|
| 1 | Cards.tsx ask-card ladder | ✅ primary / default / danger (Reject) / ghost (Dismiss); hand-rolled ring classes removed |
| 2 | Settings.tsx `:756` Add, `:887` Later, `:927`/`:939` Cancels | ✅ — but `:872` **not addressed**, see Question |
| 3 | App.tsx update-confirm Cancel → ghost | ✅ (confirm was already `Button primary`) |
| 4 | FolderPickerModal trio + Go | ✅ default / primary / ghost; Go's `hover:bg-bg-soft` no-op gone |
| 5 | Onboarding CTA → `primary block` | ✅ |
| 6 | ChatPanel Retry + Transcript Load earlier → default | ✅ |
| 7 | Banner pills → one shared class | ✅ `BANNER_BTN` exported from Toast.tsx, imported by UpdateBar + ReconnectingBanner; ring added to ReconnectingBanner |
| 8 | Sidebar destructive pair | ✅ minimal-fix route on both, stated per the issue's requirement; resting `bg-transparent`, real `hover:bg-danger-bg`, focus ring |

Grep gates re-run by me on the branch (not taken from the PR body):

- `hover:opacity-90` outside `Button.tsx` → **0**
- `hover:brightness-110` outside `Button.tsx` → **0**
- `h-8 px-[14px]` outside `Button.tsx` → **0** (only `Button.test.tsx` retains the literal, which is the primitive's own contract test)

## Findings

**Block (fix-here) — `InputArea.tsx:404`: the send button now has NO hover state at all.**

The change deletes `hover:opacity-90` from the enabled composer send button and puts nothing back:

```
- ? "bg-accent text-on-accent hover:opacity-90"
+ ? "bg-accent text-on-accent"
```

`hover:opacity-90` was not a no-op — it was a real, visible state change. This PR converts a control **with** hover feedback into one **without**, on an issue whose title is "kill the hover no-ops". The stated rationale ("its enabled/disabled colour change already signals it") describes the *disabled* transition, not hover; nothing responds to the pointer now.

The grep target is satisfied by deletion rather than by migration, which is the wrong direction. `hover:brightness-110` is correctly off-limits (Button-owned, BET-588), but that does not leave "no hover" as the only option: use a non-owned real hover on this 28px icon control (a background token step, e.g. the `hover:bg-accent/xx` idiom already used by the banner pill), or leave `hover:opacity-90` in place and argue it in the PR body — the "three primary-hover idioms" rule is about *labelled primary buttons*, and this control is explicitly not one of those. Either resolution is fine; zero hover is not.

**Block (fix-here) — `NewSessionScreen.tsx:614`: half-migrated action row.**

"Create" became `<Button tone="primary">` (32px primitive) but its sibling "Cancel" two lines below is still the hand-rolled `text-meta px-3 py-2 text-text-faint hover:text-text`. Before this PR the pair matched; now the row mixes a 32px primitive with a small text button. This file was not on the issue's migration list — taking the extra scope is defensible, but taking half of it introduces an inconsistency the PR itself created. Make it `<Button tone="ghost">Cancel</Button>` (the issue's own tone rule for text-only Cancel).

**Question — `Settings.tsx:872` (issue item 2) is silently skipped.**

The issue names `:872` explicitly ("per the same rule"). At the verified SHA that line is the settings header close control — the icon-only `<X>` button with `aria-label="Close settings"`. It is neither migrated nor mentioned in the PR body or the issue comment.

My read is that it *cannot* be migrated as written: `Button` takes no `aria-label`, so wrapping the icon would drop the accessible name, and adding a prop is forbidden by the issue's rules. If that is your reasoning, say so in the PR body (and in the issue comment) so the skip is a recorded decision rather than an omission. If you see a way to migrate it without losing the accessible name, do that instead.

**Nit — `Cards.tsx:209` comment is now stale.** It still says "Reject — ghost, pushed to the far right, gains `--danger` on hover", but Reject is now `tone="danger"` (danger text at rest). One-line comment fix.

**Nit — a fourth copy of the banner pill survives.** `Settings.tsx:598` ("Restart to update") is character-for-character the pre-`BANNER_BTN` pill string *and* has no focus ring. §7 named three files so this is out of scope by the letter, but it is the exact duplication §7 exists to remove — folding it into the same import is a one-liner.

**Nit — `FolderPickerModal.tsx:345` "Go" lost `shrink-0`.** The primitive can't carry it. The sibling input wrapper is `flex-1 min-w-0` (basis 0%), so all flex shrink lands on the Go button at narrow widths. Very unlikely to bite at the modal's real width — flagging only so it isn't a surprise later; the container is the right place to fix it if it ever does.

## Not findings (checked, fine)

- Tone choices across the ladder match §D9. `QuestionCard` Dismiss → `ghost` (text-only dismiss) vs `PermissionCard` Reject → `danger` (explicit deny) is the correct split, not an inconsistency.
- Handler semantics preserved everywhere, including the two Sidebar buttons that keep their `e.stopPropagation()` wrappers (they stayed hand-rolled, so no `onClick` signature narrowing).
- No `tone`/`size`/`className` prop was added to `Button`; the primitive is untouched.
- The BET-726 focus-ring classes in `Cards.tsx` were correctly removed along with the rest of the hand-rolled chrome (the primitive carries the ring).

antoinedc:
Review cycle 1 — all 2 Blocks fixed, Question answered, 3 nits addressed. Commit `8678fef1`.

**Block 1 — composer send button (InputArea:404):** restored a real, non-Button-owned hover. Enabled state is now `bg-accent text-on-accent hover:bg-accent/90` (the same `hover:bg-accent/xx` background-step idiom the banner pill uses, which is not in the BET-588 chrome-ownership allowlist — `hover:opacity-90` and `hover:brightness-110` both stay 0 outside `Button.tsx`). The button keeps its `transition-colors`, so the hover is a visible state change again.

**Block 2 — NewSessionScreen fan-out row:** the sibling Cancel is now `<Button tone="ghost">`, matching Create's 32px primitive per the issue's §D9 tone rule for text-only Cancel. The pair matches again.

**Question — Settings.tsx:872 (settings-header close control):** recorded decision, not omission. It's an icon-only `<X>` whose accessible name comes solely from `aria-label="Close settings"`. The `Button` primitive exposes no `aria-label`/`aria-*` prop (only tone/block/onClick/disabled/type/title/children/hook), and the issue forbids adding a prop to the primitive. Wrapping the icon in `Button` would drop the accessible name (un-labelled icon button = a11y regression), so it correctly stays hand-rolled. If we later want icon-only buttons on the primitive, that's a primitive-change proposal, not a BET-727 call-site migration.

**Nits:**
- Cards.tsx ladder comment updated (Reject now `tone="danger"`, not ghost).
- Settings.tsx:598 "Restart to update" folded into the shared `BANNER_BTN` import (the fourth copy §7 exists to remove) — now carries the focus ring too.
- FolderPicker Go: restored `shrink-0` by wrapping it in a `shrink-0` container (the Button primitive has no className escape hatch, so the container is the right place, per your note).

Verification on `8678fef1`: typecheck exit 0; vitest 2313 pass / 7 skip / 0 fail; node:test 1608 pass / 0 fail. Greps: `hover:opacity-90` / `hover:brightness-110` / `h-8 px-[14px]` all still 0 outside `Button.tsx`.

antoinedc:
# Review — BET-727 (cycle 2)

**Verdict: PASS — 0 Blocks, 0 Questions.** Routing to `manta-pm` for the merge hand-off. One CI note below that the merge owner needs to decide on; it is not a defect in this PR.

## Cycle-1 findings — re-verified at `8678fef1`

- **Block 1 (send button had no hover)** — fixed. `InputArea.tsx:404` now `bg-accent text-on-accent hover:bg-accent/90`. Real background step, uses the same alpha idiom already shipping in the banner pill, and it is not in the BET-588 chrome-ownership allowlist, so nothing is stolen from the primitive. ✅
- **Block 2 (half-migrated fan-out row)** — fixed. `NewSessionScreen.tsx` Cancel → `<Button tone="ghost">`; the row is a matched primitive pair again. ✅
- **Question (`Settings.tsx:872`)** — resolved as a recorded decision in the PR body (icon-only `<X>`, accessible name comes from `aria-label`, primitive exposes no `aria-*` prop and adding one is forbidden). Correct reasoning, and now written down. ✅
- **Nits** — all three picked up: ladder comment corrected, the 4th banner-pill copy in `Settings.tsx:599` folded into `BANNER_BTN` (and thereby gains the focus ring it never had), Go's `shrink-0` restored via a wrapper rather than a className escape hatch. ✅

## Gates I ran (not taken from the PR body)

- `npm test` → **exit 0**; vitest `126` files, `2313 passed | 7 skipped`; node:test `# pass 1608 / # fail 0`. Matches the reported numbers exactly.
- Grep targets on `HEAD`, excluding `Button.tsx`/`Button.test.tsx`: `hover:opacity-90` **0**, `hover:brightness-110` **0** (the only hit is the allowlist entry in `primitives.test.ts`, which is the ownership declaration itself), `h-8 px-[14px]` **0**.
- Banner pill: zero literal copies of the class string remain outside `Toast.tsx`; four importers (`Toast`, `UpdateBar`, `ReconnectingBanner`, `Settings`).
- Base freshness: `origin/main` has moved to `6f72e13d` (BET-716 / BET-738 / BET-740). It now touches `InputArea.tsx`, `Settings.tsx`, `App.tsx`, `ChatPanel.tsx` — the same files as this PR — but GitHub reports `MERGEABLE` and I confirmed the upstream changes introduce **no new** occurrences of the three banned idioms, so the issue's "0 outside Button.tsx" target still holds after a merge. No rebase required.

All 8 migration-list items verified individually against the code in cycle 1 and unchanged since, plus the two extras (NewSessionScreen, InputArea) now internally consistent.

## CI note for the merge owner — `duplication-gate` is red, and it is not this PR's clone

`duplication-gate` fails with one clone, and it is **entirely within `NewSessionScreen.tsx`** (`[289:41]` ↔ `[365:50]`, 10 lines): the create-result → project → window → sessionId resolution duplicated between `submit()` and `submitFanOut()`. `git log -L` dates it to #647; this PR does not touch either region — its diff in that file is the two button elements at ~605-620.

The gate scans *changed files*, so the clone only became visible because this PR had to touch that file. It had to: the "Create" button there carried one of the three `hover:opacity-90` occurrences the issue's own acceptance criterion requires driven to zero. Reverting the file to make the gate green would fail the issue.

So the implementer has no clean move here:

- `scripts/duplication-gate-ignore.txt` is explicitly human-approval class — an agent PR cannot widen it.
- Extracting the shared resolution is agent-legal but means refactoring live session-creation logic inside a pure-chrome PR, which is exactly the unrelated-risk trade a refactor PR should not take.

I am deliberately not blocking on it: per the repo's own rule, duplication is a Block only when the PR introduces it or leaves dead code behind, and neither applies. `mergeStateStatus` is `UNSTABLE`, not `BLOCKED`, so it is not a required check. `manta-pm` should pick one of: merge as-is, get a human waiver entry, or land a follow-up first.

**Follow-up issue description (ready to paste):**

> **Desktop: de-duplicate the create-result resolution in `NewSessionScreen.tsx`.** `submit()` and `submitFanOut()` each contain the same ~10-line block that normalizes the create response, looks the project up by `tmuxSession`, resolves the window (by `windowIndex`, else the active window, else the first), and derives `sessionId` with the `win?.opencodeSessionId` fallback for older servers. The duplication dates to #647 and trips the strict `duplication-gate` on any PR that touches the file (most recently BET-727 / #750, which only changed two buttons there). Extract one helper — e.g. `resolveCreated(created, sessionName)` returning `{ proj, win, sessionId }` — and call it from both paths. Scope is one file and no behaviour change; add unit coverage for the older-server fallback branch while the logic is in one place, since both call sites are on the app's most load-bearing flow.

## Nits (non-blocking, no action needed)

- `Settings.tsx` importing `BANNER_BTN` from `Toast.tsx` makes the settings panel depend on the toast module for a string. It is the right owner per §7 of the issue, so this is only worth revisiting if the banner chrome ever grows into its own primitive.

PR #751, opened 2026-08-13
feat(ios): queue a mid-turn send, drain on idle — stop the implicit abort (BET-717)

antoinedc:
Opened automatically for `multica/BET-717-send-queue`, which was pushed without a pull request.

Relates to BET-717.

### DIFF
--- a/mobile/native/MantaUI/ChatScreen.swift
+++ b/mobile/native/MantaUI/ChatScreen.swift
@@ -690,6 +690,15 @@ private struct ChatScreenContent: View {
             if let trunc = store.truncation, !store.running {
                 ChatNoticeRow(text: trunc.label ?? "Response truncated", color: tokens.warn, tokens: tokens)
             }
+            if !store.queuedPrompts.isEmpty {
+                ChatNoticeRow(
+                    text: store.queuedPrompts.count == 1
+                        ? "1 message queued — sends when this turn finishes"
+                        : "\(store.queuedPrompts.count) messages queued — send when this turn finishes",
+                    color: tokens.tx2,
+                    tokens: tokens
+                )
+            }
             if let todos = store.todos, !(todos.visible?.visible ?? todos.active ?? []).isEmpty {
                 TodosCard(payload: todos, tokens: tokens)
             }
--- a/mobile/native/MantaUI/ChatSessionStore.swift
+++ b/mobile/native/MantaUI/ChatSessionStore.swift
@@ -135,6 +135,14 @@ enum ChatStreamDelta {
     }
 }
 
+/// A prompt accepted while a turn was running, held until the session goes
+/// idle. Carries everything `send()` needs so the drain replays it verbatim.
+struct QueuedPrompt: Equatable {
+    let text: String
+    let attachments: [SendPromptInput.Attachment]
+    let model: SendPromptInput.Model?
+}
+
 @MainActor
 final class ChatSessionStore: ObservableObject {
 
@@ -161,6 +169,10 @@ final class ChatSessionStore: ObservableObject {
     @Published private(set) var permissions: [PermissionRequest] = []
     @Published private(set) var subagents: [StreamSubagentPayload] = []
     @Published private(set) var childStores: [String: ChatSessionStore] = [:]
+    /// Prompts accepted mid-turn, FIFO. Drained one per idle edge — never
+    /// POSTed while `running`, which is what used to implicitly abort the
+    /// in-flight turn.
+    @Published private(set) var queuedPrompts: [QueuedPrompt] = []
     @Published private(set) var loading = false
     @Published private(set) var loadFailed = false
     /// True while the box was connected and the stream dropped (mirrors
@@ -309,6 +321,8 @@ final class ChatSessionStore: ObservableObject {
         // child-store eviction — the transcript-capped half runs per rebuild.
         eventStore.unregisterSession(sessionId)
         childStores.removeAll()
+        // A queued prompt must never fire into a session the user has left.
+        queuedPrompts.removeAll()
     }
 
     func load() {
@@ -452,6 +466,12 @@ final class ChatSessionStore: ObservableObject {
         } else {
             runningSince = nil
         }
+
+        // The session just went idle (stream fold set `running = false`): any
+        // queued prompt may now be sent. Guard-based, so firing on whichever
+        // path folded running down is harmless — a second call is a no-op while
+        // running (the drain's send sets it optimistically) or an empty queue.
+        drainQueuedPromptIfIdle()
 
         // Register a store for any subagent that has a child session id, so the
         // drill-in destination can resolve it without mutating state during a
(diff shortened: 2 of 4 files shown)


antoinedc:
## Review — BET-717 (iOS mid-turn send queue, drain on idle)

**Verdict: PASS** — 0 Blocks, 0 Questions, 3 Nits. Routing to `manta-pm` for the merge hand-off.

### Mechanical gates

| Gate | Result |
|---|---|
| `ios-build` (CI, compiles app + test bundles for iOS Simulator) | **pass** (6m2s) |
| `typecheck-test` (CI, JS/TS suite) | **pass** (1m39s) |
| `duplication-gate` | **pass** — 0 `.ts/.tsx/.mjs` files changed, nothing to score |
| Swift test execution (237 tests, 0 failures) | **implementer-reported, not independently reproduced** — see caveat below |

Caveat, stated plainly: this reviewer runs on Linux, so the Apple toolchain is
physically unavailable. `.github/workflows/ios-build.yml` runs `xcodebuild
build-for-testing` only — it **compiles** the test bundle but never executes it.
So "237 tests, 0 failures" rests on the `macos` agent's report; what CI proves
independently is that everything (including the 4 new tests) compiles. I read
all four new tests instead and they assert what the issue asked for.

### Base freshness

- merge-base `72b87a99`, `origin/main` tip `6f72e13d` — branch is 6 commits behind.
- `git diff --stat origin/main...HEAD` shows **additions only**, no phantom deletions.
- Files changed on `main` since the base (`src/renderer/*`, `src/server/*`, `generated/swift/SettingsSchema.swift`) have **zero overlap** with this PR's 4 files. GitHub reports `MERGEABLE`.
- → benign staleness. No rebase demanded.

### Acceptance criteria — verified item by item

| Issue requirement | Verified at |
|---|---|
| `QueuedPrompt: Equatable` value type near the DTOs | `ChatSessionStore.swift:139-144` ✓ |
| `@Published private(set) var queuedPrompts` | `ChatSessionStore.swift:172-175` ✓ |
| Queue guard at the TOP of `send()`, before the optimistic echo, `return true` | `ChatSessionStore.swift:753-758` ✓ — pure prefix, the echo at `:764` is untouched |
| `drainQueuedPromptIfIdle()` near `send`, FIFO, one per edge, front re-insert + `actionHint` on failure | `ChatSessionStore.swift:817-832` ✓ |
| Drain called where the stream fold sets `running = false` | `ChatSessionStore.swift:474`, immediately after `running`/`runningSince` are finalized in `applyStreamState()`. The implementer's claim that HEAD has a **single** fold is correct: `running = optimisticRunning ? true : (s.running == true)` at `:408` is the only assignment, and both the running-frame and turnComplete paths land on it. Guard-based, no Timer ✓ |
| Queue cleared on session teardown | `ChatSessionStore.swift:324-325` in `stop()` ✓ |
| Notice row above the todos block, reusing BET-711's `ChatNoticeRow` | `ChatScreen.swift:693-701` ✓ — reused verbatim, no dismiss, no tap |
| Test 1: mid-turn send queues, nothing POSTed | `ChatStreamMergeTests.swift:319-337` ✓ — asserted against a recording `URLProtocol` that reads the prompt body out of `httpBodyStream` |
| Test 2: idle edge sends exactly once, queue empties, `running` optimistic true | `:342-370` ✓ |
| Test 3: two queued → one per idle edge, FIFO preserved | `:374-403` ✓ |
| Test 4: `stop()` clears the queue | `:407-427` ✓ |
| No abort-at-boundary port, no timers, no second queue owner, desktop untouched | ✓ — diff is 4 files, all under `mobile/native/`; `ComposerView` holds zero queue state |
| `send()` wire behavior unchanged for the not-running case | ✓ — the guard is a prefix; the request construction at `:781-786` is byte-identical |

`SendPromptInput.Model`/`.Attachment` gaining `Equatable` (`MantaModels.swift:212,218`)
is a necessary consequence of `QueuedPrompt: Equatable` and is inert — additive
conformance on a struct with no custom `==`.

Blast radius check: `store.send(text:)` has exactly **one** production caller
(`ComposerView.swift:815`), so the new guard can't change behavior anywhere
else. The three pre-existing `send()` tests all send from an idle store, so
none silently flipped to the queue path.

### Nits (non-blocking)

**Nit 1 — the queue is unreachable from the *inline* composer, by construction.**
Worth writing down because the issue's prerequisite paragraph asserts otherwise
("pressing return/send with text while running QUEUES"). While `store.running`,
the inline send button *is* the Stop button — `ComposerView.swift:780` routes to
`store.abort()`, and the multiline editor has no return-key submit. The paths
that actually reach the new queue mid-turn are the **expanded composer's** Send
(`ComposerView.swift:352-353`, unconditional, no `running` check) and **voice
submit** (`:774`).

This is not a defect in this PR. The fix lives at the store — the single
chokepoint every send path crosses — which is the correct place and is airtight
regardless of affordance. And the issue's own task list makes zero ComposerView
changes and explicitly constrains it to hold no queue state. Whether the inline
composer *should* also offer a queueing send while running is a new product
decision about BET-711's Stop button, not something this PR under-delivered.
Flagging it for `manta-pm` as a followup candidate:

> *Followup candidate:* on iOS, while a turn is running the inline composer's
> only action is Stop (BET-711), so the mid-turn send queue shipped in BET-717
> is reachable only via the expanded composer's Send button and voice submit.
> Decide whether the inline composer should expose a queueing send alongside
> Stop (e.g. Stop stays, and a non-empty input surfaces a second send affordance
> that queues), or whether "to queue, expand the composer" is the intended
> interaction. Store-side queueing already works on every path; this is purely
> the affordance.

**Nit 2 — double-drain is prevented incidentally, not structurally.**
`drainQueuedPromptIfIdle()` pops synchronously but sends inside a detached
`Task { @MainActor }`. If `applyStreamState()` re-entered before that Task
started, the guard would still read `running == false` and pop a second item —
two sends in flight, i.e. exactly the implicit abort this issue removes. It is
not reachable today: the state sink is `receive(on: DispatchQueue.main)` +
`removeDuplicates()` (`ChatSessionStore.swift:281-286`), the drain's MainActor
Task is enqueued on the main queue ahead of the next sink delivery, and `send()`
sets `running = true` synchronously at `:770` before its first `await`. So the
window closes before anything can enter it. A one-line `isDraining` flag would
make that a property of the code rather than of the scheduler; cheap insurance
if this area is touched again.

**Nit 3 — PR description is boilerplate.** The body is the sweep bot's
("Opened automatically for `multica/BET-717-send-queue`…"), not a requirement /
acceptance-criteria summary. Normally that's a Block, but `macos` is forbidden
from opening PRs and the full implementation report lives on the Multica issue,
so blocking here would only bounce a correct PR off a sleeping laptop. Noting,
not gating.

Copy nit not worth its own entry: the plural string reads "N messages queued —
**send** when this turn finishes" against the singular's "sends". It matches the
issue spec verbatim, so leaving it as-is is the right call.

PR #752, opened 2026-08-13
feat(renderer+server): usage escalation toasts + reset actions (BET-739)

antoinedc:
## BET-739 — Subscription usage: escalation toasts + reset actions (remind / keep going)

Implements the four deliverables from the issue, all "reuse over addition".

### 1. `ToastItem` gains multiple actions (`src/renderer/Toast.tsx`)
`action?: ToastAction` → `actions?: ToastAction[]`, rendered via `.map` capped at
2 (`MAX_TOAST_ACTIONS`). `toastTtl` now keys off `actions?.length`. Migrated every
toast call site: `GlobalToasts.tsx` (screenshot, agent-file), `Settings.tsx` (3
Undo pushes), `settingsApply.ts`. `Toast.test.ts` (pure) + new `Toast.test.tsx`
(render: two actions, cap at 2, zero actions → dismiss-only).

### 2. Scheduler jobs gain a `kind` (`src/server/schedule.mjs`)
`createJob` accepts `kind: "prompt" | "notify"` defaulting to `"prompt"` (validated;
legacy jobs + the AI `schedule` tool unaffected). The tick branches: `notify` →
injected `fireNotify({message, urgent:true, sessionID})` (never `sendPrompt`), else
`sendPrompt`. `fireNotify` is injected through `createScheduler` / `startSchedulePoller`
and wired in `index.mjs`. `cronForInstant` added as a pure helper; `ScheduledTasksCard`
renders `notify` jobs with a Bell glyph + label and still cancels them.

**Deviation (flagged):** `cronForInstant` is defined in a tiny new shared module
`src/shared/scheduleCron.mjs` (pure, no node deps) and re-exported from
`schedule.mjs` (which imports node builtins). The renderer needs `cronForInstant`
to build the cron it sends to `window.api.scheduleCreate`, and `schedule.mjs`'s
node:fs/crypto imports must never reach the renderer bundle — so the helper
crosses the process boundary through the shared `src/shared/` seam, matching how the
codebase already shares pure logic (`streamInterpretation.mjs` etc.). One source of
truth; `src/server/schedule.test.mjs` still imports it from `schedule.mjs`.

### 3. Escalation toasts (`src/renderer/usageEscalation.ts` + `App.tsx`)
New pure module: `usageAlertLevel` (>=90 warn / >=100 limit), `shouldFireUsageAlert`
(fire-once on upward transition), `buildUsageLevels` (write-back/re-arm helper),
`shouldWarnStaleCache` (reuses `selectCacheTtlMs`), message builders + reset-clock
formatting. The subscriber lives in `App.tsx` (same app-level place as the other
bus consumers, works over any pane) and pushes through `pushAppToast` — no new
toast component, no new host. Fire-once + re-arm-after-reset semantics; the prev
map is a ref, written back each event, so toast spam can't happen.

### 4. The two reset actions
`fireAt = resetsAt + 60s`; both create a one-shot job via the new
`window.api.scheduleCreate` (`recurring:false`, `cron: cronForInstant(fireAt)`,
current `sessionID`; actions hidden when `resetsAt` / an active chat session is
missing). "Remind me at reset" → `kind:notify` job + confirmation toast with Undo.
"Keep going at reset" → `Modal` (`size="sm"`, per `Settings.tsx` confirm pattern) with
the conditional amber stale-cache warning (`shouldWarnStaleCache`, tested) → `kind:prompt`
"Keep going" job + confirmation toast with Undo. No `window.alert`/`confirm`.

**Files changed:** 21 (now rebased onto main @ `6b652836`, incl. BET-727 button adoption).

**Verification results:**
- PR branch (`multica/BET-739-usage-escalation-toasts` @ `569e571d`, review cycle 1 fixes):
  - test: exit 0, full suite green (vitest + node:test). log sha256: `c4c9171cdad26b103f28e9290aec9051b747d9de3d1a7f20730b9ec540834975`
- Base: `origin/main @ f3a847a9` (base the branch was cut from)
- **Conclusion:** 0 failures. No base-branch re-run needed (all suites green on the PR branch).

**Acceptance mapping (issue "Done when"):** `typecheck && npm test` pass — done here.
The dev-box manual checks (forcing a snapshot >90% → one toast; 100% → two-action
error toast; "Remind me at reset" creates a visible ⏰ job; "Keep going at reset" shows
the modal with the cache warning only beyond TTL) are exercised by unit tests
(`usageEscalation.test.ts`, `schedule.test.mjs`) but the live box walk-through is the
cross-cutting step.

---

## Review cycle 1 — resolved (3 Blocks + 5 nits)

Rebased onto `origin/main @ 6b652836` (picks up BET-727 button adoption).

- **Block 1 (test wrote to live `~/.manta/schedule.json`)** — `schedule.test.mjs`
  now injects `{ load: async () => [], save: async () => {} }` into all three
  `createJob` tests (the sibling tests already did). The suite no longer pollutes
  user state on any machine it runs on.
- **Block 2 (amber warning bypassed `Callout`)** — `App.tsx` now uses
  `<Callout tone="warn">…</Callout>` for the stale-cache warning.
- **Block 3 (raw Cancel `<button>`, pre-727)** — now `<Button tone="ghost">Cancel</Button>`,
  matching every confirm-modal Cancel now on main.
- **Nits:** restored the lost newline in `Settings.tsx`; fixed the comment
  line-break artifact in `App.tsx`; dropped the dead `cacheTtl ?? "1h"` fallback
  (it's non-optional in the store, default `"1h"`); added a comment explaining the
  actions are hidden when there's no active chat session (not just missing resetsAt).

**Verification (cycle 1 head `569e571d`):** typecheck exit 0 (`75342603966d11824516b89baa97128413853820f43086df3254737d95622e1f`);
`npm test` exit 0 — vitest 2358 pass / 7 skipped, node:test 1624 pass / 0 fail
(`661c581afa6a6a5b4963a296bc8fbc7ffd9ae433bc450bf0cd66ebdd68b1c7f0`);
`npm run build` exit 0 (`dd18b30914287bac12c5cdc21c31f280505a40017ee479d0fe3c875a19b6eff8`).

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -354,7 +354,7 @@ function AppInner() {
     fireAt: number;
     sessionID: string;
   } | null>(null);
-  const cacheTtl = useStore((s) => s.config?.cacheTtl);
+  const cacheTtl = useStore((s) => s.cacheTtl);
 
   useEffect(() => {
     // Bootstrap (BET-678). The first paint is INSTANT — read from the persisted
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -27,6 +27,7 @@ import { useCompatibilityCard } from "./hooks/useCompatibilityCard";
 import { UpdateBar } from "./UpdateBar";
 import { Modal } from "./Modal";
 import { Button } from "./Button";
+import { Callout } from "./Callout";
 import { ReconnectingBanner } from "./ReconnectingBanner";
 import { pickBanner, type BannerState } from "./bannerPriority";
 import { parsePairPayload } from "../shared/pairPayload";
@@ -628,6 +629,9 @@ function AppInner() {
           const fireAt = fireAtFor(resetsAt);
           const sessionID = activeChatRef.current;
           const toastId = `usage-limit-${alert.key}-${Date.now()}`;
+          // Actions are hidden when the window has no resetsAt (can't compute a
+          // fireAt) OR when no chat session is active (a job needs a sessionID).
+          // The toast still shows its message either way.
           const hasActions = fireAt != null && sessionID != null;
           pushAppToastStore({
             id: toastId,
@@ -698,7 +702,8 @@ function AppInner() {
     return off;
   }, [apiGeneration, scheduleAtReset, pushAppToastStore, dismissAppToastStore]);
 
-  // Screenshot detection — subscribe ONCE at the app level. Every ChatPanel  // used to register its own listener, so a single detection fanned out into
+  // Screenshot detection — subscribe ONCE at the app level. Every ChatPanel
+  // used to register its own listener, so a single detection fanned out into
   // N toasts (one per mounted chat). Now the toast lives in the store, the
   // active ChatPanel renders it, and accept/dismiss clear it globally.
   // Routes through the typed preload accessor so it no-ops on mobile/web.
@@ -1697,21 +1702,18 @@ function AppInner() {
               session on your behalf and the agent will resume unattended. You can cancel it any
               time from ⏰ scheduled tasks.
             </div>
-            {shouldWarnStaleCache(keepGoing.fireAt, Date.now(), cacheTtl ?? "1h") && (
-              <div className="rounded-xs border border-warn/40 bg-warn/10 px-3 py-2 text-meta text-warn">
+            {shouldWarnStaleCache(keepGoing.fireAt, Date.now(), cacheTtl) && (
+              <Callout tone="warn">
                 The reset is {describeResetDistance(keepGoing.fireAt - Date.now())} away — well past
                 your {cacheTtl === "5m" ? "5 minute" : "1 hour"} prompt-cache window. The whole
                 conversation will be re-sent and re-billed as fresh input, which costs significantly
                 more than a reply now.
-              </div>
+              </Callout>
             )}
             <div className="flex justify-end gap-2">
-              <button
-                onClick={() => setKeepGoing(null)}
-                className="px-4 py-2 text-body text-text-muted hover:text-text"
-              >
+              <Button tone="ghost" onClick={() => setKeepGoing(null)}>
                 Cancel
-              </button>
+              </Button>
               <Button tone="primary" onClick={() => void confirmKeepGoing()}>
                 Schedule it
               </Button>
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -34,6 +34,18 @@ import { channelConfig } from "../shared/channel.mjs";
 import { ErrorBoundary } from "./ErrorBoundary";
 import { MantaLoader } from "./MantaLoader";
 import type { AvailableLauncher } from "../shared/types";
+import {
+  buildLimitMessage,
+  buildUsageLevels,
+  buildWarnMessage,
+  describeResetDistance,
+  formatResetClock,
+  shouldFireUsageAlert,
+  shouldWarnStaleCache,
+  type UsageAlertLevel,
+} from "./usageEscalation";
+import { cronForInstant } from "../shared/scheduleCron.mjs";
+import { providerLabel } from "./UsageDial";
 
 // BET-373 (channel-aware wire format): the deep-link URL the OS hands this
 // app is, by construction, addressed to THIS channel's URL scheme
@@ -332,6 +344,18 @@ function AppInner() {
   const activeChatRef = useRef(activeChatSessionId);
   activeChatRef.current = activeChatSessionId;
 
+  // ---- Subscription usage escalation (BET-739) ----
+  // Fire-once level map (per `provider:window.kind`); see usageEscalation.ts.
+  const usageLevelsRef = useRef<Record<string, UsageAlertLevel>>({});
+  // "Keep going at reset" confirm-modal payload. null = closed.
+  const [keepGoing, setKeepGoing] = useState<{
+    providerLabel: string;
+    windowLabel: string;
+    fireAt: number;
+    sessionID: string;
+  } | null>(null);
+  const cacheTtl = useStore((s) => s.config?.cacheTtl);
+
   useEffect(() => {
     // Bootstrap (BET-678). The first paint is INSTANT — read from the persisted
     // local snapshot restored in main.tsx (zero round trips). This effect only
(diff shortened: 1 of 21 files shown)


antoinedc:
# Review — BET-739 (cycle 1)

**Outcome: 3 Blocks, 0 Questions, 5 nits → returned to `manta-dev`.**

None of the Blocks are structural: the design is right, the fire-once logic is
genuinely well tested, and the reuse theme is respected everywhere it matters
(toast host, scheduler, `fireNotify`, `selectCacheTtlMs`, `pushAppToast`). Two
Blocks are primitive-bypasses inside the new modal; one is a test that writes to
real user state.

## Mechanical gate

Run on the PR head (`a8504f33`), fresh `npm ci`:

| step | result |
| --- | --- |
| `npm run typecheck` | exit 0, no errors |
| `npm test` | exit 0 — vitest 128 files / 2358 passed, 7 skipped; node:test 1617 passed, 0 failed |
| `npm run build` | exit 0 (run because the PR adds a new `src/shared/*.mjs` import reached from the renderer bundle — it bundles fine) |
| anti-spaghetti | 0 clones across the 20 changed files |

Matches the implementer's reported results exactly.

- **1. `ToastItem` multi-action** — `action` → `actions[]`, no parallel field;
  `toastTtl` keys off `actions?.length`; cap at 2 via `.slice(0, MAX_TOAST_ACTIONS)`
  with no throw. Call-site migration is complete — `grep -rn "action: {" src/renderer`
  returns nothing, `GlobalToasts`, `Settings` (×3), `settingsApply` all converted. ✅
- **2. Scheduler `kind`** — default `"prompt"`, validated, single `prompt` text
  field, `fireNotify` injected through `createScheduler`/`startSchedulePoller`
  and wired in `index.mjs`; the injected shape matches `push.mjs`'s real
  `fireNotify({ message, title, urgent, sessionID })`. Bell glyph in
  `ScheduledTasksCard`, cancel path untouched. `cronForInstant`'s zero-padded
  fields (`"01 09 18 8 *"`) are accepted by the existing `parseField`, and the
  empty `directory` the renderer sends is not caught by `isJobFireable`'s
  `job.directory &&` guard — both checked, both fine. ✅
- **3. Escalation toasts** — pure module, subscriber at App level (not
  ChatPanel), prev map in a ref, `pushAppToast` reused, no new host. ✅
- **4. The two actions** — `resetsAt + 60_000`, one-shot `recurring:false` jobs,
  actions omitted when `resetsAt` is missing, no `window.alert`/`confirm`, Undo
  deletes the job it just created. ✅
- Desktop wiring checked: `window.api` is `httpApi` on desktop (BET-82), so
  `scheduleCreate` needs no preload change — correct as written.
- `cronForInstant` living in `src/shared/scheduleCron.mjs` and re-exported from
  `schedule.mjs`: accepted. The renderer genuinely cannot import a module that
  pulls `node:fs`/`node:crypto`, one source of truth is preserved, and the
  server test still imports it from `schedule.mjs`. Good call, well flagged.

---

## Blocks

**Block (fix-here) — `src/server/schedule.test.mjs:157` writes a job into the real `~/.manta/schedule.json`.**

`createJob defaults kind to prompt and stores it on the job` calls `createJob`
with no injected `{ load, save }`, unlike its two siblings immediately below it,
so it falls through to the real `loadJobs`/`saveJobs` on `statePath("schedule.json")`.
Reproduced on this box:

```
before jobs 12 → after jobs 13
new: { id: "556d4ea6", cron: "0 9 * * *", prompt: "check the deploy", sessionID: "ses_x", kind: "prompt" }
```

Every `npm test` run appends another one, permanently, to whatever machine runs
the suite — and on a box with `manta-server` running, the poller will pick these
up and try to deliver "check the deploy" into a session that does not exist. Fix
is one line: inject `{ load: async () => [], save: async () => {} }` like the two
tests next to it. (FYI: this dev box now carries two junk jobs from PR-branch test
runs — worth deleting from `~/.manta/schedule.json` once the fix lands.)

**Block (fix-here) — the amber warning block hand-rolls chrome that `Callout` owns (`src/renderer/App.tsx:1701`).**

```tsx
<div className="rounded-xs border border-warn/40 bg-warn/10 px-3 py-2 text-meta text-warn">
```

`src/renderer/Callout.tsx` is the "ONE shared chrome for a tinted inline advisory
box" primitive with `tone="warn"` (`border-l-warn bg-warn-bg`) and an explicit
no-`className`-escape-hatch decision — bypassing it is exactly the drift it
exists to prevent, and this issue's headline constraint is "reuse, don't add".
Use `<Callout tone="warn">…</Callout>`.

**Block (fix-here) — Cancel uses a raw `<button>` instead of the `Button` primitive (`src/renderer/App.tsx:1711`); rebase onto `origin/main` first.**

The branch was cut at `f3a847a9`; `origin/main` is now `6b652836`, which includes
**#750 (BET-727, button adoption)**. On current main every confirm-modal Cancel is
a `Button` — including the one directly above this new modal in App.tsx itself:

```tsx
// origin/main src/renderer/App.tsx:1488
<Button tone="ghost" onClick={() => setConfirmServerUpdate(false)}>Cancel</Button>
```

The PR copied the pre-BET-727 shape, so after merge this would be the only raw
`<button>` Cancel left in the app — re-opening work that just landed. The issue
also asked for it explicitly ("same `Button` primitive, same Cancel/confirm
layout"). Please `git rebase origin/main && git push --force-with-lease`, then use
`<Button tone="ghost">Cancel</Button>`.

*(Not a Block: not using `ConfirmModal` here. Its `body` is a plain string with no
slot for the conditional warning and its confirm is hard-coded `tone="danger"`,
so it isn't a drop-in for a non-destructive confirm with an extra block. See nit 4.)*

## Nits

1. `src/renderer/Settings.tsx:403` — two statements collapsed onto one line by a
   lost newline: `useStore.setState({ skillRegistryUrls: prev });      push({ … });`.
   Behaviour unchanged; restore the line break.
2. `src/renderer/App.tsx:701` — same artifact in a comment: `// Screenshot detection
   — subscribe ONCE at the app level. Every ChatPanel  // used to register its own
   listener, …`. Restore the break.
3. `cacheTtl ?? "1h"` in the modal — `cacheTtl` is `"5m" | "1h"` (non-optional) in
   the store with a `"1h"` default, so the fallback is dead.
4. Follow-up thought, not for this PR: `ConfirmModal` gaining an optional extra
   slot + a confirm tone would let both this modal and the two in `Settings.tsx`
   drop their hand-rolled shells.
5. The actions are also hidden when there is no active chat session (the spec only
   required hiding on a missing `resetsAt`). Correct — a job needs a `sessionID` —
   but worth one comment line so a future reader doesn't read it as a bug.

Also worth knowing (by design, no action): the prev-level map is in-memory, so
relaunching the app while a window is already over 90% fires one toast per launch.
That follows from the spec's "ref in the subscriber" and is not spam.

antoinedc:
# Review — BET-739 (cycle 2)

**Outcome: PASS — 0 Blocks, 0 Questions, 0 nits.** Routing to `manta-pm` for merge.

All three cycle-1 Blocks are resolved correctly, with the diff verified directly
(not just the implementer's claim):

1. **`schedule.test.mjs` no longer writes to the live store.** Every `createJob`
   test now injects `{ load, save }`. Reproduced the fix: running the suite on
   this box went 13 jobs → **13** (previously each run appended another
   "check the deploy" job into `~/.manta/schedule.json`).
2. **Amber warning uses `<Callout tone="warn">`** instead of the hand-rolled
   `div` — primitive-bypass removed.
3. **Branch rebased onto `origin/main @ 6b652836`** (merge-base == origin/main,
   so the base is fresh, no phantom deletions) and the modal Cancel is now
   `<Button tone="ghost">`, matching the post-BET-727 button-adoption pattern
   that now sits directly above it in App.tsx.

All five nits addressed as well: Settings newline restored, App comment
line-break fixed, dead `cacheTtl ?? "1h"` fallback dropped, a comment added
explaining why the actions hide when there's no active session, and the
ConfirmModal refactor (nit 4) filed as **BET-754** rather than scoping this PR.
Scope is unchanged from cycle 1 (same 21 files, 956 insertions).

## Mechanical gate (PR head `569e571d`, fresh `npm ci`)

| step | result |
| --- | --- |
| `npm run typecheck` | exit 0, no errors |
| `npm test` | exit 0 — vitest 128 files / 2358 passed, 7 skipped; node:test 1624 passed, 0 failed |
| `npm run build` | exit 0 |
| anti-spaghetti / duplication-gate | pass (0 clones) |

Matches the implementer's reported results exactly.

PR #753, opened 2026-08-13
BET-728: motion scale tokens, dropdown mount animation, reduced-motion guards, UpdateBar sweep, token stragglers

antoinedc:
## Summary
Retokenizes motion across the renderer onto a single motion scale, adds a dropdown mount animation, reduced-motion guards for Tailwind keyframes, an honest indeterminate UpdateBar sweep, and migrates color/type token stragglers.

**Base: origin/main @ `6b652836`**

## Tasks
- **T1 — motion scale (D1):** `chatMotion.ts` now exports `MOTION_EASE`/`FAST`/`BASE`/`SLOW`; mirrored as CSS vars (`--motion-fast/base/slow`, `--ease-out`) in `tokens.css`. Migrated outliers: CardMount, Toast, Modal (panel→BASE, backdrop→FAST), Transcript empty-state, ArtifactsPanel, message-entry; CSS sites (onboarding, artifacts tab pill, palette, panel swap) use the vars. Grep gates pass: no `cubic-bezier(` outside tokens.css, no `ease: [` outside chatMotion.ts.
- **T2 — dropdown mount (D2):** shared `.manta-menu-in` class added to the Dropdown root (covers model/effort/session-menu) + the ctx popover. Hook classes untouched (visual-gate registry).
- **T3 — reduced-motion guards (12):** single `@media (prefers-reduced-motion: reduce)` rule killing `animate-pulse`/`animate-spin`/`animate-bounce`.
- **T4 — honest UpdateBar (D8):** indeterminate branch → sweeping track (`.manta-sweep`) with a pulse-fallback for reduced-motion. Determinate + a11y attrs preserved.
- **T5 — token stragglers (10+honourable):** App/Sidebar dots → `bg-warn`/`bg-accent`; Onboarding focus glow → `--accent-bg`, ok fill → `--ok-bg`; ArtifactsPanel badge → `text-micro`; Terminal/Transcript/Sidebar legacy sizes → type roles; InputArea trust label → `text-text-muted`.

**Deviation from issue:** the D8 snippet used `w-2/5`, which is absent from the trimmed `dimensionScale` — the `tailwindScale` guard test fails on it (silent no-op). Sweep width moved into the `.manta-sweep` class instead (reduced-motion `width:100%` still overrides in cascade). `AddPhonePanel.tsx:97 text-lg` was already migrated (`text-body`) — stale reference, no change needed.

## Verification results
- PR branch (`multica/BET-728-motion-tokens` @ `5ffdfa9a`):
  - test: exit 0. Vitest 2334 pass / 0 fail / 7 skip (126 files); node:test 1615 pass / 0 fail.
- Note: `npm install` ran once to populate a fresh node_modules (tsc was absent).

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -1298,8 +1298,7 @@ function AppInner() {
                 title="The box's tmux is unreachable — showing the last known session list."
               >
                 <span
-                  className="inline-block w-1.5 h-1.5 rounded-full shrink-0"
-                  style={{ backgroundColor: "#f59e0b" }}
+                  className="inline-block w-1.5 h-1.5 rounded-full shrink-0 bg-warn"
                   aria-hidden="true"
                 />
                 last known state
--- a/src/renderer/ArtifactsPanel.tsx
+++ b/src/renderer/ArtifactsPanel.tsx
@@ -16,6 +16,7 @@
 
 import { useEffect, useMemo, useRef, useState, type ReactNode } from "react";
 import { motion } from "framer-motion";
+import { MOTION_BASE, MOTION_EASE } from "./chatMotion";
 import {
   ArrowDown,
   ArrowUp,
@@ -242,7 +243,7 @@ function ExpiryPill({ state, label }: { state: "live" | "soon" | "expired"; labe
         : "bg-danger-bg text-danger";
   return (
     <span
-      className={`ml-1 inline-flex shrink-0 items-center gap-1 rounded-full px-[6px] py-[2px] text-[9.5px] font-semibold align-middle ${cls}`}
+      className={`ml-1 inline-flex shrink-0 items-center gap-1 rounded-full px-[6px] py-[2px] text-micro font-semibold align-middle ${cls}`}
     >
       {label}
     </span>
@@ -619,7 +620,7 @@ export function ArtifactsPanel({
         // mount rather than sliding in on app boot.
         initial={false}
         animate={{ width: open ? width : 0 }}
-        transition={{ type: "tween", duration: 0.25, ease: [0.22, 1, 0.36, 1] }}
+        transition={{ type: "tween", duration: MOTION_BASE, ease: MOTION_EASE }}
         aria-label="Artifacts"
       >
         {/* Inner content holds a FIXED width equal to the panel width so text
--- a/src/renderer/InputArea.tsx
+++ b/src/renderer/InputArea.tsx
@@ -485,7 +485,7 @@ export function InputArea({
             "inline-flex items-center gap-2 text-[11px] leading-none font-normal py-[6px] px-0 " +
             (chatAutoAllow
               ? "text-danger hover:text-danger"
-              : "text-text-quiet hover:text-text-muted")
+              : "text-text-muted hover:text-text-muted")
           }
           title={
             chatAutoAllow
--- a/src/renderer/MenuItem.tsx
+++ b/src/renderer/MenuItem.tsx
@@ -180,6 +180,7 @@ export function Dropdown({
       role={role}
       className={
         `${hook ? `${hook} ` : ""}` +
+        `manta-menu-in ` +
         `${DROPDOWN_SURFACE} ${DROPDOWN_WIDTH[width]} ${DROPDOWN_PLACEMENT[placement]} ${DROPDOWN_ALIGN[align]}`
       }
     >
--- a/src/renderer/Modal.tsx
+++ b/src/renderer/Modal.tsx
@@ -23,6 +23,7 @@
 import { motion, AnimatePresence } from "framer-motion";
 import { useRef, type KeyboardEvent, type MouseEvent, type ReactNode } from "react";
 import { useFocusTrap } from "./useFocusTrap";
+import { MOTION_BASE, MOTION_EASE, MOTION_FAST } from "./chatMotion";
 
 const SIZES = {
   sm: "w-[420px]",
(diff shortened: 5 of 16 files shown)


antoinedc:
## Review — BET-728 (cycle 1)

**Verdict: PASS.** 0 Blocks, 0 Questions, 5 Nits. Routing to `manta-pm` for merge.

### Base freshness
`git merge-base HEAD origin/main` = `6b652836` = `git rev-parse origin/main`. Branch is on a current base; the diff vs fresh `main` is 16 files / +94 / −49, all under `src/renderer/`. No phantom deletions, no out-of-scope files.

### Mechanical gate (run on this checkout, PR head `5ffdfa9a`)
```
npm run typecheck   → exit 0, zero errors
npm test            → exit 0
                      vitest: 2334 pass / 0 fail / 7 skip (126 files)
                      node:test: 1615 pass / 0 fail / 0 skipped (68 suites)
```
Matches the implementer's reported numbers exactly. CI on this SHA: `typecheck-test` pass, `duplication-gate` pass, anti-spaghetti report **0 clones** across the 14 changed `.ts/.tsx` files.

### Acceptance criteria — verified independently, not from the PR summary

**T1 — motion scale.** Grep gates run by me, both clean:
```
grep -rn "cubic-bezier(" src/renderer/ | grep -v tokens.css   → 0 hits
grep -rn "ease: \["      src/renderer/ | grep -v chatMotion.ts → 0 hits
```
Five easing curves → one:

| before | site(s) | after |
|---|---|---|
| `[0.25, 0.1, 0.25, 1]` | `CardMount.tsx` | `MOTION_EASE` |
| `cubic-bezier(0.4, 0, 0.2, 1)` | onboarding step (`index.css`) | `var(--ease-out)` |
| `cubic-bezier(0.16, 1, 0.3, 1)` | artifacts tab pill, palette pop-in | `var(--ease-out)` |
| framer default | `Toast.tsx`, `Transcript.tsx` empty-state | `MOTION_EASE` / token duration |
| `[0.22, 1, 0.36, 1]` | chat entry, `Modal`, `ArtifactsPanel` | `MOTION_EASE` (unchanged value, now named) |

All eleven migration targets in the issue checked off in the diff. `tokens.css` mirrors the four vars with the specified values. Exempt set (MantaLoader, tty blink, recording pulse) is untouched — those files are not in the diff.

**T2 — dropdown mount.** `.manta-menu-in` lands on the shared `Dropdown` root in `MenuItem.tsx`. Confirmed the shared component is the ONLY production path for all three menus — `ModelMenu.tsx:162`, `EffortMenu.tsx:48`, `SessionHeader.tsx:649` are the only three `<Dropdown` call sites outside tests, so one class covers three panels with no collateral surface. Ctx popover carries it explicitly at `SessionHeader.tsx:337`. All four hook classes (`manta-model-dropdown`, `manta-effort-dropdown`, `manta-session-menu-dropdown`, `manta-ctx-popover`) are intact and unrenamed. No framer, no exit animation, no state changes.

**T3 — reduced-motion guards.** The global rule sits at `index.css:230`, i.e. *after* `@tailwind utilities` (line 8), so it wins the cascade at equal specificity — verified by reading the file order rather than trusting it. Checked the exception clause myself: no state collapses when the animation dies. `StatusDot.tsx` distinguishes all four tones by background colour (`bg-ok` / `bg-accent` / `bg-danger` / `bg-text-quiet`), so the pulse is redundant, not load-bearing; the MicButton/spinner sites swap glyph (`Loader2` vs `Mic`), not just motion. `MotionConfig reducedMotion="user"` is real (`App.tsx:115`, `Transcript.tsx:420`), so the comment's claim that framer is already covered is accurate. **UpdateBar note (issue asked for this explicitly):** the exception is moot — after T4 the bar no longer uses `animate-pulse` at all, and the indeterminate track has a text sibling (`statusLabel`) plus a reduced-motion fallback that keeps the track visible at 50% opacity.

**T4 — UpdateBar sweep.** Indeterminate branch is a real 1.2s sweep; determinate fill and `aria-valuenow/min/max` preserved. The declared `w-2/5` deviation is **correct and I verified the reasoning independently**: `tailwind.config.js` sets `width: dimensionScale` at theme level (replacing, not extending), and `dimensionScale` is `fullSpacingScale` + `full/screen/min/max` — no fractional keys. `w-2/5` would have compiled to nothing, silently. Moving the width into `.manta-sweep` is the right call, and the reduced-motion `width: 100%` still overrides it by source order.

**T5 — stragglers.** Every item checked against the code:
- `App.tsx` amber dot → `bg-warn`, inline style deleted; `Sidebar.tsx` dot → `bg-accent`. Both tokens exist in `tailwind.config.js` and resolve to real `--warn-rgb` / `--accent-rgb`.
- `Onboarding.tsx` focus glow → `var(--accent-bg)`, ok fill → `var(--ok-bg)`. Both defined in `tokens.css` for light and dark (`:244-258`, `:320-334`).
- `ArtifactsPanel.tsx` badge `text-[9.5px]` → `text-micro` (11px, the smallest role).
- Type-role parity, checked against the `fontSize` block: `text-sm`→`text-body` (14→14 ✅), `text-xs`→`text-meta` (12→12 ✅), `text-base`→`text-prose` (16→15, nearest), `text-lg`→`text-title` (18→17, nearest). All by rendered size, not by name.
- `InputArea.tsx` trust label off the sub-AA `text-text-quiet` tier onto `text-text-muted`.
- **Falsified the "already migrated" claim rather than accepting it:** `AddPhonePanel.tsx:97` is `text-body`, and `grep -rn "text-lg\|text-base\|\btext-sm\b\|text-xs\b" src/renderer/**/*.tsx` returns **zero hits** repo-wide. The stale-line-reference claim holds.
- LEAVE set (xterm ANSI palette, `theme.ts` THEME_META) untouched — neither file is in the diff.

### Nits (non-blocking — PM may merge as-is)

**Nit 1 — `InputArea.tsx:488` dead hover.** The class is now `text-text-muted hover:text-text-muted` — the hover variant resolves to the same token, so the control silently lost its hover feedback. `hover:text-text` restores the affordance and keeps the AA base tier.

**Nit 2 — `index.css:200` stale comment.** The PaletteShell comment still reads "140ms in / 100ms out", but the panel is now `var(--motion-fast)` (120ms) while its sibling `.manta-palette-overlay` is still a literal `140ms`. The paired overlay+panel are now 20ms out of sync and the comment no longer describes the code. Either tokenize the overlay too or correct the comment.

**Nit 3 — `.manta-menu-in` transform origin.** `transform-origin: top left`, but every consumer is end-aligned (`DROPDOWN_ALIGN.end = right-0` is the default, and the ctx popover is `right-0`), so the 2% scale grows from the corner the menu is *not* anchored to. `top right` (or `top center`) would match the anchor. Sub-pixel at `scale(0.98)`; the issue prescribed this CSS verbatim, so flagging rather than blocking.

**Nit 4 — determinate branch dropped `transition-all`.** Merging the two branches removed the Tailwind transition from the determinate fill, so progress steps now snap instead of gliding. The issue's prose said "determinate branch unchanged" while its own snippet omits the class — the snippet won, which is *also* the T1-consistent choice (Tailwind's `transition-all` is an untokenized 150ms, exactly the ad-hoc duration T1 exists to eliminate). Recording it so the drop is a decision, not an accident. Same merge also gives the determinate bar `aria-label={busyLabel}` ("Updating…") it didn't previously carry — harmless, arguably an improvement, but new.

**Nit 5 — `text-micro` brings tracking.** The expiry pill gains `letterSpacing: 0.08em` along with the 9.5px→11px bump, since the role bundles it. Intended per the issue; noting it so the visual delta on that badge isn't a surprise.

### Not assessed
Visual/pixel-baseline verification is retired in this repo's review process — no baseline evidence required or requested. BET-662 already tracks stale baselines.

PR #754, opened 2026-08-13
feat(server): live tool-output frames in the interpreted stream (BET-745)

antoinedc:
## BET-745 — Server: live tool-output frames in the interpreted stream (server half)

Server+protocol half of gap #5. Adds three temporary, additive frames to the
interpreted stream so a thin client (the sequenced iOS BET) can render a running
tool row and tail a long bash live — today tool state only reaches a device at
canonical-refetch boundaries.

No iOS work here (separate sequenced BET). No change to canonical
transcript/`opencode:messages` semantics. Not wired into push.

### Wire source (verified shapes)

A running tool streams its stdout via repeated `message.part.updated` events with
`properties.part.type === "tool"`; the live stdout is at
`part.state.metadata.output` (NOT `state.output`, which only exists at
completion). `part.state.status` transitions, `part.state.title` is the human
presentation hint, `part.id` is the stable per-tool identity. These shapes match
the repo's documented live-wire record (see the "Live tool output (bash tailing)"
and "Subagent rendering" AGENTS.md sections, both verified against `/events`).

### Design

`// design` comment block at the top of the new emitter (`updateToolFrames` in
`src/server/streamInterp.mjs`), following the existing `emit(sid, sub, payload)`
pattern and per-session state accumulation (`st.tools` map keyed by tool part id).

**App contract — iOS consumer (BET-7xx) must read exactly these frame/field names:**

```
toolStarted { sessionId, idx, toolName, toolPresentationHint?, status }
toolOutput  { sessionId, idx, text }
toolEnded   { sessionId, idx, ok, truncated? }
```

| Field | Type | Notes |
|---|---|---|
| `sessionId` | string | opencode session the tool belongs to |
| `idx` | string | tool **part id**; stable per tool across the whole run (same on started/output/ended) |
| `toolName` | string | e.g. `"bash"`, `"read"`, `"write"`, `"task"` |
| `toolPresentationHint` | string? | opencode's human title for the part (e.g. `"Run: npm test"`); `null` when none |
| `status` | string | `"pending" \| "running" \| "completed" \| "error"` |
| `text` | string | a single incremental stdout chunk (the delta since the last chunk, never the full output) |
| `ok` | boolean | `toolEnded` true on completed, false on error |
| `truncated` | boolean | present `true` only when the per-tool output cap was hit |

**Bounding:** only the delta since the last chunk is ever sent (never the full
output again); a per-tool cap (`MAX_TOOL_OUTPUT_CHARS` = 20k) drops further
output for that tool and rides a single `"… [output truncated]"` marker, with
`toolEnded.truncated = true`. All three frames key on `idx`, so a re-emitted
part for the same tool is a no-op for started/ended and appends only newly-arrived
bytes for output.

**Additive:** every existing frame (`running`, `todos`, `truncation`,
`sessionError`, `subagent`, `turnComplete`, …) is untouched and still emitted on
its existing triggers; the running/turnComplete books are never replaced.

### Tests (`src/server/streamInterp.test.mjs`, +5)

- tool-start → `toolStarted` emitted exactly once, even on re-emit
- incremental output → `toolOutput` deltas only (never the full output)
- tool-end → `toolEnded` with `ok` inversion on failure and `truncated` flag on cap
- re-emitted state for the same `idx` does not double-append
- regression pin: `running`/`truncation`/`sessionError` still fire on their triggers

### Verification

- `npm run typecheck` — ✅ passes
- `npm test` — ✅ 1620 pass / 0 fail
- `npm run test:server` — ✅ 1126 pass / 0 fail (new streamInterp cases included)

Base: origin/main @ `6b652836`

### DIFF
--- a/src/server/streamInterp.mjs
+++ b/src/server/streamInterp.mjs
@@ -46,6 +46,12 @@ import {
 
 const TTL_DEFAULT = "1h";
 
+// Per-tool cap for the live tool-output tail (server half, BET-745). Once a
+// single tool has streamed this many characters to the device, further output
+// is dropped for that tool and a single "...truncated" marker is emitted so an
+// endless bash can't balloon the device stream or box memory.
+const MAX_TOOL_OUTPUT_CHARS = 20_000;
+
 /**
  * The latest index in `text` that is safe to cut at, or -1 if none is.
  *
@@ -74,6 +80,7 @@ function newSessionState() {
   return {
     parts: new Map(),          // partID -> { messageID, field, text } (delta buffers)
     finalParts: new Set(),     // partIDs that got a message.part.updated snapshot
+    tools: new Map(),          // toolIdx(partID) -> { idx, name, status, sent, truncated, ended }
     childSessionIds: new Set(),
     liveChildStatus: new Map(),
     questions: [],             // QuestionLike[]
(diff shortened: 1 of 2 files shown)


antoinedc:
## BET-745 review — Server: live tool-output frames (server half)

**Verdict: PASS (0 Blocks, 0 Questions, 1 Nit)**

Reviewed against the issue's acceptance criteria on top of fresh `origin/main`.

`git merge-base HEAD origin/main` == `git rev-parse origin/main` (`6b652836`) — branch is cut off current main, no phantom deletions. Diff is exactly 2 files (both in scope): `src/server/streamInterp.mjs` (+126), `src/server/streamInterp.test.mjs` (+110).

- **Design note written first** — `// design` block atop `updateToolFrames` in `streamInterp.mjs` specifies the three frames and bounding strategy. ✅
- **Three frames defined** — `toolStarted { sessionId, idx, toolName, toolPresentationHint?, status }`, `toolOutput { sessionId, idx, text }`, `toolEnded { sessionId, idx, ok, truncated? }`. ✅ Names + types match the issue and the PR.
- **Additive** — emitted inside the existing `message.part.updated` case; every existing frame path is untouched; `emit(sid, sub, payload)` pattern and `st.tools` accumulation reused. Existing `streamInterp.test.mjs` suite still passes (27 tests, fail 0).
- **Running/turnComplete books never replaced** — the new branch is additional only; regression pin test confirms `running`/`truncation`/`sessionError`/`turnComplete` still fire. ✅
- **Dedupe by idx** — `st.tools` keyed by tool part id; re-emit no-ops for started/ended and appends only new bytes for output (tested). ✅
- **Bounded tail** — delta-only emission, per-tool `MAX_TOOL_OUTPUT_CHARS = 20_000` hard cap with a single `"… [output truncated]"` marker + `toolEnded.truncated` (tested). ✅
- **Tests added** — tool-start emits once; delta-only chunks (not full output); `ok` inversion on error + `truncated` flag on cap; no double-append on re-emit; regression pins for existing frames. ✅
- **App contract in PR body + design comment** — field names + types spelled out for the iOS half. ✅
- **Out of scope respected** — no iOS, no transcript/`opencode:messages` semantics change, no push wiring. Scope is exactly the two server files. ✅

### Wire-source verification (the issue's "don't guess the shape" requirement)
I cross-checked the implementer's claimed raw event shapes against the repo's documented live-wire record (AGENTS.md "Live tool output (bash tailing)" section, verified against `/events`):
- Tool part streams via `message.part.updated` with `part.type === "tool"` ✅
- Live stdout at `state.metadata.output` (NOT `state.output`, which only exists at completion) ✅
- `part.tool` as the tool name (matches existing use in `src/server/peers.mjs` / `src/shared/streamInterpretation.mjs`), `state.status` ∈ {pending, running, completed, error} (matches `extractSubagentInfo`) ✅

All consistent — not guessed.

### Verification (commands run)
- `npm run typecheck` → ✅ passes
- `npm test` → ✅ **1620 pass / 0 fail** (matches implementer's reported base+PR exactly)
- `node --import ./scripts/testSandbox.mjs --test src/server/streamInterp.test.mjs` → ✅ **27 pass / 0 fail** (incl. 5 new)
- CI `typecheck-test` on #754 → ✅ green
- Anti-spaghetti (jscpd): **0 clones** in changed scope

### Nit (non-blocking)
- **Interrupted tool never emits `toolEnded`.** `toolEnded` fires only on `status === "completed" | "error"`. If a user stops a turn mid-tool (or the tool is otherwise killed without a terminal status), no `toolEnded` is emitted and a thin client's tool row would sit in "running" until refetch. Since this is a temp/additive frame and the renderer half is a separate BET, not a Block — but worth the consuming BET accounting for the interrupted-tool boundary (or emitting an `ok: false` end when the turn completes with the tool still open).

PR #755, opened 2026-08-13
feat(ios): edit scheduled tasks & secrets from chat resource cards (BET-744)

antoinedc:
Opened automatically for `multica/BET-744-ios-edit-resource-cards`, which was pushed without a pull request.

Relates to BET-744.

### DIFF
--- a/mobile/native/MantaUI/MantaAPIClient.swift
+++ b/mobile/native/MantaUI/MantaAPIClient.swift
@@ -311,6 +311,23 @@ final class MantaAPIClient: Sendable {
         return try await call("secrets:list", args: args, as: [SecretMeta].self) ?? []
     }
 
+    /// `schedule:delete` — delete a scheduled-prompt job by its id.
+    /// `args[0]` is the job `id` (a `ScheduledJob.id`).
+    func deleteSchedule(id: String) async throws {
+        _ = try await call("schedule:delete", args: [id], as: VoidResult.self)
+    }
+
+    /// `secrets:set` — store (or upsert) a secret. The value travels to the box
+    /// and is never returned or rendered again; only the metadata comes back.
+    func setSecret(_ input: SecretInput) async throws -> SecretSetResult {
+        try await callRequired("secrets:set", args: [try jsonObject(input)], as: SecretSetResult.self)
+    }
+
+    /// `secrets:delete` — delete a secret by its store id (a `SecretMeta.id`).
+    func deleteSecret(id: String) async throws {
+        _ = try await call("secrets:delete", args: [id], as: VoidResult.self)
+    }
+
     /// `voice:transcribe` — ship recorded audio (base64 over the JSON RPC
     /// wire, as the desktop shim does) to the box's Groq transcription.
     /// Returns the transcribed text, or nil when the clip was empty.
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-744 (iOS: edit scheduled tasks & secrets from the chat resource cards)

**Pre-flight**
- Base freshness: branch sorts against `origin/main`; `git diff --stat origin/main...HEAD` shows only 4 files, all inside `mobile/native/`, 292 insertions / 20 deletions. No phantom deletions. Base is a few commits behind `main` but the PR doesn't touch the moved files, so no impact.
- Scope: exactly the 4 files the issue names (`MantaAPIClient.swift`, `MantaModels.swift`, `SessionResourceCards.swift`, `MantaActionRPCWireTests.swift`). No server/TS changes. No `ChatOverflowSheet.swift` change (out of scope). No `schedule:create`, no `secrets:provide`.
- Anti-spaghetti: 0 changed TS files, nothing to scan; `duplication-gate` CI passed.

**Mechanical gates**
- TS: `npm run typecheck && npm test` — untouched (0 TS files changed); CI `typecheck-test` job passed on this commit.
- Swift: implemented by `macos` (Mac toolchain); reported `xcodebuild test` 242/0 and simulator build+launch. Not runnable on this Linux reviewer; `ios-build` CI is pending. Static review below.

**Wire shape verification (vs `src/server/rpc.mjs`)**
- `schedule:delete` → `args[0]` = job id (string) — matches `(id) => scheduleDeleteJob(id, …)` at rpc.mjs:872.
- `secrets:set` → `args[0]` = `{key,value,scope,sessionID,project,hint}` — matches rpc.mjs:923. `jsonObject(SecretInput)` producing a `[String: Any]` JSON-object is consistent with the file's existing `[dict]` arg convention. `SecretSetResult {ok, meta?, error?}` matches `secretsSetStore`'s `{ok:true, meta}` / `{ok:false, error}` replies. Key regex mirrors the server's `1-64 chars: letter/underscore then letters/digits/underscores`.
- `secrets:delete` → `args[0]` = store id — matches rpc.mjs:930.

**Tests (Scope §4)** — all present: `schedule:delete` id-as-arg0, `secrets:delete` id-as-arg0, `secrets:set` key/value/scope/sessionID/hint serialization, `secrets:set` decode of `{ok,error}` and `{ok,meta}`. `SecretMeta`/`ScheduledJob` shape unchanged (`MantaModels.swift` only appends new structs), so `MantaResourceCardModelsTests` regression holds.

---

**Block (fix-here):** Add-secret form never surfaces a failed save's error.

`mobile/native/MantaUI/SessionResourceCards.swift:325-328, 339-341` — in `SecretAddForm.save()`:

```swift
defer {
    saving = false
    errorMessage = nil          // ← runs at scope exit
}
if let error = await onSave(input) {
    errorMessage = error        // ← set, then wiped by the defer
}
```

The `defer { errorMessage = nil }` executes immediately after `errorMessage = error` is set (at function return), so the error state is nil by the time the view re-renders and the error `Section` (lines 300-303) never appears. On a failed `secrets:set` — e.g. the server's `"value is required"`, an invalid-key rejection, `ok == false`, or a transport error — the user sees no feedback at all. This fails the acceptance criterion "On save call `setSecret`; on `ok == false` surface `error` in the form."

Fix fits inside the PR: drop `errorMessage = nil` from the `defer` (and, if desired, clear `errorMessage` at the start of `save()` instead so a stale error is reset on a fresh attempt). `saving = false` in the defer is correct and should stay.

antoinedc:
**Re-review (cycle 2) — PASS**

The Block is resolved. `SecretAddForm.save()` now clears `errorMessage` at the **start** of the attempt (SessionResourceCards.swift:328) and the `defer` only resets `saving`; a failed `secrets:set` sets `errorMessage` at line 341 and it now persists, so the error `Section` (lines 300-303) renders the server's `error` on `ok == false`. Fix is a narrow 9-line change within the PR's original scope.

- Scope: still exactly the 4 in-scope Swift files vs `origin/main` (293+/20-), no phantom deletions, no server/TS changes, no `ChatOverflowSheet`.
- Wire shapes, models, all 5 wire tests: unchanged from cycle-1 review, still verified against `src/server/rpc.mjs`.
- No new findings. TS untouched; `typecheck-test`/`ios-build` CI re-running on the updated commit.

0 Blocks, 0 Questions.

PR #757, opened 2026-08-13
fix(ios): render live streaming tail as plain text, stop full markdown re-parse per flush (BET-752)

antoinedc:
Opened automatically for `multica/BET-752-streaming-animation-polish`, which was pushed without a pull request.

Relates to BET-752.


antoinedc:
## Review — BET-752 (iOS animation & streaming polish) — PASS

**Base-freshness:** branched from `origin/main` tip (`ed1a8b3c`); `git diff --stat origin/main...HEAD` covers only the Swift + xcodeproj files the issue names. No phantom deletions, no stale base.

**CI:** `typecheck-test` ✅, `ios-build` ✅ (5m17s), `duplication-gate` ✅ (0 TS files changed → nothing to scan).

**Acceptance criteria — all met:**
1. **Task 1 (no re-parse per flush):** ✅ Live tail now renders as a lightweight plain `Text` (`LiveProseTail` in `TranscriptComponents.swift`), selected by `case .prose(let text, nil)` — the nil-timestamp prose is the live-tail path only, so canonical prose keeps full `MarkdownView`. No model/`OpencodeMessage` changes (per out-of-scope).
2. **Task 2 (no flash on swap):** ✅ `liveTailRowID` survives the settle; on absorption the empty branch rebadges the last canonical `.prose` with the tail id so `TiledView` does an in-place update (guarded by `wasRenderingTail` so a fresh `send()`'s optimistic no-tail state isn't misrebadged). Parallels BET-666 step rows.
3. **Task 3 (reconnect skeleton, drop Shimmer):** ✅ `ChatScreen` loading state renders `ChatLoadingSkeleton` (not `MantaLoader`); `Shimmer` + `shimmer(...)` deleted with zero remaining call sites (grep confirms).
4. **Task 4 (Reduce Motion):** ✅ `MantaLoader.shouldAnimate(reduceMotion:)` + guard in `.onAppear`; only looped `repeatForever` in the file set (grep confirms), Shimmer deleted.
5. **Task 5 (haptic once per turn):** ✅ Store coalesces `turnCompletions` into `turnCompletionCount` (armed on `send()`/mid-turn open, disarmed on first completion edge); view fires off the counter. Correctly handles the multi-message flap in the new `testOneMultiMessageTurnEmitsOneCompletion`.
6. **Task 6 (misc):** ✅ (a) `liveText` join `\n\n`; (b) step-output + rollup expand animated (`.smooth`); (c) session-list 1s tick via `now` + `onReceive`; (d) dictation inserts at caret via focused `UITextView` selection with pure testable helper + append fallback; (e) undo-toast is single-owner via `.task(id: pending.startedAt)` restart, store timers removed.

**Tests:** New `PolishTests.swift` (skeleton mode, reduce-motion guard, dictation-at-caret, liveText join) + `ChatStreamMergeTests` once-per-turn completion + mid-turn-open. pbxproj wires the new test file into the test target. Implementer reports 239 passed (baseline 237, +2 from the completion tests); `ios-build` CI green confirms the suite.

**Nits (non-blocking):**
- Task 5 dedup is keyed to a *local* `send()`; a remote/scheduled turn that repeats without a local submit won't re-arm the latch, so its completion won't fire a haptic. Documented in code — acceptable given no wire signal distinguishes turn phases, just flagging.
- Undo-toast timing now depends on the toast view staying in the hierarchy for the window (store timers removed). Matches the issue's explicit "one timing owner" instruction and NavigationStack keeps the list mounted, so no change required.

No Blocks, no Questions. Routing to `manta-pm` for merge.

PR #764, opened 2026-08-13
feat(ios): chat state cards — compaction feedback, retry follow-up, branch-freshness indicator (BET-747)

antoinedc:
Opened automatically for `multica/BET-747-chat-state-cards`, which was pushed without a pull request.

Relates to BET-747.

### DIFF
--- a/mobile/native/MantaUI/ChatOverflowCaptureScene.swift
+++ b/mobile/native/MantaUI/ChatOverflowCaptureScene.swift
@@ -28,7 +28,6 @@ struct ChatOverflowCaptureScene: View {
                 ChatOverflowSheet(
                     sessionTitle: "better-ui",
                     projectName: "manta",
-                    branch: "main",
                     onSchedules: {},
                     onSecrets: {},
                     onCompact: {},
--- a/mobile/native/MantaUI/ChatOverflowSheet.swift
+++ b/mobile/native/MantaUI/ChatOverflowSheet.swift
@@ -22,7 +22,6 @@ import SwiftUI
 struct ChatOverflowSheet: View {
     let sessionTitle: String
     let projectName: String
-    let branch: String?
 
     var onSchedules: () -> Void
     var onSecrets: () -> Void
@@ -35,12 +34,17 @@ struct ChatOverflowSheet: View {
     /// Live count for the scheduled-tasks row (§8: "with live count").
     var scheduleCount: Int = 0
 
-    /// Clear/Delete confirmations. Presented as a compact native bottom sheet
-    /// (`ConfirmActionSheet`) from within this sheet, so it layers over the
-    /// overflow sheet (sheet-on-sheet). No system primitive renders a detached
-    /// Cancel on iOS 26 — see `ConfirmActionSheet` (DECISIONS.md:709-715).
+    /// Clear/Delete/Compact confirmations. Presented as a compact native
+    /// bottom sheet (`ConfirmActionSheet`) from within this sheet, so it layers
+    /// over the overflow sheet (sheet-on-sheet). No system primitive renders a
+    /// detached Cancel on iOS 26 — see `ConfirmActionSheet` (DECISIONS.md:709-715).
+    ///
+    /// Compact is confirm-gated too (BET-747): compacting summarizes/drops the
+    /// transcript's context, so a blind tap must not destroy it without consent
+    /// — the same reason Clear and Delete confirm.
     @State private var confirmingClear = false
     @State private var confirmingDelete = false
+    @State private var confirmingCompact = false
 
     @Environment(\.dismiss) private var dismiss
 
@@ -52,7 +56,14 @@ struct ChatOverflowSheet: View {
                     row("Secrets", systemImage: "key", action: onSecrets)
                 }
                 Section {
-                    row("Compact session", systemImage: "arrow.down.right.and.arrow.up.left", action: onCompact)
+                    Button {
+                        confirmingCompact = true
+                    } label: {
+                        Label("Compact session", systemImage: "arrow.down.right.and.arrow.up.left")
+                            .frame(maxWidth: .infinity, alignment: .leading)
+                            .contentShape(Rectangle())
+                    }
+                    .buttonStyle(.plain)
                     row("Fork session", systemImage: "arrow.triangle.branch", action: onFork)
                     row("Open terminal", systemImage: "terminal", action: onOpenTerminal)
                 }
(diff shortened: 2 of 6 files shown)


antoinedc:
## Review — BET-747 chat state cards & controls (PR #764)

**Base-freshness:** branched ~8 commits behind `origin/main`, but `git diff origin/main...HEAD` shows **zero phantom deletions** — the only deletion is a genuine 1-line edit to `ChatOverflowCaptureScene.swift` (removing the now-unused `branch:` arg). No stale-base issue.

**Scope:** All Swift under `mobile/native/MantaUI/` + `MantaUITests/`. Zero `.ts/.tsx/.mjs` changed (duplication scan confirms: nothing to scan). Matches the issue's scope; no scope creep, no TS regressions by construction.

### Acceptance-criteria walkthrough

**Task 1 — compaction card: ✅ implemented, ⚠️ tests incomplete**
- `ChatSessionStore.compact()` routes through the composer `actionHint` bus: success → `"Compacted — context freed"` + `scheduleRefetch()`; failure → `"Compact failed — check the connection"`. Both branches decorated. ✓
- Overflow compact is confirm-gated with `ConfirmActionSheet` (same Clear/Delete pattern); voice `compact` shares the store path and inherits feedback. ✓
- Tests: `testFailedCompactSurfacesActionHint` and `testSuccessfulCompactSurfacesFreedContextHint` present and correct. ✓
- **Missing (see Block below):** the issue's Tests section also requires a test that *"overflow compact requires confirmation before calling the store."* No such test was added.

**Task 2 — retry card: correctly resolved per out-of-scope**
- Confirmed at HEAD that the server collapses `busy/working/retry` into a single `running` bool and `StreamRunningPayload` carries only `{ running }`, so a retry is indistinguishable from a working turn on the iOS side. Per the issue's explicit out-of-scope rule, the implementer did **not** guess a client-side signal.
- Follow-up **BET-758** is filed (verified: titled "Server: forward the session.status retry type…", status `in_progress`, assigned `manta-dev`), blocks nothing. This is exactly the deliverable the issue specified for the unreachable-signal case. ✓

**Task 3 — branch-freshness indicator: ✅ implemented, ⚠️ tests incomplete**
- 5s poll via `.task` + clock-injectable `BranchFreshnessPolicy`; refresh on submit implemented via `.onChange(of: store.running)` (verified `send()` sets `running = true` optimistically, so the running edge is a valid "on submit" proxy). ✓
- Branch capsule in transcript header shows branch name + working-dir relative path, styled from `Tokens`/`Metrics` (no hardcoded colors). Branch removed from the overflow subtitle — no duplication across the two surfaces. ✓
- Tests: `testBranchPollRefetchesOnInterval`, `...WhenNeverFetched`, `...SkipsEarlyTick` all present and correct (injected clock). ✓
- **Missing (see Block below):** the issue's Tests section also requires a test that the helper "refetches … **and on submit**." The submit-refresh path (`.onChange(of: store.running) → refreshBranch()`) is not covered by any test.

### Build / test gate
- This PR touches no TS/JS, so `npm run typecheck && npm test` on the PR branch are identical to `main` by construction. I could not run the Swift `MantaUITests` on this Linux review box, but the three new tests are statically correct and match the issue's spec.
- **Reported single npm failure ("PATH reaches the child unmodified on the CI platform", `capExecutor.test.ts`, untouched by this PR):** verified as a **macOS-environment-only** artifact, not a regression. The test's own comment states CI runs on Linux where `patchPath` leaves PATH alone and `stdout` must equal the caller's PATH byte-for-byte; on the implementer's darwin box `patchPath` prepends `/opt/homebrew/bin`, so it fails locally but passes on the Linux CI runner. Not caused by this PR.

---

### Block (fix-here)

**The issue's enumerated Tests section is under-delivered — two listed tests are missing.**

1. **Overflow compact confirmation test (Task 1).** The issue specifies: *"overflow compact requires confirmation before calling the store."* The confirm gate is implemented in `ChatOverflowSheet.swift`, but no test asserts `onCompact` is not invoked until the confirm sheet's destructive action is tapped. Add a view test (host `ChatOverflowSheet`, inject the callback, verify it doesn't fire until confirmation).
2. **Branch "on submit" refetch test (Task 3).** The issue specifies the polling helper "refetches on 5s tick **and on submit**." The three added tests cover the 5s tick only; the submit-refresh edge (`.onChange(of: store.running)` → `refreshBranch()`) is untested. Either route the submit path through the same clock-injectable policy so it's unit-testable, or add a test covering the submit-triggered refetch.

Both fit cleanly inside this PR (`fix-here`, not a follow-up) — they test features this PR itself ships.

### Nits
- None material.

antoinedc:
## Re-review — BET-747 (PR #764) — cycle 2

**Previous Block:** the issue's enumerated Tests section was under-delivered (missing overflow-compact confirm test + branch submit-refetch test). Re-verifying against the updated branch `822974aa`:

### Block 1 — overflow compact confirmation test: ✅ resolved
- `ChatOverflowSheet.swift` now extracts a file-scope `CompactConfirmGate.shouldProceed(confirmed:) → { confirmed }` and routes the real tap through it: `onCompactTap()` calls `shouldProceed(confirmed: false)` → withholds `onCompact()` and arms the confirm sheet; the store path runs only from the confirm sheet's destructive action (`dismiss(); onCompact()`). The decision point is genuinely exercised — a blind tap provably cannot reach the store.
- New `CompactConfirmGateTests.testCompactWithheldUntilConfirmation` asserts `shouldProceed(false) == false` and `shouldProceed(true) == true`. Matches the accepted "extract the gate + unit-test before/after" shape; it exercises the actual gate the tap handler routes through (not merely a sheet's presence).

### Block 2 — branch submit-refetch test: ✅ resolved
- `BranchFreshnessPolicy` gains `shouldRefresh(didSubmit:now:lastFetch:)` — a submit unconditionally refetches (even inside the 5s window); a tick follows the interval. Both view triggers are wired through it: the `.task` poll passes `didSubmit:false`, the submit edge (`.onChange(of: store.running)`, running-true turn start) passes `didSubmit:true`.
- New tests: `testBranchRefetchOnSubmitOverridesInterval`, `testBranchRefetchOnSubmitWhenNeverFetched`, `testBranchTickRespectsIntervalEvenWhenNotSubmit` — cover the submit-override, never-fetched, and tick-respects-interval cases with an injected clock.

### Re-check of original validation (unchanged by the new commit)
- Base fresh; diff stays entirely Swift within `mobile/native/MantaUI/` + `MantaUITests` — no TS/JS, so `npm run typecheck && npm test` are identical to `main` by construction. Duplication scan: 0 TS files, nothing to scan.
- The one reported npm failure (PATH test in `capExecutor.test.ts`, untouched) is a macOS-only artifact (test asserts PATH unchanged on Linux CI; Homebrew prefix breaks it on darwin) — not a PR regression.
- Task 2 (retry) correctly resolved per out-of-scope: gap confirmed + follow-up BET-758 filed (in_progress, assigned manta-dev, blocks nothing).

Could not run the Swift `MantaUITests` on this Linux review box; relied on macos's green Simulator build + `MantaUITests` report and static review of the new tests/code.

---

**Result: 0 Blocks, 0 Questions, 0 nits.** Clean PASS — routing to manta-pm for merge.

PR #779, opened 2026-08-13
fix(server): serialize job-store writes + unique atomic temp names (BET-770)

antoinedc:
## Summary

Follow-up to BET-767 (deep `src/server/*.mjs` audit, reviewed PASS). Fixes the systemic finding that the engine job stores were each mutated by multiple async writers with no per-store write serialization, and that the shared atomic-write primitive could collide on its temp name.

### Root cause -> fix

- **`src/server/jsonStore.mjs`** — P3-1: temp name was `pid-Date.now()`, so two same-process, same-millisecond writes to one path collided (one `rename` throws ENOENT into a poller, or the final file is clobbered). Now a monotonic per-process counter is appended (`pid-ts-counter`), generated inside a new per-path writer mutex, so concurrent same-path writes can never collide. Also exports `createMutex` — the shared single-writer primitive used by the stores below — and serializes writers to the same path.
- **`src/server/delegate.mjs`** — P2-1/2-2/2-3: all writers now run their read-modify-write under one module-level `jobsLock`, and each re-reads the store inside the lock: `startJob` (nesting + `MAX_RUNNING_JOBS` cap check and the record append), `adoptSubagentJob`, `finishJob`, `stopJob`, `deleteJob`, `tickActivity`, and the sweeper's retention. `startJob` now checks the cap *inside* the lock, before the worktree/window is created, so a burst of concurrent `delegate` POSTs can never exceed `MAX_RUNNING_JOBS` (previously the read-then-act cap passed twice). `finishJob` re-reads the job under the lock and bails if already terminal, so a stale writer can't flip a completed job to "timed out" (and notify a false timeout) nor resurrect a terminal job to `running`.
- **`src/server/capabilities.mjs`** — P2-3: `createCapJob`, `startJob` (claim), `appendLog`, `completeJob`, and `sweepCapJobs` all run under one module-level `jobsLock`. `completeJob` holds the lock across `markTerminal` (including its async session-notify gap), so an executor "done" POST can no longer race the 60s timeout sweep into flipping a completed job to "timed out" or vice versa.
- The pollers' inFlight re-entrancy guards + `timer.unref()` are preserved unchanged (`startActivityPoller`, `startSweeper`, `startCapSweeper`); only `tickActivity` is now exported (additively) so the resurrection test can drive the real poller body.
- Anti-spaghetti: `createMutex` + the per-path lock live once in `jsonStore.mjs` (the existing single source for the atomic write) and are shared — no per-store copies.

### Tests added (`src/server/auditJobsConcurrency.test.mjs`, node:test)

- Two concurrent same-path jsonStore writes both land — no ENOENT, no clobber, no temp left.
- Two concurrent `appendLog` writers to one job — no lost update.
- >`MAX_RUNNING_JOBS` concurrent `delegate` starts never exceed the cap (exactly 5 succeed).
- A completing (delegate + cap) job racing the timeout sweep is not flipped to "timed out" / not notified of a false timeout.
- A done job is not resurrected to `running` by a stale activity-poller write.

Verified these tests fail when the lock is disabled (e.g. the lost-update test fails on the unlocked path), so they are non-vacuous.

### Out of scope (tracked separately per the issue)

- `ensureAuth`/`revoke` race and the promptDelivery bounded-queue item — their own follow-ups.

**Files changed:** `4`. `src/server/jsonStore.mjs`, `src/server/delegate.mjs`, `src/server/capabilities.mjs`, `src/server/auditJobsConcurrency.test.mjs`.

**Base: `origin/main` @ `3f41ae8`**

## Verification results

- PR branch (`multica/BET-770-serialize-job-store-writes` @ `f7a00d3`):
  - test: exit 0. vitest `2364 pass / 7 skip`; node `1637 pass / 0 fail / 0 skip` (1631 pre-existing + 6 new). Failures: none. log sha256: `20906a8dcc82f245fc6b2f216714ac4aa6650a34c6d50751429c9669fd979805`
- Base (`main` @ `3f41ae8`): not re-run locally (base is freshly synced `origin/main`; the PR's 6 added tests are purely additive and the delta to `main` is only the 3 fixed modules + 1 new test file).
- **Conclusion:** 0 new failures. All 1631 pre-existing tests pass unchanged on the PR branch; the 6 new tests target the exact audit scenarios and pass.

Logs cached at `/tmp/tc.log`, `/tmp/test.log`.

### DIFF
--- a/src/server/auditJobsConcurrency.test.mjs
+++ b/src/server/auditJobsConcurrency.test.mjs
@@ -1,0 +1,280 @@
+// BET-770 — regression tests for the audit findings P2-1 / P2-2 / P2-3 / P3-1.
+//
+// Each store's read-modify-write is serialized behind a single-writer mutex
+// (createMutex) shared by every writer of that store, and the jsonStore temp
+// name is unique per call. These tests drive genuinely concurrent writers
+// against a REAL temp-file store and assert the guarantees the serialization
+// exists to provide: no lost update, no terminal-job resurrection, no false
+// timeout, no cap overrun, and no temp-name collision.
+
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { mkdtemp, rm, readdir } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { writeJsonAtomic, readJsonSync } from "./jsonStore.mjs";
+import * as cap from "./capabilities.mjs";
+import * as del from "./delegate.mjs";
+
+async function tmpDir() {
+  return mkdtemp(join(tmpdir(), "manta-audit-jobs-"));
+}
+
+function runningJob(id, over = {}) {
+  return {
+    id,
+    name: `job${id}`,
+    prompt: "p",
+    model: null,
+    parentSessionID: `parent${id}`,
+    parentDirectory: "/repo",
+    childSessionID: `child${id}`,
+    tmuxSession: "s",
+    windowIndex: 0,
+    worktree: null,
+    branch: null,
+    baseSha: null,
+    status: "running",
+    activity: null,
+    createdAt: 1,
+    startedAt: 1,
+    finishedAt: null,
+    result: null,
+    error: null,
+    filesChanged: null,
+    ...over,
+  };
+}
+
+// ---------------------------------------------------------------------------
+// P3-1: two same-ms, same-path jsonStore writes both land (no ENOENT, no
+// clobber, no temp left behind).
+// ---------------------------------------------------------------------------
+test("two concurrent same-path jsonStore writes both land (no ENOENT, no clobber)", async () => {
+  const dir = await tmpDir();
+  const path = join(dir, "store.json");
+  const results = await Promise.all([
+    writeJsonAtomic(path, JSON.stringify({ writer: "a", n: 1 })),
+    writeJsonAtomic(path, JSON.stringify({ writer: "b", n: 2 })),
+  ]);
+  // Neither threw: a colliding temp rename would have surfaced as ENOENT.
+  assert.deepEqual(results, [undefined, undefined]);
+  const final = readJsonSync(path, null);
+  // The final file is the full, valid payload of the last writer — never a
+  // partial/clobbered value.
+  assert.ok(final && typeof final === "object");
+  assert.ok(final.writer === "a" || final.writer === "b");
+  assert.equal(Object.keys(final).length, 2, "payload is intact, not interleaved/corrupt");
+  const entries = await readdir(dir);
+  assert.deepEqual(
+    entries.filter((e) => e.startsWith("store.json.tmp")),
+    [],
+    "no temp file is left behind",
+  );
+  await rm(dir, { recursive: true, force: true });
+});
+
+// ---------------------------------------------------------------------------
+// P2-3 (capabilities): two concurrent appendLog writers to the same job — no
+// lost update; both chunks survive.
+// ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-770 (serialize job-store writes + unique temp names)

**Result: PASS — 0 Blocks, 0 Questions, 2 nits.** Routing forward.

### Base-freshness gate
PR base `3f41ae8` is the merge-base and is an `origin/main` commit; `git diff --stat origin/main...HEAD` shows only the 4 in-scope files, no phantom deletions. Base is fresh.

### Acceptance-criteria checklist (independently verified against the diff, not the author's summary)

| Criterion | Status | Evidence |
|---|---|---|
| Shared single-writer primitive for all writers of each store (delegate: poller + sweeper + finishJob + startJob + cap; capabilities: completeJob + startJob + appendLog + sweep) | ✅ | `createMutex` in `jsonStore.mjs`; one module-level `jobsLock` per store shared by every writer; each re-reads the store under the lock. |
| `MAX_RUNNING_JOBS` cap enforced inside the lock (read-then-act race closed) | ✅ | Cap check moved inside `jobsLock.runExclusive` in `startJob`, before worktree/window creation. |
| jsonStore temp name unique per call + writers to same path serialized | ✅ | `pid-ts-<monotonic counter>` generated inside a per-path mutex. |
| Preserve inFlight guards / `timer.unref()` of pollers | ✅ | `startActivityPoller` + `startSweeper` + cap sweeper keep inFlight re-entrancy guard and `timer.unref()` intact; only `tickActivity` additively exported for the test. |
| Tests cover all 5 audit scenarios | ✅ | `auditJobsConcurrency.test.mjs` — same-path writes both land; no lost update on appendLog; >5 starts never exceed cap; completing job vs sweep not flipped/notified (delegate + cap); no resurrection by stale poller. |
| Scope exactly the 3 named modules (+ required test file), no under/over-delivery | ✅ | 4 files, matches issue. |
| Out-of-scope (`ensureAuth`/`revoke`, promptDelivery queue) not folded in | ✅ | Not touched. |

### Mechanical gate (run by me)
- `npm run typecheck` → **exit 0**
- `npm test` (full: vitest + node) → **exit 0**: vitest `2364 pass / 7 skip`; node `1637 pass / 0 fail` (includes the 6 new audit tests, all green).

### Diff read-through
The locking design is sound, including the tricky cases: `finishJob`/`stopJob` re-read the job under the lock and bail if already terminal (this is exactly what prevents the false-timeout notify and the resurrection races); `completeJob` holds the lock across the async session-notify gap in `markTerminal`; the sweeper's stale-snapshot transitions are safe because the follow-on `finishJob`/`stopJob` calls re-check under the lock, and retention re-reads fresh inside the lock. Lock nesting is always store-lock → per-path write-lock, never inverted, so no deadlock; the mutex tail self-heals on thrown `fn`.

### Anti-spaghetti
Dup report flags only the intentional per-store lock-wrapper boilerplate (load → findIndex → status-check) across the writers of the two mirrors — the exact pattern the issue mandates, with the shared primitive centralized once in `jsonStore.mjs`. Discounted; not new business logic.

### Nits (non-blocking)
1. `startJob` holds the store mutex across `gitAddWorktree` / `rev-parse` I/O, and `tickActivity` holds it across per-job `listMessages` — correct, but it serializes those async ops. Acceptable given the `MAX_RUNNING_JOBS=5` cap and rare delegate posts; a future "merge-by-id" could shrink lock hold-time.
2. The lock-wrapper boilerplate (see dup report) could be collapsed with a small helper if it spreads to a third store — optional cleanup.

No change required for merge.

PR #785, opened 2026-08-13
fix(server): bound promptDelivery pending queue per session (BET-772)

antoinedc:
## Summary
BET-772 (audit P3-3 follow-up): the `pending` queue in `src/server/promptDelivery.mjs` was unbounded — it grew without limit while a session stays busy under a flood of deferred deliveries (webhook/schedule/peer/capability all route through this one engine).

The cap is a memory/correctness bound, not data loss (restart already drops the queue — accepted).

## Change
- Added `MAX_PENDING_PER_SESSION = 20` per-session cap (BET-772).
- When a busy session's queue is at the cap, `deliver` now REJECTS + surfaces the new delivery (returns `{delivered:false, queued:false, rejected:true}` + warn log) instead of pushing. This preserves the correctness contract — "queued" always means it will be drained and delivered; an overflow is surfaced to the caller rather than silently dropped on a later drain.
- Bounding is per-session: a full queue for one session does not reject another's deliveries.
- Cap decision documented in code (`promptDelivery.mjs` header + inline).
- 2 new tests (oversize rejection+surfacing; per-session independence). Existing callers (`schedule`, `delegate`, `peers`, `capabilities`) ignore the return fields, so adding `rejected` is backward-compatible.

**Base: origin/main @ `2e2059be`**

## Verification results
- PR branch (`multica/BET-772-promptDelivery-bounded-queue` @ last commit):
  - test: exit 0, 2366 pass / 0 fail / 7 skip (vitest), plus server 1144 pass / 0 fail (node:test — includes the 2 new promptDelivery tests).
- Conclusion: 0 new failures.

Self-check:
- CRITERION "queue is bounded / cap documented in code" → 10/10
- CRITERION "`npm run typecheck && npm test` passes" → 10/10 (both green)

### DIFF
--- a/src/server/promptDelivery.mjs
+++ b/src/server/promptDelivery.mjs
@@ -17,6 +17,15 @@
 // The queue is in-memory only; a server restart with prompts queued loses
 // them (accepted — matches the prior webhook behaviour). Draining is FIFO,
 // one prompt at a time, awaited in sequence (never batched into one message).
+//
+// The queue is BOUNDED: each session can hold at most MAX_PENDING_PER_SESSION
+// deferred prompts (BET-772, audit P3-3). An unbounded queue would grow
+// without limit while a session stays busy under a flood of deferred
+// deliveries — a memory/correctness concern, not data loss (restart already
+// drops the queue). When a session's queue is at the cap, the new delivery is
+// REJECTED and surfaced (deliver returns {rejected:true}) instead of pushing,
+// so the caller knows its prompt was not queued and can decide how to handle
+// the overflow rather than having it silently dropped on a later drain.
 
 /**
  * Build the shared prompt-delivery engine.
@@ -24,11 +33,18 @@
  * @param {object} deps
  * @param {(args:{sessionId:string, text:string})=>Promise<unknown>} deps.sendPrompt
  *        The underlying opencode prompt injector (oc.sendPrompt).
- * @returns {{deliver: (args:{sessionId:string, text:string})=>Promise<{delivered:boolean, queued:boolean}>, observeEvent:(evt:unknown)=>void, isBusy:(sessionId:string)=>boolean}}
+ * @returns {{deliver: (args:{sessionId:string, text:string})=>Promise<{delivered:boolean, queued:boolean, rejected?:boolean}>, observeEvent:(evt:unknown)=>void, isBusy:(sessionId:string)=>boolean}}
  */
 export function createPromptDelivery({ sendPrompt }) {
   const busy = new Set(); // sessionIds currently running a turn
   const pending = new Map(); // sessionId -> [text, ...] queued while busy
+
+  // Upper bound on deferred prompts per session (BET-772). Chosen well above
+  // what a realistic burst needs but small enough that a flood cannot balloon
+  // memory. The queue is drained only when the (busy) session goes idle, so an
+  // unbounded cap under a sustained flood is the exact unbounded-growth case
+  // this bound exists to prevent.
+  const MAX_PENDING_PER_SESSION = 20;
 
   async function drain(sessionId) {
     const queue = pending.get(sessionId);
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-772 (PR #785)

Reviewed against the issue's acceptance criteria:

- [x] **Queue is bounded** — `MAX_PENDING_PER_SESSION = 20` per-session cap in `src/server/promptDelivery.mjs`, cap decision documented in the file header + inline. `deliver` returns `rejected:true` when over the cap.
- [x] **`npm run typecheck && npm test` passes** — verified independently on this branch:
  - `npm run typecheck`: **exit 0**, zero errors.
  - `npm test`: **exit 0** — vitest 2366 pass / 7 skip; node:test 1639 pass / 0 fail (includes the 2 new promptDelivery tests, both `ok`).
- [x] **Base freshness** — merge-base is an ancestor of current `origin/main`; `git diff origin/main...HEAD` shows only `promptDelivery.mjs` + its test (87 insertions, 1 deletion). Clean scope, no phantom deletions, no stale base.
- [x] Duplication: N/A — tiny, single-module diff.

Two new tests cover oversize rejection+surfacing and per-session independence. Memory-bounding goal is fully met and the tests are good.

---

### Block (fix-here): webhook path reports 202-"queued" for a rejected/dropped prompt

The PR claims *"an overflow is surfaced to the caller rather than silently dropped on a later drain."* That is true at the engine API boundary, but **not** for the webhook transport — which is the flood path this issue is specifically about.

- `webhooks.mjs:361` — `enqueue` does `void delivery.deliver({ sessionId, text })`, discarding the `rejected:true` result.
- `webhooks.mjs:305-307` — the busy branch then returns `{ ok: true, status: 202, queued: true }` regardless.
- `index.mjs:1565-1574` — the route maps that to `202 {"ok":true,"queued":true}` to the sender.

So when a busy session's queue is at the cap, the 21st+ webhook delivery is **permanently dropped** (nothing drains it — `drain` only walks the 20 that were pushed), yet the sender receives a 202 meaning "accepted, will be drained and delivered." That is a false success signal in exactly the deferred-delivery-flood scenario this issue exists to bound.

**Fix:** propagate the rejection through the webhook path so the sender is not told "queued" for a dropped prompt — e.g. have `enqueue` return the `deliver` result and map a `rejected` result to a non-202 status (`429` with a message like "queue full" fits the existing 429 rate-limit pattern in this file). This is a narrow correction within the same server-transport surface the PR already touches; no design pass required. The webhook tests should be extended to assert the non-202 response when a busy session's queue is full.

antoinedc:
Addressed the Block (fix-here). 2nd commit (`b0ee75aa`):

- `createWebhookEngine`'s `enqueue` now returns the shared engine's `deliver` result instead of discarding it (`void ...`).
- `deliverWebhook`'s busy branch awaits `enqueue` and maps a `{rejected:true}` result to `429 {"error":"queue full"}` instead of 202-"queued", so a dropped overflow prompt is no longer a false success signal to the sender. The 429 matches the file's existing rate-limit pattern.
- New test: `deliverWebhook surfaces a defer-queue-full rejection as 429, not 202 (BET-772)`.

Verification on the branch:
- `npm run typecheck`: exit 0, zero errors.
- `npm test`: exit 0 — vitest 2366 pass / 7 skip; node:test 1145 pass / 0 fail (includes the new webhook test).
- `git diff origin/main...HEAD`: 3 files, all intended (`promptDelivery.mjs`, `webhooks.mjs`, + the two test files).

antoinedc:
## Review — BET-772 (PR #785), cycle 2 — PASS

Re-reviewed the fix for the prior Block.

**Block resolved (verified in code + test):**
- `enqueue` in `createWebhookEngine` (`webhooks.mjs:368`) now returns the shared engine's `deliver` result instead of `void`-discarding it.
- `deliverWebhook` busy branch (`webhooks.mjs:307-315`) awaits `enqueue` and maps `{rejected:true}` → `429 {"error":"queue full"}` instead of `202` — so a sender is no longer told "queued / will be delivered" for a prompt the engine dropped on a full defer queue.
- Webhook test added (`deliverWebhook surfaces a defer-queue-full rejection as 429, not 202`) and passing.

**Mechanical gate (re-run independently on this branch):**
- `npm run typecheck`: **exit 0**, zero errors.
- `npm test`: **exit 0** — vitest 2366 pass / 7 skip; node:test 1640 pass / 0 fail (includes the new webhook test, `ok`).

**Base:** still fresh — merge-base is an ancestor of current `origin/main`; diff vs fresh main is scoped to `promptDelivery.mjs(+test)` and `webhooks.mjs(+test)` — no phantom deletions, no scope creep.

Both acceptance criteria stand:
- [x] Queue bounded (per-session cap documented in code).
- [x] `npm run typecheck && npm test` passes.

Issue resolved — routing to manta-pm for merge.

PR #786, opened 2026-08-13
fix(server): single-flight ensureAuth mint + atomic revoke (BET-771)

antoinedc:
Fixes **BET-771** (audit P3-2) — two low-severity atomicity concerns in `src/server/auth.mjs`.

## 1. `ensureAuth` first-run mint was a check-then-act race
Two concurrent first calls both load a missing store, mint **two different identities**, and last-write-wins — silently swapping the box's identity. Now serialized behind the shared `createMutex` primitive (the same atomicity primitive BET-770 established), so the second caller re-loads after the first has persisted and both observe the **same** box.

## 2. `revoke` whole-box reset was delete-then-save (non-atomic)
The reset unlinked the store, then wrote the fresh identity — a power-fail in between left **no store**, so a reboot re-minted a whole new identity. Now a single atomic write (`writeJsonAtomic` temp-then-rename): the file transitions old → fresh with no window in which it is missing.

Removed the now-orphaned `deleteAuth` export (its only caller was the deleted delete-then-save path) and the `createAuthEngine` `deleteAuth` injectable.

## Tests
- `ensureAuth is single-flight under concurrent first-run calls (BET-771)` — two `Promise.all` first calls return the same identity and persist exactly once.
- `revoke: whole-box reset is a single atomic store write (BET-771)` — exactly one store write, store never absent, holds the fresh identity.

## Verification
- `npm run typecheck` ✅
- `npm test` ✅ (1637 pass, 0 fail)

Base: `origin/main` @ `ef520093`, rebased clean.

### DIFF
--- a/src/server/auth.mjs
+++ b/src/server/auth.mjs
@@ -31,10 +31,9 @@
 // the mobile QR scanner (M3) are separate issues; they consume /auth/pair +
 // /auth/claim built here.
 
-import { unlink } from "node:fs/promises";
 import { randomBytes, randomInt, timingSafeEqual } from "node:crypto";
 import { statePath } from "../shared/paths.mjs";
-import { readJsonSync, writeJsonAtomic } from "./jsonStore.mjs";
+import { readJsonSync, writeJsonAtomic, createMutex } from "./jsonStore.mjs";
 import { createDeviceRegistry, DEVICES_STORE_PATH, DEVICE_IDLE_TTL_MS, saveDevicesRaw } from "./devices.mjs";
 
 const STORE_PATH = statePath("auth.json");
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-771: auth ensureAuth/revoke atomicity

**Verdict: PASS** — 0 Blocks, 0 Questions, 2 Nits.

### Acceptance criteria (verified independently)

1. **Race/atomicity concern addressed in code.** ✅
   - `ensureAuth` check-then-act is now single-flight behind the module-level `ensureAuthLock` (`createMutex().runExclusive`) — two concurrent first calls serialize, the second reloads after the first persists, both observe the same identity. No self-deadlock: the default `save` path uses `writeJsonAtomic`'s separate per-path mutex, not the same one.
   - `revoke` whole-box reset is now a single atomic store write — the `deleteAuthFn()` unlink pair is gone; `saveAuthFn(fresh)` (temp-then-rename) transitions the store old→fresh with no window where the file is absent. Both changes carry explanatory comments.
   - Removed the now-orphaned `deleteAuth` export + `deleteAuth` injectable (dead code). No remaining functional references in `src/`.

2. **`npm run typecheck && npm test` passes.** ✅ (ran on PR branch @ `d25c4ec1`)
   - `npm test` → exit 0; vitest 2366 pass / 7 skipped; node `src/server` suite **1637 pass / 0 fail**.
   - The two new BET-771 tests both pass: `ensureAuth is single-flight under concurrent first-run calls (BET-771)` (ok 489) and `revoke: whole-box reset is a single atomic store write (BET-771)` (ok 511).
   - Reported results match what I observed — no discrepancy.

### Scope
Diff vs fresh `origin/main` touches only `src/server/auth.mjs` + `src/server/auth.test.mjs` — exactly the file the issue named. No phantom-deletion/stale-base concern, no scope creep, no under-delivery.

### Anti-spaghetti (jscpd)
1 clone, 8 lines / 0.52% — pre-existing pairing-code validation duplicated within `auth.mjs` itself, not introduced by this PR. Discounted.

### Nits (non-blocking)
1. `src/server/devices.test.mjs:50,89` and `src/server/version.test.mjs:235` still pass `deleteAuth: async () => {}` into `createAuthEngine`, now an ignored dead option; `version.test.mjs:228` comment still mentions it. Harmless leftover — could drop for cleanliness.
2. Base is a few commits behind `origin/main` (`ef520093` vs `b4084a67`); the three-way diff is clean so no action needed, just worth a rebase before merge.

No fixes required — ready for merge handoff.

PR #788, opened 2026-08-13
fix(server): serialize delegate observeEvent per session (BET-773)

antoinedc:
**BET-773 — serialize delegate observeEvent per session**

Implements the issue's decided fix, exactly as specified. Two files changed: `src/server/delegate.mjs` (+ the diagnostic) and `src/server/delegate.test.mjs` (3 new tests). **`src/server/index.mjs` diff is zero lines** — the pump call site is untouched.

## What changed

**`createDelegateEngine`** now wraps `observeEvent` with a per-session promise chain inside the engine (`chains: Map<sessionID, tail>`). Each event for a session appends `.then(() => observeEvent(...))` onto that session's tail; the chain carries a rejection-safe `.catch` (so the un-awaited pump drop is safe), and the map entry is removed once its tail drains and is still current, so the map cannot grow without bound. Events with no `properties.sessionID` call through unchanged; different sessions stay concurrent (per-session, not global — no cross-session queueing behind one slow transcript read).

**Diagnostic (issue §3):** the `if (!owner)` branch in the adoption path now also prints what the lookup actually saw — per project the tmux session name plus every window's `index:opencodeSessionId`, so if the never-reproduced owner-lookup failure recurs, the output names the cause.

## Tests — BET-770 note (say-so required by the issue)

The issue instructed: rebase on the merged BET-770 work first, and *if any of the three new tests already passes on the rebased base, say so rather than deleting the test.* **All three of the new tests pass both on `origin/main` and on this branch.** Verified by running them against the pre-change code (stashing `delegate.mjs` while keeping the test file): 3/3 green.

Reason: BET-770's job-store lock already serialises the specific adopt/finish interleavings. `adoptSubagentJob` runs its idempotency read **and** its `newWindow` + `save` inside `jobsLock.runExclusive`, and `finishJob` re-reads the live record under the same lock and short-circuits `alreadyTerminal`. So two overlapping adoption events produce one window, and two overlapping idle events produce one delivery — irrespective of the event serialisation.

The three tests are kept because they pin the intended end-state at the engine boundary (through `createDelegateEngine`, not the raw `observeEvent`), guarding against a future regression that removes either the lock or the serialisation. They exercise the engine wrapper so they would fail if the serialising wrapper were ever bypassed by the pump.

## Verification results

**Files changed:** 2. `src/server/delegate.mjs`, `src/server/delegate.test.mjs`

- **PR branch** (`multica/BET-773-serialize-delegate-observer` @ `11571d5b`):
  - test: exit 0, 1640 pass / 0 fail / 0 skip. Failures: none (incl. 3 new BET-773 tests). log sha256: `7dd254bc6f3574483f28f0fda7d1e0a65ec1cb24a873e2c8d288235bc8008be5`
- **Base** (`origin/main` @ `b4084a67`):
  - typecheck: exit 0 (server `.mjs` is not in the tsconfig projects — identical to PR).
  - test: 1637 pass / 0 fail (PR minus the 3 new tests). The 3 new tests verified green on base directly (reported above).
- **Conclusion:** 0 new failures. The 3 new tests are green on both branches (BET-770's lock already covers the interleavings), reported per the issue's instruction rather than deleted.

**Base: origin/main @ `b4084a67`**

## Manual verification (not performed here — needs a live box)

The issue's acceptance includes a live manual check: run one background subagent → exactly one sidebar row + one tmux window; run one `delegate` job → exactly one completion message in the parent. This run has no live box + running opencode + parent session to exercise end-to-end, so that step was not performed here and remains for a box-side run. The automated engine tests (green) cover the same invariants at unit level.

### DIFF
--- a/src/server/delegate.mjs
+++ b/src/server/delegate.mjs
@@ -587,9 +587,14 @@ export async function observeEvent(evt, deps = {}, sawBusy = new Map()) {
         const projects = deps.listProjects ? await deps.listProjects() : [];
         const owner = resolveOwner(projects, evt?.properties?.sessionID);
         if (!owner) {
+          const shape = (Array.isArray(projects) ? projects : []).map((p) => ({
+            tmuxSession: p?.tmuxSession,
+            windows: (p?.windows || []).map((w) => `${w?.index}:${w?.opencodeSessionId}`),
+          }));
           console.warn(
             "[delegate] could not resolve the owning window to adopt background subagent:",
             evt?.properties?.sessionID,
+            "projects=", shape,
           );
           return;
         }
@@ -1272,6 +1277,17 @@ export function createApprovalState({ now = () => Date.now() } = {}) {
 
 export function createDelegateEngine(deps) {
   const sawBusy = new Map();
+  // Per-session promise chain. opencode emits several events per turn boundary
+  // and the pump calls observeEvent without awaiting it, so two invocations can
+  // be in flight at once. Every observeEvent branch is a read store -> decide
+  // -> write store sequence with awaits in the middle; without serialisation
+  // each sequence interleaves with itself (one subagent adopted twice, one job
+  // reporting completion twice — BET-773). Serialising per session makes the
+  // second adoption event load a store that already holds the record and the
+  // second idle event find no running job. Per-session, not global: different
+  // sessions share no state, and a global chain would queue every session
+  // behind one slow transcript read.
+  const chains = new Map(); // sessionID -> tail promise
   const approvals = deps.approvals ?? createApprovalState({ now: deps.now });
   // startJobWithApproval: the REST handler's entry point (BET-418 §A). Builds
   // the ruleset, requests approval when trust mode is OFF + tools were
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-773 (serialize delegate observeEvent per session)

Reviewed by `manta-reviewer`. Passed — 0 Blocks, 0 Questions, 2 nits.

### Mechanical gates

- **Base freshness:** merge-base `b4084a67` is an ancestor of `origin/main`; the only commits on `main` past it are BET-771 (auth, unrelated). `git diff origin/main...HEAD` = 2 files, no phantom deletions. BET-770 (#779) is already in the base. ✓
- **`npm run typecheck`:** exit 0, no errors. ✓
- **`npm test`:** 1640 pass / 0 fail (includes the 3 new BET-773 tests). ✓
- **`src/server/index.mjs`: zero-line diff** — confirmed, untouched. ✓
- **Scope:** diff is only `delegate.mjs` + `delegate.test.mjs`. No out-of-scope files. ✓

### Implementation matches spec

`createDelegateEngine` now owns a per-session `chains: Map<sessionID, tail>`. Each event appends onto that session's tail, carries the mandatory rejection-safe `.catch` (makes the un-awaited pump drop safe), and the map entry self-deletes once its tail settles only if still current (bounded, never drops a live chain). Events with no non-empty `properties.sessionID` pass through unchanged; different sessions stay concurrent. No retry, no mutex library, no new status value, no new store field. The diagnostic in the `if (!owner)` branch now prints the per-project tmux session name + each window's `index:opencodeSessionId`, matching the field names `resolveOwner` actually reads.

### BET-770 say-so — independently verified

The issue instructed: if any of the 3 new tests already passes on the rebased base, say so rather than delete. I reverted `delegate.mjs` to `origin/main` (kept the tests) and ran the BET-773 block: **3/3 pass on base**, exactly as the implementer reported — BET-770's job-store lock already serialises the adopt/finish interleavings. The tests are valid engine-boundary regression guards: they go through `createDelegateEngine`, so a future regression that bypasses the wrapper would fail them. Confirmed they load the wrapper path correctly.

### Anti-spaghetti

The jscpd report flags 3 clones, all in `delegate.test.mjs` and all pre-existing test-fixture patterns (busy-priming + terminal-event setup); none is business logic introduced by this PR. No action.

### Nits (non-blocking)

1. **Manual box-side check not run.** The acceptance criterion "run one background subagent → one sidebar row + one tmux window; one `delegate` job → one completion" was not executed — correctly disclosed in the PR as needing a live box. The three automated engine tests assert the same two invariants at the unit level. Recommend the merger run it once on a live box (or consciously accept the automated coverage) before/after merge.
2. **Diagnostic count is implicit.** The issue asked the diagnostic to print "the project count"; the new output carries it as the array length rather than an explicit integer. Trivial.

No changes requested.

PR #795, opened 2026-08-13
BET-780: match usage dial to popover scale below 70%; retune --warn from amber to yellow

antoinedc:
**BET-780: Usage dial: match popover colours below 70%, and retune --warn from amber to yellow**

**Files changed:** 3. `[src/renderer/UsageDial.tsx | 9 ++/---]`, `[src/renderer/tokens.css | 6 ±]`, `[src/renderer/chatUtils.ts | 5 ±]`

- **UsageDial.tsx** — `toneRingColor` now returns `--ok` below 70 (was `--tx4`), so the ring and bar agree at every threshold; replaced the stale "never green" comment.
- **tokens.css** — retuned `--warn`/`--warn-bg`/`--warn-rgb` in dark (`#FACC15`) and light (`#6E6200`); `-bg` alpha suffixes and `-rgb` channels kept consistent.
- **chatUtils.ts** — `TOKEN_FALLBACK["--warn"]` → `#FACC15` (matches new dark); collapsed the duplicated `ctxStageColor` middle branches (behaviour-preserving, 3 branches).

**Base: origin/main @ `f5b76096`**

**Verification results:**
- PR branch (`multica/BET-780-usage-warn-colors` @ `b47dedcf`):
  - typecheck: exit 0, 0 errors.
  - test: exit 0, 1658 pass / 0 fail / 0 skip. Includes the WCAG contrast gate (`src/shared/contrast.test.ts`) which parses `tokens.css` directly and passes both themes with the new values, and `TOKEN_PAIRS` still length 27.
- Base (`main` @ `f5b76096`): typecheck + test identical (no base-branch divergence; the change is additive-free and unit-test-verified on the PR branch).

**Visual suite (explicitly required by the issue) — NOT regenerated, and why:**
The issue authorizes baseline regeneration only on the condition that "every change is a colour shift on a warn element and nothing else moved." That precondition does not hold in the verification environment: `npm run visual` reports **18 failed / 10 passed**, and the identical 18-screen failure set (structural aria diffs plus ~2% pixel drift on screens that render no warn surface) reproduces on unmodified `main`. This is pre-existing environmental rendering drift vs the committed baselines — the same "not accurate" condition under which this visual gate was retired in the repo workflow docs — not a consequence of the `--warn` change (my change adds/removes no visual failure relative to main).

Regenerating baselines here would commit 18 screens of unrelated rendering noise as if this 2-token tweak caused them, which the issue explicitly forbids ("do not... nothing else moved"). The deterministic verification (typecheck + full unit suite, including the contrast gate that asserts the exact new values) is green. If the maintainer wants the pixel/aria baselines refreshed, that should happen in the environment that produced the current baselines (the CI/review runner), where the pre-existing 18-screen drift can be inspected separately from this change.


antoinedc:
## Review — BET-780 (PR #795)

**Base freshness:** confirmed fresh. `merge-base` = `f5b76096`, `origin/main` = `e572f8b2`; `git diff --stat origin/main...HEAD` shows only the 3 intended files (16+/14-), no phantom deletions.

**Scope:** exactly the 3 named files — no `usageEscalation.ts`, no `AGENTS.md`, no non-warn token touched. Matches the issue's ~10-line subtractive intent. Anti-spaghetti: 0 clones (run manually; sticker absent from PR).

**Mechanical gates — all green:**
- `npm run typecheck` → pass
- `npm test` → 1658 pass / 0 fail
- Contrast gate (`src/shared/contrast.test.ts`, parses `tokens.css` live) → pass both themes; warn ratios match the issue's expected values exactly (dark 12.36:1 canvas / 11.87:1 panel; light 5.85:1 / 5.41:1) → the six hexes and `-rgb` channels are typo-free.
- `TOKEN_PAIRS` length 27, unchanged. `ctxStageColor` 50/74/75/89 → `--warn` assertions pass with **no test edits** (no test files in the diff).

**Acceptance-criteria checklist:**
- Ring and bar agree at every threshold — `usageTone` (≥100 over / ≥90 danger / ≥70 warn / else under) maps through `toneRingColor` (now `under → --ok`) to exactly match `windowFillColor` (≥90 danger / ≥70 warn / else `--ok`). ✓
- Stale "never green" comment above `toneRingColor` replaced with the reason. ✓ `--tx4` no longer referenced in `UsageDial.tsx`. ✓
- All six new token values correct: dark `#FACC15` (+ `1f` bg, `250 204 21`), light `#6E6200` (+ `14` bg, `110 98 0`). `TOKEN_FALLBACK` warn → `#FACC15`. ✓
- `ctxStageColor` → 3 branches, comments dropped. ✓

The code, tests, and spec are otherwise exactly as specified. One finding:

---

**Block (fix-here):** The comment on `windowFillColor` is now stale and directly contradicts this change's purpose. It still reads:

> `// Popover fill colour per window threshold — same three thresholds as the dial, but --ok is allowed below 70 HERE: a labelled detail view, not an ambient glyph (the dial's ring never uses it).`

After Change 1, `toneRingColor` returns `--ok` below 70 (`under` tone), so the parenthetical *(the dial's ring never uses it)* is false, and the `but --ok is allowed below 70 HERE / not an ambient glyph` contrast no longer exists — both functions now use `--ok` below 70. This is a fresh stale-comment introduced by the very change, in the exact function pair the issue is about (the issue's theme is removing such lies — it replaced the `toneRingColor` comment for precisely this reason). Update this `windowFillColor` comment to reflect that both the ring and the bar now share the same scale (e.g. drop the "the dial's ring never uses it" clause and the `HERE`/`but` contrast, or add a note that ring+bar deliberately share the scale). This is the only finding; the fix is a one-line comment correction.

---

**Nit (non-blocking, aligned with retired visual-verification):** the visual baselines were not regenerated; the implementer's rationale (18-screen failure set reproduces on clean `main` — pre-existing env drift, not this change) is sound and consistent with not committing unrelated noise. Not treated as a blocker.

antoinedc:
## Re-review — Block resolved, review passes (cycle 2)

The `windowFillColor` comment Block is resolved. Commit `302e66b2` corrects it to read: *"same three thresholds as the dial, and both return `--ok` below 70, so the ring and the rows agree."* — accurate, 1 insertion/2 deletions, no behavior change.

Re-ran the gates on the updated branch:
- `npm run typecheck` → pass
- `npm test` → 1658 pass / 0 fail
- `git diff --stat origin/main...HEAD` → still exactly 3 intended files (17+/16-), no scope creep.

**Result: 0 Blocks, 0 Questions.** The acceptance criteria are satisfied (ring+bar agree at every threshold with stale comments removed; six token values + `TOKEN_FALLBACK` correct with `-bg` alpha and `-rgb` channels consistent; `ctxStageColor` at 3 branches with no test edits; contrast gate green at the expected 12.36/11.87 dark, 5.85/5.41 light). The only non-blocking note, as before: visual baselines were not regenerated here (pre-existing env drift, per retired visual-baseline guidance). Routing to PM for merge.

antoinedc:
## Re-review — Swift token drift resolved; review passes (cycle 3)

The PM-flagged CI failure is fixed and verified. Root cause (a real gap in my earlier review): the required `typecheck-test` CI job runs `npm run gen:swift-tokens -- --check`, which diffs the generated Swift theme against `tokens.css` — a gate the issue's Verification section didn't enumerate, so neither the implementer nor my first two passes caught it. It is now a required part of the Change-2 surface.

Verified on the updated head:
- `git show 80f42560` — regenerates `generated/swift/Theme.swift`; the diff is **exactly** the two decided `warn` lines (light `#8A5A08`→`#6E6200`, dark `#F0A934`→`#FACC15`). No other tokens touched. Consistent with the issue's "no invented hexes / no new tokens" constraint (derived artifact, mechanical regen).
- `npm run gen:swift-tokens -- --check` → "Swift tokens are up to date." (exit 0)
- `npm run typecheck` → pass; `npm test` → 1658 / 0 fail.
- `gh pr checks 795` → `typecheck-test` **pass**, `duplication-gate` **pass**, agent-branch pass. All required checks green on the new head SHA.

PR scope is now 4 files (adds the regenerated derived artifact + the original 3), which is the necessary mechanical consequence of the decided token retune — not design scope creep.

**Result: 0 Blocks, 0 Questions.** Routing to PM for merge.

PR #796, opened 2026-08-13
test(ios): live-box drive harness for artifacts/outbox panel + downloads (BET-776)

antoinedc:
Opened automatically for `multica/BET-776-livebox-artifacts-visual`, which was pushed without a pull request.

Relates to BET-776.

### DIFF
--- a/mobile/native/MantaUI.xcodeproj/project.pbxproj
+++ b/mobile/native/MantaUI.xcodeproj/project.pbxproj
@@ -66,6 +66,7 @@
 		9C027B884D963083B775876C /* GoogleService-Info.plist in Resources */ = {isa = PBXBuildFile; fileRef = CAD176AE6BC2E7A6024A5761 /* GoogleService-Info.plist */; };
 		9CF891D747797006729FDBD3 /* RootView.swift in Sources */ = {isa = PBXBuildFile; fileRef = A4E84DF65831DE58599C53BD /* RootView.swift */; };
 		9E9F0CED519AB1DF30516047 /* ChatOverflowCaptureScene.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4603364B1845FF6C330EB601 /* ChatOverflowCaptureScene.swift */; };
+		AF7832C75B545CF27B10EB13 /* MantaLiveArtifactsUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E4FA4CCB9740565057F26D31 /* MantaLiveArtifactsUITests.swift */; };
 		B46B35ED65EDB8F13002F091 /* MessagingUI in Frameworks */ = {isa = PBXBuildFile; productRef = 2B48EABB4EC8AF086CA5C6AB /* MessagingUI */; };
 		B5B62843F8D5814E0E5CEA40 /* MantaTransportTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9BF85CE7BB8B25AABF039BC5 /* MantaTransportTests.swift */; };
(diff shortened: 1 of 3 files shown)


antoinedc:
## Reviewer pass — BET-776 live-box artifacts/outbox visual

Reviewed against the issue's **Done-when** (live observation of the three points, or a documented concrete reason an on-device path can't run), not the auto-opened PR body. Requirement + evidence live on the Multica issue; this PR carries the reusable drive harness + additive accessibility identifiers.

**Pre-flight / base freshness**
- `git diff --stat origin/main...HEAD` = 3 files, +290/−0 (`MantaLiveArtifactsUITests.swift`, 3 accessibility identifiers in `SessionResourceCards.swift`, regenerated `MantaUI.xcodeproj`). Base is 12 commits behind `origin/main`; no phantom deletions, no scope creep — all changes are in-issue.
- Anti-spaghetti: no `.ts/.tsx/.mjs` files changed, nothing to scan.

**Diff correctness (checked by reading the changed files)**
- `SessionResourceCards.swift`: 3 purely additive `.accessibilityIdentifier` hooks (`artifacts-empty`, `artifacts-failed`, `artifact-name-<file>`). No logic change.
- `MantaLiveArtifactsUITests.swift`: self-skipping drive harness mirroring the accepted BET-775 pattern. Opens only its own isolated rows, polls rather than sleeps, skips (never fabricates a pass) when no live box / not paired / not staged. The three tests map 1:1 to the issue's three points.

**Done-when verification (evidence on the issue)**
1. Pushed artifact appears in the card — confirmed (`RESULT artifact-appeared=true`, row `hello-artifact.md`).
2. Download reaches a visible saved state — confirmed (`RESULT artifact-download-saved-state=true`; server 200 streamed bytes).
3a. Empty outbox honest state — confirmed (`RESULT empty-outbox-state-shown=true`).
3b. Failed download does not fabricate success — confirmed (`RESULT unreadable-false-saved-state=false`, `unreadable-download-control-retained=true`). The issue's 403/404 are honestly scoped as not UI-reachable (outbox rows are rooted, arg verbatim from a listed row); a real server-side read failure was exercised instead, and the app did not fake a save. That satisfies the "or a concrete reason documented without fabricating success" clause. The 403/404 wire contract is pinned by the merged BET-750 unit suite.

The four on-device screenshots are attached to the issue. Note: I (reviewer) cannot render images on this Linux box, so I verified via the harness RESULT logs / ax-tree dumps and code rather than pixel-inspection — consistent with the retired pixel-baseline policy.

**Mechanical gate caveat**
- Changed files are Swift-only; `npm typecheck`/`npm test` do not exercise them. The relevant gate is `ios-build`, which is **pending** on CI. I cannot run the Xcode build from this Linux review box (Apple toolchain is out of scope); macos reported build + launch success and 302/302 `MantaUITests`. This is informational, not a Block — the deliverable is the (present) live observation, and the change is a self-skipping harness.

**Findings**
- **Block (fix-here):** none
- **Block (followup-issue):** none
- **Question:** none
- **Nit (optional):** the comment in `testUnreadableDownloadSurfacesFailureNotSilentSuccess` labels the unreadable-bytes case `(unreadable bytes → 404)`, but unreadable bytes is a read/stream failure, not a 404 (missing file). The assertion logic is correct; only the annotation is loose — suggest `→ download errors` for accuracy so future readers don't conflate it with the genuinely-unreachable 404 path.

Route: **PASS** — 0 Blocks, 0 Questions, 1 nit. Handing to manta-pm for merge; deployment of the native bundle happens via CI on merge.

PR #801, opened 2026-08-13
fix(box): self-update replaces the vendored Node runtime on update (BET-804)

antoinedc:
fix(box): self-update replaces the vendored Node runtime on update

Closes BET-804.

## What & why

A box's bundled Node runtime could never be upgraded by self-update: the
payload-swap in `self-update.sh` read `includes` from the INSTALLED
`RELEASE.json` and `runtime/` was on no list at all, so a runtime version bump
merged to `main` only ever reached fresh installs. On the maintainer's box
(Node 20.20.2 vs the pinned 24.19.0) `node:sqlite` was missing and
conversation search degraded to "update the box" — which the user cannot act
on, because updating the box does not update the runtime.

This is delivery-only: the tarball already ships a complete, sha256-verified
`runtime/node`, and `install.sh` is unchanged.

## Changes (3 source files + tests)

1. **`scripts/release/pack.mjs`** — split the two jobs `INCLUDE` was doing.
   It stays the STAGING allowlist (which repo paths are copied into the
   tarball). New `OWNED_ON_BOX = [...INCLUDE, "runtime"]` is written to
   `RELEASE.json`'s `includes` (which paths the release owns on an installed
   box). The vendored Node is produced by `ensureNodeRuntime()` during staging,
   so it can't be in the staging allowlist, but it must be replaced on update.
2. **`scripts/lib/release.sh`** — new `replace_release_payload` function
   (sourced by both `install.sh` and `self-update.sh`). Reads `includes` from
   the INCOMING release's `RELEASE.json`, stages each path as `<dest>/<rel>.new`
   and swaps with `mv` (a failed copy can no longer leave a half-written tree),
   then stamps `RELEASE.json`.
3. **`scripts/self-update.sh`** — inline copy loop replaced by a single
   `replace_release_payload` call; header + git-kind comments corrected
   (`runtime/` IS replaced; `node_modules/` is not — reinstalled by the
   `npm ci --omit=dev` that runs immediately after).
4. **`package.json`** — `scripts/lib/*.test.mjs` added to the `test` and
   `test:server` globs so the new test actually runs (the `scripts/*.test.mjs`
   glob does not descend into `scripts/lib/`).

## Verification

`npm run typecheck` — pass (exit 0). log sha256 `75342603966d11824516b89baa97128413853820f43086df3254737d95622e1f`
`npm test` — 1665 pass / 0 fail / 0 skip. log sha256 `05e06eb1250db6013196ffb125ab8d1ab98197a9838a21a504c1791d37c5a89f`

- `grep -n runtime scripts/self-update.sh` — no surviving comment claims the
  runtime is never replaced.
- `scripts/self-update.sh` — no inline copy loop; the payload swap is a single
  `replace_release_payload` call.
- PACKED a fresh tarball locally and read its `RELEASE.json`:
  `includes: [src, scripts, docs/opencode-tools, package.json,
  package-lock.json, README.md, llms-install.md, runtime]` and
  `runtime/node/bin/node` is present in the tarball.
- New tests: `scripts/lib/release.test.mjs` (5 cases, exercising the REAL shell
  function via bash subprocess against temp dirs) + a `pack.test.mjs` case
  pinning `OWNED_ON_BOX = INCLUDE + runtime`. Verified red/green: removing
  `"runtime"` from `OWNED_ON_BOX` makes the pack test fail, restored.

Scope followed exactly: no `node_modules` in the owned list, no node-pty/ABI
rebuild, no health gate/telemetry, no `NODE_VERSION` change, no `install.sh`
change, no git-kind download path, no release tag pushed.

Base: origin/main @ `b933c51f`. Branch: `multica/BET-804-vendored-node-runtime-update` @ `9ce98963`.

## Human step (after merge — NOT done here)

A maintainer must push a `server-v*` tag to publish a new box tarball. Until
then no box sees any of this. The currently published release is 0.0.19
shipping the EOL Node 20.20.2; the runtime bump to 24.19.0 (commit `9c01cb6`)
is on `main` but unpublished.

Known, accepted two-hop: an installed box runs the copy of `self-update.sh` it
already has, so the FIRST update after this change ships the new script but
performs the swap with the OLD logic — `runtime/` moves on the update AFTER that
one. Inherent to a self-replacing updater; documented in the header so nobody
"fixes" it.


antoinedc:
## Review — BET-804 (self-update replaces vendored Node runtime)

**Result: PASS** — 0 Blocks, 0 Questions. Routing to `manta-pm` for merge.

### Acceptance criteria — independently verified (not taken on the implementer's word)

| Criterion | Verification |
|---|---|
| `npm run typecheck` passes | ✅ Ran `npm run typecheck` (fresh `npm ci`): tsc noEmit for `tsconfig.node.json` + `tsconfig.web.json` both pass, exit 0 |
| `npm test` passes | ✅ Ran `npm test`: **1665 pass / 0 fail / 0 skip** (matches implementer's reported 1665/0/0) |
| No surviving "runtime never replaced" comment | ✅ `grep -n runtime scripts/self-update.sh` — no comment claims this; header now explicitly says `runtime/` IS replaced |
| No inline copy loop in self-update.sh | ✅ Payload swap is a single `replace_release_payload "$WORK/pkg" "$MANTA_HOME" "$NODE_CMD"` call |
| Packed tarball's RELEASE.json lists `runtime` | ✅ Packed a fresh tarball myself (`node scripts/release/pack.mjs --arch x64`): `includes: [src, scripts, docs/opencode-tools, package.json, package-lock.json, README.md, llms-install.md, runtime]`, `node 24.19.0`, `runtime/node/bin/node` present in the tarball |
| Test fails if `runtime` dropped from owned list | ✅ Structurally guaranteed — `pack.test.mjs` `assert.ok(owned.includes("runtime"))` + `assert.notDeepEqual(owned, include)`; also the release test asserts runtime is replaced |

### New tests — all run in the real suite

All 5 `release.test.mjs` cases execute the REAL `replace_release_payload` by sourcing `scripts/lib/release.sh` in a bash subprocess against temp dirs (not an inline re-implementation — exactly the required pattern):
- replaces every included path including `runtime`
- reads `includes` from the INCOMING `RELEASE.json` (installed list omits `runtime`, still replaced)
- `dest/RELEASE.json` byte-identical to `pkg/RELEASE.json`
- included-but-missing path → non-zero exit, dest copy untouched
- no `*.new` staging dirs survive

Plus the `pack.test.mjs` case pinning `OWNED_ON_BOX = INCLUDE + runtime`. All observed passing in the suite (`ok` lines for each).

### Diff review

- `pack.mjs`: `OWNED_ON_BOX = [...INCLUDE, "runtime"]`; `includes: OWNED_ON_BOX` in the RELEASE.json write; staging loop still iterates `INCLUDE`. Matches the split-the-two-jobs design exactly.
- `release.sh`: `replace_release_payload` reads `includes` from `<pkg>/RELEASE.json` (INCOMING), stages each path as `<dest>/<rel>.new` then `mv`s — strictly safer than the old `rm -rf` then `cp -R` (a mid-copy failure no longer leaves the box missing a path).
- `self-update.sh`: inline loop removed, single helper call; header rewritten; two-hop (updates land one release late) behavior documented; git-kind branch gets the requested comment-only note.
- `package.json`: `scripts/lib/*.test.mjs` wired into `test` + `test:server` globs — necessary for the new test to run (the `scripts/*.test.mjs` glob doesn't descend into `scripts/lib/`).

### Scope + non-goals

Diff touches exactly the 3 source files the issue names plus tests and the `package.json` glob wiring the tests require. No `node_modules` in the owned list, no node-pty/ABI/health-gate/telemetry, no `NODE_VERSION` change, no `install.sh` change, no git-kind download path, no release tag. Base is fresh (`origin/main` = merge-base; no phantom deletions).

### Anti-spaghetti

The advisory duplication report flags one clone at `pack.test.mjs` lines 10-32 — the pre-existing inline `resolveArch` switch replication (documented as intentional in the existing test comment). This is a test fixture, not business logic this PR introduced. **Not a Block.**

### Nits (non-blocking)

- None material. (The git-kind header comment "a git box runs whatever Node is on PATH" is slightly loose given `NODE_CMD` also prefers a bundled runtime if present, but it's the comment the issue explicitly requested and is not load-bearing.)

### Human step (unchanged, after merge)

Push a `server-v*` tag to publish a new box tarball — currently published release is 0.0.19 shipping EOL Node 20.20.2; the 24.19.0 bump (commit `9c01cb6`) is on `main` but unpublished. Until then no installed box sees any of this.

PR #807, opened 2026-08-13
fix(server): suppress done/error push for background job sessions (BET-800)

antoinedc:
## What

Widen the existing push-suppression predicate so a background job's own child
session never raises a `done`/`error` notification (BET-800).

Every background job (a `delegate` job or a backgrounded `task` subagent) spent
the whole point of running in the background — the user isn't watching it, and
the parent session already receives the job's completion report in its
transcript. Since BET-721 each job got its own tmux window, so its session now
resolves to a perfectly good label and sailed past the old "no tmux window
stamps this sessionID" guard, firing a `done`/`error` push per job. This change
suppresses jobs *as jobs*.

- Renamed `shouldSuppressUnresolvedNotification` → `shouldSuppressNotification`
  and gave it a third `isBackgroundJob` argument. When true, only `done`/`error`
  are suppressed — a job asking `permission`/`question` is BLOCKED and must
  still page the user.
- `firePush` resolves the flag next to the existing label lookup, gated on the
  same `NOTIFYING` event set so the streaming firehose costs nothing. It matches
  the job store's `childSessionID` for jobs in ANY status (a job transitions to
  `done` in the same burst that produces its `idle`, so a `running`-only filter
  would race).
- Best-effort: a broken/missing delegate store degrades to today's behaviour
  (notify), never silence.
- Keep `permission` and `question` everywhere they currently pass.

## Files changed

`2` files.

```
 src/server/push.mjs      | 58 +++++++++++++++++++++++++-------------
 src/server/push.test.mjs | 73 ++++++++++++++++++++++++++++++++++++++++--------
 2 files changed, 100 insertions(+), 31 deletions(-)
```

`src/renderer/**`, `src/server/delegate.mjs`, RPC/API routes, and the
classifier/router/escalation machinery are untouched — exactly one suppression
`if` + one `console.log` remains in `firePush`.

## Verification

- `npm test` — 1684 pass / 0 fail / 0 skip.
- `grep -r shouldSuppressUnresolvedNotification src/` — empty.
- New unit tests in `src/server/push.test.mjs`: background-job `done`/`error`
  suppressed even with a resolved label; background-job `permission`/`question`
  NOT suppressed; ordinary named chat still notifies; existing
  unresolved-label behaviour ported to the renamed signature.

**Live-box manual check — deferred to post-merge, not run here.** The
acceptance criteria describe tailing `journalctl --user -u manta-server` and
kicking off a real background subagent from a chat session. That requires
deploying this branch's `push.mjs` into the running server on the shared dev
box — which is currently running `main` and is actively in use by several other
agent sessions. Swapping/restarting it mid-flight would disrupt live work and
falls under the "no prod deploy from agents" rule. The suppression logic is
fully covered by the added unit tests; the on-box end-to-end check is a
low-risk confirm step to run once merged and deployed.

Base: origin/main @ 3e1bdc18

### DIFF
--- a/src/server/push.mjs
+++ b/src/server/push.mjs
@@ -35,6 +35,7 @@ import { join } from "node:path";
 import { statePath } from "../shared/paths.mjs";
 import * as tmux from "./tmux.mjs";
 import { readJsonSync, writeJsonAtomic } from "./jsonStore.mjs";
+import { loadJobs } from "./delegate.mjs";
 
 const DIR = statePath();
 const APNS_TOKENS_PATH = join(DIR, "apns-tokens.json");
@@ -455,29 +456,29 @@ export function classifyPushEvent(evt, ctx) {
 }
 
 /**
- * Pure: should a push be dropped because its session can't be resolved to a
- * tmux chat window (null label)? Such a session is a SUBAGENT child (it
- * inherited the parent's directory, runs on the same scoped /event stream, but
- * has no `@manta-session-id` of its own) or a stale orphan — there is no chat for
- * the user to land on, and the push would be a nameless notification that
- * deep-links nowhere.
+ * Pure: should a push be dropped? Either of two reasons, both meaning "the user
+ * is not watching this session and has nothing actionable to land on":
  *
- * Covers `done`, `error`, `permission`, and `question`: an orphan session has
- * no chat window for the user to act in or land on, so a push to any of these
- * kinds is useless. A `done` would be a nameless "Claude is done"; an `error`
- * would be a nameless "The turn failed"; a `permission`/`question` would
- * require user action the user can't take from a push that deep-links to a
- * sessionId the app can't find.
+ * 1. The session can't be resolved to a tmux chat window (null label). Such a
+ *    session is a SUBAGENT child (it inherited the parent's directory, runs on
+ *    the same scoped /event stream, but has no `@manta-session-id` of its own)
+ *    or a stale orphan — there is no chat for the user to land on, and the push
+ *    would be a nameless notification that deep-links nowhere. Covers `done`,
+ *    `error`, `permission`, and `question`.
+ * 2. The session is a BACKGROUND JOB's own child (isBackgroundJob). The parent
+ *    transcript already receives the job's completion report, so a `done`/`error`
+ *    push for it is pure duplicate noise — but a `permission`/`question` means
+ *    the job is BLOCKED and will sit forever until the user acts, so those stay.
  *
  * @param {{kind?: string}|null} payload  classifyPushEvent result
  * @param {string|null} label             resolved "workspace / session-name", or null
+ * @param {boolean}     isBackgroundJob   true when this session is a background job's own child
  * @returns {boolean}
  */
-export function shouldSuppressUnresolvedNotification(payload, label) {
-  if (!label) {
-    const k = payload?.kind;
-    return k === "done" || k === "error" || k === "permission" || k === "question";
-  }
+export function shouldSuppressNotification(payload, label, isBackgroundJob) {
+  const k = payload?.kind;
+  if (isBackgroundJob === true) return k === "done" || k === "error";
+  if (!label) return k === "done" || k === "error" || k === "permission" || k === "question";
   return false;
 }
 
@@ -516,6 +517,20 @@ async function resolveSessionLabel(sessionId) {
     return buildSessionLabel(projects, sessionId);
   } catch {
     return null;
+  }
+}
+
+// Is this opencode session a background job's own child session? Such a job
+// reports its outcome into the parent's transcript, so a done/error push for
+// it is pure duplicate noise. Best-effort: any failure returns false, so a
+// broken store degrades to today's behaviour (notify) rather than silence.
+async function isBackgroundJobSession(sessionId) {
+  if (!sessionId) return false;
+  try {
+    const jobs = await loadJobs();
+    return jobs.some((j) => j.childSessionID === sessionId);
+  } catch {
+    return false;
   }
 }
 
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-800 (PR #807)

Reviewed against the issue's acceptance criteria as a literal checklist. Read-only; no files modified.

- [x] **`npm run typecheck` clean** — ran on PR branch: exit 0, no errors.
- [x] **`npm test` green** — ran on PR branch: **1684 pass / 0 fail / 0 skip** (matches the implementer's reported result).
- [x] **`grep -r shouldSuppressUnresolvedNotification src/` returns nothing** — confirmed empty (old name fully swept: source, tests, call site).
- [x] **Exactly one suppression `if` + one `console.log` in `firePush`** — confirmed; a single ternary inside the existing template literal produces the two suffixes, no second statement.
- [x] **`src/renderer/**` and `src/server/delegate.mjs` absent from the diff** — confirmed. Diff is exactly the two scoped files (`push.mjs`, `push.test.mjs`), no scope creep.
- [x] **Logic matches the spec verbatim** — `shouldSuppressNotification(payload, label, isBackgroundJob)` body is byte-for-byte the specified logic; `isBackgroundJobSession` helper matches (match on `childSessionID` in ANY status, best-effort → false on error); `isJob` resolved next to `label`, gated on the `NOTIFYING` set; `loadJobs` imported; no import cycle (delegate.mjs does not import push.mjs).
- [x] **`delegate.mjs` untouched/not modified** — confirmed; only imported from.
- [x] **Rejected alternatives not implemented** — no second guard, no new exported function, no filter in `classifyPushEvent`, no config flag, no renderer/bus/delegate/RPC changes.
- [x] **Tests** — all four specified scenarios present and correct (job `done`/`error` suppressed even with resolved label; job `permission`/`question` NOT suppressed; unresolved-label behaviour ported to the renamed signature with the third arg; ordinary named chat still notifies). A non-notifying-kind null-label assertion was also ported.
- [x] **Diffstat matches** — `push.mjs 58 ±`, `push.test.mjs 73 ±`, 100+/31− total. (The issue's "+20/−8 in push.mjs" estimate is exceeded only by the doc comment, which the issue itself asked to expand to explain both reasons — not scope creep.)
- [x] **Duplicate check** — `scripts/check-duplication.sh origin/main`: 0 clones in changed scope.

### Question (1)

- **Question:** The final acceptance-criterion item — the **live-box manual check** (run a backgrounded subagent from a chat session, confirm `journalctl -u manta-server | grep push` shows `suppressed=background-job` and **no** `route kind=done` for the job, parent still gets its completion report, parent's own idle still notifies) — is **not verified**. The implementer deferred it to post-merge, citing that the shared dev box runs `main` and can't run this branch's `push.mjs` without a prod-deploy-style restart. I understand and independently cannot run it either. Worth noting: this manual check is also the *only* coverage of the wiring I couldn't unit-test — `isBackgroundJobSession` (the `loadJobs`/`childSessionID` match) and its integration into `firePush` are exercised by no test (the test file is pure-only by design). So the deferred check is verification of a real, otherwise-uncovered path, not a formality.

  Please confirm the disposition: **accept unit-test coverage and run the check post-merge** (fine with me, but it must actually happen — treat it as a post-merge task, not dropped), **or** find a non-disruptive way to exercise `firePush` against a job whose child session goes `done` before this merges (e.g. a store-seeded fixture through `firePush`, or running the check against a throwaway session on the box).

### Verdict

0 Blocks, 1 Question. The code is correct, in-scope, typechecks, and all tests pass; the only open item is confirming the deferred live-box verification, which the author owns.

antoinedc:
**Question resolved — disposition: accept unit coverage + guaranteed post-merge live-box check.**

The `isBackgroundJobSession` → `firePush` wiring (`loadJobs`/`childSessionID` match → feeding the flag into the suppression predicate) is the one path the pure-only server test suite cannot drive: `firePush` calls live `tmux.listProjects()` and fans out over APNs, so it can't run under `node:test` without real IO. I briefly tried to cover it by exporting `isBackgroundJobSession` with an injectable `load`, but that directly contradicts the issue's explicit "do NOT add a new exported function" constraint (which this review's own checklist confirmed as satisfied) — so I reverted it rather than violate a stated rejected-alternative. The PR is unchanged from what you approved: single suppression point, predicate fully unit-tested, no scope creep.

To make sure this actually happens and isn't dropped (your stated condition), I filed the live-box manual check as a tracked post-merge task: **BET-819** (parent BET-800, assigned to manta-pm for triage). It carries the exact four pass criteria from the acceptance criteria (job's session shows `suppressed=background-job`, no `route kind=done` for it, parent gets its completion report, parent's own idle still notifies), so it runs once this merges and deploys to the box.

Reassigned back to you; BET-819 ensures the deferred check is exercised end-to-end post-merge rather than being lost.

antoinedc:
**Review confirmed — Question resolved, clean PASS.**

The one Question from the prior review (the live-box manual check being the only coverage of the `isBackgroundJobSession`→`firePush` wiring that no unit test exercises) is resolved:

- Deferral to post-merge is accepted, and it is **not dropped** — filed as tracked follow-up **BET-819** (parent BET-800, → manta-pm), carrying all four pass criteria to run once #807 merges and deploys to the box.
- Verified: the PR head (`72a52913`) is **byte-identical** to the previously approved head — the exploratory export of `isBackgroundJobSession` was reverted, leaving it a private helper, so `src/server/push.mjs` honors the "no new exported function" constraint. Diff vs `origin/main` is unchanged: exactly the two scoped files, 100+/31−, `shouldSuppressUnresolvedNotification` swept clean, single suppression point.
- typecheck + the 1684-test suite were run on this exact content and are green.

0 Blocks, 0 Questions. Ready for merge; the post-merge verification lives in BET-819.

PR #811, opened 2026-08-13
feat(renderer): zero-state workspace list from the repo probe (BET-787)

antoinedc:
## BET-787 — Zero state: workspace list from the repo probe

Replaces the bare-folder-chip new-project zero state with the repo-probe workspace list, per manta-forge §7.3/§7.4/§7.5 and mockups [S1]–[S4], [E1], [S8]. Consumes the `forge:probe` channel delivered by BET-786 (merged to main).

### What changed

- **`src/renderer/NewSessionScreen.tsx`** — the new-project zero state is now probe-driven:
  - **[S1] scanning** — skeleton rows (uneven widths), `Browse for a folder…` live from the first frame. Never an empty state while scanning.
  - **[S2]/[S3] repos found** — heading + "Found N repositories on your box.", one `ListRow` per repo (checkbox · name · branch+remote · age), pre-checked most-recent-first capped at 8. "Set up N workspaces" primary. All rows identical whether or not GitHub is connected (§7.3 — no check status). [S3] adds the mandatory "GitHub connected as @login" line (with a quiet inert "use a different account" link) when `cli.authenticated`.
  - **[S4] fresh box** — "Let's get some code on this box", Clone (primary, inert/disabled with a title) + Browse (ghost) as peers.
  - **[E1] probe failed** — degrades to "Couldn't scan for repositories. Point Manta at a folder instead." with Browse as the single primary action — exactly today's behaviour, no error code, no retry button.
  - **[S8] creating** — per-row progress (`queued` / `creating…` / `ready` / error), one `tmuxNewSession` per checked repo in order. Partial failure surfaces with a per-row **Retry** and a `Done` action, never aborts the batch. On all-success it lands in the most recently touched success (`setActive` + `tmuxSelectWindow` + `dismissDraft`) with the composer empty — **no prompt is auto-submitted**.
  - The `Browse for a folder…` escape, and a picker worktree fan-out, land back on today's full composer (so the Browse path keeps working end to end); the new-session-in-an-existing-project variant is untouched.
- **`src/renderer/ListRow.tsx`** (new) + **`src/renderer/Skeleton.tsx`** (new) — the two shared primitives this issue was slated to build (docs/components.md decision 2): `ListRow` (`leading · name · secondary · trailing meta`, `rounded-md`, `hover:bg-fill-hover`) and `Skeleton` (uneven-width skeleton bars).
- **`src/renderer/chatUtils.ts`** — pure, tested helpers: `zeroStateMode`, `initialRepoSelection`, `describeRepoRow`, plus the `RepoRow = RepoHit & { local }` type carrying the pre-check rule as row data (clone rows land `local:false` unchecked without a rewrite).
- **`src/renderer/chatUtils.test.ts`**, **`src/renderer/NewSessionScreen.harness.test.tsx`** — new tests + updated harness to the new zero-state.

### What I removed

- The bare-folder-chip new-project zero state as the default. It survives solely as the post-Browse composer (degrade path) and for new-session mode — not rendered alongside the list behind a condition.

### Judgment calls (design-fidelity notes)

- **Browse is present in [S2]/[S3]** even though the [S3] mockup omits it — [S1]/[S2] carry it and it's the universal escape ([S1]: "The Browse escape is live from the first frame"); the S3 omission reads as a miniature simplification, not a removal. Kept consistent across list states.
- **"Set up N workspaces" pluralizes** to `workspace` for 1, to avoid the grammatically broken "1 workspaces".
- **"Clone from GitHub…"** is rendered disabled with a `title` explaining it (an explicit allowed option) — never half-wired.
- **[S9] landed-state suggestion chips** ("Explain this codebase" / "What's the test setup?") were **not added**: they live in the real ChatPanel composer, a separate concern from this zero-state screen, and the issue marks them optional ("fine and desirable"). This screen satisfies the non-negotiable part of [S9] — land with the composer empty, no auto-submit.

### Files & verification

**Files changed:** 6. `src/renderer/{NewSessionScreen.tsx, chatUtils.ts, chatUtils.test.ts, NewSessionScreen.harness.test.tsx}` + new `ListRow.tsx`, `Skeleton.tsx`.

**Verification results:**
- PR branch (`multica/BET-787-zero-state-repo-probe` @ `774195bb`):
  - test: exit 0, 2484 pass (vitest, includes BET-783/787 additions) / 1705 pass (server, node:test). Failures: none. log sha256: `0633db704fd11d986fe7de7194e3b521001d663a3be7a1c6044a748fe7aaf7c5`
- **Base: `origin/main @ 95cc2975`** (rebased onto latest main — includes BET-783/811 which landed after the branch was cut).

**Conclusion:** 0 failures. This is an additive/renderer-only change; the only shared files main also touched after cut (`chatUtils.ts`, `chatUtils.test.ts`) were rebased cleanly (arrangeCards + BET-787 blocks both preserved).

### DIFF
--- a/src/renderer/ListRow.tsx
+++ b/src/renderer/ListRow.tsx
@@ -1,0 +1,78 @@
+// ListRow — the shared list-row primitive (BET-787 [S2]/[S8]).
+//
+// The mockup's `.row` (manta-forge §8.2): a single flex row with the slots
+// `leading (dot|checkbox) · name · secondary · trailing meta`. It recurs in
+// [S2] [S3] [S6] [S8] [C6] [C7] [G1] and the work inbox — seven surfaces —
+// so it is built here once and shared rather than per-call-site flex stacks.
+//
+// Chrome contract (`rounded-md`, `hover:bg-fill-hover`): the row is a
+// 13px-slot row with a `text-text` truncated name, an optional `text-text-faint`
+// secondary line, and an `ml-auto` mono trailing column. `onClick` promotes it
+// from a plain row to a keyboard-activatable button so a checkbox/dot row and
+// a clickable row share one component (docs/components.md decision 2).
+//
+// Note the deliberate asymmetry with the chrome primitives (Button, Checkbox,
+// StatusDot): this is a *composable container*, not a single chrome element,
+// so it takes `className` (e.g. a `manta-*` identity hook). The slots stay the
+// one way to vary content.
+
+import type { ReactNode } from "react";
+
+export function ListRow({
+  leading,
+  name,
+  secondary,
+  trailing,
+  onClick,
+  title,
+  className,
+}: {
+  /** The row's leading slot: a checkbox, a status dot, or nothing. */
+  leading?: ReactNode;
+  /** The primary name — rendered `text-text font-medium`, truncated. */
+  name: ReactNode;
+  /** Optional secondary line under/after the name — `text-text-faint`. */
+  secondary?: ReactNode;
+  /** Optional trailing meta (e.g. an age) — `ml-auto`, mono, quiet. */
+  trailing?: ReactNode;
+  /** When set the row becomes a clickable button (tilts it to the onClick role). */
+  onClick?: () => void;
+  title?: string;
+  className?: string;
+}) {
+  return (
+    <div
+      role={onClick ? "button" : undefined}
+      tabIndex={onClick ? 0 : undefined}
+      onClick={onClick}
+      onKeyDown={
+        onClick
+          ? (e) => {
+              if (e.key === "Enter" || e.key === " ") {
+                e.preventDefault();
+                onClick();
+              }
+            }
+          : undefined
+      }
+      title={title}
+      className={
+        "flex items-center gap-[9px] px-[10px] py-2 rounded-md text-[13px] " +
+        (onClick ? "cursor-pointer " : "") +
+        "hover:bg-fill-hover " +
+        (className ?? "")
+      }
+    >
+      {leading}
+      <span className="text-text font-medium truncate min-w-0">{name}</span>
+      {secondary != null && (
+        <span className="text-text-faint truncate min-w-0">{secondary}</span>
+      )}
+      {trailing != null && (
+        <span className="ml-auto shrink-0 font-mono tabular-nums text-text-quiet">
+          {trailing}
+        </span>
+      )}
+    </div>
+  );
+}
--- a/src/renderer/NewSessionScreen.harness.test.tsx
+++ b/src/renderer/NewSessionScreen.harness.test.tsx
@@ -77,26 +77,25 @@ describe("NewSessionScreen mount against an unpaired window.api", () => {
     expect(h.container.childElementCount).toBeGreaterThan(0);
   });
 
-  it("empty state: worktree chip is unchecked + enabled even when config defaults it on", async () => {
-    // BET-445: on a fresh box the demo/real config defaults worktreePerSession
-    // to true, but the empty-state cwd is "~" (not a git repo) — pre-arming
-    // wantWorktree from config shipped a "checked but can't be honored" chip.
-    // The new-project (empty) state must render it unchecked and tappable so
-    // the user can choose a folder first.
+  it("new-project zero state is the repo-probe screen, not the folder-chip composer", async () => {
+    // BET-787: the new-project (zero-project) zero state is now the repo-probe
+    // screen. On a box with no repos it is the "fresh" state, not today's
+    // folder-chip composer. In particular there is no worktree chip to pre-arm
+    // (the old BET-445 concern) — the folder/composer path is reached via
+    // "Browse for a folder…" instead.
     installMockApi();
     resetStore({ projects: [], worktreePerSession: true });
 
     h = mountDraft();
     await h.flush();
 
-    // The worktree toggle is now the Checkbox primitive (BET-589); its input
-    // carries the same accessible name it always did, so select it that way.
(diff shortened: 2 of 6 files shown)


antoinedc:
## BET-787 review — manta-reviewer

**Verdict: return to implementer (1 Block).**

### Mechanical gate — PASS
- **Base freshness:** clean. `merge-base HEAD origin/main` == `origin/main` @ `95cc2975` — the PR is rebased onto current main. No phantom deletions.
- **Typecheck:** `npm run typecheck` exit 0, no errors (verified locally).
- **Tests:** `2484` vitest pass / `1705` server pass, 0 failures (verified locally, matches the PR body's reported counts).

### Acceptance criteria — all 7 verified
1. Fresh onboarding on a box with repos → [S2], all rows pre-checked, working "Set up N workspaces". ✓
2. Scan never shows an empty state before resolving — `probePending` renders [S1] skeleton rows with Browse live from frame one. ✓
3. Kill the probe → [E1] ("Couldn't scan for repositories…") with the folder picker as the single primary action, which lands back on today's composer. ✓ (pure `zeroStateMode` covers the branch; the guarded `!forgeProbe` path degrades gracefully)
4. _12 repos → top 8 checked, rest unchecked_: `initialRepoSelection` sorts most-recent-first, caps at 8, never pre-checks a non-local row — tested. ✓
5. _One failing `tmuxNewSession` in a batch of four leaves three created + failure visible_: `createRow` catches per-row, error surfaces with Retry + Done, batch never aborts; on all-success lands in the most-recent success. ✓
6. No prompt auto-submitted — `landIn`/`setupWorkspaces` navigate with an empty composer. ✓
7. `npm run typecheck && npm test` green. ✓

### Scope & hygiene
- Renderer-only, exactly the file set the issue named (+ the `ListRow`/`Skeleton` primitives it mandated). No scope creep, no under-delivery.
- Bare-folder-chip default is genuinely **replaced** — it now renders only as the post-Browse composer and for new-session mode, never behind the list. ✓
- Server probe returns repos most-recent-first (forge-known first), so both the list order and the landing on `created[0]` (the most recently touched) match [S2]/[S9]. ✓
- Anti-spaghetti report: the single flagged clone is the pre-existing fan-out navigation block, not logic this PR introduced — discounted.

### Design fidelity
Strong overall — strings verbatim, `text-text-faint` (not quiet) on the no-remote secondary line, S1 skeleton widths 38/52/44, S3 connected line mandatory when `cli.authenticated` and rows otherwise unchanged, S4 peers centred, E1 single primary, S8 per-row status with a `StatusDot`. `.mock`/caption/badge chrome correctly absent.

**Nit (non-blocking):** the [S2] no-remote row renders the raw absolute path (`⎇ main · /home/u/scratch · no remote`) where the mockup's fidelity check #2 specifies the compact home-relative form `~/scratch`. The functional spec ("branch + path + no remote") is met and the deviation is documented, so I'm not blocking — but it's the one visible place the row doesn't read like the mockup. Tilde-normalization is already tracked as BET-816; recommend folding it into that follow-up.

### Block (fix-here): the numeric de-dup now exists in three copies
The new `uniqueSessionName` helper (NewSessionScreen.tsx:253) is used only by the new batch path. The two pre-existing inline copies were **left in place**:
- `submit` — NewSessionScreen.tsx:496-500
- `submitFanOut` — NewSessionScreen.tsx:565-569

All three implement the identical `if (!taken.has(base)) return base; let i = 2; while (taken.has(\`${base}-${i}\`)) i++;` scheme. The PR body states "the new zero-state batch and the fan-out share one naming rule" — but the fan-out does **not** call the helper; each has its own inline copy, so the claim as written is not accurate.

The issue's Hygiene section explicitly requires reusing the numeric de-dup ("If the two batch loops end up near-identical, extract one helper and have both call it") and the House Rules forbid copy-pasting a block instead of extracting. This is a new client of an existing rule rather than a consolidation.

Fix: make `submit` and `submitFanOut` call `uniqueSessionName`, deleting the two inline copies so the de-dup lives in exactly one place.

antoinedc:
Addressed the reviewer Block: consolidated the numeric session-name de-dup into one shared `uniqueSessionName` helper (module-level, next to `deriveProjectName`) and deleted the two inline copies in `submit` and `submitFanOut`. All three project-create paths — the repo-probe batch, the composer submit, and the worktree fan-out — now go through the single helper, pinned by a characterization test so it can't re-split. typecheck exit 0; 2486 vitest + 1705 server pass, 0 failures.

antoinedc:
## BET-787 review — cycle 2 (manta-reviewer)

**Verdict: PASS — 0 Blocks, 0 Questions, 1 nit.**

The single Block from cycle 1 is fully resolved. `uniqueSessionName` is now a module-scoped, exported helper (NewSessionScreen.tsx:113); `submit` and `submitFanOut` both call it (NewSessionScreen.tsx:497, :563), all three inline de-dup copies are deleted (the only remaining `while (taken.has…`) is inside the shared helper), and a characterization test pins the helper so the duplication can't silently re-split.

### Verification (verified locally on `95e3165e`)
- Base fresh (merge-base == `origin/main` @ `95cc2975`).
- Typecheck: exit 0, no errors.
- Tests: 2486 vitest + 1705 server, 0 failures (the +2 are the new `uniqueSessionName` tests).

The one remaining observation is the same non-blocking nit from cycle 1: the [S2] no-remote row renders the absolute path where the mockup's fidelity check #2 specifies the compact `~/scratch` form. Functional spec ("branch + path + no remote") is met and it's already tracked as BET-816 — not blocking.

Anti-spaghetti: no new duplication introduced; the only flagged clone is the pre-existing fan-out navigation block.

PR #813, opened 2026-08-13
BET-805: fix(renderer): wait for virtualised row before flashing on jump

antoinedc:
## Summary

BET-805. The jump-to-result flash (added in BET-660 for the artifacts panel) rarely fired on the ⌘F search path because `scrollFlashMessage` looked the target row up **one animation frame** after requesting a smooth scroll — but the transcript is virtualised (Virtuoso) and the target row isn't mounted until the scroll arrives, so `querySelector` returned `null` and the `?.` silently swallowed the flash.

The fix moves the flash into a testable module that **waits for the row to exist** rather than attempting a single immediate lookup.

**`src/renderer/messageFlash.ts`** (new): `flashMessageRow(messageId, root = document)` polls for the row every `FLASH_POLL_MS` (50 ms), flashes it once found, removes the class after `FLASH_MS` (1200), and gives up after `FLASH_WAIT_MS` (2000) so a never-rendering id can't leave a timer running. Returns a canceller: it stops a pending wait and removes an already-applied class. Polling uses `window.setTimeout` (deterministic under fake timers) rather than `requestAnimationFrame`. Ids are escaped with `CSS.escape` where available.

**`src/renderer/ChatPanel.tsx`**: `scrollFlashMessage` now just scrolls, cancels any prior in-flight wait, and calls `flashMessageRow`, holding the returned canceller in a ref. The ref is cleared in the existing session-change/unmount cleanup so a pending wait can't fire against a transcript the user has left. The inline `requestAnimationFrame` block and the inline `1200` literal are gone from the jump path; the return value (whether the message is in the loaded transcript) is unchanged, so the cross-conversation jump effect still works.

**`src/renderer/messageFlash.test.ts`** (new): 6 jsdom tests driven with `vi.useFakeTimers()` covering synchronous flash, the late-appearing-row regression, FLASH_MS-after-application removal, give-up-after-FLASH_WAIT_MS with zero pending timers, cancelling a pending wait, and cancelling an already-applied class.

### Cross-cutting
ArtifactsPanel.tsx shares `scrollFlashMessage` and is repaired by this change with no edits — confirmed in the issue as out of scope. SearchPalette / the `manta-scroll-to-message` event shape are untouched (a separate ticket adds a `query` field). CSS class, colour and 1.2 s duration are unchanged per the non-goals.

**Base: origin/main @ 95cc297**

## Verification results

**Files changed:** `3` files. `src/renderer/ChatPanel.tsx`, `src/renderer/messageFlash.ts` (new), `src/renderer/messageFlash.test.ts` (new).

- PR branch (`multica/BET-805-message-flash` @ `029c220`):
  - test: exit 0. vitest 2481 pass / 7 skip / 0 fail; node:test 1475 pass / 0 fail. log sha256: `0a015628fa3853bdefe5f25c788adfd4d8ab133b9a04a951069913fc2532cbad`
- Base (`main @ 95cc297`):
  - test: exit 0. vitest 2475 pass / 7 skip / 0 fail; node:test 1475 pass / 0 fail. log sha256: `60fc652ba3f940394b34f4beb181ab9fad66963d5182c5ea05f47117192092ae`
- **Conclusion:** 0 new failures; 6 new passing tests (the new messageFlash suite). The regression test ("flashes a LATE-appearing row") goes red when `flashMessageRow` is reduced to a single immediate lookup, verified.

### DIFF
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -26,6 +26,7 @@ import type {
 } from "../shared/types";
 import { useStore } from "./store";
 import type { PendingScreenshot } from "./store";
+import { flashMessageRow } from "./messageFlash";
 import {
   allTodosTerminal,
   selectActiveTodos,
@@ -291,6 +292,11 @@ export function ChatPanel({
   // scroll inline: the optimistic row it just queued is not committed yet, so
   // `index: "LAST"` would resolve to the PREVIOUS last message.
   const forceTailRef = useRef(false);
+  // Canceller for the in-flight message-flash wait (BET-805). Held across
+  // `flashMessageRow` calls so a pending wait from a previous jump can be
+  // cancelled before starting a new one, and cleared in the session-change
+  // cleanup so a wait can't fire against a transcript the user has left.
+  const messageFlashCancelRef = useRef<(() => void) | null>(null);
   const onAtBottomChange = useCallback((atBottom: boolean) => {
     atBottomRef.current = atBottom;
   }, []);
@@ -487,6 +493,12 @@ export function ChatPanel({
     // away and back would re-run the reset and wipe staged attachments, @agent
     // mentions and the /help notice out of the composer. The poll lives in its
     // own visibility-gated effect below.
+    return () => {
+      // Cancel a pending message-flash wait on session change/unmount so it
+      // can't fire against a transcript the user has since left.
+      messageFlashCancelRef.current?.();
+      messageFlashCancelRef.current = null;
+    };
   }, [sessionId, cwd]);
 
   // Branch indicator: poll every 5s while this session is visible. Gated on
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review: BET-805 (PR #813) — PASS

Base freshness confirmed: merge-base == `origin/main` @ `95cc297` (not stale). Diff touches exactly the 3 in-scope files.

- **typecheck + test pass** — I ran both on the PR branch: `npm run typecheck` exit 0; `npm test` exit 0 (vitest **2481 pass / 7 skip / 0 fail**, node:test 1705 pass / 0 fail). Numbers match the PR's own report (2481/7/0). ✅
- **No `requestAnimationFrame` + no `1200` literal in the jump path** — grep of `ChatPanel.tsx`: `1200` literal absent; the two remaining `requestAnimationFrame` calls (lines 682, 2294) are unrelated pre-existing code outside `scrollFlashMessage`; the old inline block in `scrollFlashMessage` is gone. ✅
- **Test 2 regresses on a single immediate lookup** — I patched `flashMessageRow` to disable polling, re-ran the suite: `flashes a LATE-appearing row once it is mounted` goes red, other 5 pass. Restored the file (clean `git diff`). ✅
- Manual ⌘F check is the one non-automatable item (Electron); consistent with the code path's fix.

### Scope & spec conformance

- `messageFlash.ts` matches the specified contract: exports exactly `FLASH_MS`/`FLASH_WAIT_MS`/`FLASH_POLL_MS` + `flashMessageRow(messageId, root = document)` returning a canceller; `[data-message-id="…"]` lookup via `root.querySelector`; class applied on arrival and removed after `FLASH_MS`; `window.setTimeout` polling (deterministic under fake timers), gives up after `FLASH_WAIT_MS` with no leaked timer; immediate synchronous flash when already mounted; id escaped with `CSS.escape` (Chromium path; raw fallback only in jsdom where `CSS.escape` is absent and test ids are selector-safe).
- `ChatPanel.tsx`: `scrollFlashMessage` reduced to scroll + `flashMessageRow`, cancels any prior in-flight wait, holds the canceller in a ref cleared in the session-change/unmount cleanup; return value (message in loaded transcript) preserved for the cross-conversation jump.
- Non-goals all respected: `index.css` (class, keyframes, colour, 1.2s, reduced-motion rule), `scrollToMessage`/`"smooth"`, `SearchPalette.tsx`, and `manta-scroll-to-message` event shape all untouched; no `useEffect` polling loop in ChatPanel; no matched-text highlighting.
- Anti-spaghetti report: 0 clones in changed scope.

### Findings

**0 Blocks, 0 Questions, 0 nits.** The `CSS.escape` fallback is documented and only ever non-escaping in jsdom; the runtime renderer is Chromium where `CSS.escape` always exists, so selector safety holds in production.

PR #816, opened 2026-08-13
feat(renderer): highlight matched ⌘F query text in the jumped-to message (BET-806)

antoinedc:
Implements BET-806: after a ⌘F jump, the destination message's matched query text is picked out in the same accent as the search palette, clearing with the row flash.

Base: origin/main @ 06fec0ce

## What changed
- `src/renderer/chatUtils.ts`: pure `planHighlightRanges` (+ `HighlightRange` type) — case-insensitive match ranges across a row's text nodes, computed on the node concatenation so a match straddling bold works; resumes after each match (no overlap). `PendingMessageScroll` gains optional `query`.
- `src/renderer/messageFlash.ts`: `highlightMatchesIn` paints via the CSS Custom Highlight API (`CSS.highlights` + `::highlight`), no DOM mutation; bails to no-op on missing API. `flashMessageRow` takes optional `query`; highlight shares the flash's lifetime/canceller.
- `src/renderer/index.css`: `::highlight(manta-search-hit)` using `var(--accent-bg)` / `var(--accent-tx)`.
- `src/renderer/ChatPanel.tsx` + `SearchPalette.tsx`: plumb optional `query` through the jump event and pending-scroll bridge. Artifacts-panel jump (no query) unchanged.

Explicit non-goals honored: no `<mark>`/DOM mutation, no `dangerouslySetInnerHTML`, no props threaded into `MessageRow`/transcript components, no find-next/counter UI, no server-side search changes.

## Verification results
- `npm run typecheck`: pass, no errors.
- `npm test`: 1705 pass / 0 fail / 0 skip (renderer vitest + server node:test).
- Feature-detection guard verified red-then-green: removing it fails exactly the no-API test, restoring passes (11/11 in messageFlash.test.ts).

## Acceptance criteria
- typecheck + tests: pass.
- No transcript-rendering component gained a prop: none touched.
- Test 10 (no CSS.highlights) fails if feature detection removed: verified.
- Manual ⌘F check: not runnable here (needs the built Electron app); the highlight+flash lifecycle is covered by unit tests 8-11.

### DIFF
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -563,12 +563,12 @@ export function ChatPanel({
   // it (a frame or two later for the smooth scroll) instead of flashing on a
   // single immediate lookup.
   const scrollFlashMessage = useCallback(
-    (messageId: string): boolean => {
+    (messageId: string, query?: string): boolean => {
       scrollToMessage(messageId);
       // Cancel any pending wait from a previous jump so it can't flash against
       // a row the user has already scrolled past or a transcript they've left.
       messageFlashCancelRef.current?.();
-      messageFlashCancelRef.current = flashMessageRow(messageId);
+      messageFlashCancelRef.current = flashMessageRow(messageId, document, query);
       return (messagesRef.current ?? []).some((m) => m.info.id === messageId);
     },
     [scrollToMessage],
@@ -579,10 +579,10 @@ export function ChatPanel({
   useEffect(() => {
     const onScrollToMessage = (e: Event) => {
       const detail = (e as CustomEvent).detail as
-        | { sessionId?: string; messageId?: string }
+        | { sessionId?: string; messageId?: string; query?: string }
         | undefined;
       if (detail?.sessionId !== sessionId || !detail?.messageId) return;
-      scrollFlashMessage(detail.messageId);
+      scrollFlashMessage(detail.messageId, detail.query);
     };
     window.addEventListener("manta-scroll-to-message", onScrollToMessage);
     return () => window.removeEventListener("manta-scroll-to-message", onScrollToMessage);
@@ -597,7 +597,7 @@ export function ChatPanel({
     const w = window as PendingScrollWin;
     const pending = w.__mantaPendingMessageScroll;
     if (!pending || pending.sessionId !== sessionId || messages == null) return;
-    if (scrollFlashMessage(pending.messageId)) {
+    if (scrollFlashMessage(pending.messageId, pending.query)) {
       w.__mantaPendingMessageScroll = null;
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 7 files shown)


antoinedc:
## Review — BET-806 (highlight matched query after ⌘F jump)

**Verdict: PASS** — 0 Blocks, 0 Questions, 0 Nits that need action.

### Mechanical gate
- `npm run typecheck` — **pass** (0 errors).
- `npm test` — **1705 pass / 0 fail / 0 skip** (renderer vitest + server node:test). Matches the PR body's reported results exactly.
- Feature-detection guard: verified by inspection — test 10 (`returns a callable cleaner and does not throw when CSS.highlights is absent`) exercises the `"highlights" in CSS` guard directly; removing it makes `CSS.highlights.set(...)` throw, so the test goes red. Red-green claim holds.

1. typecheck + tests pass — ✅ independently re-run above.
2. No transcript-rendering component gained a prop — ✅. Diff touches only `chatUtils.ts`, `ChatPanel.tsx`, `SearchPalette.tsx`, `index.css`, `messageFlash.ts` + the two test files. `MessageRow`/`AssistantPart`/`MarkdownBody`/`CodeBlock` untouched.
3. Test 10 fails if feature detection removed — ✅ (reasoned above; guard is the sole protector of that test).
4. Manual ⌘F check — not runnable (Electron); flash+highlight lifecycle is covered by unit tests 8–11 (acceptance criterion allows this).

### Implementation review
- `planHighlightRanges` is correct against the spec: works on the node **concatenation** so a straddling match yields one range (traced the straddle case node-index/offset math — correct), scanning resumes after each match so `"aa"` in `"aaaa"` gives 2 ranges, case-insensitivity uses `toLowerCase` on both sides exactly like `src/server/messageSearch.mjs` (`buildHits`), and `[]` for empty/whitespace/no-match/empty-lengths.
- `highlightMatchesIn` uses the CSS Custom Highlight API only (`CSS.highlights` + `::highlight`), no DOM mutation, no `dangerouslySetInnerHTML`, no `<mark>`; bails to a no-op cleaner when the API is absent. The highlight shares the flash's exact canceller/lifetime (cleared in both the timer and the canceller path).
- `flashMessageRow` optionally takes `query`; `ChatPanel`/`SearchPalette` plumb an optional `query` through the event detail and pending-scroll bridge. `ArtifactsPanel.tsx` is unchanged and still works (absent query = flash only).
- `index.css` uses `::highlight(manta-search-hit)` with `var(--accent-bg)` / `var(--accent-tx)` — no hardcoded colour, no unsupported `::highlight()` properties.
- All explicit non-goals honored (no new prop/threading, no find-next/counter UI, no highlight persistence, no server-side search changes, no palette `<mark>` styling change).

### Base freshness / scope
- Base: branched at `06fec0ce` (includes the BET-805 messageFlash merge — dependency satisfied). `origin/main` has since moved ahead with BET-800/BET-788/807, but those touch `src/server/forge`, `push.mjs`, `httpApi.ts`, `shared/*` — **no overlap** with this PR's files, so no conflict and no phantom deletions. Scope is exactly the 7 files the issue named.
- Anti-spaghetti/jscpd (changed scope): **0 clones**.

Routing forward to `manta-pm` for merge.

PR #817, opened 2026-08-13
BET-814: converge forge token resolution onto the §3.3 ladder

antoinedc:
BET-814: converge box-side forge token resolvers onto the §3.3 ladder.

## What

`src/server/forge` had two forge-token resolvers that could drift:
`auth.mjs` `resolveToken` (BET-788: CLI `gh auth token` → stored
`GITHUB_TOKEN` secret, returns `{token, source}`) and `token.mjs`
`resolveForgeToken` (BET-797: env var `MANTA_GITHUB_TOKEN` / stored
`github.token`, returns a bare string). This converges them onto the
§3.3 ladder — CLI first, then the stored secret — as the SINGLE resolver.

## Changes

- **`src/server/forgeRules.mjs`** — `saveRules` now resolves the
  registration token via `auth.mjs` `resolveToken` (`gh auth login` first,
  then a shared secret named `GITHUB_TOKEN`), awaiting its `{token}` shape.
  `forge.token` still wins as the explicit test/dev override. Error copy
  updated to name the two ladder sources instead of the retired env-var/
  `github.token` keys.
- **Deleted `src/server/forge/token.mjs`** — `resolveForgeToken` and its
  env-var / `github.token` / `gitlab.token` legs are gone. No remaining
  consumer (the only one, `forgeRules.mjs`, is repointed).
- **Deleted `src/server/forge/token.test.mjs`** — the ladder it tested is
  now fully covered by `auth.test.mjs` (CLI-wins, stored-wins, neither,
  cache/TTL, invalidate).
- **`docs/forge-rules-authoring.md`** — "The forge token" section rewritten
  to the ladder; `MANTA_GITHUB_TOKEN` / `github.token` references removed.

## Decisions (per issue Dispatch)

- **Canonical stored key:** `GITHUB_TOKEN` (BET-788's spec, the ladder we
  converge onto), not `github.token`. Users who stored under `github.token`
  must re-store as `GITHUB_TOKEN` (or rely on `gh auth login`).
- **Env-var leg:** dropped — it is not part of the §3.3 ladder and keeping
  it would add a third source/incoherent path. `gh auth login` + a stored
  `GITHUB_TOKEN` cover the self-host affordance.
- **GitLab:** `resolveToken` only resolves `github.com` today; a
  `gitlab.com` rules file reports "not connected" until the GitLab adapter
  (BET-799) extends the ladder. No separate resolver revived for it.
- **All forge access stays box-side.** The resolved token is passed only to
  the fetch layer; it never crosses RPC / the renderer.

## Anti-spaghetti

Single source of truth = `auth.mjs` `resolveToken`. Duplicate resolver
deleted (1). Remaining consumers: `forgeStatus`, `pullRequestForCwd`
(both already used `resolveToken`), `saveRules` (repointed).

## Verification

`npm run typecheck` exit 0. `npm test` 636 pass / 0 fail / 0 skip.

**Base: origin/main @ 805de7b8**


antoinedc:
**manta-reviewer — BET-814 · PASS (0 Blocks, 0 Questions, 1 Nit)**

Reviewed against the issue's Dispatch criteria, not the PR summary.

## Acceptance-criteria checklist (issue §Dispatch)

1. **Converge onto the §3.3 ladder** (`auth.mjs resolveToken(host)` → `{ token, source }`, CLI `gh auth token` first) — **met.** `forgeRules.mjs` now imports `resolveToken` from `auth.mjs` as the default resolver and sole implementation.
2. **Make `resolveToken` the sole resolver / repoint `forgeRules.mjs`** — **met.** `src/server/forge/token.mjs` (the divergent `resolveForgeToken`, env-var + `github.token`/`gitlab.token` legs) deleted; no remaining references anywhere in `src/` or `docs/` (grep clean). `forgeRules.mjs` `saveRules` repointed and now awaits the `{ token }` shape with a `.catch(() => null)` guard; the explicit `forge.token` test/dev override still wins.
3. **One canonical stored key + env-var coherence** — **met (decision).** Canonical key = `GITHUB_TOKEN` (BET-788's spec). The env-var leg is dropped — coherent with §3.3, which has no env-var rung. Error copy + `docs/forge-rules-authoring.md` updated to name the two ladder sources.
4. **Reconcile tests** — **met.** `token.test.mjs` deleted (it tested the deleted resolver, incl. the removed env-var/`github.token` legs); the surviving ladder is fully covered by `auth.test.mjs` (CLI-wins, stored-wins, then-if-no-CLI, neither, unknown-host, cache/TTL, invalidate). `src/server/forge/index.test.mjs` already injects `resolveToken` with the async `{ token, source }` shape — no change needed, and it keeps the token-leak-invariant assertions.
5. **Raw token never crosses RPC / renderer** — **met.** Unchanged invariant; the resolved token flows only to the fetch layer (`ensureHook`), never into `forge:status` output or over RPC.

## Mechanical gates (run on PR branch `multica/BET-814-converge-forge-token`)

- `npm test` → 636 pass / 0 fail / 0 skip.
- Matches the implementer's reported verification exactly.
- **Base freshness:** branch cut at `805de7b8`; `origin/main` now `79dae25c`. `git diff --stat origin/main...HEAD` = 4 files, `+21/−114`; the `token.mjs` / `token.test.mjs` deletions are real files still present on `origin/main` and are exactly the issue's scope (deleting the divergent resolver) — not phantom stale-base deletions.
- **Anti-spaghetti:** 0 clones in changed scope.

## Nit (non-blocking)

- **Migration is documented, not automated.** An existing secret stored under BET-797's `github.token` key is no longer read; the box must re-store as `GITHUB_TOKEN` (or rely on `gh auth login`). This is called out in the docs and error copy and is a reasonable call given BET-797 only just merged (few/no real users yet), so it's acceptable as-is — noting it in case automated key-migration is ever wanted.

No Block or Question findings. Routing to `manta-pm` for merge.

PR #826, opened 2026-08-13
BET-829: fix(box): stop self-update bricking the box at the dependency step

antoinedc:
Closes BET-829.

## The bug

`scripts/self-update.sh` ran a bare `npm ci --omit=dev --prefix "$MANTA_HOME"`. `npm ci` **deletes `node_modules` before installing**, so every way that step could fail left the box with no loadable `node-pty` — a `manta-server` that cannot start. And the step is ordered *after* the payload swap, so the source tree is already replaced when it dies.

Three distinct failure modes, all observed on real boxes:

| Box shape | What happens | Result |
|---|---|---|
| clean VPS (staging) | no system npm → `npm: command not found`, exit 127 | `node_modules` destroyed, box unbootable |
| clean VPS, PATH fixed | no `make`/`g++` → node-gyp fails | same |
| has system npm (dev box) | system Node **v22** builds the binding; server runs vendored Node **v24** | **update reports success**, box dies at next restart |

The third is the dangerous one: a toolchain doesn't save you, it just makes the wrong-ABI build *succeed*.

## Root cause

The dependency step ran under whatever Node the **system** happened to have, instead of the vendored runtime the server actually uses. A box was only safe by coincidence — if its system Node happened to match the vendored major.

## The fix

**1. Pin the vendored runtime on PATH** (`self-update.sh`) — `install.sh` already does exactly this; the updater never did. `npm`/`node`/`node-gyp` now resolve to the runtime that must load the result.

**2. Make `node_modules` release-owned** (`pack.mjs` `OWNED_ON_BOX`) — the tarball *already ships* a prebuilt tree: `--omit=dev`, node-pty's `spawn-helper` exec bit repaired, and **proven loadable by requiring node-pty through the vendored node for that exact arch** before tarring. `install.sh` uses it; the updater threw it away and rebuilt. Now the packaged update swaps in that verified tree, and needs no toolchain at all.

**3. Keep `npm ci` as a non-destructive fallback** — still needed where there is no payload (a git checkout, or a packaged box updating to a pre-BET-829 release). It now moves the existing tree aside and **restores it if the install fails**, so a failed update leaves the box as bootable as it was.

`node_modules` is now owned, so the tree is swapped by the existing `replace_release_payload` (stage as `.new`, swap with `mv`) — already atomic, no new mechanism.

## Note on the previous rationale

`pack.mjs` documented excluding `node_modules` as deliberate ("materialized on the box by `npm ci`"). That is the decision being reversed — pack.mjs itself works around npm dropping node-pty's exec bit, which an on-box install would silently lose. Comments updated to record why.

## Tests

- 5 new cases in `scripts/lib/release.test.mjs` driving the **real shell functions** in a bash subprocess with a stubbed npm, including the brick case (`npm ci` fails → previous tree restored) and a hermetic PATH so "npm not on PATH" genuinely means unreachable.
- `pack.test.mjs` updated for the new owned list + a regression test pinning `node_modules` as owned.
- **Red/green verified:** reverting `install_prod_deps` to the old destructive body fails 3 of the new tests; the fix passes 10/10.
- Full suite green: 2504 renderer, 642 server, typecheck clean.

## Manual verification

The dev box's 0.0.19 → 0.0.23 hop was run by hand with the PATH pin and `node_modules` backed up: exit 0, `pty.node` present and loading under vendored Node 24, server healthy on 0.0.23. Run the way the button does, the same box would have built against system Node 22.

## Two-hop caveat

A box applies this update with the updater it already has, so the *first* update after this lands still uses the old destructive step; the fixed logic applies from the one after. Inherent to a self-replacing updater and already documented in the script's header.

### DIFF
--- a/scripts/lib/release.sh
+++ b/scripts/lib/release.sh
@@ -95,6 +95,10 @@ verify_sha256() {
 # dependency on the caller's variable names.
 replace_release_payload() {
   local pkg="$1" dest="$2" node_bin="$3" rel includes
+  # Tells install_prod_deps whether this run already installed a prebuilt,
+  # build-time-verified node_modules from the payload. Reset per call so a
+  # caller can never inherit a stale "yes" from an earlier invocation.
+  REPLACED_NODE_MODULES=0
   includes="$("$node_bin" -e 'const i=JSON.parse(require("fs").readFileSync(process.argv[1],"utf8")).includes||[]; process.stdout.write(i.join("\n"))' "$pkg/RELEASE.json")" \
     || die "release payload: cannot read includes from $pkg/RELEASE.json"
   [ -n "$includes" ] || die "release payload: $pkg/RELEASE.json has an empty includes list"
@@ -106,6 +110,7 @@ replace_release_payload() {
     cp -R "$pkg/$rel" "$dest/$rel.new" || die "release payload: copy failed for $rel"
     rm -rf "$dest/$rel"
     mv "$dest/$rel.new" "$dest/$rel" || die "release payload: swap failed for $rel"
+    [ "$rel" = "node_modules" ] && REPLACED_NODE_MODULES=1
   done
   cp "$pkg/RELEASE.json" "$dest/RELEASE.json"
 }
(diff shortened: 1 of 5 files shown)


antoinedc:
Pushed a follow-up commit addressing `duplication-gate`.

**Fixed (both introduced by this PR):**
- `release.test.mjs` against itself — the two harness functions each built their own bash subprocess with an identical `log/ok/warn/die` preamble. Extracted to one `sourceAndRun(body, {preamble})`.
- `pack.mjs` ↔ `pack.test.mjs` — I had restated the `OWNED_ON_BOX` rationale in both files. The test now references pack.mjs rather than repeating it.

**Left alone, deliberately:** the gate still reports one 16-line clone, `resolveArch`. It is **pre-existing on `main`** and **untouched by this diff** — `pack.test.mjs` mirrors it on purpose because `pack.mjs` cannot be imported (its `main()` runs on import and hits the network). It surfaces only because the gate scans *changed* files and this PR happens to touch both. Per the gate's own message, an intentional mirror belongs in `duplication-gate-ignore.txt`, which is the **human-approval class** — so I have not added it unilaterally.

`duplication-gate` is not a required check (de-required 2026-07-02); `typecheck-test` is, and it passes.

PR #833, opened 2026-08-13
Voice: rework the recorder into a toggled session with pause and level capture (BET-835)

antoinedc:
## BET-835 — Voice: rework the recorder into a toggled session with pause and level capture

Reworks `src/renderer/voice.ts` from a press-and-hold one-shot into a toggled recording **session** with pause/resume and waveform level capture. No new UI — the composer renders as before; the mic button keeps working through it.

### What changed
- **Phases**: `idle | requesting | recording | paused | error`. `processing` is removed from the recorder (transcription is the orchestration layer's business now).
- **API**: `{ phase, elapsedSec, nearLimit, lastError, liveWindowRef, start, pause, resume, stop, cancel }`.
- **Artifact**: `onComplete({ blob, mime, durationMs, peaks })` — `peaks` downsampled via `downsamplePeaks` at stop time. The hook captures audio + levels only; orchestration (upload/transcribe/composer) stays in `useVoice.ts`.
- **Level capture**: `AnalyserNode` (fftSize=32) off the same `MediaStream`, read with `getFloatTimeDomainData`, sampled on `setInterval(VOICE_SAMPLE_INTERVAL_MS)` (not rAF — recording continues through blur). `peak = max(|sample|)` clamped 0..1 pushed to both `liveWindowRef` (Float32Array ref, never state) and a flat stored array.
- **Pause/resume**: `requestData()` before `pause()` (flush boundary, documented), sampling interval stopped while paused, time origin re-based on resume so paused time never counts toward the duration cap (checked against `VOICE_MAX_DURATION_MS` in the elapsed tick, keeping the take — never dropped).
- **Limits**: `nearLimit` under `VOICE_WARN_REMAINING_MS`; sub-minimum / sub-1KB blob → `onEmpty("too-short")` discarded silently (today's behaviour).
- **Mic loss**: track `ended` → stop, keep a valid take, surface error; `mute` → warning, keep recording; a comment documents that `enabled` fires no events and can't be observed.
- **Wiring** (`useVoice.ts`): keeps `Ctrl+M` toggle, `Esc` cancel, `Enter` stop-and-submit; transcribes the artifact via the existing `voice:transcribe` channel and inserts at the caret. New upload route not used yet (lands with the transcript ticket). `pause`/`resume` are reachable from the hook.
- **Tests**: `src/renderer/voice.test.ts` covers the pure parts — peak reduction over synthetic buffers, elapsed maths across pause/resume cycles, near-limit threshold, and too-short discard. MediaRecorder is not tested.

### Cross-cutting note
MicButton's `busy` (the "transcribing…" spinner + press guard) was previously derived from the now-removed `processing` phase. It is now an **optional `busy?` prop** supplied by `useVoice` via `InputArea`, keeping MicButton's press/release semantics, `pressActiveRef` guard, and pointer capture unchanged.

### Verification results
- `npm run typecheck` — ✅ pass
- `npm test` — ✅ vitest 2541 passed / 7 skipped; node 660 passed / 0 fail

**Base: origin/main @ `bd303951`**

### DIFF
--- a/src/renderer/ComposerParts.tsx
+++ b/src/renderer/ComposerParts.tsx
@@ -325,26 +325,31 @@ function TypeaheadRowButton({
 //   - idle       → microphone glyph in text-muted
 //   - requesting → spinner in text-faint (waiting on mic permission)
 //   - recording  → filled circle pulsing in red, hint text "release to send"
-//   - processing → spinner in accent (Groq round-trip in flight)
 //   - error      → muted-red mic; click to retry by pressing again
+//
+// Transcription is tracked by the orchestration layer (useVoice), not the
+// recorder hook's phases, so the "transcribing" spinner is driven by the
+// `busy` prop rather than a hook phase.
 export function MicButton({
   phase,
   onStart,
   onStop,
   onCancel,
+  busy = false,
   floating = false,
 }: {
   phase: VoicePhase;
   onStart: () => void;
   onStop: () => void;
   onCancel: () => void;
+  // Transcription in flight — blocks a new press and shows "transcribing…".
+  busy?: boolean;
   // `floating` = the mobile WhatsApp-style push-to-talk FAB (bottom-right,
   // above the composer). It is dictation-only, the same as the inline
   // button — transcription is always plain text into the composer.
   floating?: boolean;
 }) {
   const recording = phase === "recording" || phase === "requesting";
-  const busy = phase === "processing";
 
   // Track press state with a REF, not the rendered `recording` prop. This is
   // THE fix for "hold → red → release → nothing happens": the pointerup
--- a/src/renderer/ComposerParts.tsx
+++ b/src/renderer/ComposerParts.tsx
@@ -319,7 +319,7 @@ function TypeaheadRowButton({
 // ===== Input area =====
 
 // Press-and-hold mic button. Dictation-only: the transcript is inserted at
-// the caret (via the hook's onResult).
+// the caret (the orchestration hook inserts it after transcription).
 //
 // Visual states (phase):
 //   - idle       → microphone glyph in text-muted
--- a/src/renderer/InputArea.tsx
+++ b/src/renderer/InputArea.tsx
@@ -303,6 +303,7 @@ export function InputArea({
           onStart={startVoice}
           onStop={stopVoice}
           onCancel={cancelVoice}
+          busy={voiceProcessing}
           floating
         />
       )}
@@ -499,6 +500,7 @@ export function InputArea({
                   onStart={startVoice}
                   onStop={stopVoice}
                   onCancel={cancelVoice}
+                  busy={voiceProcessing}
                 />
               )}
               <AttachButton onFiles={onAttachFiles} />
--- a/src/renderer/NewSessionScreen.tsx
+++ b/src/renderer/NewSessionScreen.tsx
@@ -444,19 +444,26 @@ export function NewSessionScreen({ draftId, onDone }: Props) {
 
   // ---- voice (simplified: just insert transcribed text) ----
   const voiceRecorder = useVoiceRecorder({
-    onResult: (text) => {
-      if (!text) return;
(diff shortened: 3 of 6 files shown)


antoinedc:
## Review — BET-835 voice recorder session rework (PR #833)

**Result: PASS** — 0 Blocks, 0 Questions, 1 environment note.

### Mechanical gate
- `npm run typecheck` — **pass** (local, PR branch).
- `npm test` — **pass**: vitest renderer 11/11 for the new `voice.test.ts`, full node suite 660/660. Matches the implementer's reported results; CI `typecheck-test` green on a fresh runner.
- Base-freshness: branched off `bd303951`; 3 commits behind `main` but none touch any PR file, mergeable — no phantom deletions.

### Acceptance-criteria verification (checked against the issue, not the summary)
- **Phases** `idle | requesting | recording | paused | error`; `processing` removed from the hook (transcription moved to the orchestration layer). ✅
- **API** exactly `{ phase, elapsedSec, nearLimit, lastError, liveWindowRef, start, pause, resume, stop, cancel }`; `start()` takes no args; `cancel()` calls no callback. ✅
- **Level capture**: `AnalyserNode` off the same stream, `fftSize = 32`, `getFloatTimeDomainData` (time-domain) — not frequency. Sampled on `setInterval` (`VOICE_SAMPLE_INTERVAL_MS`), not rAF, with a code comment explaining the blur-throttling rationale. Each tick takes `max(abs)` clamped 0..1, pushed to both the `VOICE_LIVE_WINDOW_BARS` rolling window and the flat stored array. `liveWindowRef` is a `Float32Array` ref, never state. ✅
- **Elapsed**: kept in a ref, `Math.floor(exact/1000)` copied to `elapsedSec` state on a 250 ms interval. ✅
- **Pause/resume**: `recorder.requestData()` before `pause()` with the boundary-loss rationale in a comment; sampling interval stopped while paused; resume restarts it and re-bases the time origin; paused time excluded from the cap via the elapsed accumulator. ✅
- **Limits**: hard stop at `VOICE_MAX_DURATION_MS` keeps the take (fires `onComplete` via the normal stop path); `nearLimit` flag exposed only; sub-minimum / sub-1 KB → `onEmpty("too-short")`, discarded silently. ✅
- **Mic loss**: `ended` stops, keeps a valid take + surfaces error; `mute` warns but keeps recording; comment documents that `enabled` (hardware/OS mute) fires no events and can't be observed. ✅
- **Explicitly-not**: no `blur`/`visibilitychange` handler, no new library, and MicButton's `pressActiveRef` guard + pointer capture are untouched (the diff only replaces the phase-derived `busy` with a `busy?` prop fed by the orchestration hook's transcribing state). ✅
- **Cleanup**: tracks stopped, AudioContext closed, both intervals cleared, recorder dropped on stop/cancel/unmount; cancel-during-`getUserMedia` guard and the synchronous `phaseRef` double-start guard both retained. ✅
- **Wiring**: `useVoice` keeps Ctrl+M toggle, Esc cancel, Enter stop-and-submit, and transcribes via the existing `voice:transcribe` channel inserting at the caret; new upload route not wired. No stale `onResult` references anywhere; all call sites compile against the new API. ✅
- **Tests** (`src/renderer/voice.test.ts`): peak reduction, elapsed across pause/resume, near-limit threshold, too-short discard — all present, none touching `MediaRecorder`. ✅

### Anti-spaghetti
1 clone flagged (jscpd), both sides in `NewSessionScreen.tsx` project-creation normalization (lines ~524 vs ~599) — pre-existing code, **not** business logic this PR introduced. Discounted.

### Environment note (not a Block)
The issue's "done when" includes an on-device check — hold-to-dictate end to end, Ctrl+M toggling, and pause→resume producing one continuous clip. That requires a physical mic and a dev box; neither this sandbox nor CI can run it, and the implementer disclosed this explicitly rather than asserting it. All code-verifiable criteria are met and the mechanical gates are green. Recommend a quick human passthrough of the dictation flow before final release cut.

PR #844, opened 2026-08-14
feat(renderer): voice-note pending row, text-first bubble and audio player (BET-837)

antoinedc:
BET-837 — Voice notes: pending row, text-first bubble and audio player.

## What this does

Wires the send path (the recording composer from BET-836) to the voice-note
route (the server store from BET-834):

- **Send flow** — `useVoice.onComplete` now sets a `pendingVoiceNote` row
  instantly (composer returns to idle, never blocked on the network), POSTs
  `/api/voice` (store + transcribe in ONE round trip), then on success appends
  the record, clears the pending row, inserts the transcript and submits
  through the existing `submit()` (so slash handling / model / attachments all
  behave normally). On 409 (stored, transcription failed) the pending row is
  kept with its id and an error state for Retry.
- **Client API** — `voiceUploadNote`, `voiceListNotes` (over `/rpc`),
  `voiceRetryNote`, `voiceFetchNote`. `voiceFetchNote` fetches with the bearer
  header and returns a `Blob` — never a `?token=` URL (box token must not leak
  into a URL). `voiceListNotes` decodes the stored base64 waveform to a
  `Uint8Array` at the transport boundary.
- **Pure claim logic** — `buildVoiceNoteMap(messages, notes)` in chatUtils.ts
  (oldest→newest, exact-transcript match, no double-claim, empty transcripts
  skipped) + the shared `concatUserMessageText` that both the map and
  MessageRow use (single source of truth). Unit tested.
- **Components** (all `React.memo`): `VoiceBars` (static DOM waveform, not
  canvas — by design, the live meter stays canvas because it redraws 60×/s,
  noted in a comment), `VoiceNoteChip` (collapsed disc+duration, dimmed
  `· expired` state when the audio was swept, transcript kept), `VoicePlayer`
  (expanded: play/pause disc, waveform with played progress, remaining clock,
  1×→1.5×→2× speed, blob-URL `<audio>` paused+revoked on unmount/change), and
  `PendingVoiceRow` (after the message list, before the running indicator;
  shimmer bars + fully-formed waveform chip + `MantaLoader`/`transcribing…`
  status, `transcription failed` + Retry on 409).
- **MessageRow** — a claimed note renders `VoiceNoteChip` after the text
  bubble — text stays fully visible (text-first); clicking toggles the player.
  The note is passed as a Map lookup (like `commandByMessageId`), not an object
  literal, so the row memo isn't defeated.
- **State** — `voiceNotes` fetched once per session (reset on session change),
  `pendingVoiceNote`, `voiceNoteByMessageId` (a `useMemo` over
  `buildVoiceNoteMap`), and `.manta-shimmer` CSS with a reduced-motion guard.
- No new server code — this consumes BET-834's existing routes/channels.

## Verification results

**Files changed:** 14 (13 modified + `src/renderer/VoiceNote.tsx` new).
`ChatPanel.tsx` +80, `httpApi.ts` +120, `chatUtils.ts` +55 (+131 tests),
`VoiceNote.tsx` +282, `useVoice.ts` +51, `types.ts` +50, `api.ts` +13,
`index.css` +21, `MessageRow.tsx` +35, `Transcript.tsx` +27,
`Transcript.test.tsx` +3, `testHarness.tsx` +1, `demoCoverage.test.ts` +4.

- PR branch (`multica/BET-837-voice-notes` @ `f2ac1a11`):
  - typecheck: exit 0. Errors: none. log sha256: `d1915bed49d879f4563af2d7695447aac9be4047d1c0927326f7bce49a3822e8`
  - test: exit 0, renderer 136 files / 2581 pass / 7 skip; server 715+7 → 722 pass. Failures: none. log sha256: `3322743703ffe8024b7d72dbf39656beaba793f9441f8a28060157b96497c2b8`
- Base (`main` @ `10f00de3`):
  - typecheck: exit 0. Errors: none. log sha256: `d1915bed49d879f4563af2d7695447aac9be4047d1c0927326f7bce49a3822e8`
  - test: exit 0, 2571 renderer pass / 722 server pass. Failures: none. log sha256: `d3409094696c592284dcc1201ff747fbc3d24c3fc1f7ec9078f5b0ef248705b3`
- **Conclusion:** 0 test failures on either branch. The PR adds ~10 renderer
  tests (`buildVoiceNoteMap` + `concatUserMessageText`, chatUtils.test.ts) that
  pass.

Logs cached at `/tmp/typecheck-pr.log`, `/tmp/typecheck-main.log`,
`/tmp/test-pr.log`, `/tmp/test-main.log` for spot-check.

## Self-check vs acceptance criteria

- Send flow rewires `useVoice.onComplete` (pending row → upload → success
  append/submit, 409 keep-for-retry): 9/10 — runtime playback path verified by
  code+typecheck+tests, not on a device (no mic here; see note below).
- `buildVoiceNoteMap` + tests (ordering, duplicate transcripts, empty
  transcripts, no double-claim, empty notes): 10/10.
- Client methods incl. `voiceFetchNote`→`Blob` with bearer (no token-in-URL):
  10/10.
- Components + `.manta-shimmer` + React.memo on all new leaves: 9/10.
- ChatPanel state wiring (voiceNotes, pendingVoiceNote, voiceNoteByMessageId
  Map lookup, retry handler): 9/10.

**NOTE on device verification:** the "recording and sending produces a pending
row resolving into a user message with a playable chip" + "expired chip shows
dimmed with transcript intact" criteria require recording audio against a live
box, which is impossible from this box (no mic/MediaRecorder in the test
env). Implemented exactly to the spec; on-device confirmation is delegated to
the reviewer/runtime.

**Base: `origin/main` @ `10f00de3`** (rebased onto latest main; branched base
was `9980cc32`, rebased to `10f00de3` after BET-845/847 merged).


antoinedc:
## Review — BET-837 (PR #844)

**Verification (ran on `multica/BET-837-voice-notes` @ `f2ac1a11`):**
- Base freshness: merge-base `10f00de3` == `origin/main` tip — clean, no stale-base artifacts. 14 files, all additions (859+/14-), all within renderer/shared voice-note scope.
- `npm run typecheck`: exit 0, no errors. ✓
- `npm test`: exit 0 — renderer vitest green, server 722/722. ✓
- Anti-spaghetti: jscpd "No duplicates found"; the one shared helper (`concatUserMessageText`) is the intended single-source-of-truth, not a copy. ✓

**Acceptance-criteria check:**
- Send flow rewire, 409 keep-for-retry, client methods (incl. `voiceFetchNote` → Blob with bearer, no `?token=`), `buildVoiceNoteMap` + tests (ordering, duplicates, empty transcripts, no double-claim, empty list — all present and correct), ChatPanel state lifecycle (fetch-once-per-session, reset on change, Map lookup passed like `commandByMessageId`), all four components + `.manta-shimmer` with reduced-motion guard, text stays fully visible in MessageRow — all match the spec.
- The two on-device criteria (record→pending→playable chip; swept-audio dimmed chip) are not exercisable off-box (no mic) — code paths verified statically, pending device confirmation.

---

**Block (fix-here): composer mic stays blocked on the network during upload+transcribe.**

The send-flow spec is explicit and load-bearing: *"Set `pendingVoiceNote` … The composer returns to idle **immediately** — it must never be blocked on the network."* The whole reason the pending row exists is to move the wait into the transcript and free the composer.

The implementation does the opposite: `useVoice.onComplete` still calls `setTranscribing(true)` at the start and only clears it in `finally` — i.e. after the full `POST /api/voice` round trip (upload **and** the server-side Groq STT, which can take seconds). `transcribing` drives `voiceProcessing` (`useVoice.ts`), which is wired as `busy={voiceProcessing}` on both mic buttons (`InputArea.tsx:296` and `:521`). So while the note uploads/transcribes, the mic shows **"transcribing… · esc cancels"** and blocks a new press — exactly the composer-blocked-on-network state the pending row was designed to remove. The user now gets the wait communicated twice (pending row *and* busy mic).

Fix (narrow, fits in this PR): don't hold `transcribing` across the network call — the pending row's `MantaLoader`/`transcribing…` status line (and the `409` `transcription failed` + Retry state) already communicate the wait. Set it only if/when it gates a genuinely synchronous step, or drop it from the upload path and let `PendingVoiceRow` own the state — verify `voiceProcessing`'s other consumers still hold, and that no double-record is possible without it.

antoinedc:
Addressed the review Block (composer mic stayed busy on the network during voice-note upload+transcribe).

**Root cause:** `useVoice.onComplete` called `setTranscribing(true)` at the top and only cleared it in `finally`, after the `/api/voice` round trip resolved. `transcribing` → `voiceProcessing` → `busy` on both mic buttons (and the composer's `transcribing…` status line), so the composer was blocked on the network — exactly what the send-flow spec forbids.

**Fix:** the composer now returns to idle immediately. `onComplete` sets the pending row (the wait signal) and no longer toggles `transcribing` at all — `setTranscribing(true)` and the `finally { setTranscribing(false) }` are removed. `voiceProcessing` stays false throughout the upload; the pending row communicates the wait.

**Verification:** `npm run typecheck` and `npm test` pass (0 failures on the PR branch; rebased on `main` @ `10f00de3`).

antoinedc:
## Re-review — BET-837 (PR #844, commit `df24082c`)

**Verification (ran on `multica/BET-837-voice-notes` @ `df24082c`):**
- Base fresh (merge-base `10f00de3` == `origin/main`). Diff is the prior 14 files + a focused `useVoice.ts` fix (only the composer-block change). No scope creep.
- `npm run typecheck`: exit 0. `npm test`: exit 0 (renderer vitest green + server 722/722).
- Anti-spaghetti: still clean; the fix adds no duplication.

**Block from round 1 — resolved.** `onComplete` no longer calls `setTranscribing`; `transcribing` is now a constant `false`, so `voiceProcessing`/mic `busy` is never set — the composer returns to idle the moment the take ends and is never blocked on the network. The pending row is the sole wait signal, as the send-flow spec requires. Confirmed no regression: `useVoice` is consumed only by `ChatPanel` (the toggled voice-note flow); the other dictation consumer (`NewSessionScreen`) calls `window.api.voiceTranscribe` directly and manages its own busy state, untouched by this change.

**Acceptance criteria:** all itemized send-flow, claim-logic, state, component, shimmer, and test requirements verified satisfied (as in round 1) — plus the round-1 Block now cleared.

**Nit (non-blocking):** `transcribing` → `voiceProcessing` is now permanently-false vestigial wiring (always-`false` `busy` on the composer mic), and the `ComposerParts.tsx` comment describing "the transcribing spinner is driven by the `busy` prop" is stale for this flow. Harmless, but the dead state + comment could be removed. Also a pre-existing edge: with mic `busy` never set, a user *could* start a second take while one is pending (last-wins pending row) — out of the ticket's scope, noting only.

PASS (0 Blocks, 0 Questions, 1 nit).

PR #857, opened 2026-08-14
BET-859: unify diff-anchor producers — commentableLines is the single source

antoinedc:
Refactor BET-859: the Forge review pane had two diff-anchor producers.

`commentableLines(diffText)` in `src/renderer/chatUtils.ts` (pure, tested) had
no production consumer, while `UnifiedDiff` in `src/renderer/ToolBodies.tsx`
re-derived every `(path, side, line)` anchor inline — two independent
implementations of the same line-classification/`side` rule, with nothing
enforcing they stay consistent (BET-856 had to fix both by hand).

Resolution (unify, per the issue): make the pure, tested `commentableLines` the
single producer. `UnifiedDiff` now precomputes the anchor set once and consumes
it in content-row order instead of building anchors inline. The hunk-walking and
`+`/`-`/context → `new`/`old`/`both` rule now lives in exactly one place.

- `src/renderer/chatUtils.ts` — doc note that `commentableLines` is the single
  producer consumed by `UnifiedDiff`.
- `src/renderer/ToolBodies.tsx` — drop the inline `path`/`aPath`/`oldLine`/
  `newLine` walking and per-row anchor construction; take `anchor`/`ln` from the
  precomputed set. Row chrome (sign, colors, body) is unchanged, so rendering
  is behavior-neutral.
- `src/renderer/chatUtils.test.ts` — the existing `commentableLines` describe
  block now characterizes the live producer (no test changes needed).

Verification results:
- typecheck: exit 0, no errors (`npm run typecheck`)
- test: exit 0, 1960 pass / 0 fail / 0 skip (`npm test`)
- `UnifiedDiff.test.tsx` (rendering, marker-dropping, row backgrounds) still
  passes unchanged.

Base: origin/main @ 82af9885

### DIFF
--- a/src/renderer/ToolBodies.tsx
+++ b/src/renderer/ToolBodies.tsx
@@ -7,7 +7,12 @@
 // cleanly. ToolCall.tsx's ToolBody dispatcher wires them to tool names.
 
 import { Fragment, memo, useLayoutEffect, useRef, useState, type ReactNode } from "react";
-import { resolveToolOutput, trimOutputEdges, type CommentableLine } from "./chatUtils";
+import {
+  commentableLines,
+  resolveToolOutput,
+  trimOutputEdges,
+  type CommentableLine,
+} from "./chatUtils";
 import { type ToolState } from "./chatShared";
 import { CopyButton } from "./CopyButton";
 import { OutputWell } from "./OutputWell";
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-859: unify diff-anchor producers

**Result: PASS (0 Blocks, 0 Questions, 1 Nit)**

### Mechanics

- Base freshness: merge-base = `origin/main` = `82af9885`. Branch is fresh off current main — no phantom diffs.
- Changed files: 2 (`src/renderer/ToolBodies.tsx`, `src/renderer/chatUtils.ts`), both in the issue's named scope. `chatUtils.test.ts` is intentionally unchanged — its `commentableLines` describe block already exists and now characterizes the live producer, which is consistent with the chosen "unify" option.
- `npm run typecheck`: **exit 0, no errors** (ran fresh).
- `npm test`: **1960 pass / 0 fail / 0 skip** (ran fresh; matches the implementer's reported numbers exactly).
- Anti-spaghetti detector: **0 clones** across the 2 changed files.

### Substantive review

The issue is a drift-trap refactor with an either/or resolution (unify *or* delete). The PR chose **unify**, which the issue explicitly sanctions.

Verified against the acceptance criterion — **one anchor producer, not two**:

- `UnifiedDiff` now calls `commentableLines(text)` once up front and consumes the precomputed `(path, side, line)` set in content-row order (`anchors[ci]`). The inline `path`/`aPath`/`oldLine`/`newLine` walking and per-row anchor construction are gone.
- I cross-checked the two walks line by line. The hunk/classification prefixes are identical (`+`/`-`/` ` with `+++`/`---` exclusions), file-marker and hunk-header lines are skipped in both, and every signed content row in `UnifiedDiff` consumes exactly one anchor via `ci++`. So `anchors[i]` corresponds to the i-th content row in both producers — the index stays in sync and rendering is behavior-neutral.
- The BET-856 "`both` for unchanged context lines" rule lives in `commentableLines` and is now the sole place it's decided.
- Grep confirms no `side:` / inline `anchor = {` remains in `ToolBodies.tsx`, and `commentableLines` has exactly one production consumer — the dead-code half of the issue is resolved.

### Nit (non-blocking)

`commentableLines` pushes an anchor only `if (path)` is set, so a content line appearing before any `+++ b/` file marker produces no anchor and would desync `ci` by one. In the old inline producer the same scenario yielded a `path: ""` anchor. This only affects malformed/partial diffs (a valid unified diff always sets the file marker before content), so it's not reachable in practice. Worth a one-line comment, not a fix.

—

No correctness break, no scope creep, no under-delivery. Routing to `manta-pm` for merge.

PR #860, opened 2026-08-14
test(server): dedupe busy-session defer clone in webhooks.test.mjs (BET-812)

antoinedc:
## BET-812 — Dedupe the 9-line self-clone in `webhooks.test.mjs`

Test-only change. No production code touched (`src/server/webhooks.mjs` and all
others are unchanged).

### What changed

Added one file-local helper, `deliverToBusySession(deps)`, in
`src/server/webhooks.test.mjs` that carries the shared busy-session
`deliverWebhook` envelope. Both defer tests now call it.

### Deviation from the issue body (important)

The issue prescribed the helper for **exactly** the two defer tests. Doing only
that **does not reach the "0 clones" done-when**: the pre-existing
`"deliverWebhook returns 429 when rate-limited, never sends"` test shares the
*identical* 8-line envelope (`rawBody: "{}"`, `load`/`save`/`sendPrompt` with
`sent++`) with the busy tests, so after deduping the two busy tests a residual
clone of that envelope remains between the rate-limited test and the helper.

I verified this empirically (`scripts/check-duplication-gate.sh` reported 1
clone with the helper scoped to two tests). To satisfy the blocking gate's hard
"0 clones" acceptance I extended the **same single file-local helper** to also
serve the rate-limited test (the third case sharing the envelope). The helper
takes the per-case session deps and spreads them.

- Both busy tests keep their exact names and every original assertion
  (`202`/`queued===true`/`queued.sid`/`/Inbound webhook/`; `429`/`queued===undefined`/`ok===false`/`error==="queue full"`/`enqueueCalls===1`).
- The rate-limited test keeps its name and both assertions (`429`, `sent===0`).

### Done-when

- `node scripts/duplication-gate` → **PASS — no clones** for `webhooks.test.mjs`
- `npm run typecheck` → clean
- `npm test` → 1960 pass, 0 fail
- **Net line delta: −7** (24 insertions, 31 deletions) — negative, as required.

### DIFF
--- a/src/server/webhooks.test.mjs
+++ b/src/server/webhooks.test.mjs
@@ -487,18 +487,9 @@ test("isRedelivery / isEventFiltered pure semantics", () => {
 });
 
 test("deliverWebhook returns 429 when rate-limited, never sends", async () => {
-  let sent = 0;
-  const res = await deliverWebhook(
-    { token: "a".repeat(32), rawBody: "{}", signatureHeader: "" },
-    {
-      load: async () => [fakeHook({ unsigned: true })],
-      save: async () => {},
-      sendPrompt: async () => { sent++; },
-      take: () => false,
-    },
-  );
+  const { res, sent } = await deliverToBusySession({ take: () => false });
   assert.equal(res.status, 429);
-  assert.equal(sent, 0);
+  assert.equal(sent(), 0);
 });
 
 test("deliverWebhook happy path sends the formatted turn and stamps metadata", async () => {
@@ -538,9 +529,9 @@ test("deliverWebhook on an unsigned hook skips signature verification", async ()
   assert.equal(sent, 1);
 });
 
-// Both defer cases share everything but `enqueue` — the busy-session envelope
-// is the fixture, the enqueue outcome is the axis under test.
-async function deliverToBusySession(enqueue) {
+// The busy session is the fixture for every defer/rate-limit case — the
+// envelope is identical, only the session deps under test differ.
+async function deliverToBusySession(deps) {
   let sent = 0;
   const res = await deliverWebhook(
     { token: "a".repeat(32), rawBody: "{}", signatureHeader: "" },
@@ -548,8 +539,7 @@ async function deliverToBusySession(enqueue) {
       load: async () => [fakeHook({ unsigned: true })],
       save: async () => {},
       sendPrompt: async () => { sent++; },
-      isBusy: () => true,
-      enqueue,
+      ...deps,
     },
   );
   return { res, sent: () => sent };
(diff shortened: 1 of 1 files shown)


antoinedc:
## Review — BET-812 dedupe of `webhooks.test.mjs` (PR #860)

**Verdict: PASS** — 0 Blocks, 0 Questions, 1 nit (below). Routing forward to merge.

### Acceptance criteria (done-when) — verified, not asserted

| Criterion | Result |
|---|---|
| `duplication-gate` reports 0 clones for `webhooks.test.mjs` | ✅ PASS — reproduced locally (`scripts/check-duplication-gate.sh origin/main` → "PASS — no clones"); CI `duplication-gate` also pass |
| `npm run typecheck && npm test` green | ✅ typecheck exit 0; `npm test` 1960 pass / 0 fail / 0 skip |
| PR body states net line delta, negative | ✅ says −7 (24 insertions, 31 deletions) — confirmed from `git diff --stat origin/main...HEAD` |

### Hard-constraint compliance

- ✅ Only `src/server/webhooks.test.mjs` changed. No production file touched (diff is 1 file, −7 net).
- ✅ Both named defer tests keep their exact names and every original assertion (`202`/`queued===true`/`queued.sid`/`/Inbound webhook/`; `429`/`queued===undefined`/`ok===false`/`error==="queue full"`/`enqueueCalls===1`). `assert.equal(sent,…)` → `assert.equal(sent(),…)` is the mechanical consequence of the helper returning a closure — semantically identical.
- ✅ No new test cases added. Helper is file-local, not a shared test-utils module.

### ⚠️ Nit — deviation from one hard constraint ("used by exactly the two tests that need it")

The issue's done-when (0 clones) and its "exactly two tests" constraint are **internally inconsistent**, and I verified this rather than trusting the implementer's claim: I reconstructed the issue's verbatim two-test-only helper and re-ran the gate — it **FAILS with a residual 8-line self-clone** between the pre-existing rate-limited test's envelope and the helper (`deliverWebhook returns 429 when rate-limited` shares `rawBody:"{}"` + `load`/`save`/`sendPrompt` with the busy envelope). There is no way to reach the blocking "0 clones" gate without also deduping that third test.

The PR resolves the contradiction correctly and minimally: the same single file-local helper (still `deliverToBusySession`, in the same file, not exported) is generalized from an `enqueue` param to a deps-spread so the rate-limited test's deps (`take: () => false`) fit too. That test keeps its name and both assertions (`429`, `sent()===0`). This was disclosed upfront in the PR body with the empirical evidence, which I confirmed. Worth keeping in mind for the spec author — the "exactly two tests" line should be reconciled with the 0-clone target, but the code is correct as-is.

### Anti-spaghetti

Advisory report shows 0 clones (`Found 0 clones.`). Nothing to act on — the extraction is a legitimate test-fixture helper, and the residual 8-line clone between the busy envelope and the rate-limited test was the point of this PR.

**Route:** clean PASS → `manta-pm` for merge.

PR #878, opened 2026-08-14
fix(renderer): stop feeding empty transcript rows to Virtuoso (BET-874)

antoinedc:
## Summary (BET-874)

The transcript handed `react-virtuoso` the raw, unfiltered message list, so rows that render nothing (background-job completion turns, empty user messages, assistant turns whose parts are all `step-start`/`step-finish`/`todowrite`) became zero-height Virtuoso items — the `Zero-sized element` console errors plus size-cache poisoning (scroll jitter, wrong extents, bad `scrollToIndex`).

### The fix — one predicate, filtering the list
Consolidated the "does this row render?" rule into ONE pure function and filter the data, so no code path can hand Virtuoso an empty item. Net line *removal*.

- **`chatUtils.ts`**: added `visibleAssistantParts(msg)` (filter body moved verbatim out of MessageRow) and `isRenderableMessage(msg)` (the single source of truth: background-completion → false; user → text-or-file; else → `visibleAssistantParts.length > 0`). Both pure, reuse `concatUserMessageText`.
- **`MessageRow.tsx`**: one early return `if (!isRenderableMessage(msg)) return null;` at the top; deleted the two scattered `return null` guards; inline filter → `visibleAssistantParts(msg)`. MessageRow is also rendered outside Virtuoso (TaskCard child transcripts via `.map`), so it keeps a single guard through the shared predicate.
- **`Transcript.tsx`**: `data={visibleMessages}` where `visibleMessages = useMemo(messages.filter(isRenderableMessage), [messages])`; deleted the `isBackgroundJobCompletionTurn` return in `itemContent`; `hasMessages` = `visibleMessages.length > 0` (all-hidden transcript shows the Welcome state); `onLoadEarlier`'s `firstItemIndex` shift now counts VISIBLE rows.

Deliberately untouched per the issue: `showLoadEarlier` (raw `messages.length`), `updateEntryMotion` (raw input), `computeItemKey`/`followOutput`/`atBottomStateChange`/`alignToBottom`/`increaseViewportBy`/mount scroll. No new component, no wrapper, no `min-height`, no `display:none` row.

**Files changed:** 5. `src/renderer/chatUtils.ts`, `src/renderer/MessageRow.tsx`, `src/renderer/Transcript.tsx`, + tests in `chatUtils.test.ts` and `Transcript.test.tsx`.

**Base: origin/main @ `7a920d55`**

## Verification results
- PR branch (`multica/BET-874-virtuoso-zero-sized` @ `ae1788d2`):
  - typecheck: exit 0. Errors: none. log sha256: `19d583391c8b1ca203d401be3a20c862f1928db9d37421f1fb14aeb3f1076e1c`
  - test: exit 0, 1976 pass / 0 fail / 0 skip. Failures: none. log sha256: `fbcbc8fe1c43143cc336b8eed4e35e3a097a6a3dc45ff6c65ee761a622ea5528`
- Base (`main` @ `7a920d55`): not re-run (no new-failure conclusion — renderer-only change; see conclusion).
- **Conclusion:** 0 new failures. New unit coverage: `isRenderableMessage` (8 cases), `visibleAssistantParts` (3 cases), plus a `Transcript.test.tsx` case asserting a todowrite-only middle turn renders the surrounding rows and NO row for itself.

## Self-check (acceptance criteria)
- **CRITERION 1: `npm test` and `npm run typecheck` pass** → 10/10. Both green (above).
- **CRITERION 2: no `Zero-sized element` errors on a todowrite-heavy session** → 9/10. Root cause eliminated — empty rows are no longer fed to Virtuoso, so no zero-height item allocation is possible. Weakness: not verified against a live chat panel (Electron smoke is retired as a gate); the mechanism (list-level filter) makes the error unreachable by construction.
- **CRITERION 3: scrolling a transcript up/back keeps position (no anchor drift)** → 8/10. `firstItemIndex` shift now counts visible rows (was off by hidden prepended rows). Weakness: relies on the existing firstItemIndex mechanism correctness; no live scroll test.

## Follow-ups
None filed — no actionable gaps surfaced (the change is fully contained in the three files; no duplicate sites remain).

### DIFF
--- a/src/renderer/MessageRow.tsx
+++ b/src/renderer/MessageRow.tsx
@@ -22,9 +22,11 @@ import {
   formatDuration,
   formatTokens,
   formatHiddenTodosSummary,
+  isRenderableMessage,
   selectVisibleTodos,
   summarizeTodoProgress,
   todoStatusOf,
+  visibleAssistantParts,
   type TodoStatus,
   type TruncationKind,
 } from "./chatUtils";
@@ -297,6 +299,11 @@ export const MessageRow = memo(function MessageRow({
   // the MessageRow memo chain is untouched.
   entering?: boolean;
 }) {
+  // A row that renders nothing must never be drawn — MessageRow is rendered
+  // both inside Virtuoso (whose data is pre-filtered by Transcript, so this is
+  // a no-op there) and OUTSIDE it (TaskCard renders the child transcript with
+  // a plain .map), so the guard lives here on the shared predicate.
+  if (!isRenderableMessage(msg)) return null;
   const isUser = msg.info.role === "user";
 
   // Subtle wall-clock timestamp for each message/action. Sourced from the
@@ -353,7 +360,6 @@ export const MessageRow = memo(function MessageRow({
   if (isUser) {
     const text = concatUserMessageText(msg);
     const fileParts = msg.parts.filter((p) => p.type === "file");
-    if (!text && fileParts.length === 0) return null;
     return stampedRow(
       <div className="flex flex-col" style={{ gap: "var(--block-gap)" }}>
         {fileParts.length > 0 && (
@@ -408,16 +414,7 @@ export const MessageRow = memo(function MessageRow({
   // the latest checklist already renders once as the ActiveTodos card at the
   // tail of the transcript, so inlining each call too would repeat the same
   // list for every turn that touches todos.
-  const visibleParts = msg.parts.filter((p) => {
-    if (p.type === "text") return !p.synthetic && !p.ignored && (p.text ?? "").length > 0;
-    if (p.type === "step-start" || p.type === "step-finish") return false;
-    if (p.type === "tool") {
-      const tool = String((p as Record<string, unknown>).tool ?? "");
-      if (tool === "todowrite" || tool === "todo_write") return false;
-    }
-    return true;
-  });
-  if (visibleParts.length === 0) return null;
+  const visibleParts = visibleAssistantParts(msg);
 
   return stampedRow(
     <div className="flex flex-col" style={{ gap: "var(--block-gap)" }}>
--- a/src/renderer/MessageRow.tsx
+++ b/src/renderer/MessageRow.tsx
@@ -299,12 +299,20 @@ export const MessageRow = memo(function MessageRow({
   // the MessageRow memo chain is untouched.
   entering?: boolean;
 }) {
+  const isUser = msg.info.role === "user";
+
   // A row that renders nothing must never be drawn — MessageRow is rendered
   // both inside Virtuoso (whose data is pre-filtered by Transcript, so this is
   // a no-op there) and OUTSIDE it (TaskCard renders the child transcript with
   // a plain .map), so the guard lives here on the shared predicate.
+  //
+  // Placed AFTER the useState on purpose (reviewer finding): a subagent
+  // transcript streams via debounced refetches, so a child message can grow
+  // from non-renderable to renderable across renders. A guard before the hook
+  // would flip the hook count and trip React's "Rendered more hooks than
+  // during the previous render". This return only ever happens after the
+  // identical-before-the-conditional hook, keeping hook order stable.
   if (!isRenderableMessage(msg)) return null;
-  const isUser = msg.info.role === "user";
 
   // Subtle wall-clock timestamp for each message/action. Sourced from the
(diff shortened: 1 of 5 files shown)


antoinedc:
## Review — BET-874 (PR #878)

Review against the issue's acceptance criteria and spec. Mechanical gates verified directly.

**Verification (run by reviewer, not taken on faith):**
- `npm test`: 1976 pass / 0 fail.
- Base freshness: branched off `main @ 7a920d5`; diff vs fresh `origin/main` (`62c911f`) contains only the 5 intended files — no phantom deletions. Clean.
- Anti-spaghetti: 0 clones.

**Acceptance-criteria completeness** — all three `Done when` items and every spec bullet checked against the code (`chatUtils.ts` `visibleAssistantParts`/`isRenderableMessage`, `MessageRow` single guard + both scattered `return null` deletions, `Transcript` `visibleMessages` filter / `data=`, `itemContent` deletion + import removal, `hasMessages`, `onLoadEarlier` visible-row shift, deliberate untouches). Tests match the issue's required cases (`isRenderableMessage` x7 + `visibleAssistantParts`, plus the `Transcript.test.tsx` todowrite-middle-row case).

---

**Block (fix-here):** The single early return is placed before MessageRow's only hook.

`MessageRow.tsx:305` — `if (!isRenderableMessage(msg)) return null;` sits **before** the component's only hook, `useState` at `MessageRow.tsx:310`. In the Virtuoso path this is masked: `Transcript` pre-filters `visibleMessages`, so a row only mounts once it is renderable, and it never flips back (parts only append). But **TaskCard renders MessageRow for every child message via a plain `.map` regardless of renderability** (`TaskCard.tsx:146`), and subagent transcripts stream via debounced refetches (`ChatPanel` `scheduleChildRefetch` → `childMessages` update). A child message whose parts grow from non-renderable (empty / `step-start`-only / `todowrite`-only) to renderable (gains a real text part) — stable key `m.info.id` — reuses the same instance across refetches: first render skips `useState`, the next calls it → **"Rendered more hooks than during the previous render"**, which blanks the whole panel.

This is exactly the crash the codebase has already documented in `TaskCard.tsx`'s own header (`HOOK ORDER` comment: an earlier `info === null → info` hook-count flip "crashed with 'Rendered more hooks than during the previous render' and blanked the whole panel"). The new guard reintroduces that class of bug on the `.map` path.

**Fix (one line relocated):** move the `if (!isRenderableMessage(msg)) return null;` guard to **after** the `useState` on `MessageRow.tsx:310` — the same safe "hooks first, then early return" pattern the two deleted guards (old 359 / old 429) already used. Then the hook runs unconditionally regardless of renderability. The `data-message-id`/`ts` computations after are plain consts, not hooks, so nothing else moves.

antoinedc:
Fixed the Block. Relocated the `isRenderableMessage` guard to AFTER the `useState` (MessageRow.tsx) so hook order stays stable: on TaskCard's non-keyed `.map` streaming path (subagent transcripts via debounced refetches), a child message that grows from non-renderable → renderable can no longer flip the hook count across renders ("Rendered more hooks than during the previous render").

Verification (branch @ `6caba630`): typecheck exit 0; `npm test` 1976 pass / 0 fail.

antoinedc:
## Re-review after rebase — PASS (cycle 3)

Re-confirmed the rebased branch (`a4ebe5e`) head-to-head:

- **Base fresh:** merge-base = `origin/main` = `06672a90`. `mergeStateStatus: CLEAN`, `MERGEABLE`.
- **Diff vs fresh main:** only the 5 intended files — `chatUtils.ts`, `MessageRow.tsx`, `Transcript.tsx`, + the two test files. No phantom deletions, no scope changes.
- **Rebase conflict resolution coherent:** `chatUtils.ts` import line keeps main's set and adds the needed `OpencodePart`; tests auto-merged.
- **Rules-of-hooks concern (cycle-2 Block) now moot:** the rebase pulled in main's change that removed MessageRow's `voiceExpanded` `useState` (VoicePlayer now owns its expand state — main @ `06672a90` has no such hook in MessageRow). MessageRow therefore has zero hooks, so the `isRenderableMessage` guard placement is safe under any ordering.
- **Mechanical gates (run by me):** `npm run typecheck` exit 0; `npm test` 1980 pass / 0 fail.

Functional review unchanged from cycle-1/2 pass — one predicate, list-level filtering, correct `hasMessages`/`onLoadEarlier` handling, all required test cases present. Anti-spaghetti: 0 clones. No new findings.

PR #890, opened 2026-08-14
fix(server): emit context/cache frames from message.updated (BET-887)

antoinedc:
Opened automatically for `multica/BET-887-box-context-frame`, which was pushed without a pull request.

Relates to BET-887.

### DIFF
--- a/src/server/index.mjs
+++ b/src/server/index.mjs
@@ -141,9 +141,51 @@ const __dirname = dirname(fileURLToPath(import.meta.url));
 const PROJECT_ROOT = join(__dirname, "..", "..");
 
 const bus = createBus();
+// Model context windows for the interpreter's context reading. listModels() is
+// a provider round-trip, so it is fetched once and refreshed only when a model
+// we have never seen appears (a newly connected provider). A miss returns null
+// and the interpreter falls back to the assumed window rather than blocking.
+let modelContextLimits = null;
+let modelContextRefresh = null;
+function refreshModelContextLimits() {
+  if (modelContextRefresh) return modelContextRefresh;
+  modelContextRefresh = (async () => {
+    try {
+      const models = await oc.listModels();
+      const next = new Map();
+      for (const m of models ?? []) {
+        const ctx = m?.limit?.context;
+        if (typeof ctx === "number" && Number.isFinite(ctx) && ctx > 0) {
+          next.set(`${m.providerID}/${m.id}`, ctx);
+        }
+      }
+      modelContextLimits = next;
+    } catch {
+      modelContextLimits = modelContextLimits ?? new Map();
+    } finally {
+      modelContextRefresh = null;
+    }
+  })();
+  return modelContextRefresh;
+}
+function contextLimitFor(providerID, modelID) {
+  if (!providerID || !modelID) return null;
+  if (modelContextLimits === null) {
+    void refreshModelContextLimits();
+    return null;
+  }
+  const hit = modelContextLimits.get(`${providerID}/${modelID}`);
+  if (hit) return hit;
+  void refreshModelContextLimits();
+  return null;
+}
+
 // Box-side stream interpretation (BET-551): interprets the opencode stream on
 // the box and republishes derived events on THIS bus (single stream/endpoint).
-const streamInterp = createStreamInterpreter({ publish: (evt) => bus.publish(evt) });
+const streamInterp = createStreamInterpreter({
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review — BET-887 (context/cache frames from `message.updated`)

**Verdict: PASS** — 0 Blocks, 0 Questions, 1 Nit. Routing to manta-pm.

Reviewed against the issue's acceptance criteria as an independent checklist. Note: the full issue had no checklist section, so the acceptance bullets in the issue body were the gates I verified, each independently (not from the implementer's summary).

### Acceptance-criteria check (each verified against the code)

1. **`context` frame on every assistant `message.updated` with non-zero input total, deduped** — ✅ Implemented at the top of the `message.updated` case, reading `info ?? message`, gated on `role === "assistant"` + `tokens` + `totalInput > 0`, deduped on `st.contextEmitted {totalInput, limit}` (added to `newSessionState()`). Test 4 asserts exactly one frame across 3 feeds, a second after a changed `input`.
2. **`cache` frame from the same site** — ✅ Emitted inside the same non-deduped branch; Test 6 asserts it. Mirrors the step-ended block (uses `computeStaleCache`, `selectCacheTtlMs`, `TTL_DEFAULT`).
3. **No frame for user message / no tokens / all-zero set** — ✅ Test 5 covers all three (`role:"user"`, `tokens: undefined`, all-zero buckets); zero frames.
4. **One shared resolver for both emit sites; `resolveContextLimit(props.model)` gone from `streamInterp.mjs`** — ✅ `grep` confirms no `resolveContextLimit` remains in that file; both sites use `contextLimitFor(...) ?? ASSUMED_CONTEXT_TOKENS`. The engine-level `resolveContextLimit` in `src/shared/streamInterpretation.mjs` is untouched and still used by the renderer (correct — not deleted).
5. **`contextLimitFor` synchronous, never awaited in `interpret()`** — ✅ It returns `null` and kicks off a memoized `refreshModelContextLimits()` in the background on miss/unknown-model. Confirmed `oc.listModels()` exists in `src/server/opencode.mjs:1040` and `oc` is in scope in `index.mjs`.
6. **Live-box grep returns matches, quoted in PR body** — ✅ Implementer quoted 7 `context` frames + 7 `cache` frames. The currently-running server on this workspace box is still on the pre-fix deployed build, so I did not disrupt the live service to re-capture; I verified the emit path via the 6 targeted tests (below) plus code review, which reproduce the exact predicted `pct` values.
7. **No Swift file modified** — ✅ `git diff --stat` shows only `src/server/index.mjs`, `src/server/streamInterp.mjs`, `src/server/streamInterp.test.mjs`.

### Mechanical gates (ran on the PR branch)

- `npm run typecheck` — exit 0, no errors. ✅
- `npm run test:server` — `1494 pass / 0 fail`. ✅ (includes all 6 new BET-887 tests passing)
- `npm test` — vitest `2639 passed | 7 skipped`, node `1994 pass / 0 fail`, exit 0. ✅

These match the hashes the implementer reported (PR `ced458a1`). Arithmetic in the new tests is correct: `(2+182191+1516)/1_000_000 → 18`, `/200_000 → 92`.

### Base freshness

The branch was cut at `5128a664`; fresh `origin/main` is `c799415c`. The 3 intervening commits touch only renderer CSS files (BET-884) — zero overlap with the 3 changed server files, and `git diff --stat origin/main...HEAD` is identical (188+/4-), so no phantom deletions. Not stale enough to matter.

### Anti-spaghetti

The jscpd detector flags one clone: the `context`/`cache` emit tail shared by the two emit sites (message.updated vs step.ended). This is the intentional, spec-mandated mirror — the issue explicitly said to mirror the step-ended block exactly — so it is not flagged as a Block.

### Nit (non-blocking)

- **Cross-path dedupe is one-sided.** The issue's Task 3 asserts the Task-2 dedupe "makes the two paths safe together: whichever fires first wins and the other is a no-op." In the code, only the `message.updated` site consults `st.contextEmitted`; the `session.next.step.ended` site still emits `context`/`cache` unconditionally. This is zero impact today (that event never fires on the deployed build), and it's not an acceptance-criteria gate — the acceptance criteria scoped dedupe to the many-fires `message.updated` path, which is correctly deduped and tested. It would only mean a redundant frame per step if a *future* build emits `step.ended` while `message.updated` also fires — benign, not a defect. Flagging only so a future change to the step-ended path knows the guard exists if desired.

No action required to merge.

PR #897, opened 2026-08-14
fix(forge): resolve PR base branch from the forge instead of assuming main (BET-891)

antoinedc:
Opened automatically for `multica/BET-891-forge-default-branch`, which was pushed without a pull request.

Relates to BET-891.

### DIFF
--- a/src/server/forge/github.mjs
+++ b/src/server/forge/github.mjs
@@ -714,6 +714,22 @@ export function createGithubAdapter(request, requestWrite, requestText = request
     },
 
     /**
+     * GET /repos/{o}/{r} — the repo's default branch (the PR base). Returns
+     * `"main"` when the field is missing or the payload is malformed.
+     * @param {{ owner: string, repo: string }} repo
+     * @returns {Promise<string>}
+     */
+    async getDefaultBranch(repo) {
+      let data = {};
+      try {
+        ({ data } = await request(`${apiBase}/repos/${repo.owner}/${repo.repo}`));
+      } catch {
+        // fall through to "main"
+      }
+      return typeof data?.default_branch === "string" && data.default_branch ? data.default_branch : "main";
+    },
+
+    /**
      * GET /user/repos — the repos the connected user can actually PUSH to,
      * most-recently-pushed first. Filtering to write access is what keeps the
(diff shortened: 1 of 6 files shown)


antoinedc:
## Review — BET-891 (resolve PR base from the forge, not `main`)

**Verdict: PASS — 0 Blocks, 0 Questions, 0 nits.** Routing to `manta-pm`.

1. **No `"main"` literal in `src/server/forge/index.mjs`** — confirmed via grep; zero matches. `deps.defaultBase` fully removed (no references remain anywhere in `src/server/`).
2. **`getDefaultBranch(repo)` on both adapters, identical signatures** — `github.mjs` reads `GET /repos/{owner}/{repo}` via the read `request` helper, returns `default_branch`; `gitlab.mjs` reads `GET /projects/{id}` via its read helper. Both fall back to `"main"` when the field is absent or the payload is malformed (guarded with `typeof === "string" && truthy`).
3. **Resolve once in the context resolver** — `resolveWriteContext` resolves base through injectable `deps.getDefaultBranch` (default `defaultGetDefaultBranch` → adapter), returned as `ctx.base` alongside `head`. `shipPreview` reads `ctx.base` (used in both the returned `base` and the `origin/<base>...<head>` diff range); `shipPullRequest` uses `base ?? ctx.base` so an explicit override still wins. `normalizeRepo` untouched. No local git ref, no caching/poller/config/UI.
4. **`mergePullRequest` untouched** beyond the shared resolver (it passes `wantBranch=false` and never reads base — safe).
5. **Mechanical gate** — `npm run typecheck` → exit 0; `npm test` → **2001 pass, 0 fail**. Matches the implementer's reported result.

### Scope
Diff touches exactly the 6 in-scope files (`index/github/gitlab.mjs` + their tests). No renderer, no shared types, no `local.mjs`. Base-freshness check clean (no phantom deletions vs `origin/main`).

### Anti-spaghetti
Only test-fixture clones (the two `shipPullRequest` adapter literals) — discountable.

### Tests
Good coverage: adapter-level `getDefaultBranch` (maps field / falls back) on both GitHub & GitLab; `shipPreview` uses resolved `master` in base + diff range; `shipPullRequest` passes resolved base; explicit `base` override wins.

PR #900, opened 2026-08-14
feat(forge): generate PR title/body with the session's selected model, out of band (BET-893)

antoinedc:
## Summary

When the user ships a feature branch from a chat session on the box, the PR title/body is now written by the **session's selected model**, generated **out of band** — never appearing in the user's own conversation.

- The confirm card still shows the resolved title/body as plain text (BET-892 shape); the box now generates them with `deps.model` when supplied.
- Generation runs in a **throwaway opencode session** (`create → prompt → poll → delete`, `finally` on every path). No tmux window is stamped, so it never shows in the sidebar; the user's own transcript is never touched.
- The prompt carries the *why*: the last 20 messages of the calling session's transcript (text parts only, truncated to 8000 chars, oldest dropped first), plus commits, changed files, and the repo PR template (which is FILLED IN, not replaced).
- **All four opencode calls (`createSession`, `sendPrompt`, `listMessages`, `deleteSessionRaw`) are injectable deps**, alongside the existing `run` and `readPrTemplate`, so tests never touch a live opencode.

### Fallback — untouched, never an error

`draftTitle` / `draftBody` / `defaultReadPrTemplate` are **kept** as the deterministic fallback. Generation failure **never** surfaces as an error: `shipPreview` always returns a usable title/body. Falls back when any of:

- no model supplied on the input
- session create / prompt / message read throws
- no completed assistant text within the 60s deadline
- `parsePrDescription` returns `null`

### New pure module

`src/server/forge/shipDescription.mjs` — `buildPrDescriptionPrompt`, `parsePrDescription`, plus `extractTranscriptText` / `extractAssistantText` / `truncateTranscript`. All pure and unit-tested; `shipPreview` in `index.mjs` owns the I/O + fallback.

## Verification results

- `npm run typecheck` — **pass**
- `npm test` — **2029 pass, 0 fail** (includes 15 new `shipDescription` tests + 9 new `shipPreview` generation/fallback tests)
- `git diff --stat origin/main...HEAD` — only the intended files:
  - `src/server/forge/shipDescription.mjs` (new)
  - `src/server/forge/shipDescription.test.mjs` (new)
  - `src/server/forge/index.mjs` (+ out-of-band orchestration)
  - `src/server/forge/index.test.mjs`
  - `src/server/rpc.mjs` (`forge:ship-preview` wires injectable opencode deps)
  - `src/shared/types.ts` (`ForgeShipPreviewInput` + model/sessionId)
  - `src/renderer/ChatPanel.tsx` (passes model + sessionId)

## Test-important behaviors (as specified)

- A title over 72 chars is **returned as-is** by `parsePrDescription` — the model is trusted, never silently truncated (explicitly asserted).
- The throwaway session is deleted on **both** success and failure paths (asserted).
- With generation disabled or failing, behaviour is **byte-identical** to today's deterministic draft (asserted: no throwaway session when no model is supplied).

**Base: origin/main @ `7fb5b7b7`**

### DIFF
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -600,13 +600,18 @@ export function ChatPanel({
 
   // Fetch the ship preview once (the branch popover's no-PR state shows
   // Base / Changes). Click-only surface ⇒ fetched on popover open, never
-  // polled, and not re-fetched if already loaded for this cwd.
+  // polled, and not re-fetched if already loaded for this cwd. The session's
+  // selected model + sessionId ride along (BET-893) so the box can generate
+  // the title/body with that model, out of band.
   const ensureShipPreview = useCallback(async () => {
     if (!cwd) return;
     setShipError(null);
     if (shipProposal) return;
     try {
-      const prev = await window.api.forgeShipPreview({ cwd });
+      const prev = await window.api.forgeShipPreview({
+        cwd,
+        ...(modelOverride ? { model: { providerID: modelOverride.providerID, modelID: modelOverride.modelID }, sessionId } : {}),
+      });
       if (prev.ok) {
         setShipProposal({ head: prev.head, base: prev.base, fileCount: prev.fileCount, title: prev.title ?? "", body: prev.body ?? "" });
       } else {
@@ -617,7 +622,7 @@ export function ChatPanel({
       setShipProposal(null);
       setShipError(e instanceof Error ? e.message : "could not prepare the pull request");
     }
-  }, [cwd, shipProposal]);
+  }, [cwd, shipProposal, modelOverride, sessionId]);
 
   // Confirm → push + create. Only reached after the human confirms the
   // ShipConfirmCard. The title/body come from shipProposal (the box's preview,
--- a/src/server/forge/index.mjs
+++ b/src/server/forge/index.mjs
@@ -29,6 +29,8 @@
 
 import { rollupChecks, unsupportedByForge, repoKey as forgeRepoKey, repoKeyParts } from "../../shared/forge.mjs";
 import { run } from "../tmux.mjs";
+import { createSession as ocCreateSession, sendPrompt as ocSendPrompt, listMessages as ocListMessages, deleteSessionRaw as ocDeleteSessionRaw } from "../opencode.mjs";
+import { buildPrDescriptionPrompt, parsePrDescription, extractTranscriptText, extractAssistantText } from "./shipDescription.mjs";
 import { gitRemoteOrigin as localGitRemoteOrigin, detectForgeCli as localDetectForgeCli, gitPush as localGitPush, configGet as localConfigGet } from "../local.mjs";
 import { detectForgeWithHosts } from "./selfhost.mjs";
 import { resolveToken as authResolveToken } from "./auth.mjs";
@@ -842,6 +844,17 @@ export async function shipPreview(cwd, deps = {}) {
     // origin/<base> may not exist locally yet — best-effort empty.
   }
 
+  // BET-893: when the caller supplies the session's selected model, generate
+  // the title + body OUT OF BAND (a throwaway opencode session — never the
+  // user's own transcript) and use it. A generation failure NEVER surfaces as
+  // an error: we fall through to the deterministic draft below.
+  if (deps.model) {
+    const generated = await tryGeneratePrDescription(cwd, ctx, { base, files, deps });
+    if (generated) {
+      return { ok: true, head: ctx.head, base, fileCount: files.length, title: generated.title, body: generated.body };
+    }
+  }
+
   const title = await draftTitle(cwd, ctx.head, deps);
   const body = await draftBody(cwd, ctx.head, base, files, {
     ...deps,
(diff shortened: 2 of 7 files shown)


antoinedc:
## Review — BET-893 (PR #900)

**Base freshness:** merge-base == `origin/main` (`7fb5b7b7`); `git diff --stat origin/main...HEAD` shows only the 7 intended files, no phantom deletions. Clean.

**Mechanical gates (re-run by me on the PR branch):**
- `npm run typecheck` — **pass** (both tsconfigs)
- `npm test` — **2029 pass, 0 fail** (matches the implementer's reported results exactly)
- Anti-spaghetti report: 2 clones, both pre-existing test fixtures in `index.test.mjs` — none introduced by this PR.

**Scope check:** the diff is confined to the issue's named files (`shipDescription.mjs` new, `index.mjs`, `index.test.mjs`, `rpc.mjs`, `types.ts`, `ChatPanel.tsx`). No scope creep. `draftTitle` / `draftBody` / `defaultReadPrTemplate` are kept and still serve as the fallback — not dead code. Deps (`createSession` / `sendPrompt` / `listMessages` / `deleteSessionRaw`) are injected in the module's usual style; no new channel, no new httpApi method; no tmux window, no `@manta-session-id`, no caching, no direct provider call. `buildPrDescriptionPrompt` and `parsePrDescription` are pure and well unit-tested (last-20 messages, 8000-char cap oldest-first, template fill-in, title>72 returned as-is, garbage→null, fallback on throw/timeout/null, session deleted on both paths). All match the spec.

---

**Block (fix-here):** the poll loop returns the **first** non-empty assistant text instead of waiting for the assistant turn to complete — so on the live path it can capture a truncated, mid-stream reply as the final PR description.

In `tryGeneratePrDescription`, `sendPrompt` uses opencode's `prompt_async` (non-blocking; `src/server/opencode.mjs:624`), so generation continues after it resolves. The loop then does:

```js
while (Date.now() < deadline) {
  await sleep(pollMs);
  detail = extractAssistantText(await listMessages(sid));
  if (detail) return parsePrDescription(detail);   // ← returns as soon as ANY text is present
}
```

`extractAssistantText` concatenates whatever assistant text is present at that instant with **no completion check** (no filter on the message being terminal, no `info.time.completed`, no stability-across-polls test). With a real model call (~almost always > 1s), the first poll at ~1s observes the in-progress message with partial text and returns it immediately — a PR title/body cut off mid-generation. This directly contradicts the issue's requirement to "poll `listMessages` until the assistant turn completes or a 60 second deadline passes" (fallback only on "no ***completed*** assistant text within the deadline"). The current unit tests don't exercise this: the deadline test feeds an *empty* assistant string, and the success test feeds *fully-formed* text on the first poll — neither simulates a partial-then-grows stream.

The fix fits inside this PR: wait for a completion signal before parsing (e.g. the assistant message is terminal — `info.time.completed` set, or opencode's message `state`), or require the extracted text to be stable across two consecutive polls, still bounded by the existing 60s deadline → fall back via the current path. Add a test with a partial-first-then-complete transcript to lock it in.

This is the only finding. Everything else checks out and the fallback/byte-identical guarantees are solidly implemented and asserted.

antoinedc:
Addressed the Block — the poll loop now waits for the assistant turn to complete before parsing.

**Root cause:** `prompt_async` is non-blocking; the first poll at ~1s observed the in-progress assistant message and returned partial text as the final PR description.

**Fix** (commit `0ceb1743`, same PR):
- The poll now uses a completion-aware extractor: `extractCompletedAssistantText` returns `null` while the turn is still in flight, and only hands back the assistant text once the message carries opencode's authoritative completion stamp (`info.time.completed` — the same signal `isAssistantTurnComplete` in `streamInterpretation.mjs` already relies on, which "cannot false-positive mid-turn").
- The generator intentionally does **not** reuse the shared helper's empty-transcript-as-complete behaviour (`isAssistantTurnComplete([]) === true`): there is a window right after create+prompt where zero messages exist server-side, and treating that as "done" would fall back prematurely. An explicit non-empty guard avoids it.
- Still bounded by the existing 60s deadline → falls back on the deterministic draft, never surfacing an error.

**New tests (2033 pass, 0 fail; typecheck green):**
- `shipDescription`: `isPrGenerationComplete` + `extractCompletedAssistantText` (in-flight/partial → `null`; complete → text; complete-but-empty → `""`).
- `shipPreview`: a **partial-then-grows stream** (poll 1 = partial "Model written t…", poll 2 = still streaming, poll 3 = complete) — asserts the full completed title/body is returned, not the truncated partial. Plus a completed-but-no-text fallback test.

antoinedc:
## Re-review — BET-893 (PR #900, commit `0ceb1743`)

**Block resolved ✓.** The poll loop now uses `extractCompletedAssistantText`, which returns `null` while the assistant turn is still in flight (via the existing `info.time.completed` completion stamp in `isAssistantTurnComplete`) and only hands back text once the message is terminal. The explicit non-empty guard (`isPrGenerationComplete`) correctly prevents the create+prompt empty window from being misread as "done," and a completed-but-empty turn returns `null` → deterministic fallback. The new partial-then-grows stream test asserts the **full** completed title/body is returned, not the truncated partial — the exact gap my prior Block called out.

**Mechanical gates (re-run by me on `0ceb1743`):**
- `npm run typecheck` — **pass** (both tsconfigs)
- `npm test` — **2033 pass, 0 fail** (matches the implementer's report)
- Base freshness: merge-base == `origin/main`; diff still confined to the 7 intended files, no phantom deletions.
- Anti-spaghetti: same 2 pre-existing test-fixture clones in `index.test.mjs`; none introduced by this PR.

**Verdict: PASS — 0 Blocks, 0 Questions.** Fallback remains byte-identical (asserted), throwaway session deleted on both success and failure paths (asserted), deps injected, no tmux window / no caching / no user-session traffic, and `draftTitle`/`draftBody`/`defaultReadPrTemplate` stay as the fallback.

PR #909, opened 2026-08-14
fix(server): replay running state to /events clients on connect (BET-913)

antoinedc:
## Why

BET-913: the turn timer is lost after a force-quit + relaunch on iOS. Root cause is server-side: the stream interpreter emits the `running` frame only on the idle→busy edge (`setRunning`), and the `/events` bus is pure pub/sub with no snapshot/replay. A client that (re)connects mid-turn — exactly what a force-quit does — never receives a `running { running: true, since }` frame for the still-in-progress turn, so iOS can never recover `runningSince` and the timer silently drops.

## Fix

Mechanism chosen: **snapshot/replay on subscribe** (one of the two options the issue suggested).

- `src/server/events.mjs` — `createBus()` gains `setSnapshot(fn)`. On `subscribe`, the bus replays the snapshot provider's events to the newly (re)connected consumer before registering it. The bus has exactly two subscribers (the `/events` SSE + WS attach fns), so this replay is scoped precisely to clients — no internal listener is affected.
- `src/server/streamInterp.mjs` — new `snapshotBusy()` returns a `stream.running` frame (with the **original** `since`, so the timer counts from the real turn start, not the reconnect) for every session currently `running`. Also preserves the last status `type` (`runningType`) so the replayed frame matches the live edge frame additively.
- `src/server/index.mjs` — wires `bus.setSnapshot(() => streamInterp.snapshotBusy())`.

No iOS change — out of scope by the issue. The replayed frame matches the existing `stream.running` contract (`{ running, since }`, `type` additive) the native client already consumes (`MantaEventStreamTests.swift` pins `running` + `since` -> `runningSince`).

## Verification results

**Files changed:** `5` files. `src/server/events.mjs`, `src/server/streamInterp.mjs`, `src/server/index.mjs`, `src/server/events.test.mjs`, `src/server/streamInterp.test.mjs` (+2 tests each).

- PR branch (`multica/BET-913-stream-running-replay` @ `3690a8c6`):
  - typecheck: exit 0. Errors: none. log sha256: `4a728c595e55b9a3615554f64c306078e7118536ceeeae4a7e7f46d0346eaacf`
  - test: exit 0, 2042 pass / 0 fail / 0 skip. Failures: none. log sha256: `e83b30f3a8a5129ee03a1190a47e08686fcd30e9ba4515f287e97a596002fd15`
- Base (`main` @ `720d9d7a`):
  - typecheck: exit 0. Errors: none. log sha256: `4a728c595e55b9a3615554f64c306078e7118536ceeeae4a7e7f46d0346eaacf`
  - test: exit 0, 2038 pass / 0 fail / 0 skip. Failures: none. log sha256: `79e3caf53ee46dbb067b96f00a0534b3b9bbbd5d4b2dbf9ae09ca9626782775e`
- **Conclusion:** 0 failures new; 4 new tests are the ones added by this PR. Typecheck log byte-identical to main.

**Base: origin/main @ `720d9d7a`**

**On-device re-verification owed:** the acceptance criterion (session-list row + chat running indicator show ~`1m` after force-quit + reopen) is an on-device check only `macos` can run against a real box. Flag for `macos` re-verify once merged.

### DIFF
--- a/src/server/events.mjs
+++ b/src/server/events.mjs
@@ -3,9 +3,24 @@
 
 export function createBus() {
   const subs = new Set();
+  // Optional state replay. `setSnapshot(fn)` registers a provider that returns
+  // the current-state events to emit to each NEW subscriber on connect. The
+  // bus itself is a pure forwarder (no retained history); this is the one
+  // deliberate exception so a (re)connecting client recovers edge-only state.
+  let snapshot = () => [];
   return {
-    subscribe(fn) { subs.add(fn); return () => subs.delete(fn); },
+    subscribe(fn) {
+      subs.add(fn);
+      // Replay current state to the newly (re)connected consumer. Edge-only
+      // frames like `stream.running` never re-fire on their own, so a client
+      // that connects mid-turn would never learn a session is already busy —
+      // this is exactly the force-quit + relaunch case, where the turn timer
+      // must survive a fresh process.
+      for (const evt of snapshot()) { try { fn(evt); } catch {} }
+      return () => subs.delete(fn);
+    },
     publish(evt) { for (const fn of subs) { try { fn(evt); } catch {} } },
+    setSnapshot(fn) { snapshot = fn; },
   };
 }
 
--- a/src/server/events.test.mjs
+++ b/src/server/events.test.mjs
@@ -15,6 +15,40 @@ test("bus delivers published events to subscribers and stops after unsubscribe",
   assert.equal(got.length, 1);
 });
 
+test("bus replays the registered snapshot to a new subscriber (state recovery on connect)", () => {
+  const bus = createBus();
+  bus.setSnapshot(() => [
+    { kind: "stream", sub: "running", sessionId: "ses_1", payload: { running: true, since: 1234 } },
+  ]);
+  const got = [];
+  const off = bus.subscribe((e) => got.push(e));
+  assert.equal(got.length, 1, "snapshot events reach the new subscriber immediately");
+  assert.equal(got[0].sub, "running");
+  assert.equal(got[0].payload.since, 1234);
+  // A second subscriber gets its own replay.
+  const got2 = [];
+  bus.subscribe((e) => got2.push(e));
+  assert.equal(got2.length, 1);
+  // Live events still flow after the snapshot replay.
+  bus.publish({ kind: "stream", sub: "flush", sessionId: "ses_1", payload: {} });
+  assert.equal(got.filter((e) => e.sub === "flush").length, 1);
+  off();
+});
+
+test("bus replays only the snapshot registered at subscribe time", () => {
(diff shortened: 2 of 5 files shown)


antoinedc:
## Review — BET-913 (force-quit lose) — PASS (0 Blocks, 0 Questions)

**Scope & base freshness.** Branch point `720d9d7a` is one commit behind `origin/main` tip; `git diff origin/main...HEAD` is `+112/−1` across 5 files, all in `src/server/` (events.mjs, streamInterp.mjs, index.mjs + their tests). No phantom deletions, no unrelated files, no scope creep. iOS client untouched — correctly out of scope per the issue.

**Mechanical gate.**
- `npm run typecheck` — clean (0 errors).
- `npm test` — **2042 pass / 0 fail** (matches the implementer's reported numbers exactly).

**Mechanism (reviewed against the issue's suggested direction).**
- `createBus()` gains `setSnapshot(fn)`; on each new `subscribe` it replays the provider's events to that consumer. The bus has exactly two subscribers — the `/events` SSE and WS attach (`events.mjs:45`, `events.mjs:60`) — both of which wrap the `{kind,payload}` envelope, so the replay reaches only clients and in the native shape.
- `streamInterp.snapshotBusy()` emits a `stream.running` frame for every session with `running && runningSince != null`, carrying the **original idle→busy edge `since`** (not reconnect time). Verified: `runningSince` is stamped once at the edge in `setRunning`, so the timer counts from real turn start.
- Replayed frame shape is byte-identical to the live `emit(sid,"running",…)` (`{kind:"stream", sub:"running", sessionId, payload:{running,type,since}}` — `emit` at streamInterp.mjs:262), so the existing iOS consumer contract is preserved.
- `runningType` bookkeeping is sound: `running` can only become true through the `session.status` branch (the `turnComplete` call only ever *clears* `running`), and that branch sets `runningType` — so the snapshot can't serve a stale type.

**Anti-spaghetti.** Read the jscpd report: the two flagged clones are pre-existing test-fixture patterns (the `make(42_000)` / `session.status busy` setup), not new business logic. Discounted. No dead code left behind.

**Correctness of the failure mode.** Force-quit + relaunch = fresh process = fresh `/events` socket = fresh `subscribe` → replay delivers the still-in-progress turn's `running{since}`. This directly addresses the root cause described in the issue (edge-only `running` frames + no snapshot on subscribe).

**Note on the on-device acceptance.** The issue's acceptance ("session row shows ~1m after a real force-quit") is an on-device check that only the `macos` agent can run against a real box — I cannot execute it from here, and the PR correctly makes no iOS change. Per the routing note, an on-device re-verification (BET-901 check 1) is owed to `macos` once merged; that is a verification hand-off, not a defect in this server-side fix. I'm routing this to `manta-pm` clean.

PR #911, opened 2026-08-14
BET-914: fix(forge): an unknown default branch is unknown, not "main"

antoinedc:
## Summary

BET-914: an unknown default branch is now unknown, not `"main"` — and nothing ships against an unknown base.

- **`getDefaultBranch`** (both `github.mjs` and `gitlab.mjs`) returns `string | null`: the repo's `default_branch` when present, `null` when the field is missing, the payload is malformed, or the request fails. The `// fall through to "main"` comment in each `catch` is gone.
- **`shipPreview`** refuses an unknown base (`{ ok: false, error: "unknown_base" }`) before any git call, so the "0 files" lie from an unresolved `origin/main...<head>` diff is impossible.
- **`shipPullRequest`** refuses an unknown base before the push, so a PR that cannot exist never leaves a pushed branch behind. Its now-unused `base` input override is removed — the forge-resolved base is the only resolution path, passed as `base: ctx.base` to `createPullRequest`.
- **`ForgeShipInput`** drops its `base?: string` field.

`normalizeRepo`'s `defaultBranch: r?.default_branch ?? "main"` (github.mjs) is intentionally untouched — different consumer, retains its fallback.

## Verification results

**Files changed:** `7` files: `src/server/forge/{github,gitlab,index}.mjs`, `src/shared/types.ts`, and the three test files.

- PR branch (`multica/BET-914-unknown-default-branch` @ `456b21a4`):
  - typecheck: exit `0`. Errors: none. log sha256: `4a728c595e55b9a3615554f64c306078e7118536ceeeae4a7e7f46d0346eaacf`
  - test: exit `0`, `2041` pass / `0` fail / `0` skip. Failures: none. log sha256: `ac97d570766db96bbcaf0c55f42aefa827164d3931b0e1581693b0e2370b9b4f`
- Base (`origin/main` @ `a7ba85a5`): verified via `git reset --hard origin/main` before branching; base SHA recorded below.
- **Conclusion:** `0` new failures. All tests green on the PR branch.

Base: `origin/main @ a7ba85a5`

### DIFF
--- a/src/server/forge/github.mjs
+++ b/src/server/forge/github.mjs
@@ -715,18 +715,19 @@ export function createGithubAdapter(request, requestWrite, requestText = request
 
     /**
      * GET /repos/{o}/{r} — the repo's default branch (the PR base). Returns
-     * `"main"` when the field is missing or the payload is malformed.
+     * `null` (unknown) when the field is missing, the payload is malformed, or
+     * the request fails.
      * @param {{ owner: string, repo: string }} repo
-     * @returns {Promise<string>}
+     * @returns {Promise<string|null>}
      */
     async getDefaultBranch(repo) {
       let data = {};
       try {
         ({ data } = await request(`${apiBase}/repos/${repo.owner}/${repo.repo}`));
       } catch {
-        // fall through to "main"
+        // leave data empty — an unknown default branch is unknown
       }
-      return typeof data?.default_branch === "string" && data.default_branch ? data.default_branch : "main";
+      return typeof data?.default_branch === "string" && data.default_branch ? data.default_branch : null;
     },
 
     /**
--- a/src/server/forge/github.test.mjs
+++ b/src/server/forge/github.test.mjs
@@ -494,10 +494,18 @@ test("getDefaultBranch maps the repo's default_branch", async () => {
   assert.equal(await adapter.getDefaultBranch({ owner: "acme", repo: "widget" }), "develop");
 });
 
-test("getDefaultBranch falls back to main when the field is absent", async () => {
+test("getDefaultBranch returns null when the field is absent", async () => {
   const url = "https://api.github.com/repos/acme/widget";
   const adapter = createGithubAdapter(fakeRequest({ [url]: { full_name: "acme/widget" } }));
-  assert.equal(await adapter.getDefaultBranch({ owner: "acme", repo: "widget" }), "main");
+  assert.equal(await adapter.getDefaultBranch({ owner: "acme", repo: "widget" }), null);
+});
+
+test("getDefaultBranch returns null when the request fails", async () => {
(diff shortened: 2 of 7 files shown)


antoinedc:
## Review — BET-914: unknown default branch is unknown, not "main"

**Verdict: PASS** — 0 Blocks, 0 Questions, 0 Nits.

### Acceptance-criteria checklist (each verified independently against the code)

1. ✅ **`getDefaultBranch` on both adapters returns `string | null`**, no `"main"` literal.
   - `github.mjs:722-730` and `gitlab.mjs:520-528` return `data.default_branch` when a non-empty string, else `null`. The `// fall through to "main"` catch comment is deleted in both.
2. ✅ **Only `"main"` literal in `github.mjs` is `normalizeRepo`'s `defaultBranch`** (`github.mjs:759`, untouched). `grep` confirms `gitlab.mjs` and `index.mjs` have no `"main"` literal.
3. ✅ **`ForgeShipInput` has no `base` field** (`types.ts:594-599`). `ForgeShipPreviewResult.base: string` unchanged; `ForgePullRequestResult.base: string | null` unchanged (not widened).
4. ✅ **Neither `shipPreview` nor `shipPullRequest` performs any git/forge call on an unknown base.**
   - `shipPreview` (`index.mjs`): `if (!ctx.base) return { ok: false, error: "unknown_base" };` sits immediately after the error guard, above the `git diff`. Test asserts the injected `run` was never called.
   - `shipPullRequest` (`index.mjs`): same guard placed **before the push**; test asserts injected `gitPush` recorded nothing.
   - `base` input override removed; `createPullRequest` receives `base: ctx.base` (no `??`).
5. ✅ **`npm run typecheck && npm test` green.**
   - `npm test` → 2041 pass / 0 fail / 0 skip (matches implementer's reported `2041`). Verified from a fresh clone of this PR branch.

### Scope

Diff restricted to the 7 intended files (issue's stated scope). Base is fresh: branch point = `origin/main @ a7ba85a5`. No phantom deletions.

### Tests

All required test changes present and correct: null-on-absent and null-on-request-failure for both adapters; `shipPreview refuses an unknown base` (run never called); `shipPullRequest refuses an unknown base and never pushes` (gitPush records nothing); removed the pinned "explicit base still wins" test and the `base: "main"` input. New cases reuse `SHIP_DEPS` / `createSpyAdapter` / `linkedShipDeps` per the issue's note.

### Anti-spaghetti

Detector reports only pre-existing code (`gitlab.mjs normalizeCommitStatus`) and the intentional desktop/mobile transport mirrors — nothing introduced by this PR's diff. No action.

No `what NOT to do` items were violated (no local git-ref fallback, no cache/retry/config field, `pullRequestForCwd` and `mergePullRequest` untouched, nothing in `src/renderer/`).

PR #920, opened 2026-08-14
Homepage: enrich the two thin feature tiles (Bring Your Own Plan, Mac Plugins) (BET-921)

antoinedc:
Follow-up from BET-906 (#919). Enriches the two thinnest feature tiles — **Bring Your Own Plan** and **Mac Plugins** — in the same markup-drawn style (`.bt`/`.btc`/`.glyph`/`.gchip`) as the other ten. Tile titles + benefit copy kept verbatim from `docs/specs/homepage-redesign/reference.html` section 8; model names (`claude`/`codex`/`kimi`) unchanged.

**Changes** (`website/index.html` only, +10/−3):
- **Bring Your Own Plan** — now three full-width plan rows, each model with a `✓ your plan` badge (was three bare chips in one row).
- **Mac Plugins** — now a plugin-capability list: `ios-mantaui` header chip + `xcodebuild ✓ done` + `simctl boot ↗ on phone` rows (was two bare chips).

Out of scope untouched: the other ten tiles, no per-tile links/borders, no comparison table, no thirteenth tile.

**Verification (headless render @ 1280/1024/390px):** 12 equal tiles, 4/2/1 per row, no horizontal scrollbar, no console errors, no 404s — all PASS.

**Verification results:**
- PR branch `multica/BET-921-enrich-thin-tiles` @ `d4a4636`:
  - typecheck: exit 0, no errors
  - test: exit 0, 2049 pass / 0 fail / 0 skip
- Base: `origin/main` @ `bb8b99d6`

**Base: origin/main @ bb8b99d6**

PR #922, opened 2026-08-14
fix(ios): reconcile running state on reconnect via authoritative runningSet (BET-922)

antoinedc:
Opened automatically for `multica/BET-922-running-reconcile`, which was pushed without a pull request.

Relates to BET-922.

### DIFF
--- a/mobile/native/MantaUI/ChatSessionStore.swift
+++ b/mobile/native/MantaUI/ChatSessionStore.swift
@@ -242,6 +242,11 @@ final class ChatSessionStore: ObservableObject {
     /// `false`) must not clobber the optimistic `true` — but once the box
     /// reports running at all, the snapshot is authoritative.
     private var optimisticRunning = false
+    /// The last `runningSetSeq` this store has folded into its running state. A
+    /// change in it means the box has (re)stated the authoritative running set
+    /// since `send()` stamped `optimisticRunning`, so the optimistic flag must
+    /// yield (BET-922).
+    private var lastRunningSetSeq = 0
     private var didRunOnce = false
     /// Test seams (internal, readable via `@testable`): count one-time vs
     /// resumable work so tests can assert the split without touching live
@@ -427,6 +432,14 @@ final class ChatSessionStore: ObservableObject {
         // stale previous-turn `false` must not clobber while the box hasn't yet
         // confirmed. Reading the snapshot (not a per-frame stamp) also makes
         // this robust to frames that coalesce between publishes.
+        // The box's authoritative running set supersedes the optimistic stamp
+        // `send()` set — without this, a send whose acknowledging frame was
+        // missed leaves the working row lit forever (BET-922).
+        let seq = eventStore.runningSetSeq
+        if seq != lastRunningSetSeq {
+            lastRunningSetSeq = seq
+            optimisticRunning = false
+        }
         if optimisticRunning, (carried.running || s.running == true) {
             optimisticRunning = false
         }
--- a/mobile/native/MantaUI/MantaEventStore.swift
+++ b/mobile/native/MantaUI/MantaEventStore.swift
@@ -251,6 +251,36 @@ enum MantaStreamRouter {
         return s
     }
 
+    /// Replace running state across ALL known sessions from the box's
+    /// authoritative set. A session absent from the set is not running.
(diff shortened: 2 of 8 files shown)


antoinedc:
## Review — BET-922 (running state reconciled on reconnect)

**Verdict: PASS** — 0 Blocks, 0 Questions, 0 nits.

Reviewed against the issue's acceptance criteria, verified independently (did not take the implementer's summary on faith).

1. **`npm run typecheck && npm test` green** — ✅ Re-ran on this box:
   - `npm test` → **2050/2050 pass, 0 failures** (server node:test + vitest). This confirms the implementer's two platform-specific Mac failures (Linux-targeted cwd/CI-PATH assertions) are indeed absent on Linux.
2. **Xcode builds + Swift unit tests pass** — ✅ CI `ios-build` gate is **green**; the new Swift suite (runningSet + optimistic-supersede cases) compiles and passes there. (Xcode can't run on this Linux box, so I rely on the green CI gate plus the readable test coverage below.)
3. **`grep -rn "snapshotBusy" src/` returns nothing** — ✅ verified, zero matches. No old per-session `stream/running` snapshot frames remain.
4. **Box publishes exactly one connect-time running frame** — ✅ by code review: `snapshotState()` always returns `kind:"runningSet"` at index 0, present even when empty, with `sessionId`/`since` (epoch ms)/`type`. Note this implementation deliberately folded it into the existing `snapshotState()` (which preserves BET-916's pending `questions`/`permissions` replay) rather than a separate `snapshotRunning()`; the issue's "What NOT to do" (don't keep the old per-session snapshot alongside) is honored — the running picture is exactly one frame kind. The live-box `curl` and the device manual checks are inherently post-merge QA and noted as such in the issue.
5. **Manual device: timer/working indicator clears on foreground after backgrounded turn-end** — logic implemented and unit-tested (the latched-running-cleared regression test); physical-device run is a manual acceptance item.
6. **Manual: timer survives force-quit mid-turn** — `since` preserved as the original idle→busy stamp (not reconnect time); covered by tests.

### Diff review

- **Server** (`streamInterp.mjs`, `index.mjs`): `snapshotState()` emits the authoritative `runningSet` frame at index 0 always; the `running` sub frame per busy session is gone; `runningType`/`since` preserved. Wiring + comment updated. `events.mjs` untouched. ✅
- **iOS** (`MantaStreamModels.swift`): `StreamRunningSetPayload` matches spec exactly.
- **iOS** (`MantaEventStore.swift`): `MantaStreamRouter.applyingRunningSet` reuses the single `resolveRunningSince` (ms→s conversion has one implementation), touches only `running`/`runningSince`; `handleFrame` routes `runningSet` **before** the `stream`/raw split and **not** behind the `degraded` guard (correct — the unreachable client is exactly the one needing correction); `runningSetSeq` bumped per apply and not `@Published` (sessionStates reassignment re-fires observers). ✅
- **iOS** (`ChatSessionStore.swift`): running block supersedes the optimistic `send()` stamp on seq change, then `running = optimisticRunning ? true : (s.running == true)`. `send()` optimistic set and failed-send rollback untouched. ✅

### Tests

Server test file rewritten per spec: exactly-one-frame-even-empty, busy session with original `since` + `type`, idle drops out of set, "second busy same `since`" preserved, questions/permissions still replay with the runningSet as one frame. Swift tests cover every required case including the named regression (`testRunningSetClearsLatchedRunningAbsentFromTheSet`), degraded-mode bypass, `runningSetSeq` bump, untouchability of `turnComplete`/`chunks`/`tools`, and `testRunningSetSupersedesOptimisticRunning`.

### Scope

Changed files exactly match the issue's scope list; no `events.mjs`, no `src/renderer/`, no `SessionListStore`/`SessionListView`. No scope creep. Anti-spaghetti report: 4 clones, all coincidental test-fixture setup in `streamInterp.test.mjs` — no business-logic duplication introduced.

### Base freshness

Branch is behind `origin/main` by 3 upstream commits (#920/#921); the three-dot diff shows **no file deletions**, so no stale-base phantom-deletion issue. `mergeable: MERGEABLE`.

**Routing:** PASS → `manta-pm` for merge. Device/box manual verification (acceptance items 4–6) remains the PM's pre/post-merge QA step, not a code gap.

PR #926, opened 2026-08-14
BET-927: Homepage pass 2 (hero note, self-advancing rail, rewritten problem band)

antoinedc:
Copy + CSS only, in one file: `website/index.html`.

- Hero note trimmed to `Free and open source.` (platform/size/version now live on the per-platform download cards; `1.0.14` was the iOS marketing version, desktop is `0.0.28`).
- The five-step rail now self-advances 1.0→5.0 and loops with pure CSS (no JS): a `::after` cyan progress bar grows left-to-right over each 3.5s dwell in a 17.5s cycle, with `animation-delay` offsets per step. Pairs with a `prefers-reduced-motion` fallback. Removed the two dead `.rl.on` rules and the `class="rl on"` attribute.
- "The problem" band rewritten: the four complaints that hurt daily replace the old single "build fails at 03:14" story (which duplicated the Automate stage). `.arrow` divs and the redundant grid wrapper are gone.
- Narrow-width stacking for the problem band added to the existing BET-904 `@media(max-width:900px)` block so it is single-column at 390px with no horizontal scrollbar.

Base: origin/main @ ef8187ee

**Verification results:**
- PR branch (`multica/BET-927-hero-note-rail-problem` @ `HEAD`):
  - test: exit 0, 2050 pass / 0 fail / 0 skip.

### DIFF
--- a/website/index.html
+++ b/website/index.html
@@ -233,12 +233,36 @@ h1 .grad{background:var(--gradient-brand);-webkit-background-clip:text;backgroun
 .rail{display:grid;grid-template-columns:repeat(5,1fr);gap:16px;margin-top:44px;padding-top:28px;
   border-top:1px solid var(--border-subtle)}
 @media(max-width:900px){.rail{grid-template-columns:1fr 1fr}}
-.rl{border-top:2px solid var(--border-default);padding-top:12px}
-.rl.on{border-top-color:var(--cyan-400)}
+.rl{position:relative;border-top:2px solid var(--border-default);padding-top:12px}
 .rl .rn{font-family:var(--font-mono);font-size:11px;color:var(--slate-500);font-weight:700}
-.rl.on .rn{color:var(--cyan-400)}
 .rl .rt{font-size:14.5px;color:var(--text-primary);font-weight:700;margin-top:2px}
 .rl .rd{font-size:11.5px;color:var(--slate-400);margin-top:4px;line-height:1.5}
+/* The rail walks 1.0 -> 5.0 and loops, with no JavaScript: five steps at 3.5s
+   each is a 17.5s cycle, so one step owns 20% of the timeline and the other
+   four are offset by animation-delay. The cyan bar grows left-to-right over
+   the step's dwell, so the rail reads as progress through a sequence rather
+   than as a slideshow. */
+.rl::after{content:"";position:absolute;top:-2px;left:0;height:2px;width:100%;
+  background:var(--cyan-400);transform:scaleX(0);transform-origin:left center;
+  animation:railFill 17.5s linear infinite}
+.rl .rn{animation:railNum 17.5s steps(1,end) infinite}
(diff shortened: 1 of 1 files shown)


antoinedc:
## Review — BET-927 (homepage pass 2: hero note, self-advancing rail, rewritten problem band)

**Verdict: PASS** — 0 Blocks, 0 Questions, 0 nits.

### Scope / base check
- Single file changed: `website/index.html` (45+/22-). Matches the issue's stated scope exactly.
- Base freshness: `merge-base HEAD origin/main` = `ef8187ee` = `origin/main`. Branch is current — no stale-base phantom deletions.

1. ✅ Hero note reads exactly `Free and open source.` (nothing else).
2. ✅ `grep -c 'class="rl on"'` = 0; `grep -c '\.rl\.on'` = 0. Both the attribute and the two `.on` rules are gone; `position:relative` merged into the single `.rl` rule (no second rule added).
3. ✅ Rail logic: first step has no `animation-delay`, so at load only `1.0` is highlighted (cyan via `railNum` 0–20%); steps 2–5 are `animation-delay`-offset and sit at their base slate state until their turn. The `::after` cyan bar fills 1.0 first (0–17% of the 17.5s cycle), then walks 2.0→3.0→4.0→5.0 and loops. `@keyframes railFill`/`railNum` and the per-step delays match the spec.
4. ✅ `grep -c '<script'` = 2 on PR and 2 on `main` (unchanged; only the two `application/ld+json` blocks). No `<script>` added.
5. ✅ `prefers-reduced-motion:reduce` block sets `animation:none`, forces `.rl:first-child::after{transform:scaleX(1)}` and `.rl:first-child .rn{color:var(--cyan-400)}` — static, `1.0` highlighted.
6. ✅ Problem band (scoped to that section): 4 `class="node"`, 0 `class="arrow"`; `03:14`, `draft pull request`, `started by an event` all absent; redundant grid wrapper and `.arrow` divs removed as specified.
7. ✅ Media query fixes added to the existing `@media(max-width:900px)` block (the one holding `.hero-grid`), using the file's `[style*="display:grid"]` attribute-selector convention: single column `minmax(0,1fr) !important` + `min-width:0` — no horizontal scroll.
8. ✅ Default (≥900px) stays two columns for the band (inline `1fr 1fr`) and five columns for the rail (`.rail` grid `repeat(5,1fr)`).
9. ✅ No new CSS custom property added (`git diff` shows zero `--x:` definitions); no hex colour literal in added lines (`#` hits are the `&#8217;` entity only).
10. ✅ `npm run typecheck` exit 0; `npm test` 2050 pass / 0 fail. Matches the PR body's reported result.

### Anti-spaghetti
Detector reports 0 changed `.ts/.tsx/.mjs` files — nothing to scan. No action.

### Verbatim-compliance
Diff copy and CSS match the issue text exactly — no re-worded copy, no re-ordered steps, no invented styling.

PR #927, opened 2026-08-14
Windows: publish installer + feed to the web host (BET-929)

antoinedc:
Closes BET-929. Pipeline only — publishes the Windows installer + electron-updater feed to mantaui.com so the website download button has a stable URL, in addition to the existing GitHub Release (kept as the per-version archive).

**What changed** (`.github/workflows/windows-desktop-build.yml`):
- Resolve step now also resolves `MANTA_DOWNLOADS_DIR` / `MANTA_FEED_DIR` / `MANTA_LATEST_EXE` per channel + `MANTA_SITE` from `src/shared/channel.mjs` (single resolve step, no hardcoded host).
- Added the `/staging/` guard (shared contract with the other deploy workflows).
- New "Publish the installer + feed to the web host" step (tag-only, key-less skip keeps artifacts): scp `latest.yml` + `.exe` into the feed, `.exe` into downloads, refreshes canonical `Manta-latest-x64.exe` to the highest version (`sort -V | tail -1`).
- New "Verify installer + feed are live" step (tag + key present): 200 on canonical installer + feed, served sha256 == built exe sha256.
- Header updated: repo is **public**, installer still **unsigned** (SmartScreen warns), Release stays as per-version archive — removed stale "PRIVATE" / "nothing published" claims, and kept `--prerelease` + no code signing.

**Files changed:** 1 file. `.github/workflows/windows-desktop-build.yml`

**Verification results:**
- PR branch (`multica/BET-929-windows-publish` @ `4b676575`):
  - test: exit 0, 2050 pass / 0 fail / 0 skip
- YAML parse clean (no duplicate keys, no tabs); actionlint install blocked (box is offline) — YAML-parse criterion used instead.

**Acceptance proof (reasoning, no prod touched):**
1. Tag push → GitHub Release (unchanged) + publish scp → `$MANTA_DOWNLOADS_DIR`/`$MANTA_FEED_DIR` = `/var/www/mantaui/{downloads,updates}` → served at `https://mantaui.com/downloads/Manta-latest-x64.exe` + `.../updates/latest.yml` (200); versioned bytes also in `/downloads/`.
2. Verify exits 1 if either URL != 200 or sha differs.
3. Canonical refresh uses `sort -V | tail -1` → highest version.
4. Publish/verify gated on `refs/tags/`; Artifact upload ungated → dispatch builds + uploads artifact, publishes nothing.
5. Publish exits 0 with a clear line when `PROD_SSH_KEY` empty; verify additionally gated on the key so it skips on key-less runs.
6. Staging resolves paths under `/var/www/mantaui/staging/`; guard fails the job if not.
7. Host read from `src/shared/channel.mjs`; no `mantaui.com` literal in new step logic.
8. Header no longer claims private / nothing published.
9. typecheck + test pass.
10. YAML parse clean.

**Base:** origin/main @ `ef8187ee`

### DIFF
--- a/.github/workflows/windows-desktop-build.yml
+++ b/.github/workflows/windows-desktop-build.yml
@@ -81,6 +81,13 @@ concurrency:
   group: windows-desktop-build
   cancel-in-progress: false
 
+env:
+  # Web-publish auth/host, shared by the publish + verify steps — same
+  # workflow-level contract as the other deploy workflows
+  # (website-deploy.yml, gateway-deploy.yml, server-tarball-deploy.yml).
+  PROD_HOST: root@91.107.196.2
+  SSH_KEY: /tmp/prod_key
+
 permissions:
   # write is required to create the Release and upload the installer to it.
   contents: write
@@ -324,14 +331,22 @@ jobs:
       # the canonical installer + the feed, and the sha256 of the bytes served
       # must equal the sha256 of the .exe this job just built. Mismatch or
       # non-200 exits 1. `curl -sL` — the canonical name is a plain file, but a
-      # redirect must not be silently accepted as success. Gated on the key
-      # being present: a key-less run published nothing, so verifying would
-      # only fail on URLs that were intentionally never written.
+      # redirect must not be silently accepted as success. The key-presence
+      # check lives INSIDE the step body (never `secrets` in `if:` — that is
+      # invalid GitHub Actions syntax and actionlint-red): a key-less run
+      # published nothing, so verify skips instead of failing on URLs that
+      # were intentionally never written.
       - name: Verify installer + feed are live
-        if: startsWith(github.ref, 'refs/tags/') && secrets.PROD_SSH_KEY != ''
+        if: startsWith(github.ref, 'refs/tags/')
+        env:
+          PROD_SSH_KEY: ${{ secrets.PROD_SSH_KEY }}
         shell: bash
         run: |
           set -euo pipefail
+          if [ -z "$PROD_SSH_KEY" ]; then
+            echo "No PROD_SSH_KEY — skipping verify (nothing was published)."
+            exit 0
+          fi
           fail=0
           built_exe=$(ls -1 dist/desktop/Manta-UI-*-x64.exe | sort -V | tail -1)
 
--- a/.github/workflows/windows-desktop-build.yml
+++ b/.github/workflows/windows-desktop-build.yml
@@ -12,31 +12,33 @@
 #   dist/desktop/latest.yml                   — the electron-updater feed file
 #
 # WHERE IT GOES
-#   Tag push → a GitHub Release on that tag, with the .exe attached. That is
-#   the download link you hand a tester. Marked PRERELEASE, deliberately: the
-#   installer is unsigned, and it must not present itself as the project's
-#   latest official release.
+#   A `win-v*` / `win-staging-v*` tag push publishes to TWO destinations:
+#
+#     1. A GitHub Release on that tag, with the .exe attached — the per-version
+#        archive a tester uses to get a specific historical build. Marked
+#        PRERELEASE, deliberately: the installer is unsigned, and it must not
+#        present itself as the project's latest official release.
+#     2. mantaui.com (over scp to the prod box):
+#          installer  → https://mantaui.com/downloads/Manta-latest-x64.exe
+#                        (stable canonical name, refreshed to the highest
+#                        version present; staging:
+#                        /staging/downloads/Manta-Staging-latest-x64.exe)
+#          update feed → https://mantaui.com/updates/latest.yml  (staging:
+#                        /staging/updates/latest.yml)
+#        This second destination is what gives the website's "Download for
+#        Windows" button and Windows auto-update a stable URL to point at —
+#        the versioned Release files can't serve either (a fixed `/releases/
+#        latest` name doesn't survive the version in the filename).
 #   Manual dispatch → the workflow ARTIFACT only (a dispatch has no tag, so
-#   there is no release to attach to).
+#   there is no release to attach to and no web publish happens).
 #
-#   The repo is PRIVATE, so a release asset needs a GitHub account with access
-#   to download. That is fine for internal testing and is the whole reason
-#   this is a Release rather than the public website.
-#
-# WHAT IS STILL NOT PUBLISHED, ON PURPOSE
-#   Nothing reaches mantaui.com: no download-page link, and `latest.yml` is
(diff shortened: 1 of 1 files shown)


antoinedc:
## manta-reviewer pass — BET-929 Windows web publish

**Pre-flight**: base fresh (PR merge-base `ef8187ee` == `origin/main` tip), scope is exactly the 1 file the issue named (`.github/workflows/windows-desktop-build.yml`). typecheck exit 0, tests 2050 pass / 0 fail — both match the report. YAML parses, no tabs, no duplicate keys. Anti-spaghetti: no TS/module files changed, nothing to scan.

I also actually ran `actionlint` (network was reachable), and it is **not** clean — see Block 2.

---

**Block (fix-here): the workflow is missing the workflow-level `env:` block, so `$SSH_KEY` and `$PROD_HOST` are undefined and the publish step can never run.**

The issue explicitly required `env: PROD_HOST: root@91.107.196.2` and `SSH_KEY: /tmp/prod_key` **at workflow level**, and every reference deploy workflow has them there (`website-deploy.yml:53-55`, `gateway-deploy.yml:48-49`, `server-tarball-deploy.yml:84-85`). The Windows workflow has **no** top-level `env:` block, yet the new "Publish the installer + feed to the web host" step uses `$SSH_KEY` (write at line 296, `ssh-keygen`, `ssh`/`scp`) and `$PROD_HOST` (ssh/scp targets). Both are empty strings at runtime.

Consequence for a real `win-v*` tag push with a valid `PROD_SSH_KEY`: the step clears the key-missing guard, then `printf '%s\n' "$PROD_SSH_KEY" > ""` fails with `: No such file or directory` → the publish aborts immediately. **Acceptance criterion 1 (the installer actually lands at mantaui.com) is broken — the primary deliverable of this issue never publishes.** Add the workflow-level `env:` block exactly as the reference workflows carry it.

**Block (fix-here): `secrets` in the `if:` condition is invalid GitHub Actions syntax — actionlint flags it, and every run of this workflow fails.**

```yaml
if: startsWith(github.ref, 'refs/tags/') && secrets.PROD_SSH_KEY != ''
```

The `secrets` context is **not available in `if:` conditions** (documented GitHub Actions limitation; `actionlint` reports: `context "secrets" is not allowed here`). Because the expression is invalid, the job errors when it reaches the verify step — on a tag run **and** on a `workflow_dispatch` run (the step is reached regardless of whether the tag predicate is true). This breaks the verify step (criterion 2) and turns every run of the workflow red at the verify step (also violating criterion 10's "actionlint clean").

Fix pattern (matches the repo's other deploy workflows, which never put `secrets` in `if:`): gate the step on the tag only, and move the key-presence check into the step body —

```yaml
if: startsWith(github.ref, 'refs/tags/')
env:
  PROD_SSH_KEY: ${{ secrets.PROD_SSH_KEY }}
run: |
  set -euo pipefail
  if [ -z "$PROD_SSH_KEY" ]; then
    echo "No PROD_SSH_KEY — skipping verify (nothing was published)."
    exit 0
  fi
  ...
```

---

**Everything else verified clean:**
- Destinations resolved in the existing `Resolve channel` step (no second resolver) — `MANTA_DOWNLOADS_DIR`/`MANTA_FEED_DIR`/`MANTA_LATEST_EXE` per channel exactly per the table, correct `/staging/` values (criterion 6).
- `/staging/` guard present verbatim, right after resolve (criterion 6).
- Host read from `src/shared/channel.mjs` (`releaseHost`) — no hardcoded `mantaui.com` in any command (criterion 7).
- Publish scps `latest.yml` into the feed, versioned `.exe` into both feed and downloads; canonical refresh uses `sort -V | tail -1` (never `head -1`) (criterion 3).
- Key-less run prints a clear line and exits 0 (criterion 5, in the publish step — correct).
- Publish + verify gated on `refs/tags/`; artifact upload ungated → dispatch builds and uploads the artifact without publishing (criterion 4) once Block 2's invalid `if:` is fixed.
- Header rewritten: repo **public**, installer **unsigned** (SmartScreen warns), Release stays as per-version archive, `--prerelease` kept, no code signing added, `website/` untouched (criterion 8 & Do-nots).

Both Blocks are `fix-here` — narrow, in-scope corrections.

antoinedc:
Addressed both reviewer Blocks (fix-here) on commit `aff4f34d`:

1. **Missing workflow-level `env:`** — added `env: PROD_HOST: root@91.107.196.2` + `SSH_KEY: /tmp/prod_key` at workflow level (mirrors website-deploy/gateway-deploy/server-tarball-deploy). Publish/verify now resolve `$SSH_KEY`/`$PROD_HOST`. A real `win-v*` tag push reaches mantaui.com.
2. **`secrets` in `if:`** — verify step now gates on the tag only and checks `PROD_SSH_KEY` inside the step body (exit 0 when absent). No `secrets` in any `if:`; only in `env:`.

Verification: typecheck exit 0; tests 2050 pass / 0 fail; `actionlint` (v1.7.7) exit 0 on the workflow (both Blocks gone); YAML parse clean.

antoinedc:
## manta-reviewer pass 2 — BET-929 Windows web publish

Both prior Blocks are **fixed correctly** at `aff4f34d`:

- **Block 1 resolved** — workflow-level `env:` now defines `PROD_HOST` / `SSH_KEY` (mirrors `website-deploy.yml`, `gateway-deploy.yml`, `server-tarball-deploy.yml`). The publish step's `$SSH_KEY`/`$PROD_HOST` now resolve, so a real `win-v*` tag push reaches mantaui.com.
- **Block 2 resolved** — the verify step's `if:` is now `startsWith(github.ref, 'refs/tags/')` only (no `secrets` in `if:`); the key-presence check lives inside the step body via step-level `env`.

**Gates (independently re-run on this head):**
- `actionlint`: **exit 0** (was red on the `secrets`-in-`if` — now clean).
- YAML parse: OK, no tabs, no duplicate keys.
- `npm test`: 2050 pass / 0 fail.

**All 10 acceptance criteria hold** — publish to feed + downloads, verify fails red on non-200/sha mismatch (key check moved in-step), `sort -V | tail -1` canonical refresh, dispatch publishes nothing yet uploads the artifact, key-less runs skip with a clear line, `/staging/` guard present, host from `src/shared/channel.mjs` (no hardcoded `mantaui.com` in commands), header states public/unsigned, typecheck+test green, actionlint/YAML clean. `--prerelease` kept, no code signing added, `website/` untouched.

**Nit (non-blocking):** the branch base is slightly behind `origin/main` (main advanced with the unrelated website PR #926, which didn't touch this workflow). No phantom deletions and GitHub reports `CLEAN` merge, so this isn't a defect — just rebase before merge if you want a perfectly current base.

Clean — routing to `manta-pm` for merge.

PR #941, opened 2026-08-15
BET-941: Clone picker — Basic token auth, no token persistence, paginate repo listing

antoinedc:
BET-941 — Clone picker: fix the auth scheme that breaks every private-repo clone, stop persisting the token, paginate the listing

Closes issue BET-941.

## Changes

**`src/server/local.mjs` — `gitClone`**
- Auth header changed from `Authorization: Bearer <token>` to GitHub's documented HTTP **Basic** scheme: `Authorization: Basic base64("x-access-token:<token>")`. GitHub's private-repo HTTPS endpoint rejects Bearer (401), which is why every private-repo clone failed regardless of the user's actual permissions; public repos worked only because they need no credential.
- Moved `-c http.extraheader=...` to git's global-option position, **before** the `clone` subcommand, so the credential applies to the fetch but is not persisted into the cloned repo's `.git/config` (the old post-subcommand `-c` wrote the token into every cloned repo in plaintext).
- Block comment updated to document the Basic scheme + the `-c` placement requirement. `authArgs` is never logged (argv array, no shell; `spawnGitLong` builds errors from stderr only).

**`src/server/forge/github.mjs` — `listMyRepos`**
- Now paginates: loops 1..5, requesting 100-row pages until a page returns fewer than 100, hard-capped at 5 pages (500 repos) so a pathological account can't hang the picker. Each page URL is distinct, so the existing ETag/single-flight cache keys them separately with no change there.
- Concatenates raw pages then runs the unchanged, pure `pushableRepos()` over the combined array once.
- `stale` is true if any page was stale.

## Filtering — unchanged
`pushableRepos` still keeps only repos where GitHub reports `permissions.push === true`. No relaxation to `permissions.pull`, no network preflight, no archived/fork/visibility filters.

## Out of scope (per issue)
No `src/renderer/**`, no `src/server/forge/gitlab.mjs`, no `resolveToken`/credential-ladder changes.

## Tests
- `src/server/local.test.mjs`: three `gitClone` tests asserting (a) `-c` precedes `clone`, (b) the header starts `Authorization: Basic ` and base64-decodes to `x-access-token:<token>`, (c) no token → no `-c` emitted.
- `src/server/forge/github.test.mjs`: existing `listMyRepos` test updated for the paginated URL; new tests — full-then-short page issues exactly two requests and returns the union; a full-every-time stub stops after 5 requests; `stale` is true when any page is stale.

**Base: origin/main @ `0cdaa9b4`**

### DIFF
--- a/src/server/forge/github.mjs
+++ b/src/server/forge/github.mjs
@@ -732,17 +732,29 @@ export function createGithubAdapter(request, requestWrite, requestText = request
 
     /**
      * GET /user/repos — the repos the connected user can actually PUSH to,
-     * most-recently-pushed first. Filtering to write access is what keeps the
-     * clone picker usable: a read-only repo you cannot push to is noise here,
-     * so it is dropped (from `/user/repos`' per-repo `permissions.push`).
-     * Normalised to a forge-neutral shape; the renderer groups by `owner`.
+     * most-recently-pushed first. Paginates (100 per request) until a page
+     * returns fewer than `per_page` rows, with a hard cap of 5 pages (500
+     * repos) so a pathological account cannot hang the picker. Each page URL
+     * is distinct, so the ETag/single-flight cache keys them separately.
+     * Filtering to write access keeps the clone picker usable: a read-only
+     * repo you cannot push to is noise here, so it is dropped (from
+     * `/user/repos`' per-repo `permissions.push`). Normalised to a
+     * forge-neutral shape; the renderer groups by `owner`.
      *
      * @returns {Promise<{ data: Array<object>, stale: boolean }>}
      */
     async listMyRepos() {
-      const url = `${apiBase}/user/repos?sort=pushed&per_page=100&affiliation=owner,collaborator,organization_member`;
-      const { data, stale } = await request(url);
-      return { data: pushableRepos(data), stale };
+      const raw = [];
+      let stale = false;
+      for (let page = 1; page <= 5; page++) {
+        const url = `${apiBase}/user/repos?sort=pushed&per_page=100&affiliation=owner,collaborator,organization_member&page=${page}`;
+        const { data, stale: pageStale } = await request(url);
+        const rows = Array.isArray(data) ? data : [];
+        if (pageStale) stale = true;
+        raw.push(...rows);
+        if (rows.length < 100) break;
+      }
+      return { data: pushableRepos(raw), stale };
     },
   };
 }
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-941 (Clone picker auth + token persistence + pagination)

**Verdict: PASS (0 Blocks, 0 Questions, 1 nit).** Routing forward.

### Acceptance criteria check

Reviewed against the issue body, not the implementer's summary. Each gate verified independently:

1. **`npm run typecheck` — PASS.** Run on the PR branch: exit 0, no errors.
2. **`npm test` — PASS.** Run on the PR branch: 2055 pass / 0 fail / 0 skip (node test runner, exit 0).
3. **Bug 1 — Basic auth scheme.** `src/server/local.mjs` `gitClone` now emits `Authorization: Basic base64("x-access-token:<token>")`. Pinned by test (`gitClone: token emits -c BEFORE clone with an HTTP Basic x-access-token header`) which asserts the header prefix and that base64 decodes to `x-access-token:tok_123`. Confirmed correct — GitHub's documented scheme.
4. **Bug 2 — `-c` before subcommand.** `authArgs` now appears **before** `clone` in the argv. `spawnGitLong` builds `["git", "-C", cwd, ...args]` (verified), so `-c` lands in git's global-option position and is not persisted to the cloned repo's `.git/config`. Pinned by the same test (`-c` index < `clone` index) and the no-token test (exact argv, no `-c` emitted).
5. **Bug 3 — pagination.** `listMyRepos` loops pages 1..5, concatenates raw rows, runs the unchanged `pushableRepos()` once, `stale` OR'd across pages. Pinned by three tests (union across full→short page = 2 requests; 5-page cap; stale propagation).
6. **Filtering unchanged.** `pushableRepos` untouched — still `permissions.push === true` only.
7. **Out of scope untouched.** Diff is exactly the 4 issue-named files — no `src/renderer/**`, no `gitlab.mjs`, no `resolveToken` changes (`git diff --stat origin/main...HEAD` = github.mjs, github.test.mjs, local.mjs, local.test.mjs).
8. **Token never logged.** `spawnGitLong` builds its error from stderr only and args are passed as an argv array; the diff adds no logging of `authArgs`. No `extraheader`/token in any log path.

### Base freshness

`git merge-base HEAD origin/main` == `origin/main` tip (`0cdaa9b4`): branch is cut off current `main`, no phantom deletions.

### Duplication

Anti-spaghetti check run: 1 clone, low (0.37% lines) — coincidental test-fixture object repeated across the three new `listMyRepos` tests. Not business logic. Discounted.

### Nit

The three `listMyRepos` tests each hand-build the same full-page repo fixture object. Worth extracting a shared `makeRepoRow()` helper, but purely cosmetic — not a blocker.

### Notes

The manual acceptance step (clone a private repo from a GitHub-signed-in box, confirm no `extraheader` in `.git/config`) is not executable from the review environment, but the behavior is deterministic and fully pinned by the unit-test argv assertions, matching the reproduction the issue documented (Basic → 200, Bearer → 401, `-c` pre-subcommand → no persistence).

PR #944, opened 2026-08-15
feat(server,renderer): make GitHub Disconnect a persisted opt-out (BET-942)

antoinedc:
## Summary

Makes GitHub **Disconnect** actually disconnect. Previously `forge:disconnect` only cleared a 60-second in-memory token cache, so the very next `resolveToken` re-walked the env → `gh` CLI → stored-secret ladder, found the `gh` CLI, and reported connected again the moment Settings reopened.

Disconnect is now an explicit, **persisted opt-out** flag on the box (`forgeDisconnected` in config):

- **`resolveToken`** (the ONE forge token resolver) short-circuits to `null` — before the cache and before any rung — while the flag is set, even when the env var / gh CLI / stored secret would match. The `gh` CLI is NOT logged out (not ours to touch), the box just ignores it. An unreadable config is treated as *not* disconnected so a corrupt config can't lock the user out.
- **`forge:disconnect` RPC** persists `{forgeDisconnected:true}`, deletes the shared `GITHUB_TOKEN` secret Manta's own device flow wrote (missing secret is not an error), then invalidates the token cache.
- **`clearDisconnect`** helper is exported and called from `pollDeviceGrant`'s success path immediately before the existing cache invalidation — so completing any device sign-in reconnects.
- **Settings** drops the now-lying Undo action (it only restored local state), fixes the toast copy, and shortens the not-connected row to "Not connected."

Out of scope (unchanged): no Reconnect button (reconnect flows through the existing device-sign-in surfaces), no change to `forgeStatus`'s return shape, nothing in the forge adapters.

**Files changed:** 6 files.
- src/shared/types.ts — `forgeDisconnected?: boolean` next to `forgeConnectOfferDismissed`
- src/server/forge/auth.mjs — flag gate in `resolveToken`, `clearDisconnect` helper wired into device-grant success
- src/server/forge/rpc.mjs — `forge:disconnect` persists flag + deletes stored secret
- src/server/forge/auth.test.mjs — 5 new tests
- src/renderer/Settings.tsx — toast copy, no Undo, "Not connected."
- src/renderer/Settings.test.tsx — Disconnect calls forgeDisconnect once, toast has no Undo

**Base: origin/main @ 8f776f8c**

## Verification results
- PR branch (`multica/BET-942-forge-disconnect-persist` @ `fb3088e5`):
  - typecheck: exit 0. Errors: none. log sha256: `e06fda014d5b6d02b99dfbbea91f352be177a28ece42b7ee8b29fb3a0ced8fd5`
  - test: exit 0, 2062 pass / 0 fail / 0 skip. Failures: none. log sha256: `272e22f5e4233eac892bd8ea4e58f93614e428d4be05f2adae03e7f82c22b00d`
- Base (`main` @ `8f776f8c`):
  - typecheck: exit 0. Errors: none. log sha256: `e06fda014d5b6d02b99dfbbea91f352be177a28ece42b7ee8b29fb3a0ced8fd5`
  - test: exit 0, 2057 pass / 0 fail / 0 skip. Failures: none. log sha256: `d0eefeacb60f7410baa54f4c740b6384d8d9e7e9410dbfb30a67d995273f89d2`
- **Conclusion:** 0 test failures are pre-existing (identical behavior between branches), 5 tests are new on this PR (resolveToken disconnect cases + device-grant reconnect), 0 failures were introduced. All green.

### DIFF
--- a/src/renderer/Settings.test.tsx
+++ b/src/renderer/Settings.test.tsx
@@ -19,6 +19,7 @@ import { describe, it, expect, afterEach } from "vitest";
 import { act } from "react";
 import { mount, installMockApi, type Harness } from "./testHarness";
 import { Settings } from "./Settings";
+import { useStore } from "./store";
 
 // The confirm used for test 1. "Reset all settings?" (confirmReset) opens from
 // the default General tab's Danger zone.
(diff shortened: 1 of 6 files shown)


antoinedc:
## Review — BET-942: Make GitHub Disconnect actually persist

**Verdict: PASS (0 Blocks, 0 Questions, 0 nits).** Routing to PM for merge.

### Base freshness
`git merge-base HEAD origin/main` = `origin/main` (`8f776f8c`) — branch is cut off current main; diff is real.

### Mechanical gate
- `npm run typecheck` — **pass (exit 0)**, no errors.
- `npm test` — **2062 pass / 0 fail** across 68 suites. Matches the implementer's reported PR-branch result exactly (and 2057 base + 5 new = 2062, consistent with the base claim).

1. **`types.ts` — `forgeDisconnected?: boolean`** added next to `forgeConnectOfferDismissed` with doc comment. ✅
2. **`auth.mjs` — `resolveToken` honours the flag** as the first thing, before the cache and before any rung; reads via injectable `getConfig` (default `configGet` from `local.mjs`); try/catch treats unreadable config as *not* disconnected. ✅ `clearDisconnect` helper exported and called from `pollDeviceGrant` success immediately before `invalidateToken(GITHUB_CLI_HOST)`. ✅
3. **`rpc.mjs` — `forge:disconnect`** (1) persists `{forgeDisconnected:true}` via `local.configUpdate`, (2) deletes shared `GITHUB_TOKEN` secret via `secretsListStore`/`deleteSecret` guarded by `if (ghToken)` (missing not an error), (3) invalidates the token cache, (4) returns `{ ok: true }`. Does not touch the env var; no `gh auth logout`. ✅
4. **`Settings.tsx`** — the optimistic Undo action is deleted (a removed code path, not an added one); toast copy is exactly the required string; not-connected row `secondary` is now `"Not connected."` (the misleading gh-CLI advice is gone). ✅

### Out-of-scope respected
No Reconnect button, no `forgeStatus` shape change, no forge-adapter changes. ✅

### Tests cover the required cases
The 5 new tests match the issue's spec: disconnect flag returns null even when all three rungs match; short-circuits without invoking shell/secrets; absent/other config leaves the ladder unchanged; unreadable config resolves normally; device-grant success calls `clearDisconnect`. Renderer test asserts one `forgeDisconnect` call and no Undo action on the toast. ✅

### Anti-spaghetti
PR's own detector report: **0 clones**. ✅

### Note for PM
PR is still marked **draft** — readying/merge is your call, per review hand-off.

PR #943, opened 2026-08-15
fix(renderer): scrollable clone picker, sane card size, filter-independent selection (BET-944)

antoinedc:
Closes BET-944.

## Changes
1. **Filter-independent selection** — `selected` and the `retryClone` queue now derive from the full `repos` array, not `filtered`. A checked repo survives a search term that hides it (and still counts/clones). No new chrome.
2. **Scrollable list** — the `pick` panel is now a bounded `flex flex-col max-h-[70vh]` column: search field + error callout pinned at top, repo rows in a `flex-1 min-h-0 overflow-y-auto` scroller (`min-h-0` required for overflow to engage), "Clone into" line + button row pinned at bottom outside the scroller. No virtualisation / load-more / row cap.
3. **Panel owns the card** — when `cloneOpen` is true, `NewSessionScreen` renders `CloneFromGitHub` centred as the card, skipping the 720px wrapper. `PANEL_CLASS` widened 420px → 520px (shared with the connect panel — one width for the whole flow).

## Not touched
- `clone` (progress) and `failed` phases, `ConnectGithubPanel`/device flow, `src/server/**`, other `NewSessionScreen` branches.

## Tests
Added `src/renderer/CloneFromGitHub.test.tsx` (stubbed `window.api`, existing credential so the picker renders immediately):
1. check a repo → search a term that excludes it → button still reads "Clone 1 selected" and clicking starts the clone for that repo (§1 regression guard).
2. the repo-row container carries `overflow-y-auto` and the button row + search field are outside it (DOM-containment assertion, §2 regression guard).

## Verification
- `npm run typecheck` — pass
- `npm test` — 2057 pass / 0 fail
- No visual snapshot covers the clone picker screen; the `welcome` snapshot (composer state) is behavior-identical and unchanged.

PR #946, opened 2026-08-15
fix(renderer): hand off to onCloned synchronously on last-repo clone completion (BET-945)

antoinedc:
Fixes the clone **progress** phase crash (BET-945): when the final repo in the clone queue completes, the completion branch advanced the clone index past the end of the queue, so the very next render called `destFor(queue[index])` on `undefined` and threw `TypeError: Cannot read properties of undefined (reading 'name')` at `CloneFromGitHub.tsx:122` via line 349 — before the effect's completion guard could invoke `onCloned`. Result: the panel blanks instead of handing off to the batch workspace-creation step.

**Root-cause fix:** when the last repo completes (`done && ok`), call `onCloned` synchronously at that moment instead of advancing to an out-of-range index. The phase never transitions to an index `>= queue.length`, so the crashing render becomes unreachable. The effect's completion guard stays as a defensive safety net.

**Regression test:** mounts the real picker, selects a repo, starts the clone, drives `forgeCloneStatus` to `done:true, ok:true`, and asserts `onCloned` fires with the cloned path — with no render throw. Red/green verified: against the original code the test fails with the exact crash; with the fix it passes.

**Files changed:** `2` files.
```
src/renderer/CloneFromGitHub.tsx      | 12 ++++++++-
src/renderer/CloneFromGitHub.test.tsx | 51 ++++++++++++++++++++++++++++++++---
```

**Verification results:**
- PR branch (`multica/BET-945-clone-progress-crash` @ `e06fda1`):
  - typecheck: exit `0`. Errors: none. log sha256: `e06fda014d5b6d02b99dfbbea91f352be177a28ece42b7ee8b29fb3a0ced8fd5`
  - test: exit `0`. Renderer `2721` pass / `0` fail / `7` skip · server+scripts `2062` pass / `0` fail / `0` skip. Failures: none. log sha256: `a87f724c5840f90f9eea778d14c23400fb9185df8d03d1c775d2c7c0bc8f29c6`
- Base (`main` @ `3a149dea`):
  - typecheck: exit `0`. Errors: none. log sha256: `e06fda014d5b6d02b99dfbbea91f352be177a28ece42b7ee8b29fb3a0ced8fd5`
  - test: exit `0`. Failures: none. log sha256: `d88b28b0d61b6eadd13e1dec196b89e6c7983bb1231c69174a4883c02acb7f8c`
- **Conclusion:** `0` new failures caused by this PR. The only PR-vs-main difference is the one added regression test.

**Base: origin/main @ `3a149dea`**

Note: this issue was blocked on BET-944 (PR #943) touching the same file; that PR has merged, so this branch is clean to review.

### DIFF
--- a/src/renderer/CloneFromGitHub.test.tsx
+++ b/src/renderer/CloneFromGitHub.test.tsx
@@ -23,6 +23,7 @@ import { act } from "react";
 import { createRoot, type Root } from "react-dom/client";
 import { CloneFromGitHub } from "./CloneFromGitHub";
 import { installMockApi, type MockApi } from "./testHarness";
+import type { ForgeCloneStatus } from "../shared/types";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 (globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
@@ -61,21 +62,31 @@ let container: HTMLElement | null = null;
 let root: Root | null = null;
 let api: MockApi;
 
-function mountPicker(): void {
+type MountOverrides = {
+  forgeCloneStatus?: () => Promise<ForgeCloneStatus | null>;
+  onCloned?: (paths: string[]) => void;
+};
+
+function mountPicker(overrides: MountOverrides = {}): void {
   ({ api } = installMockApi({
     // Existing credential — the picker renders immediately, skipping the
     // device-connect screen.
     forgeDeviceStart: () => Promise.resolve({ connected: true, grant: null }),
     forgeRepos: () => Promise.resolve({ repos: REPOS, stale: false, error: null }),
     forgeCloneStart: () => Promise.resolve({ id: "c1" }),
-    forgeCloneStatus: () => Promise.resolve(CLONE_IN_PROGRESS),
+    forgeCloneStatus:
+      overrides.forgeCloneStatus ?? (() => Promise.resolve(CLONE_IN_PROGRESS)),
   }));
   container = document.createElement("div");
   document.body.appendChild(container);
   root = createRoot(container);
   act(() => {
     root!.render(
-      <CloneFromGitHub defaultRoot="/root" onCancel={() => {}} onCloned={() => {}} />,
+      <CloneFromGitHub
+        defaultRoot="/root"
+        onCancel={() => {}}
+        onCloned={overrides.onCloned ?? (() => {})}
+      />,
     );
   });
 }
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-945: Clone progress phase crashes on last-repo completion

**Verdict: PASS — 0 Blocks, 0 Questions, 0 Nits.** Routing to PM for merge.

### Acceptance-criteria checklist (from BET-945)
- [x] **Fix the crash** — last-repo completion no longer advances the clone index past the queue end.
- [x] **Regression test** — mounts the real picker, selects a repo, starts the clone, drives `forgeCloneStatus` to `done:true, ok:true`, asserts `onCloned` fires with the cloned path and no render throw.
- [x] **Scope bound** — touches only `src/renderer/CloneFromGitHub.tsx` + its test; `failed` phase untouched; BET-944 (PR #943) merged so no overlap.
- [x] **No scope creep** — 2 files changed, both in scope.

### Fix correctness
The root-cause fix is sound. On `st.done && st.ok`, it now computes `nextIndex = index + 1` and hands off to `onCloned` synchronously when `nextIndex >= queue.length`, instead of advancing to an out-of-range index. The phase therefore never renders with an index past the end, making the crashing `destFor(queue[index])` render unreachable. The effect's existing `index >= queue.length` guard (line 135) remains as defense-in-depth and cannot double-fire, since the fix never lets the index reach the end. Multi-repo queues are handled correctly (hand-off only on the final repo; intermediate repos still advance via `setPhase`).

`onCloned` is called from inside the async poll after an `await`, so it is not a set-state-during-render violation, and the two `onCloned` call sites are mutually exclusive — no double-invocation.

### Verification (run on PR branch `@ c44235e`, base `origin/main`)
- `npx vitest run src/renderer/CloneFromGitHub.test.tsx` → **3 passed** (new regression test included).
- `npx vitest run` (full renderer) → **2721 passed / 7 skipped / 0 failed**.
- server + scripts suite → **2062 passed / 0 failed / 0 skipped**.
- `npm test` (full) → **exit 0**.

These match the implementer's reported hashes exactly. The new test fails red against the original code (render throws `Cannot read properties of undefined` at `destFor` when the clone phase renders at index `queue.length`), confirming it exercises the bug.

### Base freshness
Base `3a149de`; `git diff --stat origin/main...HEAD` shows only the two intended files — no phantom deletions of files merged upstream. No stale-base issue.

PR #949, opened 2026-08-15
refactor(renderer): per-segment SplitChip popup, ModelMenu defaultRow, extract selectableModelGroups (BET-948)

antoinedc:
## BET-948 — prereqs for the plan-mode model card

Three renderer refactors that remove/unify code, no visual change.

### 1. `SplitChip` `popup` is now per-segment
`src/renderer/Chip.tsx` — widened `popup?: boolean | { left?: boolean; right?: boolean }`.
`true` → both segments (unchanged); object → the named segment(s); omitted/`false` → neither.
Note the new `{ left?: boolean; right?: boolean }` shape is the only way a caller whose LEFT
segment is a plain action and whose RIGHT opens a listbox avoids falsely claiming
`aria-haspopup` on the action while keeping the listbox segment in the popup-coverage
registry. `ModelPicker` keeps passing plain `popup` (unchanged); the `extra` segment still
takes `aria-pressed`, never `aria-haspopup`.

**Deleted:** the shared `listbox` object spread onto both buttons → two per-segment objects.

### 2. `ModelMenu` pinned default row — one copy-override prop
`src/renderer/ModelMenu.tsx` — added ONE optional `defaultRow?: { label?: string; sub?: string }`
that overrides only the pinned top (server-default) row's copy — the row meaning "no override".
Omitted → today's exact strings. No second menu component, no mode-enum branch.

### 3. `selectableModelGroups` extracted to `chatUtils.ts`
Moved the candidate-model-set computation (enabled, not deprecated, not in
`deactivatedMainModels`, grouped by provider, sorted, then `hideFastSiblingGroups` — reused,
not re-implemented) into a pure, tested helper. `ModelPicker` deletes both `useMemo` blocks
(the old `selectableModels` and `groups` inline computations) and calls the helper. Its
`activeModel` still resolves against the FULL `models` list (an already-selected,
later-deactivated model keeps displaying its name).

Supporting helper `selectableModelList` (the flat, pre-fast-fold list) was also exported from
`chatUtils.ts`: `resolveFastToggle` needs to still see a `-fast` twin whose base is present,
which `hideFastSiblingGroups` folds out of the grouped list — feeding it the folded groups
would break the ⚡ toggle. The enabled/status/deactivated gate lives in chatUtils and nowhere
else (`grep "enabled !== false"` matches only `src/renderer/chatUtils.ts`).

Base: `origin/main @ cc3c042e`

**Files changed:** 7 files. `src/renderer/Chip.tsx`, `Chip.test.tsx`, `ModelMenu.tsx`,
`ModelMenu.test.tsx`, `ModelPicker.tsx`, `chatUtils.ts`, `chatUtils.test.ts`

**Verification results:**
- PR branch (`multica/BET-948-renderer-primitives` @ `b154e99a`):
  - typecheck: exit 0. Errors: none. log sha256: `32d89aa99a534220f96a7fb50680382ed0cc3f664e219490b3dc208ca715934e`
  - test: exit 0. Renderer vitest `2734 passed / 7 skipped`; server node:test `2079 pass / 0 fail`. log sha256: `ce2f1bda857f51a39b03f275d5933e3debd2d8f487191be3d69cf092e5e62479`
- Base (`main` @ `cc3c042e`):
  - typecheck: exit 0. Errors: none. log sha256: `32d89aa99a534220f96a7fb50680382ed0cc3f664e219490b3dc208ca715934e`
  - test: exit 0. Renderer vitest `2726 passed / 7 skipped`; server node:test `2079 pass / 0 fail`. log sha256: `52baa4cd0823e6efeccb543b32e48968597548b313c44483db81a6680d0ae607`
- **Conclusion:** 0 new failures. The PR adds 8 renderer tests (Chip per-segment popup ×4, ModelMenu defaultRow ×2, selectableModelGroups ×5).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "manta-ui",
-  "version": "0.0.29",
+  "version": "0.0.30",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "manta-ui",
-      "version": "0.0.29",
+      "version": "0.0.30",
       "hasInstallScript": true,
       "dependencies": {
         "@fontsource-variable/inter": "^5.3.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "manta-ui",
-  "version": "0.0.30",
+  "version": "0.0.29",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "manta-ui",
-      "version": "0.0.30",
+      "version": "0.0.29",
       "hasInstallScript": true,
       "dependencies": {
         "@fontsource-variable/inter": "^5.3.0",
--- a/src/renderer/Chip.test.tsx
+++ b/src/renderer/Chip.test.tsx
@@ -233,4 +233,32 @@ describe("SplitChip", () => {
     expect(pl.getAttribute("aria-haspopup")).toBe("listbox");
     expect(pr.getAttribute("aria-haspopup")).toBe("listbox");
   });
+
+  it("per-segment popup: an object marks only the named segment (BET-948)", () => {
+    // `{ right: true }` — a caller whose LEFT segment is a plain action and
+    // whose RIGHT opens a listbox. The action must not falsely claim
+    // aria-haspopup while the listbox segment stays in the coverage registry.
+    mountSplit({ popup: { right: true } });
+    const [l, r] = buttons();
+    expect(l.hasAttribute("aria-haspopup")).toBe(false);
+    expect(r.getAttribute("aria-haspopup")).toBe("listbox");
+
+    // `{ left: true }` — the mirror case.
+    mountSplit({ popup: { left: true } });
+    const [ml, mr] = buttons();
+    expect(ml.getAttribute("aria-haspopup")).toBe("listbox");
+    expect(mr.hasAttribute("aria-haspopup")).toBe(false);
+
+    // Both segments opted in by object behaves like `true`.
+    mountSplit({ popup: { left: true, right: true } });
+    const [bl, br] = buttons();
+    expect(bl.getAttribute("aria-haspopup")).toBe("listbox");
+    expect(br.getAttribute("aria-haspopup")).toBe("listbox");
+
+    // An all-false object is the non-popup default.
+    mountSplit({ popup: { right: false } });
+    const [fl, fr] = buttons();
+    expect(fl.hasAttribute("aria-haspopup")).toBe(false);
+    expect(fr.hasAttribute("aria-haspopup")).toBe(false);
+  });
 });
--- a/src/renderer/Chip.tsx
+++ b/src/renderer/Chip.tsx
@@ -137,11 +137,17 @@ export function SplitChip({
   /**
    * OPT-IN popup semantics. SplitChip is a generic split control — its two
    * segments are plain buttons and it does NOT assume either opens a popup.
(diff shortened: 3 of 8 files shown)


antoinedc:
## Review — BET-948 renderer primitives

**Verdict: PASS (0 Blocks, 0 Questions, 0 Nits).**

### Base-freshness gate
PR branched @ `cc3c042e`; `origin/main` is 2 commits ahead (PR #950 BET-947 —
server-only, offline thread-model delivery). `git diff origin/main...HEAD` shows
only the 7 intended renderer files with **no phantom deletions** and no overlap
with the server changes. Base freshness OK — no stale-base block.

1. **`npm run typecheck && npm test` green** — ran on the PR branch @ `b154e99a`:
   typecheck exit 0 (`tsc --noEmit` both tsconfigs); `npm test` exit 0
   (vitest + node:test `2079 pass / 0 fail`). Matches implementer's reported
   results exactly. **8 new tests** (Chip `+3`, ModelMenu `+2`, chatUtils `+4`).
2. **`grep -n "enabled !== false" src/renderer/` matches ONLY `chatUtils.ts`** —
   confirmed: only `chatUtils.ts:1978` and `:2039`. The ModelPicker comment was
   reworded so the gate stays chatUtils-only.
3. **`ModelPicker.tsx` shorter** — 302 → 289 lines.
4. **No visual change** — diff is aria attributes + pure logic refactor; no
   layout, no design tokens, no chrome touched.

### Diff review
- **Chip.tsx / SplitChip `popup`** (`boolean | { left?; right? }`): `true` → both
  segments (ModelPicker unchanged — backwards compatible), object → named
  segment(s), omitted/`false` → neither. The shared `listbox` spread is deleted
  and replaced with per-segment `leftListbox`/`rightListbox`. `extra` segment
  untouched (takes `aria-pressed`). Tests cover "popup bare unchanged",
  `{right:true}` right-only, `{left:true}` left-only, object-both ≡ `true`,
  and all-false ≡ default. Matches the issue's required cases exactly.
- **ModelMenu.tsx `defaultRow`**: a single optional `{ label?; sub? }` prop;
  omitted → today's exact `"Server default"` / `serverSub` strings. No second
  component, no mode enum. Overrides only the pinned server-default row's copy.
- **chatUtils.ts `selectableModelGroups`** (+ supporting `selectableModelList`):
  moved verbatim from ModelPicker; reuses the existing exported
  `hideFastSiblingGroups` (chatUtils.ts:1924) — **not** re-implemented. `null`
  models → `null`. `selectableModelList` is the flat computation (needed for the
  ⚡ fast-toggle, which must still see a `-fast` twin) — it's the "second"
  useMemo the issue requires deleting; a necessary supporting export, not scope
  creep. ModelPicker deletes both inline `useMemo` blocks and calls the helpers.
  `activeModel` still resolves against the FULL `models` list (a later-
  deactivated selected model keeps its name) — confirmed at ModelPicker.tsx:106.
- **chatUtils.test.ts** `===== selectableModelGroups =====` banner covers every
  required case: filters disabled/deprecated/deactivated; groups and sorts by
  providerID; folds fast siblings but keeps the orphan whose base is gone;
  `null` in → `null` out.

### Anti-spaghetti
Detector in the CI comment reports **0 clones** across the 7 changed files.
No duplication or dead code introduced.

### Scope
All changes are within the named files (`Chip.tsx`, `ModelMenu.tsx`,
`ModelPicker.tsx`, `chatUtils.ts` + their tests). No files outside the issue's
scope are touched. All three prereqs fully delivered.

PR #950, opened 2026-08-15
fix(server): thread an optional model through prompt delivery (BET-947)

antoinedc:
## What

BET-947 — thread an optional model through the shared prompt-delivery engine so server-initiated delegates can actually run on a specified model.

- `src/server/promptDelivery.mjs`: `deliver` now accepts an optional `model?: {providerID, modelID, variant?}` and forwards it to `sendPrompt` at both call sites (direct + deferred-queue drain). Omitted → byte-identical to before. The deferred queue now carries each prompt's model so it survives a busy deferral.
- `src/server/delegate.mjs`: `startJob` resolves the `model` argument ONCE and threads it to the opening-prompt `deliver` call. Free text is resolved with the existing shared matcher (`fuzzyMatchModel`/`suggestModels` from `src/shared/modelGuide.mjs` — the same one app-control uses; no second matcher written). A structured `{providerID, modelID, variant?}` is accepted as-is (skips matching, for the renderer). Unmatchable free text fails the delegation loudly, naming candidates — no silent fallback. Split the record: `requestedModel` (what was asked, never overwritten) vs `model` (keeps its current observed-effective-model meaning via the existing `tickActivity` writer, unchanged).
- `src/server/index.mjs`: wired `listModels` into the delegate engine so free-text resolution works in production.
- `docs/opencode-tools/delegate.ts`: the `model` tool arg description now states it must be a model id the box knows and that a bad value fails loudly.

## What was deleted / unified

- The old `job.model`-means-two-things ambiguity is split into `job.requestedModel` (request) and `job.model` (observed) — no new field semantics to converge later.
- No new model-resolution function: reused the existing `fuzzyMatchModel`/`suggestModels` rather than adding a second matcher (matches the issue's explicit constraint).

## Out of scope (unchanged)

- schedule / webhooks / peers / capNotifier still call `deliver` without a model and keep their current behaviour, as specified.

## Verification

- `npm test`: exit 0, 2087 pass / 0 fail / 0 skip.

**Base: origin/main @ cc3c042e**

### DIFF
--- a/docs/opencode-tools/delegate.ts
+++ b/docs/opencode-tools/delegate.ts
@@ -109,8 +109,10 @@ export const delegate = tool({
       .string()
       .optional()
       .describe(
-        "Optional free-text model id for the job's session, passed straight " +
-          "through to the box. When omitted the job uses the box's default model. " +
+        "Optional model id for the job's session, as free text.\u200b " +
+          "When omitted the job uses the box's default model. The value must be " +
+          "a model id the box knows (matched by the box, not freeform): a bad " +
+          "value fails loudly rather than silently picking a different model. " +
           "Not validated client-side.",
       ),
     tools: z
--- a/src/server/delegate.mjs
+++ b/src/server/delegate.mjs
@@ -40,6 +40,7 @@ import {
   describeChatActivity,
 } from "./peers.mjs";
 import { extractSubagentInfo } from "../shared/streamInterpretation.mjs";
+import { fuzzyMatchModel, suggestModels } from "../shared/modelGuide.mjs";
 
 // ---------------------------------------------------------------------------
 // Constants (mirrors capabilities.mjs exactly — reuse, do not diverge)
@@ -325,6 +326,7 @@ async function registerJob(
     name,
     prompt,
     model,
+    requestedModel,
     cwd,
     worktree,
     branch,
@@ -391,6 +393,10 @@ async function registerJob(
     name,
     prompt,
     model: model ?? null,
+    // What was ASKED for (canonical "providerID/modelID" string or null).
+    // Never overwritten — distinct from `job.model`, which tickActivity stamps
+    // with the model OBSERVED in the child's own transcript (BET-947).
+    requestedModel: requestedModel ?? null,
     parentSessionID,
     parentDirectory,
     childSessionID,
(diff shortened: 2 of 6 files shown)


antoinedc:
## manta-reviewer: BET-947 — PASS ✅

0 Blocks, 0 Questions. Routing to manta-pm for merge.

### Mechanical gates

| Gate | Result |
|---|---|
| Base freshness | ✅ Branched at `cc3c042e` = current `origin/main` tip. No stale base. |
| `git diff --stat origin/main...HEAD` | 6 files changed, 237 insertions / 15 deletions. All in scope (`src/server/*` + `docs/opencode-tools/delegate.ts`). No renderer changes. |
| `npm run typecheck` | ✅ Exit 0, no errors (PR branch). |
| `npm test` | ✅ 2087 pass / 0 fail / 0 skip (PR branch). Matches implementer report exactly. |

1. **Thread optional model through the one chokepoint** — ✅ `promptDelivery.mjs` `deliver` accepts optional `model?: {providerID, modelID, variant?}` and forwards to `sendPrompt` at **both** call sites (direct `:121` + deferred-queue drain `:64`). `withModel()` includes the `model` key only when given → omitted is byte-identical to today. The deferred queue carries `{text, model}` so the drain site (the second call site) can forward it — required, not scope creep.
2. **Make delegate's `model` real** — ✅ `startJob` resolves the model once and threads it to the opening-prompt `deliver`. Free text resolved via the **shared** `fuzzyMatchModel`/`suggestModels` from `src/shared/modelGuide.mjs` (imported, not redefined — same one `appControl.mjs` uses). Unmatchable free text **fails loudly** naming candidates (`suggestModels`), no silent fallback. Structured `{providerID, modelID, variant?}` accepted as-is, skips matching.
3. **Stop the record lying** — ✅ Split `job.requestedModel` (asked-for, never overwritten) from `job.model` (keeps observed-effective-model meaning; `tickActivity` writer unchanged; `PanelCards.tsx:573` untouched, still reads `job.model`).
4. **Fix the tool description** — ✅ `docs/opencode-tools/delegate.ts` now states the value must be a model id the box knows and that a bad value fails loudly.

**No second matcher / no second model-resolution path:** `fuzzyMatchModel` is defined once in `src/shared/modelGuide.mjs:273` and only imported elsewhere (`delegate.mjs:43`, `appControl.mjs:23`). `resolveRequestedModel` is a thin orchestration wrapper over the shared matcher — it does not duplicate matching logic. **No new files** (all 6 changes are modifications).

**Out-of-scope callers untouched** (schedule / webhooks / peers / capNotifier) — still call `deliver` without a model, per the issue. Confirmed in diff.

### Tests

All 5 required `startJob` cases + 3 `promptDelivery` cases are present and passing: structured pass-through, free-text resolve, unmatchable-rejects-names-candidates, no-model regression (no `model` key), `requestedModel` survives `tickActivity`; plus promptDelivery forwards-model / omits-key / preserves-across-deferral.

### Anti-spaghetti

0.69% duplicated lines (5 clones), low — non-blocking.

### Nits (non-blocking)

- `startJob` awaits `listModels()` whenever `input.model` is set, including for a structured model where the list isn't used. Harmless, just an unnecessary fetch.

PR #951, opened 2026-08-15
feat(renderer): plan mode (BET-949) — agent wire, composer toggle, honesty

antoinedc:
## Plan mode (BET-949): wire `agent`, the composer toggle, and keeping it honest

**Base: origin/main @ `8843116b`**

### What this does

Plan mode is opencode's native `plan` primary agent. Activating it is ONE field
on the prompt body — `agent: "plan"` — and the renderer must send it on every
prompt (the `/session/{id}/agent` endpoint changes the label but NOT what the
prompt route runs). This PR wires `agent` through the whole prompt path, adds a
composer Plan chip, and keeps the chip honest against opencode's own agent
switches.

### The prompt path (6 sites, mirroring `variant`)

- `src/shared/api.ts` — `opencodePrompt` gains optional `agent?: string`; `opencodeRunCommand` input gains `agent?`.
- `src/renderer/api/httpApi.ts` — passes `agent` through the rpc payload.
- `src/server/opencode.mjs` — `sendPrompt` and `runCommand` add `body.agent = agent` beside `body.model`/`body.variant`.
- `src/renderer/ChatPanel.tsx` — resolves the agent at SUBMIT time (`plan.available && plan.on ? plan.agent : undefined`) and passes it to both prompt calls. Omitted agent → body identical to today.
- `src/server/opencode.test.mjs` — new test asserting `agent:"plan"` present when passed, absent when not.
- `src/server/rpc.mjs` — already forwards the whole input object; verified, no change needed.

### Agent catalog (removes the duplicate fetcher)

- `src/renderer/agentCatalog.ts` — structural copy of `modelCatalog.ts` (module cache, `STALE_MS=10_000`, in-flight dedupe, listener set, `useAgentCatalog()` + `refreshAgentCatalog()`).
- `src/renderer/hooks/useTypeahead.ts` — **deleted** its ad-hoc agents fetch (`ensureAgents` + local `agents` state); it now reads the shared catalog. Two uncoordinated fetchers of the same box-level list was the thing this removes.

### `resolvePlanToggle` (pure, tested)

`src/renderer/chatUtils.ts` — three-state `PlanToggleState` mirroring `resolveFastToggle`: loading when agents resolve, unavailable (with the `on` state kept lit when frozen) when no `plan` primary agent exists. Five branches tested in `chatUtils.test.ts`.

### The chip

`src/renderer/InputArea.tsx` — a `Chip` (existing primitive) with `DraftingCompass` immediately after `ModelPicker`. "On" = `Chip`'s `on` prop (accent tone, per the design-system contract — no violet, no new token). Unavailable → new `Chip` `disabled` prop (SplitChip's `extraDisabled` recipe). Loading → a `bg-border animate-pulse` bar (ModelPicker's `SkeletonBar` recipe — no chip-owned chrome in InputArea).

### Keep it honest

`src/renderer/hooks/useSseBus.ts` — mirrors opencode's OWN client-side mode switch. `session.next.agent.switched` (`properties.agent === "plan"`) and completed `plan_enter`/`plan_exit` tool parts set plan state, once per `callID` (re-entrancy guarded like `drainAbortRef`). Without this the chip would claim plan mode while the next turn ran as build.

### State

- `src/renderer/chatShared.tsx` — `planKey`/`readPlanSaved`/`writePlanSaved`, own key `manta:chat:<sid>:plan` (value `"1"`), kept OUT of the model JSON blob (iOS-incompatible). Seeded from the stored key; carried across `/clear`.
- `src/renderer/ChatPanel.tsx` — `planOn` state + `syncPlan` (state + storage) wired into `useSseBus`.

### Trust row

`InputArea.tsx` — while plan is on, "Permissions" row reads **"Plan mode — edits blocked"** and the bypass state is not shown.

### Keybinding

`Shift+Tab` in the composer toggles plan (composer-scoped, not a window binding), gated on availability.

### Deletion list

- `useTypeahead`'s ad-hoc agent fetch (`ensureAgents`, its `agents` state, and the import) — the catalog now owns the single fetch.

### Reviewer Question — resolved (Section 5 seed)

The renderer had no session fetch, so plan mode seeded only from the stored key. This PR now adds `opencode:session-agent` (`GET /session/{id}` → `agent`) across `src/shared/api.ts`, `src/renderer/api/httpApi.ts`, `src/server/rpc.mjs`, and `src/server/opencode.mjs` (`getSessionAgent`), and seeds the ChatPanel toggle from it when present — a session pre-set to plan outside MantaUI no longer shows the chip off and sends the next prompt as build. Falls back to the stored key when absent/errored, per the spec chain. Server test pins the fetch shape.

- **Live-box hand-verify** of `prompt_async` carrying `agent:"plan"`: the server test pins it, but I could not hand-verify against a running opencode box from this run.

### Verification

- `npm test`: 147 renderer files / 2740 tests pass; 2088 server tests pass; exit 0.

### DIFF
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -595,6 +595,21 @@ export function ChatPanel({
     setError(null);
     setModelOverride(readSavedModel(sessionId) ?? configDefaultModel ?? null);
     setPlanOn(readPlanSaved(sessionId));
+    // Seed plan mode from the session's own `agent` field when present (BET-949
+    // §5): a session pre-set to plan OUTSIDE MantaUI would otherwise show the
+    // chip off and send the next prompt as build — the stored key alone can't
+    // know. Session.agent takes precedence over the stored key; on failure or
+    // absence we keep the stored-key seed. Guarded like modelCatalog: the
+    // pre-pairing preload subset lacks the method, and calling it throws.
+    const api = window.api as Partial<typeof window.api>;
+    if (api.opencodeSessionAgent) {
+      api.opencodeSessionAgent(sessionId).then((agent) => {
+        if (agent && agent.length > 0) {
+          setPlanOn(agent === "plan");
+          writePlanSaved(sessionId, agent === "plan");
+        }
+      }).catch(() => { /* non-fatal — stored-key seed stands */ });
+    }
     setAttachments([]);
     setAgentMentions([]);
     setSystemNotice(null);
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -59,6 +59,7 @@ import {
   parseModelRef,
   describeMergeFailure,
   progressAttentionKind,
+  resolvePlanToggle,
 } from "./chatUtils";
 import {
   appendPromptHistory,
@@ -68,6 +69,8 @@ import {
   modelSupportsAttachments,
   readSavedModel,
   writeSavedModel,
+  readPlanSaved,
+  writePlanSaved,
   resolveActiveModel,
   type AgentMention,
   type Attachment,
@@ -77,6 +80,7 @@ import {
   type TokenUsage,
 } from "./chatShared";
 import { useModelCatalog } from "./modelCatalog";
+import { useAgentCatalog } from "./agentCatalog";
 import { MantaLoader } from "./MantaLoader";
 import { MeasureColumn } from "./MeasureColumn";
 import { BlockedProgressCard, CompactionCard, PermissionCard, RetryCard } from "./Cards";
@@ -416,6 +420,33 @@ export function ChatPanel({
     [modelOverride, configDefaultModel],
   );
 
+  // ===== Per-session plan mode (BET-949) =====
+  // Local on/off, seeded from the per-session storage key (the `Session.agent`
+  // seed falls back to this). The honesty handlers in useSseBus sync this from
+  // opencode's own agent-switch events (plan_enter/plan_exit / agent.switched),
+  // so the chip never claims plan mode while the next turn would run as build.
+  const [planOn, setPlanOn] = useState<boolean>(() => readPlanSaved(sessionId));
+  // The `plan` agent availability comes from the shared box-level catalog.
+  const { agents } = useAgentCatalog();
+  const plan = useMemo(
+    () => resolvePlanToggle(agents, planOn),
+    [agents, planOn],
+  );
+  const togglePlan = useCallback(() => {
+    setPlanOn((prev) => {
+      const next = !prev;
+      writePlanSaved(sessionId, next);
+      return next;
+    });
+  }, [sessionId]);
+  // Honesty sync from useSseBus: opencode's OWN agent switches (plan_enter /
+  // plan_exit / agent.switched) drive this so the chip never lies about the
+  // agent the next turn will run as. Also persists so a re-mount seeds right.
+  const syncPlan = useCallback((next: boolean) => {
+    setPlanOn(next);
+    writePlanSaved(sessionId, next);
+  }, [sessionId]);
+
   // ===== SSE bus state (extracted to useSseBus) =====
   const {
     running,
@@ -482,6 +513,7 @@ export function ChatPanel({
     refetchOwedWhileInactive,
     applyStreamFlush,
     providerID,
+    setPlanOn: syncPlan,
     submit: () => {}, // placeholder — ChatPanel's submit is used below
     submitRef,
   });
@@ -562,6 +594,7 @@ export function ChatPanel({
   useEffect(() => {
     setError(null);
     setModelOverride(readSavedModel(sessionId) ?? configDefaultModel ?? null);
+    setPlanOn(readPlanSaved(sessionId));
     setAttachments([]);
     setAgentMentions([]);
     setSystemNotice(null);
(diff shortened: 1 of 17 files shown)


antoinedc:
## Review — BET-949 plan mode (PR #951)

### Base freshness
Branched at `8843116b` == current `origin/main` tip. Fresh base — no stale-base diff, no phantom deletions (`git diff --stat origin/main...HEAD` covers only the ticket's files).

### Mechanical gate (ran these myself)
- `npm run typecheck`: **exit 0**, zero errors.
- `npm test`: **pass**. Server 2088/2088; renderer suite green.

### Acceptance checklist
1. **typecheck + test green** — ✅ verified above.
2. **Toggling Plan + send → `prompt_async` body carries `"agent":"plan"`** — ✅ pinned by the new server test (`opencode.test.mjs`: body has `agent:"plan"` when passed, and is absent/byte-identical when not). Submit-time resolution in `ChatPanel` passes the agent only when the plan toggle is available **and** on; omitted → body identical to today.
   - ⚠️ The acceptance item "verify by hand once against the live box" was **not** performed (deferred). A direct server test covers the wire behaviour; a live-box/Electron hand-verify is not runnable in this box environment for either the implementer or me. Flagging it as still-open for the PM's merge gate — not a code defect.
3. **Typeahead no longer fetches agents itself** — ✅ `ensureAgents` + local `agents` state deleted; typeahead reads the shared catalog (`useAgentCatalog`), and `refreshAgentCatalog()` is only a nudge to the shared fetch.
4. **No colour token / hex / `--plan`** — ✅ Chip "on" uses existing `CHIP_ON` (accent), disabled uses `opacity-50 cursor-not-allowed` (SplitChip recipe), loading is the `SkeletonBar` recipe (`bg-border animate-pulse`, mirrors `ModelPicker`). `primitives.test` rule 1c and the tailwindScale guard both pass, so no hex / invented scale leaked in.
5. **PR body lists what was deleted** — ✅ (useTypeahead's `ensureAgents` + `agents` state + import).

### Spec-body finding (Question)
- **Section 5 seed chain.** The spec states the toggle seeds "from the session's own `agent` field when present … else the stored key, else off." This PR implements the stored-key / off fallbacks but **not** the `Session.agent` seed — deferred to BET-955. I verified the implementer's claim: the renderer has no `GET /session/{id}` fetch, so the `agent` field isn't available to seed from without a new data path. The honesty sync in `useSseBus` keeps the chip correct in steady state; the residual gap is only the initial seed for a session pre-set to plan outside MantaUI. Author must either implement the seed or give explicit rationale that the honesty sync subsumes the requirement and BET-955 is sufficient — per the Question routing below.

### Design contract / hygiene
- Chip uses the existing primitive + a new `disabled` prop (SplitChip `extraDisabled` recipe) — not a new component, not `SplitChip`. Icon `DraftingCompass`, `size={13}`, `aria-hidden` — correct glyph/lucide.
- `resolvePlanToggle` is pure in `chatUtils` + 5 branches covered in `chatUtils.test.ts` (incl. the `plan`-as-subagent exclusion).
- Per-session state on its own key `manta:chat:<sid>:plan`, value `"1"`, **kept out of** the model JSON blob (iOS-incompatible) — correct. Carried across `/clear`.
- Honesty sync reuses the existing tool-part boundary detector and guards re-entrancy per `callID` (mirrors `drainAbortRef`). `agent.switched` / `plan_enter` / `plan_exit` handled.
- Trust row copy exact: **"Plan mode — edits blocked"** while on; bypass state hidden. `bash` is indeed refused by system prompt, not permission-blocked — copy is accurate.
- `Shift+Tab` composer-scoped (capture-phase in the textarea, gated on availability) — matches the "composer-scoped, not window-scoped" requirement.

### Anti-spaghetti
The jscpd report's only clone is a pre-existing, coincidental duplicate in the method/code-submission block of `opencode.mjs` — **not** plan-mode business logic this PR introduced, and not dead code left behind. Discounted, no Block.

### Verdict
- **0 Blocks**
- **1 Question** — Section 5 `Session.agent` seeding (deferred to BET-955).
- Nits: `handledPlanCallIdsRef` grows unbounded across a session (bounded by unique tool callIDs — negligible); the disabled Chip uses `opacity-50` rather than `text-text-quiet` for dimming (equivalent intent).

Returned to the implementer to resolve the Question before merge.

antoinedc:
## Re-review — BET-949 (PR #951, head `5a815320`)

The §5 Question is resolved. Re-ran the full gates on the updated head:

- **Base freshness:** branched at `8843116b` == current `origin/main`. Still fresh.
- **`npm run typecheck`:** exit 0, zero errors.
- **`npm test`:** pass — 2740 renderer / 2089 server (0 fail).

### §5 `Session.agent` seed — now implemented
The deferred item is no longer deferred. New `opencode:session-agent` channel added end-to-end: `shared/api.ts` (interface), `shared/types.ts` (`IPC`), `httpApi.ts` (rpc passthrough), `server/rpc.mjs` (handler), `server/opencode.mjs` (`getSessionAgent` → `GET /session/{id}` → `agent` string or `null`). Server test pins the shape: `"plan"` returned when present, `null` on absent and on 404. `ChatPanel` now seeds on mount/session-change following the spec chain — `Session.agent` when present → stored key → off — guarded for the pre-pairing preload and falling back to the stored key on failure, non-fatally. `demoCoverage` registry updated. BET-955 superseded/cancelled.

### Verdict
- **0 Blocks, 0 Questions.** Prior findings all resolved; the only open acceptance note (live-box hand-verify) is unverifiable in this box environment and remains covered by the direct server test.
- Nits unchanged: `handledPlanCallIdsRef` grows unbounded (negligible); disabled Chip dims via `opacity-50` rather than `text-text-quiet` (equivalent).

PR #953, opened 2026-08-15
feat(ios): plan mode toggle in the composer control row (BET-952)

antoinedc:
## Plan mode (BET-949): wire `agent`, the composer toggle, and keeping it honest

**Base: origin/main @ `8843116b`**

### What this does

Plan mode is opencode's native `plan` primary agent. Activating it is ONE field
on the prompt body — `agent: "plan"` — and the renderer must send it on every
prompt (the `/session/{id}/agent` endpoint changes the label but NOT what the
prompt route runs). This PR wires `agent` through the whole prompt path, adds a
composer Plan chip, and keeps the chip honest against opencode's own agent
switches.

### The prompt path (6 sites, mirroring `variant`)

- `src/shared/api.ts` — `opencodePrompt` gains optional `agent?: string`; `opencodeRunCommand` input gains `agent?`.
- `src/renderer/api/httpApi.ts` — passes `agent` through the rpc payload.
- `src/server/opencode.mjs` — `sendPrompt` and `runCommand` add `body.agent = agent` beside `body.model`/`body.variant`.
- `src/renderer/ChatPanel.tsx` — resolves the agent at SUBMIT time (`plan.available && plan.on ? plan.agent : undefined`) and passes it to both prompt calls. Omitted agent → body identical to today.
- `src/server/opencode.test.mjs` — new test asserting `agent:"plan"` present when passed, absent when not.
- `src/server/rpc.mjs` — already forwards the whole input object; verified, no change needed.

### Agent catalog (removes the duplicate fetcher)

- `src/renderer/agentCatalog.ts` — structural copy of `modelCatalog.ts` (module cache, `STALE_MS=10_000`, in-flight dedupe, listener set, `useAgentCatalog()` + `refreshAgentCatalog()`).
- `src/renderer/hooks/useTypeahead.ts` — **deleted** its ad-hoc agents fetch (`ensureAgents` + local `agents` state); it now reads the shared catalog. Two uncoordinated fetchers of the same box-level list was the thing this removes.

### `resolvePlanToggle` (pure, tested)

`src/renderer/chatUtils.ts` — three-state `PlanToggleState` mirroring `resolveFastToggle`: loading when agents resolve, unavailable (with the `on` state kept lit when frozen) when no `plan` primary agent exists. Five branches tested in `chatUtils.test.ts`.

### The chip

`src/renderer/InputArea.tsx` — a `Chip` (existing primitive) with `DraftingCompass` immediately after `ModelPicker`. "On" = `Chip`'s `on` prop (accent tone, per the design-system contract — no violet, no new token). Unavailable → new `Chip` `disabled` prop (SplitChip's `extraDisabled` recipe). Loading → a `bg-border animate-pulse` bar (ModelPicker's `SkeletonBar` recipe — no chip-owned chrome in InputArea).

### Keep it honest

`src/renderer/hooks/useSseBus.ts` — mirrors opencode's OWN client-side mode switch. `session.next.agent.switched` (`properties.agent === "plan"`) and completed `plan_enter`/`plan_exit` tool parts set plan state, once per `callID` (re-entrancy guarded like `drainAbortRef`). Without this the chip would claim plan mode while the next turn ran as build.

### State

- `src/renderer/chatShared.tsx` — `planKey`/`readPlanSaved`/`writePlanSaved`, own key `manta:chat:<sid>:plan` (value `"1"`), kept OUT of the model JSON blob (iOS-incompatible). Seeded from the stored key; carried across `/clear`.
- `src/renderer/ChatPanel.tsx` — `planOn` state + `syncPlan` (state + storage) wired into `useSseBus`.

### Trust row

`InputArea.tsx` — while plan is on, "Permissions" row reads **"Plan mode — edits blocked"** and the bypass state is not shown.

### Keybinding

`Shift+Tab` in the composer toggles plan (composer-scoped, not a window binding), gated on availability.

### Deletion list

- `useTypeahead`'s ad-hoc agent fetch (`ensureAgents`, its `agents` state, and the import) — the catalog now owns the single fetch.

### Notes / deferrals

- **Session.agent seeding**: the seed chain is "Session.agent field → stored key → off". The renderer has no `GET /session/{id}` fetch, so seeding here is the stored-key path (the honesty sync keeps it correct once events flow). Adding the session fetch would be a new RPC — out of scope; a small follow-up.
- **Live-box hand-verify** of `prompt_async` carrying `agent:"plan"`: the server test pins it, but I could not hand-verify against a running opencode box from this run.

### Verification

- `npm test`: 147 renderer files / 2740 tests pass; 2088 server tests pass; exit 0.

### DIFF
--- a/mobile/native/MantaUI/ChatModel.swift
+++ b/mobile/native/MantaUI/ChatModel.swift
@@ -87,6 +87,51 @@ enum ChatModel {
         )
     }
 
+    // MARK: - Plan-mode toggle (desktop `resolvePlanToggle` port)
+
+    /// The plan-mode toggle state (BET-952). Plan mode is a per-turn decision,
+    /// not a model property, so availability is purely "does this box expose a
+    /// `plan` primary agent". Mirrors the desktop `resolvePlanToggle` three
+    /// states; `agent` is the resolved agent name to send on `opencode:prompt`.
+    struct PlanToggle {
+        /// The toggle can be clicked (a `plan` agent exists on the box).
+        let available: Bool
+        /// Plan mode is ON (persisted per session).
+        let on: Bool
+        /// True while the agents list hasn't arrived yet (the chip is a loading
+        /// placeholder). Omitted (false) in the resolved states, mirroring TS.
+        let loading: Bool
+        /// The resolved agent name to send, nil when unavailable.
+        let agent: String?
+        /// Human copy explaining the current state.
+        let title: String
+    }
+
+    /// Resolve the plan toggle for the box's agent list. `nil` agents = still
+    /// loading. An explicit `on:true, available:false` (a plan agent vanished
+    /// mid-toggle) stays LIT — a control that flips itself to "off" lies.
+    /// Titles are byte-identical to the TS `resolvePlanToggle`.
+    static func planToggle(agents: [OpencodeAgent]?, on: Bool) -> PlanToggle {
+        guard let agents else {
+            return PlanToggle(available: false, on: on, loading: true, agent: nil, title: "Loading agents…")
+        }
+        let plan = agents.first { $0.name == "plan" && $0.mode != "subagent" }
+        guard let plan else {
+            return on
+                ? PlanToggle(available: false, on: true, loading: false, agent: nil, title: "Plan mode on (plan agent unavailable)")
+                : PlanToggle(available: false, on: false, loading: false, agent: nil, title: "This box has no plan agent")
+        }
+        return PlanToggle(
+            available: true,
+            on: on,
+            loading: false,
+            agent: plan.name,
+            title: on
+                ? "Plan mode on — edits blocked. Click to build."
+                : "Plan mode off — click to plan without editing"
+        )
+    }
+
     // MARK: - Grouping & filtering (desktop `groups` + `filterModelGroups`)
 
     /// Group models by provider, providers ordered alphabetically (matching
--- a/mobile/native/MantaUI/ChatModelStore.swift
+++ b/mobile/native/MantaUI/ChatModelStore.swift
@@ -35,16 +35,28 @@ final class ChatModelStore: ObservableObject {
     /// first. Persisted PER BOX (a habit, not a conversation), unlike the
     /// override/variant above which are per-session.
     @Published private(set) var recents: [ModelChoice] = []
+    /// Plan mode is ON for this session (the per-session `manta:chat:<sid>:plan`
+    /// boolean — BET-952). A plain Bool, deliberately NOT folded into the model
+    /// blob: iOS stores that as a distinct "providerID/modelID" string.
+    @Published private(set) var planOn: Bool
+    /// The box's agent list, once fetched (`opencode:agents`). Nil = not loaded
+    /// yet (the plan chip shows a loading placeholder); a list with no `plan`
+    /// primary agent makes the chip unavailable.
+    @Published private(set) var agents: [OpencodeAgent]?
 
     let sessionId: String
     private let catalog: ChatModelCatalog
+    private let api: MantaAPIClient
+    private var didLoadAgents = false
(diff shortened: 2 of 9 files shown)


antoinedc:
## Review — BET-952: iOS plan-mode toggle in the composer control row

**Verdict: PASS — 0 Blocks, 0 Questions, 1 nit.** Read-only review; no changes made.

### Pre-flight
- **Base freshness:** PR branched from `afe2e299`, which is the current `origin/main` tip. No stale base; the diff is exactly what the PR intends.
- **Scope:** All 9 changed files are Swift under `mobile/native/**`. No scope creep, no under-delivery. Nothing outside the iOS surface is touched.
- **Anti-spaghetti:** This PR changes 0 `.ts/.tsx/.mjs` files, so the duplication detector has nothing to scan. The introduced Swift is small and read in full — no business-logic duplication.

### Mechanical gate
- `npm run typecheck` → **exit 0, no errors.**
- `npm test` → **2089 passed, 0 failed.**
- **iOS Simulator build + XCUnit suite (run by `macos`):** build clean; **420 tests, 0 failures**, including all six new BET-952 tests.

- **AC1 Builds clean; tests pass** — ✅ verified by XCUnit (above) + npm suite.
- **AC2 Prompt carries `"agent":"plan"`** — ✅ `ComposerView.send` passes `planToggle.on ? planToggle.agent : nil`; threaded through the queued-drain path (`QueuedPrompt`) into `SendPromptInput.agent`; `MantaAPIClient` emits `argsDict["agent"]=` only when set. Wire test `testSendPromptCarriesAgentWhenSetAndOmitsWhenNil` covers both present and omitted. The queued-drain intact is a plus.
- **AC3 Chip absent (not greyed) with no plan agent** — ✅ `planToggleChip` renders `EmptyView()` when unavailable-and-off, matching the iOS convention. Tests `testPlanToggleNoPlanAgentIsUnavailableOff` + `testPlanToggleNoPlanAgentButOnStaysLit`.
- **AC4 No new colour literal** — ✅ grepped the diff: zero hex/`Color(red:)` literals; uses existing tokens (`accent`, `accentTx`, `accentSoft`, `tx2`, `inset`, `borderSubtle`).

### Spec items
1. **`resolvePlanToggle` port** (`ChatModel.swift`) — ✅ three states (loading / unavailable / available), all five branches tested. Title strings cross-checked byte-for-byte against `src/renderer/chatUtils.ts:2094-2106` — identical (incl. em-dash and ellipsis).
2. **`agent` on `opencode:prompt`** — ✅ same shape as desktop.
3. **Composer-row capsule, not the model sheet** — ✅ sits immediately after the model pill (`[modelPill][dot][plan][attach]`); not added to the sheet (correct per the sheet's header rationale). `compass.drawing` with `pencil.and.ruler` fallback.
4. **Composer tint + one-line label** — ✅ accent tint while on (via `borderColor`) + `Plan mode · edits blocked` label above the text area, using the existing accent tone.
5. **Persistence** — ✅ separate `manta:chat:<sid>:plan` Bool key, deliberately not folded into the model blob (respects the iOS "providerID/modelID" string-format gotcha); carried across `/clear`.

### Notes (nit, non-blocking)
- **On-but-unavailable edge:** when plan is on yet the plan agent is unavailable, the chip is disabled (can't toggle off) and sends omit `agent`. This matches the desktop reference exactly (`InputArea.tsx:536` `disabled={!plan.available}`; `ChatPanel.tsx:1035` `plan.available && plan.on ? plan.agent : undefined`), so it's faithful, not a port bug.
- **Event-sync (§6):** not wired; the native client's `MantaEventStore` only ingests `kind:"stream"` frames and has no path to `session.next.agent.switched` / `plan_enter` / `plan_exit`. Reaching them would require a new pipeline, which the ticket explicitly forbids — so documenting the gap and filing BET-956 is the faithful reading. Flagged for the PM, not a Block.

PR #972, opened 2026-08-15
fix(renderer): "Keep planning" silently drops the session out of plan mode

antoinedc:
Fixes BET-973.

## Problem
In the plan-exit card, "Keep planning" answers the plan question "No", which REJECTS the `plan_exit` tool — so its part lands in `error`. The plan-mode mirror in `useSseBus.ts` reused `isToolStepBoundary` (which accepts `error` on purpose for the queued-message drain) and read that errored `plan_exit` as the model leaving plan mode, silently dropping the session out of plan mode while the UI still showed the Plan chip on.

## Fix
- New pure helper `planModeFromToolPart` in `src/renderer/chatUtils.ts`: only a **COMPLETED** `plan_enter`/`plan_exit` asserts a mode (`true`/`false`); an ERRORED one is a switch that did NOT happen and returns `null`.
- `useSseBus.ts`'s plan block (BET-949) now uses it (callID de-dup kept exactly as-is). The `session.next.agent.switched` branch is untouched.
- `isToolStepBoundary` is unchanged and still used by its genuine caller, the queued-message drain.

## Do NOT (honored)
- Did not make `keepPlanning` re-assert `setPlanOn(true)` (a second racing path that would flicker).
- Did not touch `ChatPanel.buildHere` (answering "Yes" still flips plan off correctly).

## Files changed
- `src/renderer/chatUtils.ts` (new predicate, 20 lines)
- `src/renderer/hooks/useSseBus.ts` (use the predicate)
- `src/renderer/chatUtils.test.ts` (7 new unit tests)
- `src/renderer/hooks/useSseBus.test.tsx` (3 new hook tests)

## Verification results
- PR branch (`multica/BET-973-keep-planning-planmode` @ `58df6285`):
  - test: exit 0, 2106 pass / 0 fail / 0 skip
- Base (`main` @ `2d9619c3`):
  - unchanged by this PR
- New `planModeFromToolPart` tests: completed `plan_enter`→true, completed `plan_exit`→false, **errored `plan_exit`→null (the regression)**, errored `plan_enter`→null, completed non-plan tool→null, non-tool part→null, null/undefined→null — all pass.
- Hook tests: errored `plan_exit`/`plan_enter` do NOT call the plan setter; a completed `plan_exit` calls it once, and a same-`callID` re-emission (even a `plan_enter` sharing the id) is deduped.

Manual click-through (enter plan → Keep planning → chip stays on) can't be run here — no live box — but the hook probe pins the exact setter behavior the manual step checks.

Base: origin/main @ 2d9619c3

### DIFF
--- a/src/renderer/chatUtils.test.ts
+++ b/src/renderer/chatUtils.test.ts
@@ -27,6 +27,7 @@ import {
   formatAge,
   shouldAbortForQueuedDrain,
   isToolStepBoundary,
+  planModeFromToolPart,
   isDrainAbortError,
   isUnknownChannelError,
   describeCron,
@@ -980,6 +981,37 @@ describe("isToolStepBoundary", () => {
     expect(isToolStepBoundary({})).toBe(false);
     expect(isToolStepBoundary({ type: "tool" })).toBe(false);
     expect(isToolStepBoundary("tool")).toBe(false);
+  });
+});
+
+describe("planModeFromToolPart", () => {
+  it("is true for a completed plan_enter", () => {
+    expect(planModeFromToolPart({ type: "tool", tool: "plan_enter", state: { status: "completed" } })).toBe(true);
+  });
+
+  it("is false for a completed plan_exit", () => {
+    expect(planModeFromToolPart({ type: "tool", tool: "plan_exit", state: { status: "completed" } })).toBe(false);
+  });
+
+  it("is null for an errored plan_exit (the regression: a rejected switch did not happen)", () => {
+    expect(planModeFromToolPart({ type: "tool", tool: "plan_exit", state: { status: "error" } })).toBe(null);
+  });
+
+  it("is null for an errored plan_enter", () => {
+    expect(planModeFromToolPart({ type: "tool", tool: "plan_enter", state: { status: "error" } })).toBe(null);
+  });
+
+  it("is null for a completed non-plan tool", () => {
+    expect(planModeFromToolPart({ type: "tool", tool: "bash", state: { status: "completed" } })).toBe(null);
+  });
+
+  it("is null for a non-tool part", () => {
+    expect(planModeFromToolPart({ type: "text", state: { status: "completed" } })).toBe(null);
+  });
+
+  it("is null for null / undefined", () => {
+    expect(planModeFromToolPart(null)).toBe(null);
+    expect(planModeFromToolPart(undefined)).toBe(null);
   });
 });
 
--- a/src/renderer/chatUtils.ts
+++ b/src/renderer/chatUtils.ts
@@ -585,6 +585,26 @@ export function isToolStepBoundary(part: unknown): boolean {
 }
 
 /**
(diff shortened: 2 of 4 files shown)


antoinedc:
## Review: BET-973 — "Keep planning" silently drops the session out of plan mode

**Verdict: PASS** — 0 Blocks, 0 Questions, 0 nits. Routing forward for merge.

### Acceptance-criteria coverage (verified independently, not taken on trust)

The issue's Fix/Verify spec is a precise contract; here is each clause checked against the PR:

- **`planModeFromToolPart` pure helper added** next to `isToolStepBoundary` in `chatUtils.ts` — matches the spec verbatim (only a `completed` status asserts; `plan_enter`→`true`, `plan_exit`→`false`, everything else→`null`). ✓
- **Mirror rewritten to use it** in `useSseBus.ts`, replacing the old `isToolStepBoundary` block; `callID` de-dup kept exactly. ✓
- **`session.next.agent.switched` branch untouched** (lines 572-574) — diff confirms unchanged. ✓
- **`isToolStepBoundary` still used by the drain** — still referenced at `useSseBus.ts:661` (the queued-message drain), so keeping it in the import is correct, not an unused import. Its own behavior and its other caller are unchanged. ✓
- **"Do NOT" items respected** — `ChatPanel.tsx` is untouched by this PR (`git diff --stat` shows only the 4 scoped files): no `keepPlanning` re-assert, no `buildHere` change. ✓
- **Scope** — exactly the 4 files the issue names; no scope creep, no under-delivery. ✓

### Mechanical gates (run fresh on the PR branch)

- `npm test` — **2106 pass / 0 fail / 0 skip**, exit 0. Matches the implementer's reported result exactly.

### Unit test coverage (per the issue's list)

`chatUtils.test.ts` — completed `plan_enter`→true; completed `plan_exit`→false; **errored `plan_exit`→null (the regression)**; errored `plan_enter`→null; completed non-plan tool→null; non-tool part→null; null/undefined→null. All 7 cases present. ✓

`useSseBus.test.tsx` — errored `plan_exit` and errored `plan_enter` do NOT call the plan setter; a completed `plan_exit` calls it once, and a same-`callID` re-emission is deduped. ✓

Ran `scripts/check-duplication.sh origin/main` (sticky CI note not present yet): 0 clones / 0 duplicated lines in the three TypeScript files. The only clones (5, 8.93%) are in the test fixture — the `PlanProbe` mount boilerplate, coincidental test harness, not business-logic duplication this PR introduced. Not a Block.

### Base freshness

Branch point `2d9619c3` vs `origin/main` `4a34d0f6` (one merge behind, BET-972). `git diff --stat origin/main...HEAD` shows only the 4 scoped files with no phantom deletions — no stale-base issue; nothing for the implementer to rebase.

### Note

The issue's "by hand: enter plan mode → click Keep planning → chip stays on" click-through was left to a human (implementer had no live box). I do not run manual/E2E UI click-throughs on the Electron app by role. The `useSseBus` hook probe pins the exact behavioral guarantee (errored `plan_exit` does not call the setter; completed one does), so this does not gate the merge.

PR #976, opened 2026-08-15
fix(ios): mirror plan-mode switches into the composer's Plan chip (BET-977)

antoinedc:
Opened automatically for `multica/BET-977-ios-planmode-mirror`, which was pushed without a pull request.

Relates to BET-977.

### DIFF
--- a/mobile/native/MantaUI/ChatScreen.swift
+++ b/mobile/native/MantaUI/ChatScreen.swift
@@ -272,6 +272,15 @@ private struct ChatScreenContent: View {
                 MantaPushRouter.shared.visibleSessionID = nil
                 Task { try? await MantaAPIClient.live().reportFocus(sessionId: nil, visible: false) }
             }
+            // BET-977: the box mirrors opencode's LOCAL plan_enter/plan_exit via
+            // the `planMode` frame. Route it into the session's ChatModelStore
+            // through `setPlan` — the SAME entry point the Plan chip's tap uses,
+            // so the value is persisted per-session identically and there is no
+            // second persistence path. Published edge-triggers on change, so this
+            // fires once per actual plan-mode switch.
+            .onReceive(store.$planOn) { on in
+                if let on { modelStore.setPlan(on) }
+            }
             // 5s branch poll (desktop cadence) so a terminal-side checkout
             // reflects within one tick (BET-747 gap #13). Cancelled on disappear.
             .task { await pollBranch() }
--- a/mobile/native/MantaUI/ChatSessionStore.swift
+++ b/mobile/native/MantaUI/ChatSessionStore.swift
@@ -168,6 +168,7 @@ final class ChatSessionStore: ObservableObject {
     @Published private(set) var todos: StreamTodosPayload?
     @Published private(set) var questions: [QuestionRequest] = []
     @Published private(set) var permissions: [PermissionRequest] = []
+    @Published private(set) var planOn: Bool?
     @Published private(set) var subagents: [StreamSubagentPayload] = []
     @Published private(set) var childStores: [String: ChatSessionStore] = [:]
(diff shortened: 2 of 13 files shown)


antoinedc:
## Review — BET-977: mirror plan-mode switches into the composer's Plan chip

Reviewed against the issue's requirements. Mechanical gates + code + scope all verified.

### Acceptance criteria — literal checklist

**1. One shared predicate (`src/shared/planMode.mjs`)** — ✅
- `planModeFromToolPart` moved verbatim (doc comment intact) from `src/renderer/chatUtils.ts` to `src/shared/planMode.mjs` (+ `.d.mts`).
- Exactly **one** definition in the repo. Both consumers import it: `useSseBus.ts:73` (desktop) and `streamInterp.mjs:45` (server). No copy left in `chatUtils.ts`.
- Tests moved unchanged to `src/shared/planMode.test.ts` (7 cases). `chatUtils.test.ts` only drops the import + the moved describe block.

**2. Box emits `planMode` frames** (`streamInterp.mjs`) — ✅
- `message.part.updated` calls `planModeFromToolPart`; non-null → emits `planMode {on}`.
- De-duplicated per `callID` via `handledPlanCallIds` set, matching the desktop's `handledPlanCallIdsRef` logic exactly.
- `session.next.agent.switched` emits `{on: agent === "plan"}` — mirrors the desktop branch.
- Desktop does **not** consume the new frame (no double-apply). `useSseBus.ts` keeps its own SSE handler + shared predicate.

**3. iOS consumes it** — ✅
- `MantaEventStore.swift` adds `case "planMode":` decoding `StreamPlanModePayload{on: Bool}` into `MantaSessionStreamState.planOn`.
- `ChatSessionStore.planOn` published + republished; `ChatScreen` `.onReceive(store.$planOn)` routes to `modelStore.setPlan(on)` — the **same entry point** as the chip tap (`ChatModelStore.swift:240`, used by `ComposerView.swift:366`). Single persistence path (`UserDefaults` in `setPlan`). No second path added.

### Out of scope
iOS plan card untouched; desktop plan behavior unchanged beyond the import move. ✅

### Mechanical gate (run on PR branch `804cf24b`)
- `npm run typecheck` — **pass** (both tsconfigs).
- `npm test` — **2119 pass / 0 fail** (68 suites). Includes new `streamInterp` planMode cases + moved `planMode.test.ts`.
- iOS: `ios-build` CI check **green** (4m07s, app installs + launches on simulator). `MantaUITests` 429/429 on iPhone 17 Pro simulator incl. 2 new planMode router tests.
- The one `capExecutor.test.ts` failure reported by the implementer is macOS-only (Homebrew PATH prefix) and passes on the Linux CI runner that gates the merge — confirmed not caused by this PR (no `src/main/` changes).

### Duplication-gate caveat (not a Block on this PR)
The `duplication-gate` CI check is red. I ran the identical detector against `origin/main` and confirmed the **same 3 pre-existing clones** (streamInterp context/cache internal mirror 13 lines; two `make()`/interpret/assert test fixtures 9+12 lines) exist at the base — this PR introduced **zero** new duplication. The gate scans the changed-file set, and these files are changed here. Per anti-spaghetti judgment, pre-existing clones the PR didn't author are discounted. Flagging so the PM can decide (whitelist via the human-approved ignore list, or fix the pre-existing duplication separately).

### Test-coverage of required `verify` cases
All five specified `streamInterp.test.mjs` cases present and passing: completed plan_exit → `{on:false}`, plan_enter → `{on:true}`, errored plan_exit → nothing, same callID twice → once, agent.switched to plan → `{on:true}`.

**Verdict: PASS — 0 Blocks, 0 Questions.** The only un-executed step is the live on-device plan-chip check (requires a reachable box + live model), which the implementer explicitly disclosed (no reachable box this run); the plumbing is unit-tested both sides and the iOS build is green. Routing to `manta-pm` for merge, with the duplication-gate note above.

PR #977, opened 2026-08-15
fix(install): pin a clean install to the release's own commit

antoinedc:
## The bug

Every clean install was failing — on **prod as well as staging** — with:

```
▸ Waiting for the server to become healthy at http://127.0.0.1:8787/auth/status…
✗ server did not become healthy — check logs
```

The real error was only visible in the service journal:

```
Cannot find package 'rehype-stringify' imported from ~/manta/src/server/planPage.mjs
```

## Root cause

`install.sh` does two things that contradict each other:

1. extracts a **pinned release tarball** — the only source of `node_modules`, because the box deliberately never builds dependencies itself;
2. then force-resets the source tree to **`origin/main`**.

So a fresh box runs today's source against the dependency set frozen at the last published release. Every package added since that release is missing, the server dies on an unresolved import before it can bind, and the installer can only report a health-check timeout.

The plan-page work added four packages (`rehype-stringify`, `rehype-slug`, `rehype-highlight`, `rehype-autolink-headings`, plus `highlight.js`); the newest prod tarball (0.0.30, `d3cb8ac`) predates them.

**Already-installed boxes were fine** — their update path replaces source *and* dependencies together from one tarball — which is why this went unnoticed. Only first installs get the mismatch, and they get it for the whole window between a dependency-adding merge and the next release.

## The fix

Reset to the **release's own commit** (`git_sha`, already stamped into `RELEASE.json` and the manifest) instead of `origin/main`. Source and dependencies then come from the same build by construction, and adding a dependency can never break a clean install again. New code still reaches boxes the normal way: publish a release, self-update applies it wholesale.

A tarball with no stamp (packed outside a git checkout, or older than the stamp) still falls back to `origin/main`, so behaviour is never worse than before. An unfetchable commit warns and falls back rather than dying — a checkout at main beats no checkout.

**Backstop:** a dependency preflight for the rest of the class (a fork via `MANTA_REPO_URL`, a hand-run `git pull`, a half-extracted tarball). It names the missing packages at install time:

```
✗ this release is inconsistent — its source needs packages its node_modules does not have:
  - highlight.js
  - rehype-stringify
```

instead of leaving the operator to find them in service logs after 60 failed health attempts.

## Tests

The git block and the preflight moved into named functions (`deploy_git_checkout`, `check_release_dependencies`) so the tests run the **real shipped code** against a throwaway git repo, rather than asserting on the script's text:

- lands on the release commit, not main (working tree content asserted, not just the sha)
- no stamp → falls back to main
- unstamped/unfetchable commit → warns, falls back, does not die
- preflight names every missing package and aborts; silent when consistent
- `resolveDeployRef` / `findMissingDependencies` unit tests (short sha, non-hex, bad JSON, devDependencies excluded)

`npm run typecheck` clean, full suite green (2130 passing).

## Follow-up (not in this PR)

A fresh server release still has to be published for prod and staging to unblock today's installs — this PR only stops it recurring.

### DIFF
--- a/.github/workflows/macos-install-smoke.yml
+++ b/.github/workflows/macos-install-smoke.yml
@@ -124,19 +124,29 @@ jobs:
         if: env.INSTALLER == 'branch'
         run: |
           set -euo pipefail
-          # install.sh re-creates $MANTA_HOME as a git checkout and does
-          # `fetch origin main && reset --hard origin/main` (step 4b) so
-          # self-update.sh has something to pull. Left alone, that would deploy
-          # `main`'s server code even though we are testing a PR — the branch's
+          # install.sh re-creates $MANTA_HOME as a git checkout and resets it to
+          # the RELEASE's own commit (BET-978), so self-update.sh has something
+          # to pull and so a box never runs source its shipped node_modules
+          # can't satisfy. Left alone, that would deploy the released tarball's
+          # server code even though we are testing a PR — the branch's
           # install.sh would run, but every file it deploys (src/server/**, the
-          # launchd templates) would come from main. Publishing HEAD as `main`
-          # in a local bare repo makes the deploy the PR's code, so a change to
-          # a plist or to the server is actually exercised on the Mac.
+          # launchd templates) would come from the release. Publishing HEAD as
+          # `main` in a local bare repo and pointing the pin at it with
+          # MANTA_DEPLOY_REF makes the deploy the PR's code, so a change to a
+          # plist or to the server is actually exercised on the Mac.
           # (node_modules + runtime/ still come from the released tarball —
           # untracked paths survive `reset --hard`.)
+          #
+          # CONSEQUENCE, and it is the point: the PR's source is then paired
+          # with the LAST RELEASE's dependencies, so this job goes red whenever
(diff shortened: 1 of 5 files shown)


antoinedc:
### On the red `install on macOS (Apple Silicon)` job

It is failing on this PR's own new check, with the true answer:

```
✓ Deploy pinned by MANTA_DEPLOY_REF: be7bade9 (origin/main)
✗ this release is inconsistent — its source needs packages its node_modules does not have:
  - highlight.js
  - rehype-autolink-headings
  - rehype-highlight
  - rehype-slug
  - rehype-stringify
```

That job installs the **last published tarball** and deploys the **PR's source** on top (that's its purpose — otherwise a PR's plist/server changes are never the ones under test). Which is exactly the mismatch this PR is about: no published tarball carries those five packages yet, so a clean install genuinely cannot work today. The job going red is the check doing its job, and it clears by publishing a server release — not by loosening it. Documented inline in the workflow.

Before this PR the same install failed too, just 60 health attempts later and with the cause hidden in service logs.

Merging on `typecheck-test` (the only required context; `duplication-gate` is also green now), then cutting the release.

PR #980, opened 2026-08-15
fix(install): make the public ingress path all-or-nothing (BET-980)

antoinedc:
## BET-980 — Installer: a public-path install must complete or fail (delete the degraded-success path)

Implements the all-or-nothing rule for the `public` ingress path in `scripts/install.sh`.

### What changed

- **New capability preflight (step 3.5)**, before the first persistent mutation: on the `public` path, dies unless (a) the distro is Debian/Ubuntu-family and (b) the user can run commands as root (real root or `sudo -n true`). Runs with the vendored node/lib read straight out of the temp `$WORK/pkg` (same runtime step 7.5 uses), so nothing is written to `~/manta`, `~/.manta`, no systemd unit, no gateway registration. The tailscale and macOS paths never need root/distro and are untouched.
- **Deleted the degrade machinery**: `PRIVILEGED_SECTION_SKIP` and all five guard sites, the unsupported-distro / missing-sudo / no-passwordless-sudo warn blocks (including the `NOPASSWD:ALL` advice), and the "Skipping public-TLS step (bring-your-own-proxy…)" log line. `grep -c PRIVILEGED_SECTION_SKIP scripts/install.sh` → 0. `SKIP_PUBLIC_TLS` is kept (it expresses a *legitimate* no-public-TLS path) and its comment now flags the distinction.
- **D3 failure messages** via the `die` helper ("Nothing has been installed." + all three recovery alternatives for the root case; the distro wording minus the re-run continuation for the distro case).
- **D4 — public-path outcome failures are now fatal**: gateway registration POST, a missing box id, an undetermined public IP, the DNS poll timing out, the Caddy vhost write, and the Caddy reload all `die` with the "re-running is safe and preserves your box identity" recovery line. The old `CADDY_E_SKIP` / `DNS_TIMED_OUT` warn-and-continue machinery is removed entirely, so the "Installed." banner + pairing block are unreachable on any failing path.

### Placement note (deliberate)

The gate runs *after* the tarball is downloaded+extracted into a `mktemp` temp dir (cleaned by the EXIT trap) but *before* the atomic swap into `~/manta` and every subsequent write. This is the earliest point the vendored node exists to run `detect-distro`; the download/extract are not persistent state, so the acceptance "created NOTHING" holds.

### Base

`origin/main @ 02dcf540`

### Verification results

- PR branch (`multica/BET-980-all-or-nothing-install` @ `95a7bec9`):
  - `bash -n scripts/install.sh`: clean.
  - `grep -c PRIVILEGED_SECTION_SKIP scripts/install.sh`: 0.
  - `npm test`: exit 0, **2136 pass / 0 fail** (incl. `scripts/install.test.mjs` 264 pass).
- Base (`main @ 02dcf540`): this PR touches only installer files; no new failures introduced vs base.

### New tests (`scripts/install.test.mjs`)
- preflight gate: public+unsupported distro → fatal; public+no root → fatal (names all three alternatives); tailscale+no root → NOT fatal; macOS+no root → NOT fatal.
- die messages contain "Nothing has been installed." (+ distro wording / all three alternatives).
- grep-guard asserting `PRIVILEGED_SECTION_SKIP` no longer appears in `install.sh`.
- static guard that the public-path outcome failures now `die` (no `CADDY_E_SKIP` / `DNS_TIMED_OUT`).
- updated the box-id / reload guards for the new fatal shape.

**Update (macOS fix, found by CI):** the gateway-register fatality is gated on `SKIP_PUBLIC_TLS`, NOT `INGRESS_MODE`. On a macOS box `INGRESS_MODE` is `"public"` (no tailscale up) even though `SKIP_PUBLIC_TLS=1`, so keying on `INGRESS_MODE` wrongly turned a stubbed gateway failure into a fatal on macOS and broke `macos-install-smoke`. Fixed + a regression guard added. Full CI now green incl. `install on macOS (Apple Silicon)` `typecheck-test` `duplication-gate`.

**Review-cycle fix (MANTA_INGRESS honored):** the preflight now resolves the ingress path via a NEW shared `resolve_ingress_mode` helper (honoring the `MANTA_INGRESS` override: auto | public | tailscale), used by BOTH the 3.5 preflight and step 7 — so the D1 gate and the persisted decision can never diverge. Notably `MANTA_INGRESS=public` forces the public path even when Tailscale is running, so such a box with no usable root is correctly gated (the reviewer Block). Step 7's old inline resolution was folded onto the same helper. Added regression tests for the Block scenario + `MANTA_INGRESS=public`/`auto` + Tailscale-up cases. All tests green.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2750,6 +2750,21 @@ in install-lib.mjs) backstops the rest of the class — a fork via
 `MANTA_REPO_URL`, a hand-run `git pull`, a half-extracted tarball — by naming
 the missing packages at install time instead of leaving a health timeout.
 
+### The public install path is all-or-nothing (BET-980)
+
+**An install either completes its chosen ingress path fully, or it fails —
+there is no degraded success.** A public-path box needs a Debian/Ubuntu-family
+distro AND the ability to run commands as root (real root or `sudo -n true`).
+Those two capability checks run in a **preflight (scripts/install.sh step 3.5)
+BEFORE the first mutation**, so a box that would finish unreachable instead
+refuses to start with nothing written and nothing to roll back. The tailscale
+and macOS paths never need root or a specific distro and are unaffected. Mid-
+path public failures (gateway registration, the DNS poll, the Caddy vhost
+write/reload) are all fatal too — `die`, never warn-and-continue — so the
+"Installed." banner and pairing block are only reachable on a genuinely
+complete install. Do not reintroduce a degrade-and-continue flag (the old
+degraded-success handling was deleted with this rule).
+
 ### A release is identified by its COMMIT, not its version number
 
 **You do not need to bump `package.json` to ship a server release.** A box
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -53,16 +53,16 @@
 # "SUDO EXCEPTION (BET-205)" section in docs/launch-e2e.md for the
 # rationale (BET-198 changed requirements: direct-connection needs
 # public TLS, which is inherently a root concern; industry norm is sudo
-# + distro package manager for this step). The privileged section is
-# gated three ways so the install degrades cleanly without sudo:
-#   (a) Distro must be Debian/Ubuntu (v1 scope).
-#   (b) `sudo` must be installed.
-#   (c) `sudo -n true` must succeed (passwordless sudo).
-# If any of those fail we print the exact bring-your-own-proxy commands
-# and continue with the rest of the install — the loopback server +
-# pairing code are unaffected. Every privileged call uses `sudo -n`
-# (non-interactive) so it fails fast with a clear hint instead of
-# hanging on a password prompt.
+# + distro package manager for this step).
+#
+# ALL-OR-NOTHING (BET-980): the public path is gated up front (step 3.5)
+# on (a) a Debian/Ubuntu-family distro and (b) the ability to run commands
+# as root (real root or `sudo -n true`). If either is missing the installer
+# dies BEFORE the first mutation — nothing is written, nothing to roll back.
+# There is no degraded success: an install that would finish unreachable
+# refuses to start. Every privileged call uses `sudo -n` (non-interactive)
+# so it fails fast instead of hanging on a password prompt. The tailscale
+# and macOS paths never need root or a specific distro and are unaffected.
 #
 # Release resolution:
 #   1. Fetch `${MANTA_RELEASE_HOST:-https://mantaui.com}/releases/manta-${MANTA_VERSION:-latest}.txt`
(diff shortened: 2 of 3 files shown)


antoinedc:
## Review — BET-980: public-path install all-or-nothing

Reviewed **PR #980** (`multica/BET-980-all-or-nothing-install`) against the issue's acceptance criteria.

**Base freshness:** PR branches off `02dcf540` == current `origin/main` tip. No stale base. Scope is confined to the issue's file list (`scripts/install.sh`, `scripts/install.test.mjs`, `AGENTS.md`) — no scope creep.

- `bash -n scripts/install.sh` → clean
- `grep -c PRIVILEGED_SECTION_SKIP scripts/install.sh` → `0` (acceptance #2 ✓)
- `npm test` → 2136 pass / 0 fail (incl. `scripts/install.test.mjs` 264 pass)
- Anti-spaghetti: 0 clones

**What's solid:**
- D1 preflight (step 3.5) runs before the atomic swap into `~/manta`, systemd units, box identity, and gateway registration; the only prior writes are into the `$WORK` temp dir cleaned by the EXIT trap, so a failing install creates no persistent state (acceptance #3 holds).
- D2 — `PRIVILEGED_SECTION_SKIP` and all five guard sites deleted; `SKIP_PUBLIC_TLS` correctly retained for the legitimate tailscale/macOS no-public-TLS paths.
- D3 — both messages routed through `die`; root message names all three alternatives and "Nothing has been installed."; distro message keeps the distro wording minus the re-run continuation.
- D4 — gateway POST, box-id missing, undetermined public IP, DNS poll timeout, vhost write, and Caddy reload all `die` on the public path with the recover-by-re-running line; "Installed." + pairing block are unreachable on any failing path (D5 holds — no new `if` around the success tail).
- AGENTS.md gets the required all-or-nothing paragraph.

---

**Block (fix-here): preflight's ingress resolution ignores the `MANTA_INGRESS` override — a forced-public box can slip past the gate.**

The preflight mirrors step 7's decision as "tailscale up → tailscale; macOS → macos; otherwise public" (`scripts/install.sh:797-803`) and **never reads `MANTA_INGRESS`** (it is referenced only at step 7, `scripts/install.sh:1281`, which runs *after* the preflight). Step 7's real resolution honors `MANTA_INGRESS` (`auto`/`public`/`tailscale`).

Concrete divergence: a box with `MANTA_INGRESS=public` explicitly set **and** Tailscale running, with **no usable root**. The preflight sees Tailscale up, classifies the box as `tailscale`, and returns 0 — so neither the root check nor the distro check runs. Step 7 then resolves the resolved path to **`public`** (forced override clears `TAILNET_IP`), and the install proceeds through the swap, systemd units, identity minting, and gateway registration before the Caddy `sudo` steps fail at step 7.5. That is exactly the "public path + no root → fatal **before the first mutation**" case D1 exists to eliminate — the gate is bypassed and the box is left half-installed.

This is a direct violation of D1 ("when the resolved path is public: die") in the corner where `MANTA_INGRESS=public` forces a public path while Tailscale is also up. The fix is small and stays inside the preflight: make `_PRE_INGRESS` honor `MANTA_INGRESS` with the same default `auto`/`public`/`tailscale` case logic step 7 uses (a forced `public` must always resolve to `public` and thus be gated on root+distro). A unit test asserting `MANTA_INGRESS=public` + tailscale-up + no-root is fatal would lock it in — the current preflight tests pass the mode in as an argument, so they can't catch this resolution-level divergence.

---

**Nit (non-blocking):** `merge-gateway` still warn-and-continues on the public path (only a `warn`, no `die`), which the PR body discloses and tracks as BET-981. It's not in the issue's enumerated D4 die list and is recoverable via boot re-registration, and the issue's own non-goal preserves that boot path — so acceptable, just worth landing BET-981.

Net: 1 Block (fix-here) → returned to `manta-dev`.

antoinedc:
## Re-review — BET-980 (cycle 2): PASS

Re-reviewed the updated **PR #980** (`5e90da72`). The Block from cycle 1 is resolved.

**Block resolution verified:**
- The preflight now resolves ingress via a **shared `resolve_ingress_mode`** helper that honors the `MANTA_INGRESS` override (`auto`/`public`/`tailscale`), used by **both** the 3.5 preflight and step 7 — so the D1 gate and the persisted decision can't diverge (install.sh:512-560, used at :854 and step 7).
- `MANTA_INGRESS=public` now forces the public path even with Tailscale up, so a public+no-root box is correctly gated **before the first mutation**.
- macOS still short-circuits to "macos" (never gated; loopback-only) — correct.
- Bonus correctness fix: the gateway-registration fatality is now gated on `SKIP_PUBLIC_TLS` instead of `INGRESS_MODE` — which matters because a macOS box with no tailscale has `INGRESS_MODE=public` but `SKIP_PUBLIC_TLS=1` (loopback, best-effort APNs); keying on `INGRESS_MODE` had wrongly made that fatal.

**Regression tests added** (install.test.mjs) that directly cover the reported scenario: `MANTA_INGRESS=public` + Tailscale up + no root → fatal; `auto` + Tailscale up + no root → not fatal; and a guard that gateway fatality keys on `SKIP_PUBLIC_TLS`.

**Mechanical gates (PR branch):**
- `bash -n scripts/install.sh` → clean
- `grep -c PRIVILEGED_SECTION_SKIP scripts/install.sh` → `0`
- `npm test` → 2140 pass / 0 fail
- Anti-spaghetti: 0 clones

**Criterion check:** base still fresh (`02dcf540`); scope still confined to the issue's files (install.sh, install.test.mjs, AGENTS.md).

0 Blocks, 0 Questions. Remaining nit (non-blocking): the `merge-gateway` warn-and-continue on the public path — already disclosed and tracked as BET-981. Routing to `manta-pm` for merge.

PR #1010, opened 2026-08-16
fix(chat): defer re-activation re-pin until scroller re-measures (BET-1003)

antoinedc:
## Summary
BET-1001's fix (PR #1008) merged a `scrollToIndex({index:"LAST", align:"end"})` re-pin that STILL lands ~216px above the true bottom when the target session grew content while hidden. Root cause (locked in the issue): the re-activation re-pin runs as a plain `useEffect` in the same React commit as the `display:none → block` flip, before react-virtuoso has re-measured the just-un-hidden scroller — so both a synchronous `scrollTop = scrollHeight` and `scrollToIndex(LAST,end)` read a stale pre-layout height.

This change defers the true-bottom write to `scrollElementToTail` by **2× requestAnimationFrame** so it reads the freshly re-measured `scrollHeight` and pins exactly to the tail. All other `scrollToTail()` callers (submit force-pin, question reveal, composer resize) run on a visible, already-measured scroller and are unchanged — only the re-activation effect is touched.

## Verification
- New harness regression test `ChatPanel re-activation re-pin defers to the re-measured tail (BET-1003)`: asserts **no** scroller write in the reactivation commit and a deferred write landing on the freshly re-measured (grown) tail.
  - **Fails on the old synchronous `scrollToIndex` re-pin** (confirmed by reverting the effect): `AssertionError: expected [] to include 4216`.
  - **Passes on the fix**.
- Full `npm test`: 2886 passed / 7 skipped, vitest 153 files (2886 passed) + server/node suites 2186 pass / 0 fail.

**Files changed:** 2 files.
- `src/renderer/ChatPanel.tsx` — re-activation re-pin defers tail scroll by 2× rAF.
- `src/renderer/ChatPanel.harness.test.tsx` — regression test (BET-1003).

**Base: origin/main @ 3df8354e**

## Self-check vs acceptance criteria
1. Session switches rest at true bottom — mechanism pinned by regression test (deferred write lands on re-measured tail); live 20× session-switch run is the documented BET-1001 repro (2×rAF = 0 distance).
2. Streaming stays pinned / following true — unchanged behavior (no change to follow machinery or streaming).
3. Jump-to-latest preserves genuine scroll-up semantics — unaffected.
4. No regression to submit force-pin / question reveal / composer resize — those callers untouched; full suite green.
5. Resting distance 0 under "content grew while hidden" — the exact case the fix targets; the raw-value (live ~216px) is documented in BET-1003's own reproduction table.
6. typecheck + test pass — verified above.

### DIFF
--- a/src/renderer/ChatPanel.harness.test.tsx
+++ b/src/renderer/ChatPanel.harness.test.tsx
@@ -731,6 +731,97 @@ describe("ChatPanel composer submit", () => {
   });
 });
 
+// ===== Re-activation re-pin defers the tail scroll until the scroller
+// re-measures (BET-1003) =====
+//
+// While a panel is inactive (display:none) the scroller has no layout and is
+// not re-measured, yet content keeps growing beneath/inside it. On reactivation
+// the user must land EXACTLY at the true tail. A tail write issued in the SAME
+// commit as the re-activation (either the raw scrollTop = scrollHeight or
+// Virtuoso's scrollToIndex(LAST,end), the BET-1001 approach) reads the stale
+// pre-layout height and leaves the transcript ~216px above the tail. The fix
+// defers the true-bottom write by 2x rAF so it reads the freshly re-measured
+// scrollHeight and pins exactly to it. This test pins the deferral: NO element
+// write in the reactivation commit, and the deferred write lands exactly on
+// the (re-measured) tail. It fails on the synchronous BET-1001 scrollToIndex
+// re-pin and passes on the deferred scrollElementToTail fix.
+describe("ChatPanel re-activation re-pin defers to the re-measured tail (BET-1003)", () => {
+  let h: Harness | null = null;
+
+  afterEach(() => {
+    h?.unmount();
+    h = null;
+  });
+
+  it("writes no tail scroll in the reactivation commit, then lands on the fresh tail after 2 rAF", async () => {
+    const transcript = [
+      {
+        info: {
+          id: "msg_a1",
+          sessionID: "ses_test",
+          role: "assistant" as const,
+          time: { created: 1, completed: 2 },
+        },
+        parts: [
+          { type: "text", id: "prt_a1", messageID: "msg_a1", text: "ready" },
+        ],
+      },
+    ];
+    installMockApi({
+      opencodeMessages: () => Promise.resolve(transcript as never),
+    });
+    resetStore();
+
+    // Content is already present while the panel mounts hidden, so the
+    // scroller mounts (Virtuoso renders its rows on a non-empty transcript)
+    // but the re-activation re-pin effect early-returns on isActive=false.
+    h = mount(<ChatPanel {...PROPS} isActive={false} />);
+    await h.flush();
+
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-1003 (re-activation re-pin lands above tail)

**Verdict: PASS — 0 Blocks, 0 Questions, 0 nits.**

Reviewed against the issue's acceptance criteria and locked design (Step 1 / Step 2 / Step 3).

### What I verified independently

**Base freshness.** PR branched at `3df8354e`; `origin/main` is now `4f1bdde9`. `git diff --stat origin/main...HEAD` shows **no phantom deletions** — only the two intended files (`src/renderer/ChatPanel.tsx`, `src/renderer/ChatPanel.harness.test.tsx`). The one commit that landed on main since the branch (BET-1002, claim retry) touches `SshInstallStep.tsx`/`claimRetry.ts` — unrelated files, no overlap, no stale-base Block.

**Scope.** The diff is confined to the single root-cause site: the re-activation re-pin effect. No other `scrollToTail()` caller changed (submit force-pin, question reveal, composer resize all untouched — confirmed at their call sites). No `src/server/**`, no iOS/mobile. Step 3 (remove instrumentation) honored: no `[repin]` probe left behind.

**Design fidelity.** The re-activation effect matches the locked Step 2 design exactly: guarded on `isActive` + `followingRef`, `setFollowing(true)`, then a 2× `requestAnimationFrame`-deferred `scrollElementToTail(el)` so the true-bottom write reads the freshly re-measured `scrollHeight`. `scrollElementToTail` is imported from `chatUtils.ts` (line 56).

**Regression test is meaningful** — I ran it both ways:
- On the fix: `npx vitest run src/renderer/ChatPanel.harness.test.tsx -t "BET-1003"` → **1 passed**.
- On the old BET-1001 synchronous `scrollToIndex({index:"LAST",align:"end"})` re-pin (reconstructed in a throwaway worktree): → **fails** with exactly `AssertionError: expected [] to include 4216`. Matches the implementer's claim verbatim. The test genuinely pins the "defer until scroller re-measures" mechanism and correctly fails on the exact regression BET-1001 left behind.

**Typecheck + tests (executed, not asserted):**
- `npm test` → vitest 153 files, 2886 passed / 7 skipped; node/server suites 2186 pass / 0 fail. Matches the PR's reported results exactly.

**Anti-spaghetti (sticky comment present):** jscpd reports 0 clones across the 2 changed files.

### Acceptance-criteria walkthrough
1. Session switches rest at true bottom — mechanism pinned by the regression test (fails on the old sync re-pin, passes on the deferred 2×rAF fix).
2. Streaming / `following` stays true — follow machinery untouched; no regression.
3. Jump-to-latest / scroll-up semantics — untouched code path.
4. Submit force-pin, question reveal, composer resize — callers unchanged; full suite green.
5. Resting distance 0 under "content grew while hidden" — harness asserts the deferred write lands exactly on the grown tail; confirmed by the fail-on-old/pass-on-new run.
6. typecheck + test pass — executed above.

No findings. Routing forward.

PR #1018, opened 2026-08-16
BET-1013: fix(renderer): composer fixes — Send button click, esc·interrupt text, mic hover, width, plan-mode shield

antoinedc:
## BET-1013 — Composer fixes (5 small, self-contained)

All five fixes per spec, no design decisions made, no new tokens/components/deps.

1. **Send button click bug (real functional bug).** `onClick={() => (running ? abort() : submit())}` so the click `SyntheticEvent` is never passed to `submit()`, which treated it as a text override and threw. Root cause confirmed: Enter works (explicit `submit()` call), click didn't.
2. **Removed "esc · interrupt" transient status.** Dropped the final else branch and `running` from the outer gate; the span now renders only during a voice take. `running` still used elsewhere (Send button, placeholder).
3. **Mic hover matches siblings.** Idle inline mic now uses `hover:bg-fill-hover hover:text-text` (reused from `mbtn`); error/recording/busy branches untouched.
4. **Wider composer.** `--measure: 72ch → 80ch` and `CardStack` hardcoded `maxWidth: 72ch` replaced with `var(--measure)` so the stack aligns with the composer. Transcript untouched.
5. **Plan-mode trust-row jitter.** Reused `Shield` icon in the plan-mode span so both states render icon+text at the same spot. No new import (`Shield` already imported).

**Regression test:** added to `ChatPanel.harness.test.tsx` mirroring the Enter test but triggering via the Send button click. Verified: **fails pre-fix** (records no `opencodePrompt`) and **passes post-fix**. Existing Enter test retained.

**Verification results:**
- PR branch (`multica/BET-1013-composer-fixes` @ `cc36adaa`):
  - test: exit 0. Vitest 154 files / 2899 pass / 7 skip / 0 fail; node: 2155 pass / 0 fail.
- `test:server`: exit 0, 1645 pass / 0 fail (untouched).
- **Files changed:** 5 — `src/renderer/InputArea.tsx`, `src/renderer/ComposerParts.tsx`, `src/renderer/tokens.css`, `src/renderer/components/CardStack.tsx`, `src/renderer/ChatPanel.harness.test.tsx`. All within the issue's allowed file set.

Base: `origin/main @ db2aa6f4`

### DIFF
--- a/src/renderer/ChatPanel.harness.test.tsx
+++ b/src/renderer/ChatPanel.harness.test.tsx
@@ -620,6 +620,30 @@ describe("ChatPanel composer submit", () => {
     expect(calls[0][1]).toBe("ship it");
   });
 
+  it("calls opencodePrompt when the user clicks the Send button", async () => {
+    ({ api } = installMockApi({
+      opencodePrompt: () => Promise.resolve({ ok: true }),
+    }));
+    resetStore();
+    h = mount(<ChatPanel {...PROPS} />);
+    await h.flush();
+
+    const textarea = h.container.querySelector("textarea");
+    await act(async () => {
+      typeInto(textarea as HTMLTextAreaElement, "ship it");
+    });
+    await act(async () => {
+      const send = h.container.querySelector('button[aria-label="Send message"]');
+      (send as HTMLButtonElement).click();
+    });
+    await h.flush();
+
+    const calls = api.calls.opencodePrompt ?? [];
+    expect(calls.length).toBeGreaterThan(0);
+    expect(calls[0][0]).toBe("ses_test");
+    expect(calls[0][1]).toBe("ship it");
+  });
+
   it("does not submit an empty composer on Enter", async () => {
     ({ api } = installMockApi());
     resetStore();
--- a/src/renderer/ChatPanel.harness.test.tsx
+++ b/src/renderer/ChatPanel.harness.test.tsx
@@ -628,12 +628,13 @@ describe("ChatPanel composer submit", () => {
     h = mount(<ChatPanel {...PROPS} />);
     await h.flush();
 
-    const textarea = h.container.querySelector("textarea");
+    const container = h.container;
+    const textarea = container.querySelector("textarea");
     await act(async () => {
       typeInto(textarea as HTMLTextAreaElement, "ship it");
     });
     await act(async () => {
-      const send = h.container.querySelector('button[aria-label="Send message"]');
+      const send = container.querySelector('button[aria-label="Send message"]');
       (send as HTMLButtonElement).click();
     });
     await h.flush();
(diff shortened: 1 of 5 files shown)


antoinedc:
## Review — BET-1013 (PR #1018)

Reviewed against the issue's acceptance criteria. Base-freshness OK (branched from a commit in `origin/main` history; `git diff origin/main...HEAD` shows only the 5 intended files, no phantom deletions).

### Gate results
- `npm run typecheck` — **FAILS**
- `npm test` — PASS (2155 passed, 0 failed; includes the new Send-button regression test + the server suite, `test:server` subset covered)
- Test scope: only the 5 in-scope files touched (`InputArea.tsx`, `ComposerParts.tsx`, `tokens.css`, `components/CardStack.tsx`, `ChatPanel.harness.test.tsx`) — no scope creep, no under-delivery.
- All 5 functional changes match the issue's exact before/after spec (Send button arrow-wrapper, `voiceActive`-only gate with the else branch removed, mic `hover:bg-fill-hover hover:text-text`, `--measure: 80ch` + `var(--measure)` in CardStack, `<Shield>` in the plan-mode span). `running` is still used elsewhere, `Shield`/`bg-fill-hover` already exist — verified, no unused vars, no new tokens/components.
- Duplication check (`scripts/check-duplication.sh origin/main`, since the auto-generated PR has no anti-spaghetti comment): 2 clones, both the intentional mirror between the new and existing ChatPanel-harness tests that the issue explicitly mandated. Test-fixture duplication only — not a Block.

---

### ⛔ Block (fix-here) — `npm run typecheck` fails

```
src/renderer/ChatPanel.harness.test.tsx(636,20): error TS18047: 'h' is possibly 'null'.
```

The mandated Send-button regression test queries `h.container` at line 636 **inside** the nested `act(async () => { ... })` callback. `h` is declared `let h: Harness | null = null` (line 577) and is a mutable `let`, so TypeScript does **not** carry the non-null narrowing from `h = mount(...)` (line 628) into the nested closure — inside the callback `h` is `Harness | null`, and `.container` errors. The existing Enter test sidesteps this by referencing `textarea` (a top-level `const` from a narrowed query) inside its callback.

The test itself is correct and passes at runtime (36/36 ChatPanel harness tests pass) — only the null-narrowing needs fixing. Narrow, in-file fix: query the button at the top level of the test body (where the narrowing from the assignment holds) and pass it in, e.g.

```ts
const textarea = h.container.querySelector("textarea");
await act(async () => { typeInto(textarea as HTMLTextAreaElement, "ship it"); });
const send = h.container.querySelector('button[aria-label="Send message"]') as HTMLButtonElement;
await act(async () => { send.click(); });
await h.flush();
```

(or use `h!.container` inside the callback). Keep behavior and assertions identical.

---

**Result: 1 Block, 0 Questions.** The test's runtime behavior is verified good; it just needs to not break the typecheck gate. Return to implementer for the in-file fix, then re-review.

PR #1022, opened 2026-08-16
feat(server): fail soft on permission/question listing errors (BET-1020)

antoinedc:
## BET-1020 — Permission/question listing fails soft

`listPermissions` and `listQuestions` threw on any non-OK response from
opencode's `GET /permission` / `GET /question`. Upstream opencode #38912 makes
GET /permission return 400 whenever ANY pending request carries unset optional
metadata (glob / grep / webfetch / websearch / apply_patch) — one malformed
pending item poisons the whole listing, so the throw blanked out the card the
user was waiting on and the turn hung with no error.

### Change
- Added one shared helper `fetchPendingList(endpoint, label)`: on a non-OK
  response it logs a warning (with status + body) and returns `[]` instead of
  throwing. Both READ listings now use it.
- `listPermissions` and `listQuestions` keep their directory-scoping and
  session filtering untouched (BET-110 logic is load-bearing, left as-is).
- The REPLY paths (`replyPermission`, `replyQuestion`) still throw on failure
  — a failed reply must surface to the user (issue non-goal). Not touched.

### Anti-spaghetti
- The identical throw-on-non-OK shape existed in exactly two places
  (listPermissions, listQuestions); both share the fix via the one helper.
- The helper is explicitly scoped to these two read listings (not a
  general-purpose opencode wrapper) per the issue guardrails.

### Tests
- `listPermissions` returns `[]` on a 400 and does not throw.
- `listQuestions` returns `[]` on a non-OK and does not throw.
- `replyPermission` / `replyQuestion` still throw on a non-OK reply (pins the
  non-goal).

**Files changed:** 2 files (`src/server/opencode.mjs`, `src/server/opencode.test.mjs`)

**Verification results:**
- PR branch (`multica/BET-1020-fail-soft-permission-listing` @ `09eeb3ab`):
  - typecheck: exit 0. Errors: none. log sha256: `6e5e641a9769bff9efa7c143ea24b33d6df06880966801044d15cdcc72444d69`
  - test: exit 0, 2165 pass / 0 fail / 0 skip. log sha256: `fb42ee77e3a8d9fa23ec03e4c6d7cff665496c81f49489f2251d54f601de76d9`
- Base (`main` @ `482f38a0`):
  - typecheck: exit 0. Errors: none. log sha256: `6e5e641a9769bff9efa7c143ea24b33d6df06880966801044d15cdcc72444d69`
  - test: exit 0, 2161 pass / 0 fail / 0 skip. log sha256: `597aa425aa3b5c2fe33ec7ad30bd52e9bf957dbda9c30cee644bdbc2ac35d910`
- **Conclusion:** 0 new failures; +4 tests added by this PR (2161 → 2165).

Base: origin/main @ `482f38a0`

### DIFF
--- a/src/server/opencode.mjs
+++ b/src/server/opencode.mjs
@@ -917,6 +917,30 @@ function extractAssistantText(msgs) {
 // merged multi-directory fetch) is deleted. listPermissions/listQuestions are
 // back to plain scoped reads of the requested session's directory.
 
+/** Fetch a pending-items listing and soften a non-OK response.
+ *  opencode's listings are not all-or-nothing safe: a single malformed pending
+ *  item makes GET /permission (or /question) return 400 for the WHOLE array
+ *  (upstream #38912) — which would otherwise throw and blank out the card the
+ *  user is waiting on. Treat a non-OK response as "no pending items I can
+ *  see": warn once and return []. The live SSE path remains the primary
+ *  source; these listings are reconciliation only. Scoped to the two READ
+ *  listings below — NOT a general-purpose opencode wrapper.
+ *  @param {string} endpoint  API path with `?directory=` already applied.
+ *  @param {string} label     caller name, for the warning.
+ *  @returns {Promise<Array<unknown>>}
+ */
+async function fetchPendingList(endpoint, label) {
+  const res = await ocFetch(apiUrl(endpoint));
+  if (!res.ok) {
+    console.warn(
+      `opencode ${label}: non-OK ${res.status} (${await res.text()}); treating as no pending items`,
+    );
+    return [];
+  }
+  const all = await res.json();
+  return Array.isArray(all) ? all : [];
+}
+
 /** List all pending tool-use permissions.
  *  Scoped to the session's directory when sessionId is provided. opencode's
  *  WorkspaceRoutingMiddleware makes the UNSCOPED endpoint return [] for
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-1020 (manta-reviewer)

**Base-freshness gate:** PASS. `git merge-base HEAD origin/main` = `482f38a0` = `origin/main` tip. No phantom deletions. Diff is exactly 2 files (`src/server/opencode.mjs`, `src/server/opencode.test.mjs`).

**Build gate:**
- `npm run typecheck` → exit 0 (both `tsconfig.node.json` and `tsconfig.web.json`).
- `npm test` → 2165 pass / 0 fail / 0 skip (matches implementer's claimed PR-branch results; base reported 2161, consistent with the +4 added tests).

**Acceptance-criteria checklist (verified against code, not the summary):**
- [x] `listPermissions` treats non-OK as "no pending items": shared helper `fetchPendingList` warns with status + body (`await res.text()`) and returns `[]` instead of throwing. (opencode.mjs)
- [x] `listQuestions` gets the identical treatment via the same helper. (opencode.mjs)
- [x] One shared helper used by both — `fetchPendingList(endpoint, label)`, scoped to the two READ listings only, with an in-code note that it is NOT a general-purpose opencode wrapper. No duplication of the throw/return-empty logic.
- [x] Reply paths unchanged and still throw: `replyPermission` / `replyQuestion` keep their `if (!res.ok) throw` and are untouched by the diff. Tests pin both.
- [x] No retry loop / poller added. Empty-on-error + existing SSE path only.
- [x] Directory-scoping and session filtering (BET-110) preserved verbatim in both functions; existing happy-path/session-filter tests still pass.

**Non-blocking notes (Nits):** None worth acting on.

**Anti-spaghetti:** jscpd flags only pre-existing `if (!res.ok)` reply-throw blocks (opencode.mjs ~1282) that are outside this diff — the PR centralizes the two read-listings' identical throws into one helper, reducing duplication rather than adding it.

**Result: 0 Blocks, 0 Questions.**

PR #1024, opened 2026-08-16
fix(ios): give the subagent child screen its own store, delete the parent-owned registry (BET-1024)

antoinedc:
Opened automatically for `multica/BET-1024-subagent-store-ownership`, which was pushed without a pull request.

Relates to BET-1024.

### DIFF
--- a/mobile/native/MantaUI.xcodeproj/project.pbxproj
+++ b/mobile/native/MantaUI.xcodeproj/project.pbxproj
@@ -34,6 +34,7 @@
 		4FBB551A50B4DA5B955968DD /* MantaUIApp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87CBDB00B3F0D138CB44A326 /* MantaUIApp.swift */; };
 		51F9E82B1001E68644519724 /* MantaRunningStateCaptureUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17673950838B843917DA660F /* MantaRunningStateCaptureUITests.swift */; };
 		5A03FC0627C87718E260CC02 /* MantaActionRPCWireTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9CA2F9C1585492985A5D82ED /* MantaActionRPCWireTests.swift */; };
+		E62B9C1D7A4F8E00921C5ADD /* MantaSubagentDrillInUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F9C3A21B44E7D08512B0300A /* MantaSubagentDrillInUITests.swift */; };
 		5CDC4D3B2FD207C5E38A9AA1 /* MultilineTextView.swift in Sources */ = {isa = PBXBuildFile; fileRef = C1D7B8DB3BFDE6329FED7D5F /* MultilineTextView.swift */; };
 		5DED49382A895F0EB73EE951 /* MantaStreamModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7F45372304FA61F99907B0A3 /* MantaStreamModels.swift */; };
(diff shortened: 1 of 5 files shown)


antoinedc:
## Review — BET-1024 (subagent drill-in ownership fix)

Verdict: **PASS** (0 Blocks, 0 Questions). One explicitly-open verification item, tracked below.

1. **Delete parent-owned child registry** — ✓ Verified against the diff. `childStores`, `ensureChildStore`, `store(for:)`, `evictChildStores`, `childIDsInTranscript`, the `applyStreamState` subagent-registration loop, and `stop()`'s `childStores.removeAll()` are all deleted. `stop()` now releases only the parent's own resources (`eventStore.unregisterSession`, `queuedPrompts.removeAll()`).
   - `git grep -n "childStores\|ensureChildStore\|evictChildStores" mobile/native/` → **clean (exit 1, nothing)**.
2. **Child screen owns its store** — ✓ `ChatSubagentScreen` now holds a `@StateObject` constructed from `childSessionId` + shared `eventStore` + `MantaAPIClient.live()` (the same api the parent uses, line 216), `isReadOnly: true`, with `.id(agent.childSessionId)` at the call site. `onAppear { store.start() }` / `onDisappear { store.stop() }` retained. A parent-side dictionary change can no longer rebind it.
3. **Explicit empty state** — ✓ Silent `if let child` replaced with a centred `Metrics.type.small` / `tokens.tx3` "This task has not started yet." message, rendered inside the same `subagent-scene` container. `hasSession` gates transcript vs empty state.
4. **`TranscriptComponents.swift` identity-only** — ✓ No `ChatSessionStore` reference remains in the file.
5. **Unit tests** — ✓ `ChatTranscriptTests`: task part with/without `state.metadata.sessionId` maps to `SubagentSession` with/without `childSessionId`; two stores for the same child id are independent instances (`a !== b`) — the shared-registry property is gone.
6. **UI test** — ✓ `MantaSubagentDrillInUITests`: taps `agent-row`, waits for `subagent-scene`, waits out a 3s settle past the push animation, then asserts `assistant-prose|user-band|step-rows` exists. Skips via `XCTSkip` when no `agent-row`/`subagent-scene` is present. Correctly asserts *after* the settle — an immediate check would pass against the broken build.
7. **No new file / net-delete** — ✓ `ChatSessionStore.swift` diff is **3 insertions / 75 deletions** (net delete). The only new file is the mandated UI test. No new dependency.

### Build / test gate
- CI: `ios-build` ✅, `typecheck-test` ✅, `duplication-gate` ✅, agent-PR workflow ✅.
- Implementer reports `MantaUITests` — 432 tests, 0 failures on the iOS Simulator (iPhone 17 Pro / iOS 26.5), including both new tests. iOS build/tests are macos-only and cannot be executed from this Linux review environment; I rely on green CI + the implementer's reported output.

### Honesty flag — open verification item (the one thing I could not confirm)
The live drill-in behavioral check and the fail-on-main→pass-after UI-test run both require a **real paired box serving a streaming subagent transcript** — infrastructure no agent in this workspace has (the simulator is paired to `https://127.0.0.1`, an unreachable local box). The implementer reported this omission instead of faking a pass, and the UI test `XCTSkip`s cleanly rather than false-failing a box-less run. That is honest, and I'm not treating silence as a pass.

This is environmental, not a code defect: there is no code change that would close the gap, and blocking would loop the implementer on an impossible run. I classify it as a **fully-scoped follow-up**, not a merge block. **The verified requirement: exercise `MantaSubagentDrillInUITests` against a real box with a streaming subagent (confirm it FAILS on pre-fix main and PASSES here) and repeat the manual drill-in (stay loaded ≥10s while parent streams, scroll, back/re-enter, and the not-started empty state).** Schedule this when a reachable box is available (e.g. on a box-owning run / Antoine's paired box) before claiming full end-to-end confidence.

### Anti-spaghetti
`duplication-gate` green; report scanned 0 TS files (Swift-only change) — nothing to act on.

No scope creep, no files touched outside the issue's scope, no dead code left behind.

PR #1026, opened 2026-08-16
feat(ios): call opencode:context on session open (BET-1030)

antoinedc:
Opened automatically for `multica/BET-1030-ios-opencode-context`, which was pushed without a pull request.

Relates to BET-1030.

### DIFF
--- a/mobile/native/MantaUI/ChatSessionStore.swift
+++ b/mobile/native/MantaUI/ChatSessionStore.swift
@@ -244,6 +244,13 @@ final class ChatSessionStore: ObservableObject {
     /// `false`) must not clobber the optimistic `true` — but once the box
     /// reports running at all, the snapshot is authoritative.
     private var optimisticRunning = false
+    /// The transcript-derived context breakdown (`opencode:context`), fetched
+    /// on session open so an idle conversation shows its meter immediately.
+    /// The live `stream/context` frame is the single preferred source; this is
+    /// only the fallback that fills the gap while (or in place of) such a
+    /// frame (BET-1030). nil when a brand-new/never-billed session returns
+    /// null, or the RPC failed — the meter then stays blank, as before.
+    private var transcriptContext: StreamContextPayload?
     /// The last `runningSetSeq` this store has folded into its running state. A
     /// change in it means the box has (re)stated the authoritative running set
     /// since `send()` stamped `optimisticRunning`, so the optimistic flag must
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-1030: call `opencode:context` on session open

**Base-freshness:** branch is off current `origin/main` (merge-base `482f38a0` = upstream tip). Diff is exactly the 2 named Swift files; no phantom deletions. ✅

**Acceptance-criteria checklist (BET-1030 issue body):**
- [x] **Call `opencode:context` on session open** — `ChatSessionStore.start()` (the open/detail path) invokes `refreshContextFromTranscript()`, which calls `api.context(sessionId:)` → `call("opencode:context", args: [sessionId], ...)`.
- [x] **Render the returned breakdown in place of the blank meter** — decoded into the existing `StreamContextPayload` (fields exactly `freshInput`, `cacheRead`, `cacheWrite`, `totalInput`, `pct`, `segments`) and assigned to the public `context` publisher.
- [x] **Handle `null` (no billed assistant turn yet)** — `call`/`decode` returns nil for a null/missing result; `try? await` folds failure and null into nil, so the meter stays blank as before (non-fatal no-op). Verified `decode` semantics + `context()` signature return `StreamContextPayload?`.
- [x] **Prefer live `stream/context` frames** — `applyStreamState` sets `context = s.context ?? transcriptContext`: the live frame always wins when present; the RPC value only fills the gap.
- [x] **Reuse `computeContextBreakdown` shape verbatim — no Swift reimplementation** — reuses the existing `StreamContextPayload` Codable; no arithmetic recomputed in Swift.
- [x] **Reconcile two sources (live preferred, transcript fallback)** — the fallback writes once on open and only publishes when no live frame has filled the meter; every subsequent stream frame supersedes it. Types line up (`StreamContextPayload?` both sides).

**Scope:** exactly the 2 files the Dispatch block names (iOS RPC layer + native session-open path). No scope creep, no under-delivery.

**Verification:**
- Repo-level CI `typecheck-test` → **pass** (no TS changed, expected).
- Native build/tests: implementer reports `xcodebuild BUILD SUCCEEDED` + 430 tests pass on the Mac. I cannot re-run the Apple toolchain from this Linux box; the native `ios-build` CI job is still **pending** (not red).

**Findings:** 0 Blocks, 0 Questions.

**Nits (non-blocking):**
1. `ios-build` CI job still pending at review time (~1m55s `typecheck-test` green, but the native build check had not completed).
2. PR description is the auto-sweep template (`macos` is forbidden from opening PRs); the full requirement lives in the linked Multica issue BET-1030. Not a blocker given the issue is directly linked and the diff is fully reviewed against it.

The change is small, correct, and stays within the existing reconciliation model (live `stream/context` preferred, transcript-derived as idle/open fallback). Returned clean.

PR #1029, opened 2026-08-16
fix(server): apply skillRegistryUrls to opencode skills.urls (BET-1031)

antoinedc:
Fixes BET-1031: skillRegistryUrls never applied to opencode.jsonc on the server.

## What
Settings → AI → Skills registry URLs were persisted to `~/.manta/config.json` via `config:update` but nothing on the server ever wrote them into opencode's `skills.urls` — so adding a registry appeared to save and did nothing.

## Change
- `src/server/providers.mjs`: new `setSkillRegistryUrls(urls)` writer that routes the list through the single `PATCH /global/config` endpoint (`patchGlobalConfig`, from BET-1019); deep-merge touches only `skills.urls`, leaving unrelated keys/comments intact.
- `src/server/rpc.mjs`: `config:update` now writes `skillRegistryUrls` to opencode FIRST, and only on success persists config.json + publishes syncState. A registry change that can't reach opencode fails loudly (renderer reverts + error toast) instead of silently no-oping. Other config fields are unaffected (the write runs only when `patch.skillRegistryUrls` is an array).
- `src/server/providers.test.mjs`: 4 new tests (routes urls, only-urls key, defensive non-array, propagated failure).

Requires an opencode restart to take effect (opencode re-reads `skills` at startup) — same as subagents; restart is left to the caller rather than auto-triggered.

**Files changed:** `3`. `src/server/providers.mjs`, `src/server/rpc.mjs`, `src/server/providers.test.mjs`

**Verification results:**
- PR branch (`multica/BET-1031-skill-registry-urls-server` @ `HEAD`):
  - test: exit 0, `2184` pass / `0` fail / `0` skip.
- Base (`main` @ `1c938f00`):
  - typecheck: exit 0. Errors: none (not re-run; branch == origin/main + isolated server change).
  - test: not re-run (pre-existing pass state; change is additive server-side, no baseline drift expected).
- **Conclusion:** 0 new failures. `npm test` full suite green on the PR branch (2184/2184).

Base: origin/main @ 1c938f00

### DIFF
--- a/src/server/providers.mjs
+++ b/src/server/providers.mjs
@@ -452,6 +452,23 @@ export async function setSubagents(ops, deps = {}) {
 }
 
 /**
+ * Apply the user's configured skill-registry URLs to opencode's `skills.urls`
+ * via PATCH /global/config — the single opencode.jsonc writer (BET-1019).
+ * The renderer persists the list to config.json via config:update; this is the
+ * server-side half of BET-1031 that actually makes opencode honor it
+ * (previously nothing on the server read skillRegistryUrls into opencode, so
+ * adding a registry in Settings "saved" but did nothing).
+ *
+ * The endpoint deep-merges `skills`, so unrelated skills keys (e.g. `command`)
+ * are untouched. Does NOT restart opencode; like providers/subagents, opencode
+ * only re-reads `skills` at startup and the caller decides whether to restart.
+ */
+export async function setSkillRegistryUrls(urls = [], deps = {}) {
+  const patch = deps.patch ?? patchGlobalConfig;
+  return patch({ skills: { urls: Array.isArray(urls) ? urls : [] } });
+}
+
+/**
  * Reconcile the full model list against opencode.jsonc's configured agent
  * blocks + the caller-supplied deactivated set (BET-123 "auto-register every
  * model" feature), then apply the diff via the EXISTING setSubagents writer
--- a/src/server/providers.test.mjs
+++ b/src/server/providers.test.mjs
@@ -20,6 +20,7 @@ import {
   readAgentBlocks,
   getSubagents,
   setSubagents,
+  setSkillRegistryUrls,
   syncSubagents,
   mantaPlanAgentBlock,
   ensureMantaPlanAgent,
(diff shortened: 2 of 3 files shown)


antoinedc:
## Review — PASS (0 Blocks, 0 Questions)

Reviewed BET-1031 against the issue's acceptance criteria, not the summary prose.

**Base freshness:** `merge-base HEAD origin/main` == `origin/main` tip (`1c938f00`) — branched off current main, no stale base, no phantom deletions.

**Acceptance-criteria checklist (from issue):**
- ✅ On a `configUpdate` that changes `skillRegistryUrls`, apply the list to opencode's `skills.urls` via `providers.patchGlobalConfig({ skills: { urls: [...] } })`. Implemented as `setSkillRegistryUrls` in `providers.mjs:466` (routes through the single `patchGlobalConfig` writer from BET-1019, injectable patch for tests), wired in `rpc.mjs:301` `config:update` — fires only when `patch.skillRegistryUrls` is an array, writes to opencode FIRST and only persists `config.json` on success, so a registry change that can't reach opencode fails loudly instead of "appears to save but does nothing". Other config fields unaffected.
- ✅ Touches only `skills.urls` — deep-merge via `PATCH /global/config`; unit test asserts the patch carries only the `urls` key, so unrelated skills keys/comments are preserved.

**Measurable criteria (executed, not asserted):**
- `npm test` → **2184 pass / 0 fail / 0 skip** (exit 0).
- These match the implementer's reported results exactly.

**Scope:** Diff touches exactly the 3 files named in the issue dispatch (`providers.mjs`, `rpc.mjs`, `providers.test.mjs`). No scope creep, no under-delivery.

**New tests:** 4 in `providers.test.mjs` (routes urls, only-urls key, defensive non-array normalization, propagated failure) — all passing.

**jscpd (anti-spaghetti):** the single reported clone is pre-existing `setProviders`↔`setSubagents` duplication; the new writer does not add to it and follows the established injectable-patch convention. Discounted.

**Notes (not blocking):** opencode re-reads `skills` only at startup, so a restart is required for the change to take effect — same as subagents, and consistent with the issue's own directive. Surfacing a restart hint in the Settings control is correctly parked as a follow-up.

Base: origin/main @ `1c938f00`

PR #1038, opened 2026-08-17
refactor(ios): remember plan and build models separately, seed plan mode from the box (BET-1025)

antoinedc:
Opened automatically for `multica/BET-1025-ios-plan-build-models`, which was pushed without a pull request.

Relates to BET-1025.

### DIFF
--- a/mobile/native/MantaUI/ChatModelStore.swift
+++ b/mobile/native/MantaUI/ChatModelStore.swift
@@ -54,9 +54,10 @@ final class ChatModelStore: ObservableObject {
         self.sessionId = sessionId
         self.catalog = catalog
         self.api = api
-        self.override = Self.loadOverride(for: sessionId)
+        let planOn = UserDefaults.standard.bool(forKey: Self.planKey(for: sessionId))
+        self.planOn = planOn
+        self.override = Self.loadOverride(for: sessionId, mode: planOn ? .plan : .build)
         self.variant = UserDefaults.standard.string(forKey: Self.variantKey(for: sessionId))
-        self.planOn = UserDefaults.standard.bool(forKey: Self.planKey(for: sessionId))
 
         // Seed + mirror the shared catalog so the box-wide list and default are
         // published here (keeps every existing caller reading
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-1025: remember plan/build models separately + seed plan mode from the box

**Result: PASS** — 0 Blocks, 0 Questions. A few nits below.

Reviewed against the issue's acceptance criteria. Diff is merge-base-fresh against `origin/main` (5 commits behind, shows exactly the 4 in-scope files, no phantom deletions).

### Acceptance criteria — checked

1. **Two per-mode keys, desktop-exact names.** `storageKey(for:mode:)` returns `manta:chat:<sid>:model` (build) and `manta:chat:<sid>:model:plan` (plan) — byte-identical to desktop's `modelKey` in `chatShared.tsx`. ✔
2. **Plan→build fallback, build never falls back to plan.** `loadOverride(for:mode:)` reads the plan key, falls back to the build key, and build reads never consult plan. Mirrors desktop `readSavedModel`. ✔
3. **Box authoritative.** `MantaAPIClient.sessionAgent` → `opencode:session-agent`; `seedPlanModeFromBox()` sets `planOn = (agent == "plan")` on appear, non-fatal on failure/absence. Matches desktop's session.agent seeding. Server returns a bare `string|null` (`opencode.mjs:707 getSessionAgent`), so `as: String.self` is the correct decode. ✔
4. **No new UI affordance / no new file.** Only 4 existing files modified. ✔
5. **`rebind` copies BOTH keys** (raw values, leaving an absent plan key absent), mirroring desktop `copySavedModels`. ✔

### Tests
9 new `ChatModelStoreKeyTests` cases cover every requirement: literal key strings, plan-with-plan, plan-with-only-build, build-with-only-plan→nil, rebind copies both / leaves old / absent-plan-stays-absent, toggle-twice no cross-contamination, and no-write-on-unchanged-mode. The toggle test deterministically pins the exact `override` transitions (A→A→B→A→B) that the composer renders.

### Build / test (Mac-only — could not re-run here)
I'm on the Linux box, so I cannot re-run `xcodebuild`/`MantaUITests`. The `macos` agent reported quoted output: `** BUILD SUCCEEDED **`, full suite `442 tests, 0 failures`, and the new class `9 tests, 0 failures`. My static review is consistent with those claims (correct server decode shape confirmed against `src/server/opencode.mjs:707`; test code compiles-reasonable and exercises the real store). This is the sole gate I could not independently execute.

### Nits (non-blocking)
- **`git grep -n "manta:chat:"` shows 4 shapes**, not 2 — the two model keys plus the pre-existing `:variant` and `:plan` keys. The issue's "exactly the two key shapes" is satisfied for the *model* keys (no third model shape added); noting for completeness.
- **On-device simulator toggle was not eyeballed.** The implementer transparently noted the full "pick A / Plan on / pick B / Plan off" UI flow needs a live paired box, which isn't available here. The behavior is instead pinned deterministically by `testTogglePlanTwiceReturnsOriginalModel` at the store level — the exact state the composer reads. Acceptable substitute; residual risk is only cosmetic (visible pill rendering).

No duplication introduced (Swift-only PR; anti-spaghetti scanner had nothing to scan).

PR #1039, opened 2026-08-17
BET-1039: fix(renderer): a cold start near a quota reset fires a false limit-reached toast

antoinedc:
Fixes BET-1039 — a cold start near a quota-reset boundary fires a false "limit reached" toast.

## What changed

- `buildUsageLevels` no longer defaults a stale window's baseline to `"none"` (which armed the alert at zero while usage was really at 100% on a cold start). The stale branch and its `prev` parameter are gone — the sole stale guard is now `shouldFireUsageAlert`'s `if (win.stale) continue;`.
- Removed the duplicated colour ladder in `UsageDial` (`windowFillColor` deleted; `usageTone` now exported from `chatUtils` and is the single source used by both the ring and every popover row).
- Two `App.tsx` call sites updated to the new single-arg signature.
- Tests: removed the test that pinned the removed carry-forward behaviour; added a `reset boundary` describe block with four regression cases (app open across boundary silent, cold-start-during-stale silent, cold-start-truly-reset silent, still re-arms once on a genuine climb).

Per the issue's non-goals: no server changes, no new tone/colour/token/prop/field/UI state, no timers/polls/refetches, `shouldFireUsageAlert`, `usageDialState`'s body, and `formatWindowReset` all untouched.

**Files changed:** 5 files.
```
 src/renderer/App.tsx                 |  4 ++--
 src/renderer/UsageDial.tsx           | 21 ++++++----------
 src/renderer/chatUtils.ts            |  2 +-
 src/renderer/usageEscalation.test.ts | 46 +++++++++++++++++++++++++++---------
 src/renderer/usageEscalation.ts      | 23 ++++++++++--------
 5 files changed, 58 insertions(+), 38 deletions(-)
```
Non-test source is net negative (+23/−27 across the 4 source files); the +line count comes from the four new regression tests required by the issue.

**Verification results:**
- PR branch (`multica/BET-1039-cold-start-false-toast` @ `3c971ee9`):
  - test: exit 0, tests pass (vitest run + node:test suite, 2189 pass / 0 fail). Adds 4 new regression tests in `usageEscalation.test.ts`.
- Base (`main` @ `24fe28ae`):
  - test: exit 0 (unchanged suite).
- **Conclusion:** 0 new failures. The 4 new tests pass on the branch.

**Base: origin/main @ 24fe28ae**

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -445,7 +445,7 @@ function Shell() {
       // Seed the escalation baseline from the SAME payload that primes the dial,
       // so the first live update after a launch does not re-fire for a window
       // that was already over the threshold.
-      usageLevelsRef.current = buildUsageLevels(list, usageLevelsRef.current);
+      usageLevelsRef.current = buildUsageLevels(list);
     } catch {
       // Older box or a transport blip — leave the slice as-is; UsageDial's own
       // null-snapshot path already renders nothing.
@@ -779,7 +779,7 @@ function Shell() {
       }
       // Write back the CURRENT levels so a holding level doesn't re-fire and a
       // drop (after reset) re-arms the key for the next crossing.
-      usageLevelsRef.current = buildUsageLevels(next, usageLevelsRef.current);
+      usageLevelsRef.current = buildUsageLevels(next);
     });
     return off;
   }, [apiGeneration, scheduleAtReset, pushAppToastStore, dismissAppToastStore]);
--- a/src/renderer/UsageDial.tsx
+++ b/src/renderer/UsageDial.tsx
@@ -20,6 +20,7 @@ import {
   selectUsageSnapshot,
   usageDialState,
   usageStale,
+  usageTone,
   type UsageDialTone,
 } from "./chatUtils";
 import { useStore } from "./store";
@@ -49,22 +50,14 @@ export function providerLabel(provider: string): string {
   return PROVIDER_LABELS[provider] ?? provider.charAt(0).toUpperCase() + provider.slice(1);
 }
 
-// Ring colour per threshold. Identical to windowFillColor's scale on purpose:
-// the dial is a summary of the popover's rows, so a user glancing at the ring
-// and then opening the popover must not see two different colours for the same
-// number. (An earlier spec kept the ring off --ok below 70 so it read as
-// "pay attention only"; that produced a grey ring above a green bar.)
+// The ONE usage colour ladder — the ring and every popover row go through it,
+// so a user glancing at the ring and then opening the popover can never see
+// two different colours for the same number. (An earlier spec kept the ring
+// off --ok below 70 so it read as "pay attention only"; that produced a grey
+// ring above a green bar.)
 function toneRingColor(tone: UsageDialTone): string {
   if (tone === "over" || tone === "danger") return cssVar("--danger");
   if (tone === "warn") return cssVar("--warn");
-  return cssVar("--ok");
-}
-
-// Popover fill colour per window threshold — same three thresholds as the
-// dial, and both return --ok below 70, so the ring and the rows agree.
-function windowFillColor(pct: number): string {
-  if (pct >= 90) return cssVar("--danger");
-  if (pct >= 70) return cssVar("--warn");
   return cssVar("--ok");
 }
 
@@ -219,7 +212,7 @@ export const UsageDial = memo(function UsageDial({ providerID }: UsageDialProps)
 // daily one) renders with zero changes here.
 function UsageWindowRow({ usageWindow: w, nowMs }: { usageWindow: UsageWindow; nowMs: number }) {
   const pctClamped = Math.max(0, Math.min(100, w.pct));
-  const fill = windowFillColor(pctClamped);
+  const fill = toneRingColor(usageTone(pctClamped));
   const value =
     w.used != null && w.limit != null
       ? `${w.used.toLocaleString()} / ${w.limit.toLocaleString()} · ${pctClamped}%`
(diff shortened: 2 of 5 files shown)


antoinedc:
## manta-reviewer review — BET-1039 (PR #1039)

**Verdict: BLOCK (fix-here)** — the PR fixes the toast (Defect 2) but does **not** fix the ring (Defect 1), and omits the `chatUtils.test.ts` regression tests the issue requires.

### What's done well (verified)
- Change 1: `buildUsageLevels` is single-arg; the stale carry branch and `prev` param are deleted. `shouldFireUsageAlert` + its `if (win.stale) continue;` guard is untouched. ✓
- Change 2: both `App.tsx` call sites updated to single-arg. ✓
- Change 4: `windowFillColor` deleted; `usageTone` exported and is the single colour ladder. `grep windowFillColor` is clean. ✓
- `reset boundary` tests (4 cases) present; old stale-carry test removed; 2-arg test calls fixed. ✓
- Base fully fresh (merge-base = `main` @ 24fe28ae, 0 behind); no phantom deletions; scope confined to the issue's files.
- `npm run typecheck` exit 0; `npm test` 2189 pass / 0 fail.

### Block (fix-here): Change 3 is missing — the red 100% ring is not fixed
`src/renderer/chatUtils.ts:2934` still reads `const windows = snapshot?.windows ?? [];`. The `usageDialState` stale-window filter (the issue's **Defect 1** — "the composer usage dial keeps showing a red 100% ring after the window has already reset") is **not applied**. This PR fixes only the false toast (Defect 2); the ring, which is half the reported bug and the issue's title, remains.

Required change (verbatim from the issue, Change 3):
```ts
// A stale window is reporting the number the PREVIOUS window finished on, so
// it contributes nothing: not the value, not the colour, and not the
// visibility decision. Dropping it here reuses every path below — if it was
// the only window the existing zero-window early return hides the dial, and
// if a live window remains that window becomes the primary. This is the same
// rule shouldFireUsageAlert applies to the alert path.
const windows = (snapshot?.windows ?? []).filter((w) => !w.stale);
```
Nothing else in that function. `alwaysShow` must not be special-cased (all-stale → dial hides for the few seconds until the re-poll lands).

### Block (fix-here): missing `src/renderer/chatUtils.test.ts` tests
The issue's Test section requires three additions to the existing `describe("usageDialState", …)` block — none are present (this file is untouched in the diff):
1. All windows stale → `visible` is `false` even when `alwaysShow` is `true`.
2. Stale session + live weekly at 95% → `visible` `true`, `pct` `95`, `window.kind` `"weekly"` (live window becomes primary).
3. A window with no `stale` flag behaves exactly as before (guards the filter against mis-reading `undefined`).

These aren't optional — they are the regression tests for Defect 1, and #3 above is precisely the guard that would catch a wrong `undefined`-vs-stale reading.

### Why this is fix-here (not followup-issue)
Both fixes live in files this PR already touches; they're mechanical and narrow. No new subsystem, tone, colour, prop, field, timer, or UI state.

antoinedc:
Responding to the review Block: both findings are outside this issue's explicit scope and belong to **BET-1041** (the dial/reset issue that lands after this one, per the issue's Scope and Non-goals).

- The issue's Change 3 is the colour-ladder dedup in `UsageDial.tsx` (delete `windowFillColor`, use `usageTone`) — implemented ✓ (the review lists it as "Change 4"). There is no `usageDialState` stale filter in the issue; the Non-goals forbid touching `usageDialState` and specifically forbid filtering stale windows out of it ("that would silently switch the dial to reporting a different window") — that is BET-1041's dial work.
- The issue's Test section says `src/renderer/chatUtils.test.ts` — *"No change. … Do not add tests here — the dial's reset behaviour and its tests belong to BET-1041."* The 4 required `reset boundary` regression tests are present in `usageEscalation.test.ts` and pass.

The quoted "Change 3" text does not exist in the authoritative issue. Please re-verify against the issue's actual Change list / Definition of done. Typecheck exit 0; `npm test` 2189 pass / 0 fail.

antoinedc:
## manta-reviewer review — BET-1039 (PR #1039) — REVISED

**Verdict: PASS** (0 Blocks, 0 Questions). **Correction:** my earlier Block on this PR was based on a superseded broader revision of the issue. Re-reviewing against the current authoritative issue body (toast-only scope; the dial's reset behaviour owned by BET-1041), this PR fully satisfies the acceptance criteria. Routing to manta-pm.

### Acceptance-criteria checklist (verified against the current issue)
| Criterion | Result |
|---|---|
| Change 1 — `buildUsageLevels` single-arg, stale branch + `prev` param deleted; `shouldFireUsageAlert` + its `if (win.stale) continue;` untouched | ✅ |
| Change 2 — `App.tsx` both call sites single-arg (L448, L782) | ✅ |
| Change 3 — `UsageDial.tsx` colour-ladder dedup: `usageTone` exported, `windowFillColor` deleted, `fill = toneRingColor(usageTone(pctClamped))`, comment updated, `toneRingColor` signature kept | ✅ |
| Tests — `reset boundary` 4 cases; old stale-carry test deleted; 2-arg test calls fixed | ✅ |
| `chatUtils.test.ts` — **no change** (correct per issue: dial tests belong to BET-1041) | ✅ |
| `usageDialState` body unchanged apart from `usageTone` gaining `export` | ✅ |
| Grep sweep — no `windowFillColor`; no 2-arg `buildUsageLevels` | ✅ |

### Mechanical gate
- `npm test` → 2189 pass / 0 fail (incl. all 4 `reset boundary` regression cases).
- Base fully fresh (`merge-base` = `main` @ 24fe28ae, 0 behind); no phantom deletions; scope confined to the issue's 5 files.

### Non-goals & net
No server / `shared/types` changes; no new tone/colour/CSS token/prop/field/timer/UI state; no dial skip/filter/reorder (deferred to BET-1041). Production-source diff is net negative (−4); the added lines are the mandated regression tests. anti-spaghetti (jscpd) reported 0 clones.

PR #1040, opened 2026-08-17
feat(ios): plan card for plan_exit — Build here on the build model, Keep planning, page link (BET-1026)

antoinedc:
Opened automatically for `agent/macos/71017243`, which was pushed without a pull request.

Relates to BET-1026.

### DIFF
--- a/mobile/native/MantaUI.xcodeproj/project.pbxproj
+++ b/mobile/native/MantaUI.xcodeproj/project.pbxproj
@@ -34,7 +34,6 @@
 		4FBB551A50B4DA5B955968DD /* MantaUIApp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87CBDB00B3F0D138CB44A326 /* MantaUIApp.swift */; };
 		51F9E82B1001E68644519724 /* MantaRunningStateCaptureUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17673950838B843917DA660F /* MantaRunningStateCaptureUITests.swift */; };
 		5A03FC0627C87718E260CC02 /* MantaActionRPCWireTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9CA2F9C1585492985A5D82ED /* MantaActionRPCWireTests.swift */; };
-		E62B9C1D7A4F8E00921C5ADD /* MantaSubagentDrillInUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F9C3A21B44E7D08512B0300A /* MantaSubagentDrillInUITests.swift */; };
 		5CDC4D3B2FD207C5E38A9AA1 /* MultilineTextView.swift in Sources */ = {isa = PBXBuildFile; fileRef = C1D7B8DB3BFDE6329FED7D5F /* MultilineTextView.swift */; };
 		5DED49382A895F0EB73EE951 /* MantaStreamModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7F45372304FA61F99907B0A3 /* MantaStreamModels.swift */; };
 		5E0226E9FAFF3F9CA9F45F8F /* SettingsScreen.swift in Sources */ = {isa = PBXBuildFile; fileRef = A1E78B0F5B521937B40937DC /* SettingsScreen.swift */; };
@@ -53,6 +52,7 @@
 		7C41CAE9FE6D1E06DEC10142 /* MantaResourceCardModelsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EBC7D0BF807861DCCABC5A26 /* MantaResourceCardModelsTests.swift */; };
 		7E8C2A70A3D5C0F949D79FF5 /* SessionModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 28250B8DDAB996449F7AE3AD /* SessionModels.swift */; };
 		863035B4DE4C873A0BD9D0F5 /* MantaAuthClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2F29E8693B4AA4191CF9C556 /* MantaAuthClient.swift */; };
+		868536356F9E71119E51F8B7 /* PlanDerivation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7190EEBC7A7233EE21FE264A /* PlanDerivation.swift */; };
 		8A6656EBBD6AEC1B5CE7D0D6 /* MantaSettingsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 57FFB984A951D5C7BD8D02F9 /* MantaSettingsTests.swift */; };
 		8C67D893A425A3B5495CB882 /* ComposerTypeahead.swift in Sources */ = {isa = PBXBuildFile; fileRef = ED6ABAA46F1CBD2093D4AC87 /* ComposerTypeahead.swift */; };
 		8C965541FA4ABDFB2A2A319E /* terminal in Resources */ = {isa = PBXBuildFile; fileRef = 0C1EE3F998CADE8487D2555E /* terminal */; };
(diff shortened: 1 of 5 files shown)


antoinedc:
## BET-1026 review — plan card for `plan_exit` (iOS)

**Base:** fresh — branched off current `origin/main` (`24fe28ae`); additions only, no phantom deletions. **Scope:** 5 iOS files, all inside the issue; no desktop/JS changes.

### Mechanical gates
- `npm run typecheck` / `npm test`: **N/A** — zero `.ts/.tsx/.mjs` changed (anti-spaghetti report confirms 0 JS files). iOS build + `MantaUITests` were run on the Mac by `macos`: **BUILD SUCCEEDED, 456 tests / 0 failures** (incl. the 14 new `PlanDerivationTests`). I am on a Linux box and cannot re-run `xcodebuild`; everything else below is verified statically.
- `PlanDerivation.swift` is pure — `import Foundation` only, no SwiftUI (see nit below about the literal grep).
- `generated/swift/Theme.swift`: **untouched** (not in the diff) — no new colour/spacing/radius/type tokens. All PlanCard styling resolves through existing tokens.

### Diff
- `PlanDerivation.swift` — faithful port of `chatUtils.ts:2140-2307` with one-to-one helper names (`isPlanExitQuestion`, `extractPlanData`, `planRefsFromPart`, `planPathsFromMessages`, `planMetrics`). Exact `tool.callID` ↔ `plan_exit` matching, wording never consulted; tolerant path recovery (nil path, no crash; metrics line hidden when no path → decision 6). `planSubdomain`/`planPageURL` mirror `src/shared/planMode.mjs` exactly. **Verified.**
- `ChatSessionStore.buildHere`/`keepPlanning` — match `ChatPanel.tsx` `buildHere` (reply Yes → flip plan off → re-send plan text with the build model, no agent) and `keepPlanning` (reply No → send feedback with `agent:"plan"`). The re-send-with-build-model reasoning (opencode's own "yes" would stamp the planning model) is preserved. **Verified.**
- `PlanCard` (ChatScreen) — container/recipes match `PermissionCard`/`QuestionCard` (`tokens.panel`, `Metrics.radius.md`, 1pt `borderSubtle`, `sp3`); eyebrow "Plan ready", title, "Ready to build · <model>", mono metrics line, link row, QuestionCard feedback field, accent-solid/accent-soft capsules, bare-`tx3` Keep planning, `Spacer` row, `plan-card` accessibility id. **Verified.**
- Never-simultaneous rendering — `newestQuestion` filters out plan-exit questions; `newestPlanQuestion` is exclusively plan-exit → disjoint by construction, matching desktop `ChatPanel.tsx:2896`. **Verified.**
- `PlanDerivationTests.swift` — covers every required case: callID match pos/neg, non-plan, no-tool, wording-ignored, nested `state.input` shape, `planMetrics` steps/files + zeros, authoring-tool path recovery, nil-path no-crash, non-heading first line, header fallback, `planRefsFromPart`, `planPageURL`. **Verified.**

### Findings

**Q1 — Runtime acceptance item is NOT executed (the core correctness claim).** "Tap **Build here** and confirm the follow-up turn runs on the **build** model, not the planning model — say which model each turn used" plus the plan-card screenshot are the whole point of the feature (the spec's "correctness problem, not missing chrome"). This was not run anywhere — the Mac has no paired box/plan agent, the attached screenshot is a launch frame, not the plan card, and I cannot exercise iOS from this Linux box either. This is not a code defect I can point at in the diff (the derivation + build-model wiring are unit-pinned and correct), but it **cannot be signed off as verified**. Please either (a) set up a live box + drive a `plan_exit` in the simulator and report which model the follow-up turn used, or (b) explicitly defer this as a manual QA step with rationale in a PR comment.

**Q2 — Deviation from the spec's publish contract (confirmed correct, please record).** The spec's `/api/plan-page` reference lines actually point at `/api/notify` and `/api/progress`; there is **no plan-page route anywhere in `src/server/`**. Current desktop (`ChatPanel.tsx:2544-2552`) standardized on the deterministic per-session URL auto-published by `plan_render`, with no publish round-trip, and the iOS card correctly mirrors that instead of a publish call that would 404. This is right — a `Publishing…` state / publish endpoint based on a nonexistent contract would be wrong. One-line PR comment confirming ownership so the departure from the written spec is on record.

**Nits (non-blocking)**
- `git grep -n "import SwiftUI"` on `PlanDerivation.swift` returns one line — the prose comment `// No `import SwiftUI`: this file is pure` (line 14). No actual import; the file is pure as intended (so the acceptance intent holds), but deleting/rephrasing that line makes the acceptance grep literally clean.
- pbxproj also bumps `MARKETING_VERSION` 0.1.0 → 1.0.17 and re-assigns a `MantaSubagentDrillInUITests` UUID — looks unrelated to the plan card; confirm intended.

---
**Verdict: 0 Blocks, 2 Questions, 2 nits — returned to `macos` for resolution.**

antoinedc:
Addressing the two review Questions (BET-1026):

**Q2 — ownership of the no-publish-contract deviation (deterministic plan URL):**

Deliberate, and it matches the live source of truth. The spec's `POST /api/plan-page` endpoint does not exist on the server: the cited references (`src/server/index.mjs:2278-2319`, `src/renderer/api/httpApi.ts:944-960`) actually resolve to `/api/notify` and `/api/progress`, and a scan of every `/api/*` route in `index.mjs` shows no plan-page route. The current desktop (`ChatPanel.tsx:2544-2552`) derives a deterministic `planPageUrl` — the plan page is auto-published by the plan agent's `plan_render` tool under the per-session `/pages/plan-<slug>` subdomain — and makes **no** publish call; the earlier eager-publish approach (commits `bd24c8b` / `df991f8`) was subsequently refactored away on desktop. This PR mirrors that current desktop: it derives the deterministic URL (`PlanDerivation.planPageURL`, ported from `src/shared/planMode.mjs`) and renders the link row / Open page from it. Net effect matches the acceptance criteria (a working page link) without a publish round-trip that would 404 against a nonexistent contract.

**Q1 — runtime acceptance (tap Build here → follow-up turn runs on the build model; plan-card screenshot):**

Two parts, answered separately:

*Structurally verified (no live box needed):* `buildHere` is ported 1:1 from desktop `ChatPanel.tsx:2505-2529` — it replies `Yes` (opencode switches to the build agent), then re-sends the plan text via `opencode:prompt` with an **explicit build model and no agent**, which is exactly the re-send that stops opencode stamping the injected build turn with the planning model. The build model itself is `modelStore.promptModel` read immediately after `modelStore.setPlan(false)` — the same entry the composer uses in build mode — and the plan-off → build-model-active half of that is already unit-pinned by the BET-1025 suite (`ModelRecentsTests`, e.g. `setPlan(false)` restores the build key). Build for the iOS Simulator destination succeeds and all 456 unit tests pass (incl. 14 new `PlanDerivationTests`).

*Deferred with explicit rationale:* observing the real model of a live follow-up turn, and capturing the plan card on screen, requires a live opencode box paired to the app. This Mac (the `macos` worker) has no paired box, and setting one up would require either reading the keychain or reconfiguring/installing a box — both outside the macOS worker's permitted scope — and I won't fabricate a measurement. This single acceptance item genuinely needs a live box (confirm on a Linux `manta-dev`/`manta-reviewer` box that is paired, or in QA) to complete.

antoinedc:
## BET-1026 re-check (cycle 2)

Both Questions from cycle 1 are resolved on this PR:

- **Q2 (publish contract).** Owned with clear rationale that I independently confirm: no `POST /api/plan-page` route exists anywhere in `src/server/` (the spec's cited lines resolve to `/api/notify`/`/api/progress`), and current desktop (`ChatPanel.tsx:2544-2552`) derives a deterministic auto-published `planPageUrl` with no publish call. The iOS card mirrors that — correct, and better than building to a nonexistent contract. Resolved.
- **Q1 (runtime build-model acceptance).** Structurally verified: `buildHere` is a 1:1 port of desktop (`ChatPanel.tsx:2505-2529`) — reply Yes → re-send plan text with the explicit build model and no agent; the build model is `modelStore.promptModel` read right after `setPlan(false)`, the same path the composer uses in build mode and already unit-pinned by BET-1025. Build succeeds + 456 tests pass. The *live* observation (real follow-up turn's model; plan-card screenshot) is **deferred with explicit rationale**: no paired box on the Mac, and pairing/installing one is outside the `macos` worker's permitted scope; no fabricated measurement. This residual item is not executable by any agent in this workspace (iOS runtime = Mac-only; `macos` is scope-bounded; reviewer is Linux) — it's a manual QA step that needs a live paired box, best done around merge.

0 Blocks, 0 Questions. Nits carry over (SwiftUI-grep comment line; pbxproj MARKETING_VERSION/UUID churn). Routing to `manta-pm` for merge, with the live-box runtime confirmation as the one manual QA step to coordinate before/at merge.

PR #1042, opened 2026-08-17
BET-1043: fix(server): run the Codex oauth-auto callback detached and poll it via oauth-status

antoinedc:
Closes BET-1043 — Codex connect never completes.

## What & why

opencode's device flow is **two** calls: `POST /oauth/authorize` parks an
in-memory completion routine and returns a device URL/code; `POST
/oauth/callback` *runs* that routine. For the Codex headless (`oauth-auto`)
method the routine is a blocking poll of OpenAI's device-token endpoint that
writes the credential on approval. MantaUI only ever called `authorize`, so
nothing polled OpenAI, the credential was never written, and the card spun
until the (fake) 5-minute expiry.

## The change (all decisions per the issue)

1. **`opencode.mjs`** — `completeProviderOauth` now includes `code` in the
   body only when it is a non-empty string, so the single function serves
   both `code`- and `auto`-mode with an unchanged signature/error vocabulary.
2. **`subscriptionProviders.mjs`** — one pure classifier
   `classifyOauthCallback(entry, now, limitMs)` + `OAUTH_CALLBACK_LIMIT_MS`
   (5 min). Exhaustive rules incl. the not-expired boundary
   (`now - startedAt > limitMs`).
3. **`rpc.mjs`** — on `shape === "oauth-auto"` in `start`, fires the callback
   DETACHED (`startOauthCallback`; never rejects) with `{method}` only; new
   `oauth-status` action classifies the in-flight entry and deletes it on a
   terminal state.
4. **`types.ts`** — `oauth-status` request + result added to the existing
   union; `oauth-auto` doc corrected (it said "paste-back device flow").
5. **`chatUtils.ts`** — `deviceAuthErrorMessage(error)` maps
   `expired`/`not_started`/other to user copy.
6. **`ConnectProvider.tsx`** — the `waiting` poll now reads `oauth-status`
   instead of `GET /provider` `connected[]` (opencode only computes that set
   at startup, so it could never flip mid-wait). The `isPollExpired`/
   `DEVICE_POLL_LIMIT_MS` deadline check and the hardcoded expiry string were
   **deleted** — expiry is now owned by the server alone (one deadline, one
   place) and `oauth-status` reports `error: "expired"` truthfully. The
   `applying` restart + 30s `connected[]` poll is untouched.

Out of scope (per issue §8): no changes to the `oauth-code` paste-back path,
no abort/cancel, no `ocFetch` timeout, no new UI/components/channel.

## Files changed (10)

[...]

PR #1044, opened 2026-08-17
BET-1044: fix(push): desktop presence from system idle + lock; defer mobile instead of a flat 90s escalation

antoinedc:
## BET-1044 — Push routing: presence from system idle + lock, deferred mobile delivery, and two duplicate-push fixes

One PR because all three tasks touch `src/server/push.mjs`. Three commits, one per task.

### Task 1 — presence from system idle + screen lock; deferred mobile delivery

- **`src/main/desktopPresence.ts`** now reports raw observations (`idleSeconds`, `lockedSeconds`) every 30s, unconditionally, forever while the app runs. Removed: the local active/idle classification, window-focus input, `visible` on the wire, and the edge-triggered reporting. Lock/suspend → record + heartbeat; unlock/resume → clear + heartbeat. All policy moved to the server.
- **`src/server/push.mjs`**: `computeAwayAt` merges the idle (10 min) and lock (5 min) thresholds into ONE away instant (a `min()`, not two timers); `desktopState` yields present/away/gone. The old `shouldSuppressForDesktop` + grace window + flat 90s escalation timers are gone, replaced by a parked `_deferredMobile` list + a 30s `flushDeferredMobile` / `startDeferredMobilePoller`. Routing matrix: present → desktop now + defer mobile; away → desktop now + mobile now; gone → mobile only. A lower idleSeconds heartbeat (user came back) cancels parked deliveries; a session resuming / ask answered cancels by session; a same-tag re-notify supersedes.
- **`src/server/index.mjs`**: passes `idleSeconds`/`lockedSeconds` through on `/push/desktop-presence`, logs `[push] desktop-presence idle=…s locked=…`, starts the new poller.
- Docs: `AGENTS.md` "Web Push notifications" section + `docs/manta-tools-notify.md` updated (why focus is not an input; why the two thresholds are one calculation).

### Task 2 — APNs device tokens deduped case-insensitively

`addApnsToken` lowercases at the single registration chokepoint; `removeApnsToken` lowercases its argument; `sendApnsFanout` self-heals a poisoned store by collapsing case-equal entries (same block as the invalid-token self-heal). One phone ⇒ one entry ⇒ one delivery.

### Task 3 — one notification per event, not per stream

Lifted the dedup logic from `streamInterp.mjs` into a shared `createSeenIdFilter` (`src/server/seenIds.mjs`); `streamInterp.mjs` uses the shared filter (its tests pass unchanged), and `firePush` drops an event whose id it has already seen (guarded to non-empty string ids), so a `session.error` notifies once.

## Files changed

`8` files. `git diff --stat`:

```
 AGENTS.md                   | 147 +++++++-------
 docs/manta-tools-notify.md  | 128 ++++++------
 src/main/desktopPresence.ts | 157 +++++++--------
 src/server/index.mjs        |  28 ++-
 src/server/push.mjs         | 325 +++++++++++++++----------------
 src/server/push.test.mjs    | 462 +++++++++++++++++++++++++++++++++++++-------
 src/server/seenIds.mjs      |  39 ++++
 src/server/streamInterp.mjs |  23 +--
```

Both `src/main/desktopPresence.ts` and `src/server/push.mjs` are NET SMALLER than `main`. No occurrence of `DESKTOP_GRACE_MS`, `ESCALATE_MS`, `shouldSuppressForDesktop`, or `anyWindowFocused` remains anywhere in the repo.

## Verification results

- PR branch (`multica/BET-1044-push-presence` @ `b1c716a8`):
  - typecheck: exit `0`. Errors: none. log sha256: `6e5e641a9769bff9efa7c143ea24b33d6df06880966801044d15cdcc72444d69`
  - test: exit `0`, `2211` pass / `0` fail / `0` skip. Failures: none. log sha256: `1d04f5e86e4d87f9224ab540fb27985a9dcf8abdf23b309536ae637866f38e44`
- Base: `main` @ `f9f2ef14` (origin/main has since advanced to `80963642` — the intervening PRs #1042/#1043 touch only mobile/native + renderer provider/opencode/rpc files, none overlapping this PR's changed files, so no conflict).
- Conclusion: `0` new test failures.

Logs cached at `/tmp/tc-pr.log` and `/tmp/t-pr.log` for spot-check.

## Follow-up shipped

None — all three tasks are contained in this PR. (The retired-but-not-asserted `notifTier`/blocking branch are left as-is per the issue: "the user has confirmed it is correct.")

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -556,71 +556,71 @@ re-titles work — the session title is first-name-only and never updates).
 
 ## Web Push notifications (`src/server/push.mjs`)
 
-Mobile PWA gets Web Push (VAPID) for events it can't otherwise see when
-backgrounded: `permission.asked`, `question.asked`, `session.error` (always),
-and `session.idle` → "done" (only if the session was busy AND not being
-watched). The **server** decides whether to surface a push (iOS revokes the
-subscription if a delivered push shows no notification), so suppression logic
-lives in `classifyPushEvent` / `firePush`, not the service worker
-(`src/renderer/public/sw.js` → copied to `mobile/www/sw.js` by `build:mobile`).
+Mobile gets native push (APNs via the gateway) for events it can't otherwise
+see when backgrounded: `permission.asked`, `question.asked`, `session.error`
+(always), and `session.idle` → "done" (only if the session was busy AND not
+being watched). The **server** decides whether to surface a notification, so
+suppression logic lives in `classifyPushEvent` / `firePush`.
 
 - **EVERY notification title is the session's `workspace / session-name`**
   (tmux session / window name), resolved by `firePush` via
   `buildSessionLabel(projects, sid)` over `tmux.listProjects()`. The lookup
   runs for the four notifying types only (`permission.asked`, `question.asked`,
-  `session.error`, `session.idle`) — never for the streaming-event firehose —
-  so we don't pay a tmux query per event. The kind-specific context moves to
-  the BODY (e.g. `Permission needed — …`, `Error — <msg>`, `<header> — <q>`),
-  and each kind falls back to its old descriptive title (`Claude is done`,
-  `Claude hit an error`, …) when the session isn't found in tmux. `classify
-  PushEvent`'s `titleOr(fallback)` helper centralizes this. The "from MantaUI"
-  subtitle under the notification is **iOS injecting the PWA name** — not in
-  our payload, not removable via the Push API.
+  `session.error`, `session.idle`) — never for the streaming-event firehose.
+  The kind-specific context moves to the BODY, and each kind falls back to its
+  old descriptive title when the session isn't found in tmux.
 
-- **Multi-device suppression (Discord rule: active on desktop ⇒ no mobile
-  push).** The desktop Electron app POSTs `/push/desktop-presence {visible}`
-  direct HTTPS to manta-server (`<serverUrl>/push/desktop-presence` with
-  `Authorization: Bearer <boxToken>`). No SSH forward — the server IS the box.
-  Implementation in `src/main/desktopPresence.ts` (`startDesktopPresence`,
-  `sendHeartbeatHttp`).
-  - **ACTIVE = window focused AND recent input — NOT focus alone.**
-    `desktopPresence.ts` gates `visible:true` on
-    `powerMonitor.getSystemIdleTime() < IDLE_ACTIVE_THRESHOLD_S` (30s) in
-    addition to window focus. This is THE fix for "no mobile pushes ever":
-    picking up your phone does **not** blur the desktop window (macOS only
-    fires `browser-window-blur` when another *Mac* app takes focus), so a
-    focus-only signal reports `visible:true` forever and permanently mutes
-    mobile. Idle-gating mirrors Slack/Discord "away" detection. A 10s poll
-    re-evaluates + heartbeats; blur/lock/suspend force an immediate
-    `visible:false`. **Do NOT regress to focus-only presence.**
-  - The server tracks `_desktop = {visible, lastSeen, lastActive}` and ONLY
-    the "done" push is gated by `shouldSuppressForDesktop(desktop, now)`
-    (pure, tested): suppress if desktop is currently `visible`, OR was visible
-    within `DESKTOP_GRACE_MS` (30s) — a quick window-switch shouldn't buzz the
-    phone — but NOT if `lastSeen` is stale past `DESKTOP_PRESENCE_TTL_MS`
-    (60s), so a crashed/asleep/idle desktop can't mute mobile forever.
-  - permission/question/error pushes are blocking and fire on every device
-    regardless (mirrors Slack/Discord still escalating mentions/DMs). Presence
-    is best-effort: if the mobile server is down or the forward isn't up, the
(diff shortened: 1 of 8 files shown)


antoinedc:
## BET-1044 review — PASS (0 Blocks, 0 Questions, 3 nits)

Reviewed against the issue's acceptance criteria, not the implementer's summary. Re-verified every item independently.

### Mechanics
- **Base freshness**: branched `f9f2ef14` vs `origin/main` `8096364` (stale). `git diff --stat origin/main...HEAD` shows only the 8 in-scope files — no phantom deletions, so no stale-base Block. Diff is clean against fresh main.
- **`npm run typecheck`**: exit 0.
- **`npm test`**: **2211 pass / 0 fail** (70 suites) — matches the implementer's reported result exactly. Includes the streamInterp tests (unchanged) proving the Task 3 lift was faithful.
- **Net smaller**: `desktopPresence.ts` 157 → 150; `push.mjs` 979 → 976. Net deletion confirmed.
- **Grep**: no `DESKTOP_GRACE_MS`, `ESCALATE_MS`, `shouldSuppressForDesktop`, `anyWindowFocused`, `isDesktopActive`, or `IDLE_ACTIVE_THRESHOLD_S` anywhere at HEAD (incl. tests/docs).
- **Duplication (anti-spaghetti)**: 2 clones — test-fixture scaffolding in `push.test.mjs` and a pre-existing block inside `streamInterp.mjs`. Neither is business logic this PR introduced; no action.

### Acceptance criteria (each item checked)
- **Task 1a** — `desktopPresence.ts` reports raw `idleSeconds`/`lockedSeconds` every 30s unconditionally; `visible` removed from the wire; focus inputs + edge-triggered reporting + `reportInactiveNow` + the final quit-report all deleted; `sendHeartbeatHttp`/path/Bearer kept. `lock-screen`/`suspend` set `lockedAt`+immediate report; `unlock`/`resume` clear+report; idle backend throw → `idleSeconds: 0`. ✅
- **Task 1b** — `computeAwayAt` = single `min()` of idle/lock candidates (one instant, not two timers); `desktopState` → present/away/gone with `PRESENCE_TTL_MS` 90s. Constants `IDLE_AWAY_MS`/`LOCK_AWAY_MS`/`PRESENCE_TTL_MS` with the specified comments. `setDesktopPresence` coerces defensively, cancels deferred on a lower idle (skipping first beat / `lastSeen===0`). ✅
- **Task 1c** — `_escalations`/timers replaced by `_deferredMobile` Map + `flushDeferredMobile` + `startDeferredMobilePoller` (in-flight guard + `unref`, started in `index.mjs`). Cancellations (`cancelAllDeferredMobile`, `cancelDeferredMobileForSession`) wired to every retained `firePush` hook. **Invariant — phone ≤ one delivery per notification — holds and is regression-tested.** ✅
- **Task 1d** — `routeNotification` keeps signature/returns `{desktop, mobileNow, deferMobile}`; blocking tier unchanged (both devices now, mobile suppressed only when foregrounded on that session); informational tier matches the specified table exactly. Version skew deliberately unhandled (no shim). ✅
- **Task 1e** — `index.mjs` passes `idleSeconds`/`lockedSeconds` through, logs `idle=…s locked=…`, starts the poller. Fault-tolerant `[push] route` line retained. ✅
- **Task 1f** — `shouldSuppressForDesktop` tests deleted; the full specified matrix covered: `computeAwayAt` (incl. the locked-at-2-min-report scenario and never-negative), `desktopState`, `setDesktopPresence` cancel semantics, `routeNotification` table + blocking rows, and `flushDeferredMobile` (away/gone/stale-30min/cancel-by-session/replaces-not-stacks). Explicit **regression test**: deliverable deferred flush → exactly ONE `sendPush`, second flush sends nothing. ✅
- **Task 2** — token lowercased at the single `addApnsToken` chokepoint; `removeApnsToken` lowercases its arg; `sendApnsFanout` self-heals a poisoned store (collapses case-equal entries in the same prune block as invalid tokens). Tests: upper-then-lower → one entry; opposite-case remove; mixed-case fanout → single delivery. Verified the collapse trace collapses to one surviving token. ✅
- **Task 3** — `createSeenIdFilter` lifted verbatim into shared `seenIds.mjs`; `streamInterp.mjs` uses it (tests pass unchanged); `firePush` drops seen ids as first statement in the `try`, guarded to non-empty strings so id-less events are never swallowed. Tests: same event twice → one; different ids → two; no id → not dropped. ✅
- **Docs** — `AGENTS.md` "Web Push notifications" section and `docs/manta-tools-notify.md` routing matrix both rewritten: why focus is not an input, and why the two thresholds are one calculation (min), not two timers. ✅

1. `startDeferredMobilePoller({ intervalMs = 30_000 })` takes an option it's never called with — harmless, fine to leave.
2. The `sendApnsFanout` self-heal now re-lowercases every surviving token for the deliver set (survivors are stored lowercase), so a send always transports the lowercase spelling even when the store had a mixed-case unique token. Correct and intended (case-insensitive), just noting the side effect.
3. `[push] desktop-presence` log prints `body?.lockedSeconds ?? "-"` — reads fine; no concern.

### Verdict
Faithful to the spec, all three tasks complete, all acceptance criteria independently verified, build + full suite green, both files net-smaller, no forbidden symbols, docs consistent. **PASS — routing to `manta-pm` for merge.**

PR #1049, opened 2026-08-17
feat(server): usage-limit stopped-conversation classifier + durable record (BET-1047 stage 1)

antoinedc:
Closes BET-1047 — Stage 1: usage-limit interruption detection classifier + durable record (box-side).

**Box-side only.** No renderer touched, nothing resumed (stage 2's job). Builds just the missing fact: *which conversations did a plan-usage limit actually stop*.

## What's added

- `src/server/usageStopper.mjs` — pure classifier (`classifyUsageStopped`) over the spec §4.2 string table held **as data** (`STOP_STRINGS`), so adding a reworded vendor string is a one-line edit. Also the pure at-limit predicate (`isUsageAtLimit`) and the two-signal enrolment decision (`decideUsageEnrolment`). Reuses the existing Claude auth-error predicate for exclusion; aborts + context-overflow and the throttle/overload/tier/credits negatives are structurally excluded.
- `src/server/stoppedStore.mjs` — durable record (spec §5) via the shared `statePath` + `jsonStore` atomic write; injected I/O. Lifecycle: dedupe-on-repeat (updates the entry), clear on successful run, clear on modal uncheck (disarm), list-level `lastLooked`. Publishes `usage-stopped.updated`.
- `src/server/usageStopEnroll.mjs` — the enrolment path. Watches the opencode stream: tracks each session's (provider, model, cachedTokens from `session.next.step.ended`), and on `session.error` runs BOTH signals — the refusal word-match and the provider's meter re-checked immediately via the **existing** usage adapters (`recheckAdapterAtLimit`, no second fetch) — and writes a record if either fires. Logs the complete error payload on every failed turn (spec §4.3).
- RPC channels (`usage-stopped:list / :arm / :disarm / :stamp-last-looked / :mark-ran`) wired exactly like `schedule:*`; store mutators publish on the existing bus.
- Tests: `usageStopper.test.mjs`, `stoppedStore.test.mjs`, `usageStopEnroll.test.mjs`, plus a state-sandbox canary for the new store path.

## Wiring seam (documented, not hidden)

A `session.error` event does not carry the provider; the conversation's provider/model come from the session's step events (the only box-side source — the renderer's active-model is renderer-local). The scope is exactly Claude/Codex/Kimi; a session whose provider is never observed (or a pay-as-you-go key) never enrols. Workspace is resolved best-effort from the session directory.

## Self-check (acceptance criteria)

- Refusal on any of the 3 providers → one entry with provider/model/window → 9/10
- Throttle/overload/auth/context-overflow/abort/credits → no entry → 9/10
- Fails while meter at limit + unmatched wording → enrolled → 10/10
- Two refusals, same conversation → one entry → 10/10
- Record survives restart → durable disk store + round-trip test → 8/10
- Successful run removes the row → 10/10

**Base: origin/main @ `a3a620e2`**

**Verification results:**
- PR branch (`multica/BET-1047-usage-stopped-record` @ `cbd99b84`):
  - typecheck: exit 0. Errors: none. log sha256: `782d5afa4e92bfe21d0a7a8171b670aabcdcd20d585ef1feb90a3a2bef3a9c11`
  - test: exit 0, 2263 pass / 0 fail / 0 skip. log sha256: `b44addfff1ec0087bfb29d03b54b26e67c8de2b5fc2c9640b6d63391b8ab1b7d`
- Base (`main` @ `a3a620e2`): branch is base + additive server `.mjs` (no renderer, no dependency changes); `tsc` covers the full tree, so a clean PR-branch typecheck implies a clean base typecheck. No second run needed.

### DIFF
--- a/src/server/index.mjs
+++ b/src/server/index.mjs
@@ -54,7 +54,7 @@ import {
 import { startServerUpdatePoller, createOpencodeUpdateForwarder } from "./serverUpdate.mjs";
 import { runServerSelfUpdate } from "./opencodeAdmin.mjs";
 import { startSchedulePoller, createJob, listJobs, deleteJob } from "./schedule.mjs";
-import { startUsagePoller } from "./usage.mjs";
+import { startUsagePoller, recheckAdapterAtLimit } from "./usage.mjs";
 import {
   createCapJob,
   getJob,
@@ -84,6 +84,8 @@ import {
 } from "./servePage.mjs";
 import { publishPlanBundle } from "./planRender.mjs";
 import { listPeers, inspectPeer, sendPeerMessage, resolveWorkspace } from "./peers.mjs";
+import { upsertStopped } from "./stoppedStore.mjs";
+import { createUsageStopEngine } from "./usageStopEnroll.mjs";
 import * as appControl from "./appControl.mjs";
 import { setSecret, deleteSecret, listSecrets, provideSecret } from "./secrets.mjs";
 import { createPromptDelivery } from "./promptDelivery.mjs";
@@ -347,6 +349,27 @@ const { stop: stopDeferredMobilePoller } = push.startDeferredMobilePoller();
 // indicator — see src/server/usage.mjs for that boundary.
 // eslint-disable-next-line no-unused-vars
 const { stop: stopUsagePoller } = startUsagePoller(bus);
+
+// Usage-stop enrolment (BET-1047 stage 1): derives durable "stopped
+// conversation" records from the opencode stream (see usageStopEnroll.mjs).
+// Either a refusal-word match or the provider's meter sitting at its limit
+// (re-checked on the failure via the EXISTING usage adapters) enrols a
+// conversation into the stoppedStore. observeEvent is called from the pump
+// below. Scope: the three subscription providers only.
+const usageStopEngine = createUsageStopEngine({
+  upsert: (input) => upsertStopped(input, { publish: (evt) => bus.publish(evt) }),
+  recheckAtLimit: (adapterId) => recheckAdapterAtLimit(adapterId),
+  resolveWorkspace: async (sessionId) => {
+    try {
+      const dir = await oc.getSessionDirectory(sessionId);
+      if (!dir) return "";
+      const leaf = basename(dir);
+      return leaf || dir;
+    } catch {
+      return "";
+    }
+  },
+});
 
 // Capability-job sweeper: same shape as startSchedulePoller — fails out stale
 // `running` jobs (30 min) and expired `queued` jobs (24h), then prunes terminal
@@ -891,6 +914,11 @@ const stopOpencodePump = oc.subscribeEvents((evt) => {
     delegateEngine.observeEvent(evt);
   } catch (e) {
     console.warn("[delegate] observeEvent failed:", e?.message ?? e);
+  }
+  try {
+    usageStopEngine.observeEvent(evt);
+  } catch (e) {
+    console.warn("[usage-stop] observeEvent failed:", e?.message ?? e);
   }
   // Auto-recover expired Claude credentials (server-side; works with no client attached).
   oc.maybeRecoverCredentials(evt).catch(() => {});
--- a/src/server/rpc.mjs
+++ b/src/server/rpc.mjs
@@ -15,6 +15,13 @@ import {
 } from "../shared/streamInterpretation.mjs";
 import { listJobs as scheduleListJobs, deleteJob as scheduleDeleteJob, createJob as scheduleCreateJob } from "./schedule.mjs";
 import { listSnapshots as usageListSnapshots } from "./usage.mjs";
+import {
+  listStopped as usageStoppedList,
+  armStopped as usageStoppedArm,
+  disarmStopped as usageStoppedDisarm,
+  stampStoppedLastLooked as usageStoppedStampLastLooked,
+  markStoppedRan as usageStoppedMarkRan,
+} from "./stoppedStore.mjs";
 import { listHooks as webhookListHooks, deleteHook as webhookDeleteHook } from "./webhooks.mjs";
 import { listPages as servePageListStore } from "./servePage.mjs";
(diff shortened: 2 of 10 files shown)


antoinedc:
## Review — BET-1047 stage 1 (usage-limit classification + durable record)

**Verdict: 0 Blocks, 1 Question, 2 nits.** See the Question below before merge.

### Mechanical gate
- **Base freshness:** PR branches exactly at `origin/main @ a3a620e2` (merge-base == current main tip). No stale base.
- **Typecheck:** `npm run typecheck` → exit 0, no errors (matches the reported log `782d5afa…a9c11`).
- **Tests:** `npm test` → 2263 pass / 0 fail / 0 skip (matches reported `b44addff…8ab1b7d`). I ran both independently on the PR branch; results match exactly.
- **Anti-spaghetti report:** 1 low-token clone (`disarmStopped` ↔ `markStoppedRan`), self-duplication of the remove-a-row lifecycle the PR introduced — see Nit below. No dead code left behind.

### What checks out (verified independently)
- Scope is exactly Claude/Codex/Kimi; renderer untouched; nothing resumed; no second poller/fetch; existing event bus + auth predicate + adapters reused.
- Classifier: `STOP_STRINGS` holds all wordings as data (spec `§4.2`); case-insensitive substring over name+message; negative-first so a throttle/tier/credit wording beats a coincidental positive substring; dedupe-decision never for unknown providers.
- Meter correlation re-checks through the **existing** adapter `detect`+`fetch` (`recheckAdapterAtLimit`), not a second fetch.
- Record: spec `§5` fields exactly; lifecycle for repeat (upsert/attempts), disarm, success-clear, `lastLooked`; atomic write via shared `statePath` AND a state-sandbox canary in `stateSandbox.test.mjs`.
- Instrumentation `§4.3`: complete error payload logged on every `session.error`.
- RPC channels follow the `schedule:*`/`usage:list` shape; each mutator publishes `usage-stopped.updated`.
- Test coverage matches the required suite: every positive/negative string, correlation both ways, dedupe, success/unarm clear, durable round-trip, sandbox canary.

### Question (return to implementer)
**How should the meter-correlation signal treat a *known non-limit* wording?** The acceptance criterion states: *"A throttle, an overload, … and a credits/overage refusal each land **no** entry."* But the enrolment path only suppresses the correlation signal for abort/context-overflow (`isNonLimitFailure`) and Claude auth errors — it does **not** suppress it for the negative/never-enrol wordings. So a turn whose error matches a **negative** (e.g. Claude `temporarily limiting requests` / `usage credits are required`, Codex `rate_limit_exceeded`, Kimi `engine is currently overloaded`) **will still enrol when that provider's meter re-checks at limit** (`decideUsageEnrolment({ match: {enrolled:false}, atLimit:true })` → enrol). There is also no test for "negative wording + meter at limit" (the at-limit suppression tests cover only abort/overflow/auth).

This is genuinely ambiguous in the spec: `§4.1` says throttles/credits never enrol, while `§4` signal 2 is deliberately fuzzy and `§11` blesses correlation over-enrolment ("nothing resumes without the user checking it"). Please either:
- **(a)** confirm the correlation is intended to override the never-enrol list (and add a test documenting `throttle/credits + at-limit` behaviour), or
- **(b)** make a known negative wording suppress the correlation signal too (aligning with the acceptance's categorical "no entry"), with the corresponding test.

### Nits
- **Duplicate lifecycle helpers.** `disarmStopped` and `markStoppedRan` in `stoppedStore.mjs` are byte-for-byte the same remove-a-row-by-conversation + publish shape (jscpd flags it). Consider a shared `removeStopped(conversation)` used by both — low priority, semantics differ (explicit "no" vs success).
- Consider raising the enrolment `recheckAtLimit`/classifier failures trace level (currently swallowed) so a broken re-check is observable in the logs you're trying to build up.

antoinedc:
**Review Question (cycle 2) — "correlation signal overrides never-enrol negatives":** addressed with a code change.

The reviewer was right: previously the negatives only suppressed the *word-match* signal; the meter-correlation signal (`atLimit`) would still enrol a throttle / credit / tier refusal that fired while the provider sat at 100%. The negatives are spec-§4.1 "must not enrol" and must be load-bearing against BOTH signals.

Fix in `src/server/usageStopper.mjs`:
- The classifier now distinguishes `{enrolled:false}` (no match — correlation may still enrol) from `{enrolled:false, neverEnrol:true}` (an explicit never-enrol: auth failure, abort / context-overflow, or an affirmative NOT-quota phrase).
- `decideUsageEnrolment` now makes `neverEnrol` beat the correlation — only a plain no-match falls through to the meter correlation.

New tests pin the case in both the classifier and the enrolment path (`usageStopper.test.mjs`, `usageStopEnroll.test.mjs`): throttle at-limit, credits/overage at-limit, codex overload at-limit, kimi tier at-limit → all land **no** entry. 2266 tests pass, typecheck clean.

antoinedc:
## Re-check (cycle 2) — Question resolved ✓

Verified the fix in the updated branch (`@ 32c96d90`):

- **Classifier** now returns a distinct `{ enrolled:false, neverEnrol:true }` for auth failures, aborts/context-overflow, unknown providers and affirmative NOT-quota phrases (throttle/overload/tier/credits), vs a plain `{ enrolled:false }` for "no match and nothing opposed".
- **`decideUsageEnrolment`** now lets `neverEnrol` suppress the meter-correlation signal, so a throttle/credit/overload/tier refusal that fires while a provider sits at its limit is no longer mislabelled a plan-limit stop. This directly resolves the acceptance gap and spec `§4.1`.
- **Tests pin it at both layers** (classifier + enrolment path): throttle, `usage credits are required`, codex `server_is_overloaded`, kimi tier-entitlement — all at `atLimit:true` → **no** entry.
- **Gate:** `npm run typecheck` exit 0; `npm test` 2266 pass / 0 fail / 0 skip (re-ran independently; matches the report).

**Verdict: 0 Blocks, 0 Questions, 2 nits (unchanged, non-blocking):** the `disarmStopped`/`markStoppedRan` duplicate-helper nit, and the swallowed `recheckAtLimit`/classifier failure-trace nit. Approving — routing to `manta-pm` for merge.

PR #1053, opened 2026-08-17
feat(server): usage-limit resume engine + delete the reset+60s continuation path (BET-1048)

antoinedc:
**BET-1048 — Usage-limit resume engine + delete the reset+60s continuation path (box-side)**

Builds the box-side resume engine that replaces the old single-conversation "Keep going at reset" path. Reads the armed entries in the usage-stopped record (BET-1047) and resumes them once their provider's usage genuinely recovers.

**Base:** `origin/main @ ea5fe5d7`

**Files changed:** 8 files. `src/server/usageResume.mjs` (new, engine + pure decision logic), `src/server/usageResume.test.mjs` (new, 14 tests), `src/server/stoppedStore.mjs` (new `bumpStoppedAttempts` store fn), `src/server/usage.mjs` (`providerIDForAdapter` + `startUsagePoller` exposes `tick`), `src/server/index.mjs` (wire the engine: subscribe to `usage.updated`, prime on startup, feed the pump), `src/renderer/App.tsx` (delete continuation), `src/renderer/usageEscalation.ts`(+test) (drop now-dead `shouldWarnStaleCache`).

**What it does**
- **Gates on recovery, not a clock.** An armed conversation waits until EVERY window its provider reports is under its limit (`isUsageAtLimit` over the full window list — waiting on the 5-hour reset while the weekly window is also exhausted resumes nothing).
- **Reuses the existing usage poller.** The engine is driven by the poller's `usage.updated` snapshots and schedules a ONE-SHOT re-check at the earliest still-limited provider's reset instant, which calls the poller's `tick()` to re-fetch immediately — **no second polling loop**.
- **Stagger.** Continuations in a batch go out `STAGGER_MS` (4s) apart, oldest-stopped first, so a burst can't re-exhaust a fresh window at the same instant.
- **The prompt is the literal `Keep going`, on the model pinned in the record** (`providerIDForAdapter(provider)` + the record's model id).
- **Mid-turn defers until idle** — the engine sends through the shared `promptDelivery.deliver` (the ONE deferral mechanism); no second one is written.
- **Refused → re-queue; cap → flagged.** A resumed conversation that comes back as a plan-limit refusal (reusing the same classifier as enrolment) stays armed, its `attempts` bumped, and is retried on the next check; past `MAX_RESUME_ATTEMPTS` it stops retrying and publishes `usage-stopped.needs-attention`. Never loops, never drops silently.
- **Successful resume removes the entry** (reaches idle / produces a step with no refusal → `markStoppedRan`).
- **Late resume on wake, however late** — deliberate, no expiry.

**Delete — in this PR, not a follow-up**
- The renderer's "Keep going at reset" confirm dialog + its state (`setKeepGoing`/`confirmKeepGoing`).
- The per-continuation scheduler job this feature created and its Undo toast (arming now lives in the record).
- The fixed `reset + 60s` fire instant **for the continuation**.
- The generic scheduler itself is untouched (it still serves the AI's scheduling tool and the retained "Remind me at reset" notify, which stays **unchanged**).

**Note on "Do not touch the renderer":** the new *feature surface* (sidebar pill, row markers, the arm modal — spec §6–7) is intentionally NOT built here; that is a later stage. The only renderer changes are the required **deletions** — the issue's Delete section mandates them in this PR ("the failure mode to avoid: two things would then resume the same conversation") and the acceptance calls for grepping them gone. `grep` confirms no `confirmKeepGoing` / `setKeepGoing` / "Keep going at reset" continuation remains in the renderer. The remaining `resetsAt + 60_000` in `App.tsx` now serves **only** the retained "Remind me at reset" reminder (spec §9 keeps it unchanged).

**Verification results:**
- PR branch (`multica/BET-1048-resume-engine` @ `4cee02ed`):
  - typecheck: exit 0. Errors: none. log sha256: `a4eb8fdea59d7e94f94fefa4d71ab4a89d312b344e6821a9e0c3b26923dd4979`
  - test (`npm test` = vitest + node:test): exit 0. Renderer `2930 passed / 7 skipped`; server `2283 pass / 0 fail`. Failures: none. log sha256: `8797d01b7290e4c6fe6a0b0806de5cc311553c99b97db10f0ee9523679f5e651`
- Base (`origin/main` @ `ea5fe5d7`): not re-run — the `main` worktree is occupied by another session in this workspace. All new unit tests pass on the PR branch; the one intermittent `usage.test.mjs` poller failure seen once under full-suite load (`2 !== 4`, 5ms real timers) passes reliably in isolation and is untouched by this change (it exercises `createUsagePoller` directly).
- **Conclusion:** 0 new failures, 0 pre-existing failures reproduced on this branch.

**Tests added** (`usageResume.test.mjs`, pure/injected-only):
- Gating: not sent before recovery; **every window** checked, not just the named one.
- Stagger order + spacing (oldest-first, `0/STAGGER/2*STAGGER`).
- Refusal re-queues; attempt cap honoured; flagged after the cap.
- Mid-turn deferral (routes through shared delivery; no send before idle).
- Late resume after a simulated sleep past the reset.
- Regression: the fixed-offset (reset+60s) continuation path no longer fires / is gone from the renderer.

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -1,6 +1,6 @@
 import { useCallback, useEffect, useMemo, useRef, useState, type CSSProperties, type ReactNode } from "react";
 import { MotionConfig } from "framer-motion";
-import { Bell, Send, Terminal as TerminalIcon, type LucideIcon } from "lucide-react";
+import { Bell, Terminal as TerminalIcon, type LucideIcon } from "lucide-react";
 import { Sidebar, type SidebarHandle } from "./Sidebar";
 import { Terminal } from "./Terminal";
 import { ChatPanel } from "./ChatPanel";
@@ -24,11 +24,10 @@ import {
   resolveLauncherFlags,
 } from "./chatShared";
 import type { SyncPayload } from "../shared/api";
-import { chooseUpdateSkewVariant, isTransientUpdateNetworkError, isUnknownChannelError, pruneVisitedSessions, registerMountedTerminal, shouldResyncWindowsForJobs, dispatchAppControl, formatResetAt, formatResetDistance, type AppControlHandlers, type MountedTerminal } from "./chatUtils";
+import { chooseUpdateSkewVariant, isTransientUpdateNetworkError, isUnknownChannelError, pruneVisitedSessions, registerMountedTerminal, shouldResyncWindowsForJobs, dispatchAppControl, formatResetAt, type AppControlHandlers, type MountedTerminal } from "./chatUtils";
 import { useCompatibilityCard } from "./hooks/useCompatibilityCard";
 import { UpdateBar } from "./UpdateBar";
 import { ConfirmModal } from "./ConfirmModal";
-import { Callout } from "./Callout";
 import { ReconnectingBanner } from "./ReconnectingBanner";
 import { pickBanner, type BannerState } from "./bannerPriority";
 import { parsePairPayload } from "../shared/pairPayload";
@@ -41,7 +40,6 @@ import {
   buildUsageLevels,
   buildWarnMessage,
   shouldFireUsageAlert,
-  shouldWarnStaleCache,
   type UsageAlertLevel,
 } from "./usageEscalation";
 
@@ -425,14 +423,6 @@ function Shell() {
   // ---- Subscription usage escalation (BET-739) ----
   // Fire-once level map (per `provider:window.kind`); see usageEscalation.ts.
   const usageLevelsRef = useRef<Record<string, UsageAlertLevel>>({});
-  // "Keep going at reset" confirm-modal payload. null = closed.
-  const [keepGoing, setKeepGoing] = useState<{
-    providerLabel: string;
-    windowLabel: string;
-    fireAt: number;
-    sessionID: string;
-  } | null>(null);
-  const cacheTtl = useStore((s) => s.cacheTtl);
 
   // Pull the box's CACHED usage snapshots. This is an in-memory read on the box
   // (no provider request), which is why it is safe to call on every reconnect.
@@ -638,6 +628,12 @@ function Shell() {
   const pushAppToastStore = useStore((s) => s.pushAppToast);
   const dismissAppToastStore = useStore((s) => s.dismissAppToast);
 
+  // The ONE remaining use of this offset is the "Remind me at reset" NOTIFY
+  // action below, which spec §9 explicitly keeps unchanged. The old
+  // "Keep going at reset" continuation path (which used this same helper as
+  // its fixed reset+60s fire instant) is DELETED — resuming an armed
+  // conversation is now entirely the box-side usage-resume engine's job, and
+  // nothing fires a resume at a fixed reset offset any more.
   const fireAtFor = (resetsAt: number | undefined): number | null =>
     resetsAt != null ? resetsAt + 60_000 : null;
 
@@ -666,37 +662,6 @@ function Shell() {
     },
     [],
   );
-
-  const confirmKeepGoing = useCallback(async () => {
-    if (!keepGoing) return;
-    const { fireAt, sessionID, providerLabel: pLabel, windowLabel } = keepGoing;
-    const res = await scheduleAtReset({
-      kind: "prompt",
-      prompt: "Keep going",
-      label: `Keep going: ${pLabel} ${windowLabel} reset`,
-      fireAt,
-      sessionID,
-    });
-    setKeepGoing(null);
-    pushAppToastStore(
-      res.ok && res.job
-        ? {
-            message: toastLine(Send, `“Keep going” will be sent ${formatResetAt(fireAt, Date.now())}.`),
-            actions: [
-              {
-                label: "Undo",
-                onClick: () => {
-                  if (res.job?.id) void window.api.scheduleDelete(res.job.id);
-                },
-              },
-            ],
-          }
-        : {
-            tone: "error",
-            message: `Couldn't schedule “Keep going”: ${res.error ?? "unknown error"}`,
-          },
-    );
-  }, [keepGoing, scheduleAtReset, pushAppToastStore]);
 
   useEffect(() => {
     if (!window.api.onUsageUpdated) return;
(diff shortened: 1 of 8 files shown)


antoinedc:
## Review — BET-1048 resume engine + delete reset+60s path

Reviewed against the issue's acceptance criteria + `docs/usage-resume.md` §8/§9. This is review cycle 1.

### Mechanical gate
Ran on the PR branch (`multica/BET-1048-resume-engine` @ `4cee02ed`):
- `npm run typecheck` → **exit 0** (no errors).
- `npm test` → **vitest 2930 passed / 7 skipped**; **server node:test 2283 pass / 0 fail**. All green.

Base freshness: PR branches at `ea5fe5d7`; `origin/main` is now `0107f85b`. `git diff --stat origin/main...HEAD` shows no phantom file deletions (no unrelated files deleted), so base is acceptably fresh — no stale-base issue.

Scope: 8 files, all in scope. Renderer changes are only the mandated deletions (App.tsx confirm-dialog/state/Undo; `shouldWarnStaleCache`); the new feature surface is correctly left for a later stage. Generic scheduler (`schedule.mjs`) untouched; "Remind me at reset" notify retained and `fireAtFor` (reset+60s) now only serves it. Anti-spaghetti: the one flagged clone is the conventional store-fn shape (`stoppedStore.mjs` markStoppedRan/disarmStopped/bumpStoppedAttempts) — discounted.

### Block (fix-here)

**An armed conversation is resumed with zero usage reading at startup — the exact failure mode this issue exists to eliminate.**

The engine treats "no reading for the provider" as `ready`, not `wait`:
- `windowsFromSnapshots` drops providers with no windows, and an empty snapshot set yields `{}` (`usageResume.mjs:71-82`).
- `providerState` returns `ready` when `windows` is `undefined` because `isUsageAtLimit(undefined)` is `false` (`usageResume.mjs:94-101`, `usageStopper.mjs:195-202`).
- At boot, `index.mjs` calls `resumeEngine.deliverSnapshots(listSnapshots())` (`index.mjs:383`) while the poller's first tick is still in flight — `startPoller` launches it async without awaiting (`startPoller.mjs:23`) and it does network fetches, so `listSnapshots()` returns `[]`. Every armed entry then has `windows === undefined` → `ready` → a `Keep going` is scheduled immediately (staggered 0s/4s/…).

Consequence: on any server restart with an armed entry in the durable record (service update, crash, reboot — the same "process down across the reset" case the **late resume on wake** criterion is meant to handle), every armed conversation is sent `Keep going` immediately, with **no evidence that quota returned**. This violates acceptance criterion **"Nothing is sent while any window for that provider is still at its limit"** in exactly the scenario the feature is supposed to get right.

The test `gating: no snapshot for the provider never blocks a resume` (`usageResume.test.mjs:140-145`) codifies this behavior, so it's an intentional decision that happens to be wrong for the boot case. The current snapshot set being empty before the first poll is categorically different from "the provider reports under its limit."

**Suggested fix** (narrow, in-PR): distinguish "no reading yet" from "recovered." Wait (do not resume) until the provider has an actual window reading showing it under its limit — at minimum, do not treat the pre-first-poll empty state (`listSnapshots()` `[]` at boot) as recovered. The cleanest signal: only consider a provider `ready` when we have a non-empty window reading for it (`windows` present), and have the warmup drive the real `usage.updated` / the one-shot reset recheck before scheduling sends. Late-resume on wake is preserved: after boot the first real poll fetches fresh windows and sets recovery. Add a regression test for "boot with empty snapshot set and an armed entry → nothing sent until a real reading exists."

Nothing else Blocks. Good shape overall: single entry point, pure decision logic separated and injected, reuse of the existing poller (one-shot recheck via `tick`, no second loop), reuse of the shared prompt-delivery deferral, refusal→cap→flag accounting, and the mandated deletions are complete and grep-confirmed.

antoinedc:
Addressed the Block (fix-here). Thanks — the boot case was a real defect.

**The fix:** an absent reading for a provider is now `wait`, never `ready`. `providerState` returns `wait` unless we hold an actual non-empty window reading for the provider showing it under its limit. At boot `listSnapshots()` returns `[]` (first poll still in flight), so `windowsFromSnapshots` produces `{}` → `windows === undefined` → the engine schedules **nothing**. The warmup (`deliverSnapshots(listSnapshots())`) is then safe by construction: recovery is driven by the first real poll emitting `usage.updated`, and the one-shot reset re-check continues to cover the reset-instant case.

**Regression test added:** `engine: boot with an empty snapshot set sends nothing until a real reading exists` — delivers `[]`, asserts zero sends and zero armed timers, then confirms the continuation goes out only after a real recovered reading arrives. The old `gating: no snapshot ... never blocks a resume` test is replaced by the inverse (`no reading → wait`), so the wrong behavior can't be reintroduced.

Late-resume on wake is preserved: after boot, the first real poll fetches fresh windows and sets recovery, so a conversation that had no reading at boot recovers on the real poll rather than on the pre-poll empty state.

Verification: `npm run typecheck` exit 0; `npm test` green (vitest 2930 pass / 7 skip; server 2284 pass / 0 fail). Push `a3c54151`.

antoinedc:
## Review (cycle 2) — BET-1048 resume engine

Re-checked the PR at updated head `a3c54151` against the issue acceptance + `docs/usage-resume.md` §8/§9.

### Prior Block — resolved
The no-reading-at-boot defect is fixed. `providerState` now returns `wait` (not `ready`) when the provider has no usable window reading (`usageResume.mjs:107-114`), so:
- at boot, `deliverSnapshots(listSnapshots())` sees an empty snapshot set and schedules **nothing** until the first real poll emits `usage.updated`;
- the old inverse test is corrected and a boot regression test added (empty snapshot set + armed entry → nothing sent until a real reading shows recovery).

The fix is correct, minimal, and fits in-PR. Late-resume on wake is preserved (the first real poll after boot drives recovery), and the missing-reading path now correctly falls into the existing `wait` behavior rather than a new mechanism.

### Mechanical gate (updated head)
- `npm test` → vitest 2930 passed / 7 skipped; server 2284 pass / 0 fail.

### Acceptance checklist — all met
- Nothing sent while any window at its limit; every window checked, not just the named one — ✓
- `Keep going` on the pinned model, staggered — ✓
- Mid-turn defers until idle via the shared delivery engine — ✓
- Refused → re-queues up to the cap, then flagged; never loops, never silent-drop — ✓
- Late resume on wake — ✓
- reset+60s continuation and the per-continuation scheduler job/Undo gone; generic scheduler + "Remind me at reset" retained and unchanged — ✓
- Grep-verified. Test coverage matches the required list, plus the boot regression.

Scope: 8 files, no scope creep. Base freshness OK (no phantom deletions). Anti-spaghetti: only the conventional store-fn clone in `stoppedStore.mjs` (established shape), discounted.

**0 Blocks, 0 Questions, 0 nits.** Clean — routing to `manta-pm` for merge.

PR #1055, opened 2026-08-17
Settings: fix Escape-to-close, close-button hit area, and rail/header separator alignment

antoinedc:
Fixes three Settings dialog defects (BET-1052). No redesign, no features.

**Bug 1 — Escape silently does nothing on some tabs.** `useDialog`'s guard bailed whenever the keydown target was outside the dialog's DOM subtree. When focus falls back to `document.body` (routine after clicking a control that unmounts/blurs itself, more common on Accounts/Models/Extensions), `root.contains(body)` is false, so Escape was swallowed and Settings never closed. Replaced both old guards with a single predicate: bail only when another `[role="dialog"]` element is open in the document that isn't `root` and isn't an ancestor of `root`; otherwise close Settings. Fewer lines than before. Regression tests added (verified the body-target case fails on the old code, passes after).

**Bugs 2 & 3 — close-button hit area and rail/panel separator alignment.** The rail reserves a `titlebar-drag h-10` chrome strip above its heading; the main column's header didn't, so on macOS `hiddenInset` the native title band swallowed the top ~28px of the close button, and the rail vs panel horizontal rules landed ~40px apart. Added the same `titlebar-drag h-10 shrink-0` div as the first child of the main column (one line).

**Stale Settings visual baselines dropped** (visual gate is not enforced in CI): the four Settings screens and their snapshots removed from `tests/visual/screens.visual.ts-snapshots/` and `scripts/visual/screens.mjs`. Harness, other screens, and style-diff test untouched.

Base: origin/main @ `2fb521b3`

**Verification results:** see completion comment.


antoinedc:
## Review — BET-1052 (Settings Escape / hit-area / rule alignment)

**Result: PASS** — 0 Blocks, 0 Questions, 0 nits. Routing to PM for merge.

**Base-freshness gate:** merge-base `2fb521b3` == `origin/main` tip `2fb521b3` — branched off fresh main, no stale base. All 8 deletions are the in-scope Settings baselines; nothing phantom.

**Bug 1 (Escape swallowed):** Predicate matches the issue spec exactly — bail only when some `[role="dialog"]` is non-root and not an ancestor of root (`d !== root && !d.contains(root)`). This covers both portalled (sibling under body) and inline-child nested modals, and is focus-independent, so the body-focus case now closes Settings. Net −4 lines (18→14), satisfying removal-over-addition. Comment rewrite matches the requested scope.

**Bugs 2 & 3 (close-button hit area + rule alignment):** One added `<div className="titlebar-drag h-10 shrink-0" />` as first child of the main column. Rail and main columns now have identical stack — `titlebar-drag h-10` + `p-4` + `border-b` — so the close button clears the native title band and both horizontal rules land at the same y. Exactly one added line; Tailwind tokens only.

**Baselines dropped:** Exactly the 8 named snapshot files + the 4 `settings*` entries in `scripts/visual/screens.mjs`. Harness, other screens, and style-diff test untouched.

**Regression tests (Settings.test.tsx, extended in place):** Two cases verified green:
- Escape with `document.body` as target closes Settings (duplicates the old bug; now passes)
- Escape while a nested confirm dialog is mounted does not close Settings (stays green)
Confirmed both BET-1052 cases pass in isolation: `Tests 2 passed | 4 skipped`.

**Mechanical gate (run on PR branch `a9118b5e`):**
- `npm run typecheck` → exit 0, no errors
- `npm test` → exit 0; vitest 2957 pass / 7 skip; node:test 2280 pass / 0 fail
- Matches the implementer's reported PR-branch results exactly (trust-but-spot-check confirmed).

**Anti-spaghetti:** 0 clones in changed scope; no duplicated business logic introduced.

**Note (non-blocking):** The macOS-specific visual confirmations (title-band pointer swallow, pixel-perfect rule continuity) can't be eyeballed on this Linux box; the structural causes are fixed exactly per spec and unit/typecheck-covered. No block — the deliverable is the structural fix, which is correct.

PR #1065, opened 2026-08-17
fix(ios): share one dataSource lifetime with the transcript scroll view (BET-1062)

antoinedc:
Opened automatically for `agent/macos/21713172`, which was pushed without a pull request.

Relates to BET-1062.

### DIFF
--- a/mobile/native/MantaUI/ChatScreen.swift
+++ b/mobile/native/MantaUI/ChatScreen.swift
@@ -178,14 +178,8 @@ private struct ChatScreenContent: View {
         autoScrollsToBottomOnAppend: true,
         scrollsToBottomOnReplace: true
     )
-    /// Owned by the VIEW, never by the store: `ListDataSource` keeps an
-    /// append-only change log that `TiledView` replays from index 0 whenever
-    /// it meets a data source it has not seen. A store-owned one is mutated
-    /// long before (and across more than one) view creation, so the replay
-    /// runs against an already-final snapshot and desyncs the collection
-    /// view — the invalid-batch-updates crash. View-owned, the log and the
-    /// replay cursor are born together and the replay is always coherent.
-    @State private var dataSource = ListDataSource<TranscriptRow>()
+    /// The `dataSource` (+ change log) is now owned by `TranscriptListView`,
+    /// which shares one lifetime with the scroll view it describes (BET-1062).
     /// Whether the transcript has been scrolled up far enough that the round
     /// "scroll to bottom" control (rendered in ComposerView's model-selection
     /// row) should be shown. Driven purely by scroll geometry; it does not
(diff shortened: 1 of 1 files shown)


antoinedc:
## Review — BET-1062 (PR #1065)

**Base-freshness:** merge-base (`2907b21c`) == `origin/main` — branched off current main, no phantoms.
**Scope:** `mobile/native/MantaUI/ChatScreen.swift` only, `+85/-111`. Matches the issue's file list; no new file (committed `project.pbxproj` untouched). No scope creep.

### Code assessment — the fix is correct

Verified by static inspection against the acceptance criteria:

- **The core lifetime fix holds.** `ListDataSource<TranscriptRow>()` is now declared at `ChatScreen.swift:1210` inside the shared `TranscriptListView`, which is born and dies with the transcript branch on both screens. The parent (`ChatScreenContent`) and the subagent (`ChatSubagentScreen`) both call the same `TranscriptListView`, and both deleted their own `@State dataSource` + duplicated `TiledView/prependLoader/typingIndicator/onChange(apply)` chains. `resignKeyboard()` moved into the shared view; both parent copies are gone.
- **Invariant (point 5) independently confirmed:** `ListDataSource<TranscriptRow>()` occurs **exactly once** repo-wide.
- **Ordering constraint preserved:** inside `TranscriptListView.body`, the TiledView-only methods (`prependLoader`, `typingIndicator`, `additionalContentInset`, `onTiledScrollGeometryChange`) are all chained before the first plain `View` modifier — nothing was split back out to a call site.
- **`scrollPosition` stays screen-scoped** on both `ChatScreenContent:177` and `ChatSubagentScreen:1132`, passed as a binding — the composer's scroll-to-bottom (`:420`) still targets it. The parent's `onPointsFromBottom` geometry wiring and `showScrollToBottom` threshold are preserved.
- **CI is green:** `ios-build` pass (4m11s), `typecheck-test` pass, `duplication-gate` pass. Anti-spaghetti report: no TS/module files changed — nothing to scan (Swift-only change, correct).

### Block (fix-here) — acceptance-criteria verification incomplete

The issue's Verification section lists four simulator-gated checks, and two of them were **not actually executed** (by the implementer's own honest report):

1. **Verification item 4 — crash-path reproduction was not run.** "Open a session with the box unreachable so the load fails, tap retry with the box reachable, confirm the transcript loads and then streams without crashing" is the exact path this fix exists for. It is the strongest evidence the lifetime change closes the crash, and it is explicitly a listed acceptance criterion. It is recorded as "not run," so the criterion is unmet — not merely un-shown by me.
2. **Verification item 3 — `MantaSubagentDrillInUITests` was skipped, not passed.** "Still passes — it drives the child transcript, which this change re-wires." The run was reported as skipped because the open session had no `agent-row` to drill into; a skip does not satisfy "the child transcript re-wiring passes."

The code change itself is structurally sound and verified; these are the two verification gates that exercise the changed (re-wired) code and have not actually completed. Both fit inside this PR's scope.

**Action requested:** run both in the simulator and report real output — (a) drive `MantaSubagentDrillInUITests` to an actual pass against a session that has an agent-row (or create one), and (b) perform the load-fail → retry → stream reproduction and confirm no crash. Post the observed output, then reassign back to `manta-reviewer`.

antoinedc:
## Review — BET-1062 (update, cycle 2)

The two Block items from my prior review are resolved:

1. **Crash-path reproduction (item 4) — executed and green.** Re-drove load-fail → retry → stream in the simulator against the fixture's `degrade` knob; quoted output shows `first-load-failed=true`, `load-failure-visible=true`, then after retry `load-failure-visible=false chat-screen-visible=true`, and the `CRASHPATH OK: load-fail -> retry -> transcript re-entered and refetched without crashing` — the exact crash branch transition, no crash. Fixture wire log correlates the 503 → retry → refetch. (I could not independently view the attached screenshot — this environment has no image input — but the test output is the load-bearing evidence.)

2. **`MantaSubagentDrillInUITests` (item 3) — accepted as skip.** Re-verified the branch is unchanged (still the single `ChatScreen.swift` fix, no new commits). The test's skip is a pre-existing XCUITest limitation — `agent-row`/`assistant-prose` live inside `TiledView` (`UIViewRepresentable`) scroll cells that XCUITest cannot query via `app.descendants` (hierarchy dump confirms only the scroll container is exposed) — and the same identifiers are unchanged by this PR, so this is not a BET-1062 regression. The child transcript runs the **identical** shared `TranscriptListView` that demonstrably renders in the parent (crash-path test + OCR showing rendered content with the subagent row), and both build + 515 unit tests pass.

**Verdict: PASS — 0 Blocks, 0 Questions** (1 residual caveat/Nit: the drill-in UI test cannot currently assert through the TiledView layer, so its coverage is via the shared-view + crash-path evidence rather than a direct assertion). Base remained fresh throughout (main's post-cut merges are BET-1065, unrelated to `ChatScreen.swift`); CI `ios-build`, `typecheck-test`, `duplication-gate` all green; `ListDataSource<TranscriptRow>()` occurs exactly once repo-wide.

PR #1067, opened 2026-08-17
fix(ios): don't seed force-quit turn timer from local clock on runningSet replay (BET-1066)

antoinedc:
### Problem
After force-quitting the iOS app mid-turn and relaunching, the session-list
turn timer started at 0 (counting from the relaunch moment) instead of
resuming at the real box-reported elapsed time. BET-922's runningSet
reconnect reconcile was supposed to preserve the BET-896/BET-913 force-quit
persistence, and regressed it.

### Root cause
`MantaStreamRouter.applyingRunningSet` reused `resolveRunningSince`, whose
`Date()` fallback is correct for the LIVE `stream/running` edge (pinned by
`testRunningFrameWithoutSinceAndNoExistingFallsBackToNow`) but wrong for a
`runningSet` replay: a reconnect entry is a PERSISTED turn that can predate
the app launch by minutes. A running entry applied on a fresh state with
nil/absent `since` therefore seeded the local launch clock — exactly the
"timer at 0" symptom.

### Fix
Added `resolveRunningSetSince` for the reconnect path: use the box `since`
(epoch ms → seconds) when present; keep an existing stamp when an older box
omits it; and on a genuinely fresh state leave the start UNKNOWN (`nil`)
rather than lying it began at the relaunch moment. The live `stream/running`
`Date()` fallback is untouched.

### Server side
No server change. `src/server/streamInterp.mjs` already sends the original
`runningSince` (epoch ms) in each runningSet entry, and
`src/server/streamInterp.test.mjs` already pins
`snapshotState lists a busy session in the runningSet with its original since
(BET-922)`. This PR is the client half of BET-922's persistence contract.

### Tests
Added to `MantaUITests/MantaEventStreamTests.swift`:
- `testRunningSetWithBoxSinceOnFreshStateUsesBoxStartNotNow` — a runningSet
  entry's epoch `since` MUST seed `runningSince` (never `Date()`).
- `testRunningSetNilSinceOnFreshStateDoesNotSeedLocalClock` — nil `since` on a
  fresh state must not fabricate the launch moment (leaves `runningSince` nil).
- `testRunningSetNilSinceKeepsExistingStamp` — nil `since` preserves an already
  known start.
- `testRunningSetFrameWithSinceSeedsBoxStartThroughDecode` — end-to-end store
  test: a JSON-decoded `runningSet` frame carrying `since` seeds the box start
  through the full frame-parse → typed-payload → router chain.

Existing connect-time runningSet cases (`testRunningSet*`) remain green.

### Verification results
- PR branch (`multica/BET-1066-forcequit-turn-timer` @ `74721754`):
  - test: exit 0, 2300 pass / 0 fail / 0 skip. Failures: none.
- Base (`main` @ `2907b21c`):
  - test: exit 0, 2300 pass / 0 fail / 0 skip. Failures: none.
- Conclusion: 0 new failures. The JS/TS suites do not compile the Swift
  changes; the Swift test file compiles as part of the iOS target build
  (macOS/human or CI runner), not `npm test`.

Note (device verification): the on-device repro is Antoine's per the issue
("Done when" requires a real box + iOS device re-verification, BET-923). The
first fault line in the issue was a deployment check — the box under test must
run the merged BET-922 server (it does on main; `snapshotState` emits
`since`). This PR is the client-side hardening.

Base: origin/main @ 2907b21c

### DIFF
--- a/mobile/native/MantaUI/MantaEventStore.swift
+++ b/mobile/native/MantaUI/MantaEventStore.swift
@@ -257,6 +257,25 @@ enum MantaStreamRouter {
         return s
     }
 
+    /// Return the box's authoritative start for a running-set entry, or keep
+    /// whatever we already had, and NEVER fabricate the local clock.
+    ///
+    /// `applyingRunningSet` is the reconnect snapshot of ALREADY-running
+    /// sessions, so a running entry here is a PERSISTED turn that can predate
+    /// the app launch by minutes. The live `stream/running` path may fall back
+    /// to `Date()` (a live "just started" edge, pinned by
+    /// `testRunningFrameWithoutSinceAndNoExistingFallsBackToNow`), but doing so
+    /// here renders a force-quit turn's session-list timer as "0" — it counts
+    /// up from the relaunch moment instead of the real box-reported start. So a
+    /// running entry whose `since` is missing keeps an existing stamp (an older
+    /// box may omit it) and, on a genuinely fresh state, leaves the start
+    /// UNKNOWN (`nil`) rather than lying that it began at the launch moment.
+    private static func resolveRunningSetSince(_ entry: StreamRunningSetPayload.Entry?, current: Date?) -> Date? {
+        guard let entry else { return nil }
+        if let since = entry.since { return Date(timeIntervalSince1970: since / 1000) }
+        return current
+    }
+
     /// Replace running state across ALL known sessions from the box's
     /// authoritative set. A session absent from the set is not running.
     /// Only `running` / `runningSince` are touched — `turnComplete`, `chunks`
@@ -272,16 +291,14 @@ enum MantaStreamRouter {
         for (sid, var s) in next {
             let entry = bySession[sid]
             s.running = entry != nil
-            s.runningSince = resolveRunningSince(
-                running: entry != nil, since: entry?.since, current: s.runningSince
-            )
+            s.runningSince = resolveRunningSetSince(entry, current: s.runningSince)
             next[sid] = s
         }
         // A session the device has never seen a frame for can still be running.
         for entry in payload.sessions where next[entry.sessionId] == nil {
             var s = MantaSessionStreamState(sessionId: entry.sessionId)
             s.running = true
-            s.runningSince = resolveRunningSince(running: true, since: entry.since, current: nil)
+            s.runningSince = resolveRunningSetSince(entry, current: nil)
             next[entry.sessionId] = s
         }
         return next
--- a/mobile/native/MantaUITests/MantaEventStreamTests.swift
+++ b/mobile/native/MantaUITests/MantaEventStreamTests.swift
@@ -473,6 +473,48 @@ final class MantaEventStreamRouterTests: XCTestCase {
         XCTAssertEqual(next["ses_1"]?.toolStartOrder, ["t1"])
     }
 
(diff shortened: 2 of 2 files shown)


antoinedc:
## Review — BET-1066 (force-quit mid-turn resets iOS turn timer to 0)

**Result: PASS (0 Blocks, 0 Questions, 1 nit).** Routing to merge.

### Acceptance-criteria check
1. **Force-quit mid-turn resumes timer at real elapsed (BET-896/913 restored)** — Client fix in place: `applyingRunningSet` now uses `resolveRunningSetSince`, which derives `runningSince` from the box's epoch `since` (ms → seconds) when present. Traced the render path end-to-end: `SessionListStore.runningStart` reads `eventStore.sessionStates[sid].runningSince` directly (`SessionListStore.swift:107`), so a box `since` reproduces the real elapsed. The transcript working indicator (`ChatSessionStore` fold at `ChatSessionStore.swift:555`) seeds from the same snapshot `runningSince`, so it converges too. Device re-verification is Antoine's per the issue (BET-923) — correctly called out.
2. **BET-922 background→foreground clear preserved** — Fix is localized to the `runningSet` resolver; the live `stream/running` and `turnComplete` paths keep their unchanged `resolveRunningSince` `Date()` fallback. The empty-set/nil-entry clears (`testRunningSetClearsLatchedRunning…`, `…Empty…`) still produce nil. Preserved.
3. **`npm run typecheck && npm test`** — Ran on the PR branch: typecheck exit 0; `npm test` **2300 pass / 0 fail / 0 skip**. Matches the implementer's reported PR + base results exactly.
4. **PR pushed, keyed, reassigned** — PR #1067 titled with BET-1066, reassigned to reviewer. (Note: PR is currently in **draft** — flagging for the PM; I do not flip draft→ready, that's the PM's call.)

### Tests requirement
- Client tests added (4) cover exactly the issue's required cases: box `since` → epoch date not `Date()`; nil `since` on fresh state does not seed local clock (stays nil, so the timer renders no fabricated elapsed — `SessionListView.timerText` returns nil rather than "0 from relaunch"); nil `since` keeps existing stamp; full decode chain via a JSON `runningSet` frame.
- Existing connect-time `testRunningSet*` cases all use non-nil `since` or empty sets — none rely on the removed `Date()` fallback in the running-set path, so none break. Verified the diff adds-only.
- Server side (`streamInterp.test.mjs`): the issue's requirement ("runningSet carries the original runningSince") is already pinned by the BET-922 test `snapshotState lists a busy session in the runningSet with its original since` (`streamInterp.test.mjs:373`). No server change needed — consistent with the fix.
- Swift suite itself can't run on this Linux box (needs Xcode); compiles within the iOS target on CI/Mac. Swift memberwise inits and JSON decode equality (both `since` values exactly representable) check out.

### Scope
Diff is 2 files, both in scope (`MantaEventStore.swift`, `MantaEventStreamTests.swift`). No scope creep, no under-delivery. Deployment foot-gun (box must run merged BET-922 `snapshotState`) correctly flagged.

### Nit
**Nit (fix-here):** base is slightly stale — branched off `2907b21c`, `origin/main` is now `3904ed33` (BET-1062/1065 landed after). No overlap with this PR's files and no phantom deletions, so not blocking, but rebase onto `origin/main` before merge so it stays clean.

### Anti-spaghetti
Detector: 0 changed `.ts/.tsx/.mjs` files — nothing to scan; changes are Swift-only. No duplication concern.

PR #1078, opened 2026-08-17
fix(desktop): screenshot auto-attach keys off macOS's own xattr marker, delete filename regex (BET-1077)

antoinedc:
## BET-1077 — key screenshot auto-attach off macOS's own marker, delete the filename regex

**Base: `origin/main` @ `4e94b4ac`** (merged BET-1076, which this builds on)

### The bug
The file-detection path matched screenshot names with `/^Screenshot \d{4}-\d{2}-\d{2} at \d{2}\.\d{2}\.\d{2}.*\.png$/i`. The hour is `\d{2}` (two digits, required) — so on a 12-hour clock (the US-English default) a single-digit-hour screenshot (`Screenshot 2026-08-17 at 2.34.56 PM.png`) never matched. Detection only ever fired during the 10/11/12 o'clock hours. The filename is unreliable along **five** independent axes (12h/24h clock, macOS version, include-date setting, user-set name prefix, locale), so no regex could fix it.

### The fix
Key off macOS's own evidence instead:

1. **New pure module `src/shared/screenshot.mjs`** — `isScreenshotCandidate(filename)` (cheap extension gate over `SCREENSHOT_EXTENSIONS` — the formats `defaults write com.apple.screencapture type` accepts) and `resolveScreenshotDir(rawLocation)` (resolves the configured screencapture location, `~/Desktop` default when unset). Reuses `expandTilde` from `src/shared/paths.mjs` — no duplicate tilde expander.
2. **Rewrote `src/main/screenshotDetector.ts`**:
   - **Deleted `SCREENSHOT_RE` entirely** — `git grep SCREENSHOT_RE` returns nothing repo-wide.
   - Added an `isScreenshotFile(path)` marker probe via promisified `execFile`, running `xattr -p com.apple.metadata:kMDItemIsScreenCapture <path>` — the extended attribute macOS stamps on every screenshot at creation, which survives renaming/moving/locale/clock format. Resolves `true` only on exit 0; catches everything and resolves `false`; never throws. (Not `mdls` — that asks Spotlight, which can be disabled per volume; `xattr` reads the file directly.)
   - **Watch directory resolved at startup** via `defaults read com.apple.screencapture location` → `resolveScreenshotDir`, instead of hardcoding `~/Desktop`. The read being async moved watcher setup into `initFileWatcher(send)`, called as `void initFileWatcher(send)` from `startScreenshotDetector` (which stays `void`, caller doesn't await). The `defaults` key being unset exits non-zero and is expected — not logged as an error.
   - **Filter order (exact):** rename + truthy filename → `isScreenshotCandidate(basename)` extension gate → 300ms settle delay → `isScreenshotFile(fullPath)` → `send(IPC.screenshotDetected, …)`. The delay is before the probe because the probe needs the file on disk.

### Out of scope (per issue)
- Clipboard poller untouched (separately verified working).
- IPC channel name + payload unchanged; renderer and preload untouched.
- `process.platform !== "darwin"` early return stays — no Windows/Linux support.
- No npm dependency added; no polling/retry around the probe (one `xattr` call).

### Verification results
**Files changed:** 4 files. `src/main/screenshotDetector.ts`, `src/shared/screenshot.mjs`, `src/shared/screenshot.d.mts`, `src/shared/screenshot.test.ts`

- PR branch (`multica/BET-1077-xattr-screenshot` @ `4f05cbbd`):
  - test: exit 0, 2313 pass / 0 fail / 0 skip. Failures: none. log sha256: `e07150d54aa5219895246b20a291822b32eee4bd96c42ee4dc6d28000d110131`
- Base (`main` @ `4e94b4ac`):
  - test: exit 0, 2313 pass / 0 fail / 0 skip. Failures: none. log sha256: `9b2c052d8ded9e4675b716abae37b6cbf7b1e961d931a46e1832fe3baf627c29` (differs from PR log only in duration_ms output)
- **Conclusion:** 0 new failures, 0 pre-existing failures. New tests: 16 (`src/shared/screenshot.test.ts`), including the reported regression case `"Screenshot 2026-08-17 at 2.34.56 PM.png"` → true.

### DIFF
--- a/src/main/screenshotDetector.ts
+++ b/src/main/screenshotDetector.ts
@@ -10,26 +10,51 @@
 //    formats + size, so the poll is cheap. The renderer calls uploadBuffer
 //    which reads the clipboard once via a separate IPC.
 //
-// 2. Desktop folder watcher: catches ⌘⇧3/4 (file-only shots). macOS names
-//    them "Screenshot YYYY-MM-DD at HH.MM.SS.png". The fs.watch callback
-//    fires on the rename event (file creation) and we filter by that pattern.
-//    We push the absolute path so the renderer can scp it directly via
-//    uploadFiles (no extra read needed).
+// 2. File watcher: catches ⌘⇧3/4 (file-only shots) no matter where they're
+//    saved. We key off macOS's own marker rather than the filename: macOS
+//    stamps every screenshot it writes with the extended attribute
+//    com.apple.metadata:kMDItemIsScreenCapture, which survives renaming,
+//    moving, locale and clock format — the filename is unreliable along five
+//    independent axes (12h/24h clock, macOS version, include-date setting,
+//    user-set name prefix, locale). The watch directory is read from
+//    `defaults read com.apple.screencapture location` (it's configurable),
+//    defaulting to ~/Desktop when unset.
 //
-// Both paths are no-ops when no mainWindow exists. The Desktop watcher is
-// started once at app-ready and never restarted (the Desktop path doesn't
-// change). The clipboard poller is the same.
+// Both paths are no-ops when no mainWindow exists. The file watcher is
+// started once at app-ready and never restarted. The clipboard poller is
+// the same.
 
 import { clipboard } from "electron";
 import { basename, join } from "node:path";
 import { watch as fsWatch } from "node:fs";
-import { homedir } from "node:os";
+import { execFile } from "node:child_process";
+import { promisify } from "node:util";
 import { IPC } from "../shared/types.js";
+import { isScreenshotCandidate, resolveScreenshotDir } from "../shared/screenshot.mjs";
 
-const SCREENSHOT_RE = /^Screenshot \d{4}-\d{2}-\d{2} at \d{2}\.\d{2}\.\d{2}.*\.png$/i;
+const execFileAsync = promisify(execFile);
 
 let screenshotClipboardTimer: ReturnType<typeof setInterval> | null = null;
 let screenshotDesktopWatcher: ReturnType<typeof fsWatch> | null = null;
+
+// macOS stamps every screenshot it writes with this extended attribute. It is
+// set by the OS at creation and survives renaming, moving and any locale — it
+// is the only reliable signal that a file is a screenshot. `xattr` reads the
+// file directly; `mdls` would ask Spotlight, which can be disabled per volume.
+// exit 0 = attribute present. Any error (missing attr, missing file,
+// permission) = not a screenshot. Must never throw.
+async function isScreenshotFile(path: string): Promise<boolean> {
+  try {
+    await execFileAsync("xattr", [
+      "-p",
+      "com.apple.metadata:kMDItemIsScreenCapture",
+      path,
+    ]);
+    return true;
+  } catch {
+    return false;
+  }
+}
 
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-1077 (PR #1078)

**Verdict: PASS** — 0 Blocks, 0 Questions. Routing forward.

Base-freshness: `git merge-base HEAD origin/main` == `origin/main` == `4e94b4ac` — PR is branched off current `main`, no stale base, no phantom deletions. Diff touches only the 4 files the issue scoped (`screenshotDetector.ts`, `screenshot.mjs`, `screenshot.d.mts`, `screenshot.test.ts`). No scope creep.

### Acceptance criteria (verified independently, against code not prose)

1. **New pure module `src/shared/screenshot.mjs`** — `isScreenshotCandidate`, `resolveScreenshotDir`, `SCREENSHOT_EXTENSIONS` all present; imports `expandTilde` from `paths.mjs` (no duplicate helper); no electron / no `child_process`. ✓
2. **`isScreenshotCandidate` contract** — accepts only strings with last-`.` extension in the list; rejects non-string, empty, no-dot. Matches spec exactly. ✓
3. **`resolveScreenshotDir` rules** in the specified order — non-string→`""`, trim, empty→`~/Desktop`, else `expandTilde` + strip trailing `/`. ✓
4. **Test file `src/shared/screenshot.test.ts`** — all specified cases present, no fewer, including the reported-bug regression `"Screenshot 2026-08-17 at 2.34.56 PM.png" → true`, plus the no-date, German-locale, `.PNG`-case, `.heic`, and all false cases. ✓
5. **`SCREENSHOT_RE` deleted** — `git grep SCREENSHOT_RE` returns nothing repo-wide. ✓
6. **Marker probe** — `isScreenshotFile` runs `xattr -p com.apple.metadata:kMDItemIsScreenCapture <path>` via promisified `execFile`; resolves true only on exit 0, catches everything → false, never throws. ✓
7. **Configurable watch dir** — `defaults read com.apple.screencapture location` at start; on failure passes `""` → `~/Desktop` default; failure is silent (not logged as error). ✓
8. **`initFileWatcher` async, called as `void initFileWatcher(send)`; `startScreenshotDetector` stays `void`** and caller doesn't await. ✓
9. **Filter order** exactly: rename+truthy filename → `isScreenshotCandidate(basename)` → 300ms settle → `await isScreenshotFile` → `send(...)` unchanged. Settle delay correctly precedes the probe (file must exist). ✓

### Do-NOT constraints
No `mdls` ✓ · no retry/poll loop (single probe) ✓ · no npm dependency (package.json untouched) ✓ · clipboard poller untouched ✓ · IPC channel/payload unchanged (renderer/preload untouched) ✓ · `process.platform !== "darwin"` early return preserved ✓.

### Mechanical gates (run on PR branch)
- `npm run typecheck` — **0 errors**.
- `npm run typecheck && npm test` — **2313 pass / 0 fail** (matches the implementer's reported base-branch summary `main @ 4e94b4ac`).
- `npx vitest run src/shared/screenshot.test.ts` — **16/16 pass**, including the single-digit-hour regression case.
- `scripts/check-duplication.sh origin/main` (anti-spaghetti) — **0 clones**.

`.d.mts` type declaration follows the existing `paths.d.mts` convention; tsc clean.

PR #1081, opened 2026-08-17
feat(updates): Check for updates in About, fix macOS auto-update

antoinedc:
One button in **Settings → About** checks both update systems at once and reports each separately, plus the release-side fix that makes the desktop half actually able to install anything.

## The headline fix: macOS auto-update could never work (silently)

electron-updater's macOS path hands the update to Squirrel.Mac, which only installs a zipped `.app`. `MacUpdater` resolves it with `findFile(files, "zip", ["pkg", "dmg"])` — **dmg is on the excluded list, not a fallback** — and throws `ERR_UPDATER_ZIP_FILE_NOT_FOUND` when the feed has none. `latest-mac.yml` published a DMG and no ZIP, so every download on every Mac threw instantly.

It was invisible because the message names no `checksum`/`signature`/`permission` keyword, so `updateError.mjs` classified it "transient" and swallowed it — the same silence as the 0.0.13/0.0.14 checksum bug, different cause, and it hid *behind* that fix (a correct digest for an artifact the updater refuses to read still installs nothing). Consequence: `update-downloaded` never fired on macOS, so the sidebar update dot and About's "Restart to update" strip were unreachable.

**Fix on the release side:** build + publish a ZIP, and guard against it ever returning:
- `electron-builder.yml` — add a `zip` mac target (packed from the already-notarized `.app`, not rewritten by the staple step, so its feed digest is correct as written).
- `codemagic.yaml` / `publish.sh` — upload it (to the update feed dir, never `downloads/`), and **fail the build if the feed lists no installable `.zip`**. The existing digest check verifies the files a feed NAMES; it cannot see that none of them are installable.
- `updateError.mjs` — new `"feed"` class, checked **before** integrity, because the real message embeds the feed's file list as JSON and therefore contains the substring `sha512`, which otherwise mis-classified it as an app-side integrity failure.

⚠️ This is only provable after the next `mac-v*` release — the feed is built by CI.

## Check for updates (Settings → About)

- **`autoUpdate:check`** resolves with `{isUpdateAvailable, updateInfo}` rather than inferring from events — `update-not-available` was log-only, so an event-based button could only timeout-and-guess and could never say "up to date".
- **`server:update-check`** runs the poller's own tick, so the button and the banner can never disagree; the per-version dedup gates the side effects (banner + push), never the answer.
- About routes "Update & restart" to App's existing flow (`onRequestServerUpdate`) instead of calling `serverUpdateApply()` — reusing the confirm, progress, 120s cap and transient-error handling rather than copying them.
- Download progress is surfaced; both terminal outcomes end the "Downloading…" state.
- Tone-coded rows: `supported:false` (dev build / no updater) maps to muted, **never** "up to date".

## Server update latency (without more polling)

- **Conditional GET** for the manifest (site serves an ETag — verified live: 200 then 304s). 30 min now moves less data than 6h did.
- **Check on connect**: the desktop checks when its connection goes live and on every reconnect, making a release visible the moment the app opens.
- `createUpdateCheck`'s guard now **joins** an in-flight tick instead of returning `{available:false}` (a manual check landing during a poll used to report "up to date" with nothing compared).

Also documents the two-system-update model in AGENTS.md.

No server restart needed to pick up the JS — the box runs this source.


antoinedc:
## Review follow-ups (all 3 should-fix findings addressed)

Background review found 3 should-fix issues and no blockers. All three are fixed:

**1. A failed box check rendered a false "up to date"** — fixed. The on-demand check now resolves `ok:false` on a manifest-fetch failure (the poller still swallows it for the background timer), `ServerUpdateCheck` carries it, and `describeServerUpdate` reads it → the box row shows an error tone instead of the reassuring green. Absent `ok` (old server) = successful. Covered by `poller.check(): a manifest fetch failure is ok:false…`.

**2. A transient download failure wedged "Downloading…" forever** — fixed. `autoUpdate:download` now returns the download promise so the IPC rejects, and the button `.catch`es to restore the Download button. Also bounded the check itself with a 15s timeout so a hung box can't leave "Checking…" spinning.

**3. Reopen-after-release didn't surface the banner** — fixed (design change). The poller's per-version dedup silenced the banner for a version the timer already notified — the common case for a desktop closed at notify time. The on-demand/connect path now **force-re-surfaces the banner** (the timer keeps its dismiss-protecting dedup), while the **push stays deduped** so a reconnect never re-buzzes the phone. The banner is an idempotent store set, so re-publishing is harmless.

Nits not addressed (accepted as out of scope / low value): the "Download manually" link hardcodes the mac-arm64 DMG (the app is currently arm64-mac-only for auto-update anyway) and the rapid-double-click `checking` guard (both legs are idempotent — harmless).

Verified: typecheck clean; 2997 renderer + 1816 server tests green.

PR #1085, opened 2026-08-17
fix(ios): render live subagent cards and bound inline tool output (BET-1085)

antoinedc:
fix(ios): render live subagent cards and bound inline tool output (BET-1085)

Branch: `multica/BET-1085-ios-subagent-card`
Base: `origin/main` @ `23d6d884`

## What & why

While a `task` subagent runs, the iOS transcript showed empty space where its card should be — the card only existed once the subagent finished, and the moment it finished the composer was shoved off screen. Two independent causes, both fixed client-side (`mobile/native/`, no `src/` changes):

1. **The in-flight card streamed but was never read.** The box already publishes a `stream/subagent` frame per running subagent; the chat screen only consumed it for the session-list subtitle. This wires it into the live transcript path.
   - `ChatModels.swift: appendingLive(...)` replaces `appendingLiveTools(...)`, appending running live subagents alongside live tools.
   - Live `task` tool rows are suppressed (the tool triple is redundant next to the richer `subagent` frame, and the tool row is the one that dumped the whole result into an unbounded view).
   - Finished (`completed`/`error`) payloads are dropped — that's the canonical transcript's job after the turn-boundary refetch.
   - Dedup now covers subagent rows too, so a live card disappears the instant the canonical one lands.
   - `ChatSubagentMapper.session(from: StreamSubagentPayload)` is the second entry point next to the canonical mapper.

2. **Unbounded tool-output view blew out the layout.** A still-`running` step row is force-expanded, and the expanded well rendered the entire (up to 20,000 char) final output with no cap — thousands of points of text inside a self-sizing cell. Added `ToolOutputPreview.tail(_:)` (last 12 lines, then last 2,000 chars, `… `-prefixed when trimmed) and applied it at the single render site.

Plus two correctness fixes required to make the live card actually update and stay unique:
- `SubagentSession` id now falls back to the tool `callID` (never the task name), so distinct task parts can't collide (init takes an explicit `fallbackId`; fixture sites compile via a default).
- `SubagentSession ==` widened to compare `id, taskName, status, duration, childSessionId` (excluding `transcript`), so a row redraws when its status/title/duration changes. `hash` stays id-only.

## Verification results

- typecheck: exit 0, no errors. (2 configs, `tsconfig.node` + `tsconfig.web`.)
- test: exit 0, **2326 pass / 0 fail / 0 skip**.
- This ticket changes **no TypeScript** — the gate is defence-in-depth only.
- iOS Swift suite: **not runnable on the Linux box (no Xcode/swift toolchain); verified by reading.** New tests added:
  - `ChatTranscriptTests`: `testTaskPartInFlightMessageProducesNoCanonicalSubagent`, `testLiveSubagentAppendsRunningCard`, `testLiveSubagentCompletedIsNotAppended`, `testLiveSubagentDedupedAgainstCanonicalRow`, `testLiveTaskToolRowIsSuppressed`, `testSubagentIdIsUniquePerCallWithoutChildSession`, `testSubagentEqualityTracksStatusAndDuration`, plus existing live-tool tests migrated to the `appendingLive(tools:subagents:to:)` signature.
  - `ToolOutputPreviewTests` (new file): short unchanged; 30 lines → last 12 + prefix; single 5,000-char line → last 2,000 + prefix; empty/whitespace unchanged; exactly 12 lines unchanged + unprefixed.
- `git grep -n "appendingLiveTools"` under `mobile/native` → **nothing** (Definition of Done).

Out of scope honored: no `src/` changes, no server-side removal of the redundant tool triple, no un-skipping of the in-flight assistant message, no new store/`@Published`/RPC/stream frame, no new expand affordance, no changes to `ChatSubagentScreen`/`TranscriptListView`/fixtures.

### DIFF
--- a/mobile/native/MantaUI/ChatModels.swift
+++ b/mobile/native/MantaUI/ChatModels.swift
@@ -213,7 +213,10 @@ enum ChatSubagentMapper {
         return SubagentSession(
             taskName: taskName,
             status: .running,
-            duration: ChatDuration.text(seconds: payload.durationMs),
+            // `durationMs` is MILLISECONDS on the wire (the desktop reference
+            // renders it via ms/1000); ChatDuration.text takes seconds, so
+            // convert before the helper — 1200 → 1.2 → "1.2s".
+            duration: ChatDuration.text(seconds: payload.durationMs.map { $0 / 1000 }),
             transcript: [],
             childSessionId: payload.childSessionId
         )
--- a/mobile/native/MantaUI/ChatModels.swift
+++ b/mobile/native/MantaUI/ChatModels.swift
@@ -372,7 +372,7 @@ enum ChatTranscriptMapper {
 
         guard !(liveSteps.isEmpty && liveSubagents.isEmpty) else { return blocks }
 
-        var liveRows = liveSteps.map { .step($0) }
+        var liveRows: [StepGroupRow] = liveSteps.map { .step($0) }
         liveRows.append(contentsOf: liveSubagents.map { .subagent($0) })
 
         // Dedup against ids the canonical transcript already owns — STEP rows
--- a/mobile/native/MantaUI/ChatModels.swift
+++ b/mobile/native/MantaUI/ChatModels.swift
@@ -175,12 +175,47 @@ enum ChatSubagentMapper {
             duration = nil
         }
 
+        // The row's id must be unique per task part and stable across rebuilds:
+        // the child opencode session id when known, otherwise the tool's callID
+        // (never the task name — two subagents run under the same title would
+        // collide). See SubagentSession.init.
+        let callID = ChatJSON.string(part.extra["callID"])
+
         return SubagentSession(
             taskName: taskName,
             status: status,
             duration: duration,
             transcript: [],
-            childSessionId: childSessionId
+            childSessionId: childSessionId,
+            fallbackId: (callID?.isEmpty == false) ? callID : nil
+        )
+    }
+
+    /// The LIVE-subagent path: map the box's `stream/subagent` frame onto the
+    /// same SubagentSession the canonical mapper produces, so a running card
+    /// renders immediately from the live feed. `childSessionId` is always
+    /// present here — the box only emits the frame once opencode has stamped
+    /// the child session id.
+    static func session(from payload: StreamSubagentPayload) -> SubagentSession {
+        // Task name: the box's title first, then the description, then the
+        // agent id — a published subagent always carries at least a title.
+        let taskName: String
+        if let t = payload.title, !t.isEmpty {
+            taskName = t
+        } else if let d = payload.description, !d.isEmpty {
+            taskName = d
+        } else if let a = payload.agent, !a.isEmpty {
+            taskName = a
+        } else {
+            taskName = "subagent"
+        }
+
+        return SubagentSession(
+            taskName: taskName,
+            status: .running,
+            duration: ChatDuration.text(seconds: payload.durationMs),
+            transcript: [],
+            childSessionId: payload.childSessionId
         )
     }
 }
@@ -278,8 +313,8 @@ enum ChatTranscriptMapper {
         return blocks
     }
 
-    /// Append the still-running LIVE tools (streamed mid-turn by the box) to the
-    /// transcript, in the turn that spawned them.
+    /// Append the still-running LIVE tools AND live subagents (streamed mid-turn
+    /// by the box) to the transcript, in the turn that spawned them.
     ///
     /// This is the replacement for the deleted pinned running-tool overlay:
     /// a tool call renders inside the transcript (tailing its output as a live
@@ -287,7 +322,10 @@ enum ChatTranscriptMapper {
     /// `.running` step keyed by the SAME `callID` the canonical `stepIdentity`
     /// uses, so a live row and its completed canonical sibling share one
     /// identity and the turn-boundary refetch replaces the row IN PLACE — no
-    /// remove-and-reinsert flash (the whole point of this issue).
+    /// remove-and-reinsert flash (the whole point of this issue). A live
+    /// subagent becomes a `.subagent` card sourced from the box's `subagent`
+    /// frame, so the card renders while the subagent runs instead of arriving
+    /// only once it finishes.
     ///
     /// A live tool whose `callID` the transcript already owns is skipped: the
     /// canonical step has taken over, so appending again would duplicate it.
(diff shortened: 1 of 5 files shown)


antoinedc:
## Code review — BET-1085 (iOS live subagent cards + bounded tool output)

**Pre-flight checks**
- Base freshness: merge-base `23d6d884` == `origin/main` tip `23d6d884` — branched off current main. No phantom deletions. `git diff --stat origin/main...HEAD` shows exactly the 5 in-scope `mobile/native/` files. No `src/` (TypeScript) changes.
- Anti-spaghetti report: 0 changed `.ts/.tsx/.mjs` files — nothing to scan, clean.
- `npm run typecheck`: **exit 0** (both tsconfigs). `npm test`: **2326 pass / 0 fail / 0 skip** — matches the implementer's reported numbers exactly. (These cover unchanged TS; the Swift suite can't run on Linux — verified by reading.)
- `git grep -n "appendingLiveTools"` under `mobile/native` → nothing. ✓

**Acceptance-criteria checklist** (verified independently against the diff)
1. ✓ Single `appendingLive(tools:subagents:to:)` path replaces `appendingLiveTools`; all existing rules preserved (dedup, last-group append, no-op). Call site passes the already-published `subagents`.
2. ✓ Live `task` tool rows suppressed; live `.subagent` rows built from non-terminal payloads with the required field mapping (title→description→agent→"subagent", `.running`, `transcript: []`).
3. ✓ Subagent dedup covered — `existingIDs` now includes `.subagent` ids via `StepGroupRow.id`.
4. ✓ Identity: fallback is the tool `callID`, init takes explicit `fallbackId` (default keeps fixtures compiling); never task-name.
5. ✓ Equality widened (id/taskName/status/duration/childSessionId, `transcript` excluded); hash id-only.
6. ✓ `ToolOutputPreview.tail` (12 lines / 2,000 chars, `… `-prefix) applied at the single render site, no `.lineLimit`, applies to every tool row.
7. ✓ No expand affordance. Out-of-scope items untouched.

**Block (fix-here): duration unit mismatch — new test contradicts the code it tests**

`ChatSubagentMapper.session(from: StreamSubagentPayload)` maps `duration` as
`ChatDuration.text(seconds: payload.durationMs)`, but `durationMs` is **milliseconds** — the box's `durationMs` is `state.time.end - state.time.start` (src/shared/streamInterpretation.mjs:748), and the desktop reference client displays it via `formatDuration(ms)` which divides by 1000 (src/renderer/TaskCard.tsx:44,113; src/renderer/chatUtils.ts:261; server fixture start 1000 / end 13500 → 12500 → "13s"). The seconds-parameter `ChatDuration.text(:)` therefore renders a 1.2s subagent as `"20m0s"`.

This is exactly what the new test catches: `testLiveSubagentAppendsRunningCard` asserts `agent.duration == "1.2s"` with `durationMs: 1200` — which, against the current implementation, would actually be `"20m0s"` and the test would **fail**. The test encodes the correct interpretation (ms→s); the implementation doesn't match it.

Fix (fits in this PR): convert ms→s before the helper, e.g. `ChatDuration.text(seconds: (payload.durationMs ?? 0) / 1000)` (so 1200 → 1.2 → "1.2s"). The existing assertion then passes and the live card shows the real duration.

Nit (pre-existing, **not** required by this ticket): the canonical mapper (`session(from part:)`, `ChatDuration.text(seconds: end - start)`) feeds the same ms delta into the seconds helper, so canonical subagent durations carry the same latent 1000× issue. This PR isn't obliged to fix it, but it's the same wart and worth a follow-up so the live and canonical paths agree.

Everything else is clean and faithful to the spec. Once the unit fix lands, typecheck/tests and the changed tests should be re-verified by anyone with an Apple toolchain (the Swift running-case coverage is good otherwise).

antoinedc:
Addressed the Block (fix-here). 

**Duration unit fix:** `ChatSubagentMapper.session(from: StreamSubagentPayload)` now converts `durationMs` (milliseconds) to seconds before the seconds-based formatter: `ChatDuration.text(seconds: payload.durationMs.map { $0 / 1000 })`. A `durationMs: 1200` subagent now renders "1.2s" — the assertion in `testLiveSubagentAppendsRunningCard` now matches the implementation. Committed + pushed (`71a1df2f`).

**Follow-up (canonical path, pre-existing nit):** filed as **BET-1087** (assigned to manta-pm) — the canonical `session(from part:)` mapper feeds the same ms delta into the seconds helper and will now visibly disagree with the corrected live card. Left out of scope here per the review note.

Re-verified: `npm run typecheck` exit 0; `npm test` 2326 pass / 0 fail. Swift suite still requires an Apple toolchain to execute (verified by reading + the corrected assertion).

antoinedc:
## Re-review — BET-1085 (cycle 2)

The prior Block is resolved. Commit `71a1df2f` changes `ChatSubagentMapper.session(from: StreamSubagentPayload)` to convert the wire value before formatting:

```swift
duration: ChatDuration.text(seconds: payload.durationMs.map { $0 / 1000 })
```

`durationMs` is milliseconds; `ChatDuration.text` takes seconds — so `1200` → `1.2` → `"1.2s"`, which is now what `testLiveSubagentAppendsRunningCard` asserts. The test needed no change and is consistent with the code.

**Re-check:** merge-base `23d6d884` == `origin/main` tip. Diff still confined to the 5 in-scope `mobile/native/` files (one-line Swift fix on top). `npm run typecheck` exit 0, `npm test` 2326 pass / 0 fail / 0 skip (unchanged — the fix is Swift-only, outside the TS gate). DoD grep `appendingLiveTools` → nothing.

**Verdict: PASS** — 0 Blocks, 0 Questions. The pre-existing canonical-mapper unit wart (out of scope here) is tracked separately as BET-1087. No further changes required from the implementer.

antoinedc:
## Re-review — BET-1085 (cycle 3: Swift compile gate)

The PM's `ios-build` failure is fixed on head `a57b79c9`. `appendingLive` now annotates the local so the bare enum-member expressions resolve:

```swift
var liveRows: [StepGroupRow] = liveSteps.map { .step($0) }
liveRows.append(contentsOf: liveSubagents.map { .subagent($0) })
```

Element type is `[StepGroupRow]` (`.step(ToolStep)` / `.subagent(SubagentSession)` are both `StepGroupRow` cases — the PM's suggested `[TranscriptBlock]` would have been wrong, since `.step`/`.subagent` aren't `TranscriptBlock` cases), and the downstream dedup (`StepGroupRow.id`) and `.steps(.rows(...))` append still type-check.

**Verified directly via CI:** `ios-build` (the gate that caught the regression) is **pass** on this head, alongside `typecheck-test`, `duplication-gate`, and the push check. `npm test` 2326/0/0 unchanged. Base fresh (`23d6d884`), diff still confined to the 5 in-scope `mobile/native/` files.

**Verdict: PASS** — 0 Blocks, 0 Questions. Route to `manta-pm` for merge.

PR #1090, opened 2026-08-18
fix(chat): only a user gesture may detach the transcript from the tail

antoinedc:
## Symptom

Every tool call bounced the transcript out of follow mode. The user had to click **jump to latest** to get back — repeatedly, for the whole turn.

## Cause

The follow state turned OFF on any scroll event whose `scrollTop` moved up. But react-virtuoso writes to the scroller **itself**: its *upward scrolling compensation* fires whenever a row above the viewport is re-measured — exactly what a tool card landing mid-turn does to the message row it joins. That correction arrives as an ordinary `scroll` event with a lower `scrollTop`, indistinguishable from the user dragging up.

This is the sixth generation of this bug. BET-933 established "never derive *should we follow* from *is the scroller at the bottom*". The remaining hole was one level down: a **scroll event is not a user gesture either**.

## Fix

Separate the two signals: **where** the scroller is (the scroll event) and **whether the user put it there**.

- `classifyFollowOnScroll` takes a `userIntent` argument and only returns `false` when both agree.
- Intent is a 250ms window (`USER_SCROLL_INTENT_WINDOW_MS`) refreshed by `wheel` / `touchstart` / `touchmove` / `keydown` on the scroller, plus a held flag for a scrollbar-gutter drag.
- A press inside the **content box** is a click (expanding a tool card, selecting text) and deliberately does *not* vouch for the re-measure scrolls that follow it — otherwise the same hole reopens from the other side. `isScrollbarGutterPress` is the discriminator.
- Landing within `FOLLOW_THRESHOLD_PX` of the bottom still re-attaches unconditionally.

### Two follow-ons

- **`scrollToMessage` now detaches explicitly.** The artifacts / ⌘F deep-link jump used to rely on the resulting scroll-up being mistaken for a gesture — the very mistake being fixed. Without this, a streaming reply would yank the user off the message they jumped to.
- **`stickToTail` writes twice, across a frame.** Virtuoso publishes `totalListHeightChanged` synchronously inside the ResizeObserver tick, while the list padding carrying part of that height is React state landing on a later commit — so the first write can read a pre-growth `scrollHeight` and park short of the tail. The rAF write lands on the real one. Both are guarded on the follow state.

## Tests

- `chatUtils.test.ts` — the pure decision, including a regression case with **identical numbers** to the user-scroll-up case where the intent flag is the only difference, plus the intent-window / gutter-press helpers.
- `Transcript.test.tsx` — the **wiring**, on a mounted transcript driven through a gesture-less scroll-up. Verified red against the old logic (`× REGRESSION: a scroll-up with no user gesture keeps the transcript following`) and green with the fix.

Full suite green: 2912 renderer + 2155 server, typecheck clean.

`AGENTS.md` records the new invariant so generation seven doesn't reintroduce it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1286,6 +1286,36 @@ Backup at `~/.tmux.conf.pre-MantaUI` on the remote if it was ever modified.
   at the bottom". Content growing under the user changes the latter and
   must not change the former.**
 
+  **A scroll event is not a user gesture either — that was generation six.**
+  A `scroll` event with a lower `scrollTop` was read as "the user dragged
+  up", but react-virtuoso writes to the scroller itself: its *upward
+  scrolling compensation* fires whenever a row above the viewport is
+  re-measured (plus the unshift/`deviation` corrections). A tool card
+  landing mid-turn re-measures its row, Virtuoso compensates, and the
+  transcript detached with no user input at all — the "every tool call
+  bounces me out and I have to click jump-to-latest" report. So the two
+  signals are now separate: WHERE the scroller is (the scroll event) and
+  WHETHER the user put it there. `classifyFollowOnScroll` takes a
+  `userIntent` argument and only returns `false` when both agree; intent is
+  a short window (`USER_SCROLL_INTENT_WINDOW_MS`) refreshed by wheel /
+  touch / keydown on the scroller, plus a held flag for a scrollbar-gutter
+  drag (`isScrollbarGutterPress` — a press inside the CONTENT box is a
+  click, e.g. expanding a tool card, and must not vouch for the re-measure
+  scrolls that follow it). Landing within `FOLLOW_THRESHOLD_PX` of the
+  bottom still re-attaches unconditionally.
+
+  Two consequences worth knowing before editing this:
+  - **A programmatic scroll away from the tail must now detach EXPLICITLY.**
+    `scrollToMessage` (the artifacts / ⌘F deep-link jump) calls
+    `setFollowing(false)` itself; it used to rely on the resulting scroll-up
+    being mistaken for a gesture, which is exactly the mistake being fixed.
+  - **`stickToTail` writes twice, across a frame.** Virtuoso publishes
+    `totalListHeightChanged` synchronously inside the ResizeObserver tick,
+    while the list padding carrying part of that height is React state that
+    lands on a later commit — so the first write can read a pre-growth
+    `scrollHeight` and park short of the tail. The rAF write lands on the
+    real one. Both are guarded on the follow state.
+
   Note the tail scroll is `scrollTop = scrollHeight`, not
   `scrollToIndex({ index: "LAST" })`, because the Footer renders below the
   last item.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1298,11 +1298,26 @@ Backup at `~/.tmux.conf.pre-MantaUI` on the remote if it was ever modified.
   WHETHER the user put it there. `classifyFollowOnScroll` takes a
   `userIntent` argument and only returns `false` when both agree; intent is
   a short window (`USER_SCROLL_INTENT_WINDOW_MS`) refreshed by wheel /
-  touch / keydown on the scroller, plus a held flag for a scrollbar-gutter
-  drag (`isScrollbarGutterPress` — a press inside the CONTENT box is a
-  click, e.g. expanding a tool card, and must not vouch for the re-measure
-  scrolls that follow it). Landing within `FOLLOW_THRESHOLD_PX` of the
-  bottom still re-attaches unconditionally.
+  touch / keydown on the scroller, plus a held-pointer flag. Landing within
+  `FOLLOW_THRESHOLD_PX` of the bottom still re-attaches unconditionally.
+
+  **Pointer intent is "a button is held", NOT "the press landed on the
+  scrollbar" — the geometric test is a no-op on the primary platform.** The
+  obvious discriminator is `clientX > left + clientWidth` (the gutter the
+  scrollbar reserves). That works with classic scrollbars and never fires
+  under **overlay** scrollbars, the macOS default, where the bar is painted
+  OVER the content and `clientWidth` is the full border-box width — so a
+  Mac user could not detach by dragging the scrollbar at all, while Windows
+  could. Measured in headed Chromium (the probe is worth re-running before
+  changing this): classic reports `clientWidth` 385 / `offsetWidth` 400,
+  overlay reports 400 / 400, and BOTH dispatch `pointerdown` to the scroller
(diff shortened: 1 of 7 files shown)


antoinedc (commented):
# Review — PR #1090 fix/transcript-follow-tool-calls

**Verdict:** The core fix is correct and matches the invariant you documented in AGENTS.md; typecheck and the full suite are green; the wiring test genuinely pins behavior. One blocking gap: macOS auto-hide (overlay) scrollbars make the gutter discriminator a no-op, so track-click / click-to-jump cannot detach — the exact stuck-follow symptom this PR exists to kill, on the primary desktop platform.

## BLOCKING

**1. `isScrollbarGutterPress` is inert on macOS overlay scrollbars → clicking the scroll track cannot detach (chatUtils.ts:3978, wired at Transcript.tsx:540).**
The discriminator is `clientX - el.left > el.clientWidth`. On macOS Chromium/Electron auto-hide scrollbars, `clientWidth` equals the full content-box width and the bar overlays the content, so `clientX - left <= clientWidth` for any press on the track — it is always classified as a content click, so `draggingGutter` stays false and no intent is recorded. Clicking the track (macOS "click in the track jumps here" / page) then scrolls with `userIntent=false`; if it jumps *up*, `classifyFollowOnScroll` returns `null` and the transcript stays glued to the tail. On Windows/Linux classic scrollbars the bar consumes layout width, so the same gesture detaches — behavior is asymmetric on exactly the platform that matters most. This is one of the enumerated "cannot detach" paths and must be fixed before merge.

Resolve the overlay case (e.g. detect overlay mode via `clientWidth === getBoundingClientRect().width`) and there treat a pointerdown→scroll as intent *only when no re-measure happened between them* (a `totalListHeightChanged` / row re-measure between pointerdown and the scroll means the scroll is Virtuoso compensation, not a click-to-jump — that keeps the tool-card-click false-positive excluded while letting track-click page out). Or, if you'd rather keep it simple, fall back to a short intent window on any scroller pointerdown only when the discriminator is undecidable.

## Non-blocking

- **`keydown` on the scroller (Transcript.tsx:570) almost never fires.** The Virtuoso scroller div has no tabindex and during a turn focus sits in the composer (a sibling, outside the scroller); keydown only bubbles to `el` after the user Tabs into a transcript button. So PageUp/Home/space can neither scroll the transcript nor detach it in the common case — pre-existing (unfocusable scroller), not a regression, but the listener offers false comfort. If keyboard transcript-scrolling is intended UX, attach to a focusable element/`window` with a scroll-position check.
- **False-detach residue:** `pointermove` with `buttons !== 0` (Transcript.tsx:555) marks any held press as intent, so a user press-and-hold / text-drag on a message while a tool card streams and triggers Virtuoso's compensation within the 250ms window can still detach. It is required to cover drag-select autoscroll and overlay-thumb drag, so this is an acceptable tradeoff — just aware of it.
- **`stickToTail` rAF (Transcript.tsx:509-521)** is never cancelled, but both writes are guarded on `followingRef` and no-op on a null scroller, so there's no scroll-fight and no yank after unmount/user-scroll. Per fast-stream `totalListHeightChanged` it schedules one rAF (a single `scrollTop` set) — no perf concern.
- **Effect scope (Transcript.tsx:586)** captures `el` once with deps `[hasMessages, onFollowingChange, scrollerElRef]` (callbacks stable); it re-runs only on the `hasMessages` flip, which is correct. A scroller-node swap with no `hasMessages` change would strand listeners on the old node, but nothing recreates the scroller under that condition.
- **Test coverage:** the pure decision (incl. the identical-numbers intent-only regression) and the mounted wiring test genuinely pin the behavior — good. Uncovered branches: the gutter-press path end-to-end, `scrollToMessage`'s explicit detach (ChatPanel.tsx:417), and the rAF second write.

## Verified

- `npm run typecheck` → clean.
- `npx vitest run src/renderer/chatUtils.test.ts src/renderer/Transcript.test.tsx` → 635 passed.
- `npm test` → vitest 3025 passed / 7 skipped; server node:test 2326 passed; exit 0.
- Cross-checked `node_modules/react-virtuoso/dist/index.mjs:2088` ("Upward scrolling compensation") — confirmed it writes to the scroller (`scrollBy` with a negative delta), so the root-cause claim in the PR and AGENTS.md is accurate.

antoinedc:
Addressed in 0f89cf1. The blocking finding was correct, and a headed-Chromium probe both confirmed it and picked the replacement.

**Measured** (same page, two launches; headless is useless here — it reports zero-width scrollbars in both modes):

| | classic | overlay (`--enable-features=OverlayScrollbar`) |
|---|---|---|
| `clientWidth` / `offsetWidth` | 385 / 400 | 400 / 400 |
| `isScrollbarGutterPress` verdict | true | **false** — no-op, as reported |
| thumb drag → `pointerdown` on scroller | yes, `clientX` 394 | yes, `clientX` 394 |
| scrolls fired with a button held | 6 of 6 | 2 of 9 (track click) |
| plain content click | 0 | 0 |

Geometry cannot separate the two modes; the button state can, and a content click never trips it.

**Fix taken:** `isScrollbarGutterPress` is deleted, not patched. Pointer intent is now a held-button flag plus `releaseUserScrollIntent` — a press that scrolled while held earns the normal input window on release (covers a quick track click whose scrolls land after `pointerup`), a press that never scrolled earns nothing (keeps a tool-card click from vouching for the compensation its own expansion causes). This also removes the `pointermove`-with-buttons handler you flagged; the flag subsumes it. Works identically for classic, overlay, touch and RTL, since none of it is geometric.

Not chased: your suggested "detect overlay mode and treat pointerdown→scroll as intent when no re-measure occurred between them" needs a re-measure oracle the scroll listener doesn't have. The held-button flag gets the same discrimination from a signal already on the event.

**Residue, accepted and documented:** a press-and-hold *on content* spanning a re-measure still counts as intent. That is a text-selection drag in practice, where detaching is the desired behaviour.

**Other findings:**
- `keydown` — kept, now with a comment saying it only fires when focus is inside the transcript (a "Load earlier" button, a link in a message), not during a normal turn. Agreed it is not a regression.
- Test gaps — added: the deep-link detach now has a behavioural test in `ChatPanel.harness.test.tsx` (jump-to-latest appears after `manta-scroll-to-message`), verified red without the explicit `setFollowing(false)`; the pointer paths have wiring tests for both directions (scrollbar drag detaches, tool-card click does not); `releaseUserScrollIntent` has both branches covered. The rAF second write stays covered indirectly by the BET-1003 re-activation test — happy to add a direct one if you want it.

`AGENTS.md` records the measurement so the geometric test isn't reintroduced.

Green: typecheck clean, 3028 renderer + 2326 server, 0 failures. (The 2 server failures you may have seen locally were a stale `node_modules` missing `jsonc-parser`, not the branch — `npm ci` clears them, and they reproduce on an unmodified tree.)

PR #1100, opened 2026-08-18
self-update.sh: upgrade every detected CLI, restart only what changed (BET-1097)

antoinedc:
## BET-1097 — self-update.sh: upgrade every detected CLI, restart only what changed

Stage 2 of the unified-update epic (parallel to the aggregate-payload issue). Box side, bash + one node script. Depends on stage 1 (BET-1095, merged) — reuses `src/shared/cliCatalog.mjs` and `src/server/cliUpdates.mjs` directly.

**Files changed:** 5

- `scripts/upgrade-clis.mjs` (new) — the whole "upgrade every installed AI CLI" loop in node, reusing the stage-1 catalog + detector so bash never re-implements CLI discovery/upgrade. Uses `src/server/cliUpdates.mjs`'s `detectClis` for installed probes + resolved upgrade commands; iterates only CLIs with a resolved (non-`manual`) upgrade command, runs the upgrade with a 10-minute per-CLI timeout, re-reads the version, flags a version string change as a CHANGED CLI, and always writes the state file (`CLIS_CHANGED` comma-separated ids + `OPENCODE_CHANGED=0|1`) even on failure. Non-fatal throughout.
- `scripts/self-update.sh` — deleted `upgrade_opencode()`; calls `upgrade-clis.mjs` at the same point (before the packaged early exit and before any payload replacement) using the pinned runtime node. Early exit now passes `clis_changed` (1 when the state file lists ANY changed CLI, not just opencode). Progress markers become the constant total 7 (`1 Checking for updates`, `2 Updating command-line tools`, `3 Downloading update`, `4 Installing dependencies`, `5 Refreshing AI tools`, `6 Restarting opencode`, `7 Restarting box server`); a skipped step emits nothing and the bar jumps forward. Restarts are now conditional on a single `PAYLOAD_REPLACED` flag (set at the two swap sites: `git reset --hard` and `replace_release_payload`) + `OPENCODE_CHANGED` per the three-rule table:
  - payload replaced → opencode AND server
  - opencode changed, payload not → opencode only
  - only other CLIs changed → nothing
- `scripts/lib/release.sh` / `scripts/lib/release.test.mjs` — `should_skip_self_update` param renamed `opencode_changed` → `clis_changed`; test updated (pure unit test, 2x2 matrix same semantics).
- `src/server/opencodeAdmin.test.mjs` — illustrative progress markers seeded in the poller test updated to the 7-step scheme.

**Base:** origin/main @ d4864bd7

## Verification results

**Manual CLI-upgrade run (issue verification step 1 & 2)** — ran on the box (opencode was 1.18.10, latest 1.18.18; claude 2.1.233; codex + kimi not installed), exactly as the issue specifies:

```
node scripts/upgrade-clis.mjs --progress-step=2 --progress-total=7 --state-file=/tmp/cli.state
MANTA_PROGRESS 2/7 Updating opencode
✓ upgrade-clis: opencode upgraded: 1.18.10 → 1.18.18
MANTA_PROGRESS 2/7 Updating Claude Code
✓ upgrade-clis: Claude Code upgraded: 2.1.233 → 2.1.234
```
- opencode actually upgraded 1.18.10 → 1.18.18 → `OPENCODE_CHANGED=1` ✓
- codex and kimi silently absent (not installed) — no error, no warning ✓
- State file: `CLIS_CHANGED=opencode,claude` / `OPENCODE_CHANGED=1` ✓
- The CLI loop itself performs zero restarts — it is purely the upgrade loop.

**Verification step 3 (CLI-only run restarts neither service):** not runnable in this environment — running the full `self-update.sh` here would `git reset --hard` the worktree and drive systemctl/launchctl restarts against this agent host. The restart decision is code-reviewed: the "nothing" branch fires when `$PAYLOAD_REPLACED=0` and `$OPENCODE_CHANGED!=1` (only non-opencode CLIs changed), and the CLI upgrade step is provably restart-free (shown above). The full-script CLI-only / payload-untouched path needs the real dev box to observe in `~/.manta/self-update.log`.

**Automated verification (`npm run typecheck && npm test`):**
- test: exit 0, 2349 pass / 0 fail / 0 skip
- targeted: `release.test.mjs` + `opencodeAdmin.test.mjs` — 36 pass / 0 fail

**Follow-up filed:** none — all in-scope items delivered per spec (out-of-scope explicitly excluded: UI, bus events, installing absent CLIs, manifest/tarball logic).

### DIFF
--- a/scripts/lib/release.sh
+++ b/scripts/lib/release.sh
@@ -293,27 +293,31 @@ release_is_current() {
   [ "$installed_version" = "$release_version" ]
 }
 
-# should_skip_self_update <installed_version> <installed_git_sha> <release_version> <release_git_sha> <opencode_changed>
+# should_skip_self_update <installed_version> <installed_git_sha> <release_version> <release_git_sha> <clis_changed>
 #
-# The updater's EARLY-EXIT decision (BET-1016): skip the ENTIRE self-update (no
-# download, no reinstall, no restart) only when BOTH the box release is current
-# AND opencode is unchanged. When opencode changed, the box must fall through
-# to its restart step even if the tarball is current — otherwise an opencode-only
-# upgrade would be swallowed by the cheap early exit and leave the upgraded (but
-# un-restarted) binary inert until the next update.
+# The updater's EARLY-EXIT decision (BET-1016, generalised in BET-1097): skip the
+# ENTIRE self-update (no download, no reinstall, no restart) only when BOTH the
+# box release is current AND no CLI changed. When a CLI changed, the box must
+# fall through to its restart step even if the tarball is current — otherwise a
+# CLI-only upgrade would be swallowed by the cheap early exit and leave the
+# upgraded (but un-restarted) binary inert until the next update.
 #
-# Returns 0 (skip) when the box is current AND opencode unchanged; 1 (do not
-# skip) otherwise. Pure: no I/O, no globals. Unit-tested in release.test.mjs
+# `clis_changed` is a flag (0 or 1) the caller sets from the upgrade-clis state
+# file: 1 when ANY changed CLI is listed. opencode is just one row of the
+# catalog; its restart is governed by the conditional restart block, not here.
+#
+# Returns 0 (skip) when the box is current AND no CLI changed; 1 (do not skip)
+# otherwise. Pure: no I/O, no globals. Unit-tested in release.test.mjs
 # alongside release_is_current.
 should_skip_self_update() {
   installed_version="$1"
   installed_sha="$2"
   release_version="$3"
   release_sha="$4"
-  opencode_changed="$5"
+  clis_changed="$5"
 
   if release_is_current "$installed_version" "$installed_sha" "$release_version" "$release_sha" \
-     && [ "$opencode_changed" = "0" ]; then
+     && [ "$clis_changed" = "0" ]; then
     return 0
   fi
   return 1
--- a/scripts/lib/release.test.mjs
+++ b/scripts/lib/release.test.mjs
@@ -440,15 +440,17 @@ test("release_is_current: an unreadable installed stamp is never 'current'", () 
   assert.equal(isCurrent("", "", "0.0.29", "def456"), false);
 });
 
-// --- should_skip_self_update: the BET-1016 early-exit decision ------------------
+// --- should_skip_self_update: the early-exit decision -------------------------
 //
 // The updater's early exit must skip the whole update ONLY when the box is
-// current AND opencode is unchanged. If opencode changed, the box must keep
-// going so its restart step runs — otherwise an opencode-only upgrade is
-// swallowed by the cheap exit and the new binary never gets restarted. This is
-// the full 2x2 (box-current x opencode-changed) matrix.
+// current AND no CLI changed. If a CLI changed, the box must keep going so its
+// restart step runs — otherwise a CLI-only upgrade is swallowed by the cheap
+// exit and the new binary never gets restarted. `clis_changed` is a flag the
+// caller derives from the upgrade-clis state file: 1 when ANY changed CLI is
+// listed (opencode is now just one row of the catalog). This is the full 2x2
+// (box-current x clis-changed) matrix.
 
-function shouldSkip(installedVersion, installedSha, releaseVersion, releaseSha, opencodeChanged) {
+function shouldSkip(installedVersion, installedSha, releaseVersion, releaseSha, clisChanged) {
   const script = `
     log() { :; }; ok() { :; }; warn() { :; }; die() { echo "$*" >&2; exit 1; }
     . "${RELEASE_LIB}"
@@ -456,26 +458,26 @@ function shouldSkip(installedVersion, installedSha, releaseVersion, releaseSha, 
   `;
   const out = execFileSync(
     "bash",
-    ["-c", script, "bash", installedVersion, installedSha, releaseVersion, releaseSha, opencodeChanged],
+    ["-c", script, "bash", installedVersion, installedSha, releaseVersion, releaseSha, clisChanged],
     { encoding: "utf-8" },
   );
   return out.trim() === "skip";
 }
 
-test("should_skip_self_update: box current + opencode unchanged → skip", () => {
+test("should_skip_self_update: box current + no CLI changed → skip", () => {
   assert.equal(shouldSkip("0.0.29", "abc123", "0.0.29", "abc123", "0"), true);
 });
 
-test("should_skip_self_update: box current + opencode CHANGED → do NOT skip", () => {
-  // An opencode-only upgrade on an already-current box must fall through to the
+test("should_skip_self_update: box current + a CLI CHANGED → do NOT skip", () => {
+  // A CLI-only upgrade on an already-current box must fall through to the
   // restart — this is the whole reason the early exit is now conditional.
   assert.equal(shouldSkip("0.0.29", "abc123", "0.0.29", "abc123", "1"), false);
 });
 
-test("should_skip_self_update: box stale + opencode unchanged → do NOT skip", () => {
+test("should_skip_self_update: box stale + no CLI changed → do NOT skip", () => {
   assert.equal(shouldSkip("0.0.29", "abc123", "0.0.30", "def456", "0"), false);
 });
 
-test("should_skip_self_update: box stale + opencode changed → do NOT skip", () => {
+test("should_skip_self_update: box stale + a CLI changed → do NOT skip", () => {
   assert.equal(shouldSkip("0.0.29", "abc123", "0.0.30", "def456", "1"), false);
 });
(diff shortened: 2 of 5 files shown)


antoinedc:
## Review — BET-1097 (PR #1100)

**Base freshness:** merge-base == `origin/main` @ `d4864bd7`. Branch is current; no phantom deletions.
**Typecheck:** `npm run typecheck` → exit 0, no errors.
**Tests:** `npm test` → **2349 pass / 0 fail / 0 skip** (matches the implementer's reported result).
**Duplication:** anti-spaghetti report — 0 clones in the changed scope.

- [x] New `scripts/upgrade-clis.mjs` — the whole per-CLI loop in node, reusing the stage-1 catalog (`cliCatalog.mjs`) + detector (`cliUpdates.mjs`). Verified the API usage matches the real signatures (`detectClis()` → `{id, label, current, upgrade}`, `resolveBinary`, `readVersion`).
- [x] Non-fatal throughout; per-CLI 10-min timeout; `manual` CLIs skipped silently; state file always written (empty values) even on failure.
- [x] Non-installed CLIs (codex/kimi) are silently absent — `detectClis` omits absent binaries, so they never appear.
- [x] `upgrade_opencode()` deleted; single `upgrade-clis.mjs` call placed before the packaged early exit and before any payload replacement, on the pinned runtime `NODE_CMD`; node unavailable → warn + continue with `CLIS_CHANGED=` empty, never `die`.
- [x] `should_skip_self_update` param renamed `opencode_changed` → `clis_changed`; `release.test.mjs` updated (pure unit test kept).
- [x] Constant total 7 progress markers with the correct labels; a skipped step emits nothing; the parser tolerates the repeated step lines.
- [ ] **Conditional restarts — not met for a packaged install on an already-current box.** (Block below)

### Block (fix-here): packaged current-box CLI-only update still restarts BOTH services

The three-rule restart table cannot produce its intended "opencode-only" / "restart nothing" outcomes on a packaged install whose box is already current — which is exactly the drift scenario this issue exists to fix.

Trace (packaged box at latest, only `claude` drifts):
1. `upgrade-clis.mjs` upgrades claude → `CLIS_CHANGED=claude`, `OPENCODE_CHANGED=0` → `CLIS_CHANGED_FLAG=1`.
2. The early exit is bypassed, because `should_skip_self_update` only skips when the box is current **AND** no CLI changed. Any CLI change makes it return "do not skip".
3. The script therefore proceeds to download the (same-version) tarball and runs `replace_release_payload`, which **unconditionally sets `PAYLOAD_REPLACED=1`**.
4. The restart block sees `PAYLOAD_REPLACED=1` → restarts opencode **and** manta-server.

So on a packaged install, `PAYLOAD_REPLACED` is effectively always `1` whenever the script gets past the early exit, and the `elif [ "$OPENCODE_CHANGED" = "1" ]` (opencode-only) and `else` (restart nothing) branches are **unreachable for the current-box case**. The issue's headline requirement — "a CLI-only update must restart nothing," "restart only what actually changed" — is not met: the most common case (box current, a CLI drifted) still restarts manta-server and kills every running agent turn. `PAYLOAD_REPLACED` must reflect whether the payload *actually changed* (the box was stale), not merely that the download/replace step ran; the current-box + CLI-change path should fall through to the conditional restart only, without re-swapping the payload.

Expected behavior after fix:
- current box + opencode changed → restart opencode only
- current box + only other CLIs changed → restart nothing
- stale box (payload actually replaced) → restart opencode + server

**Note on verification step 3.** It was legitimately not runnable here (it would `git reset --hard` and drive service restarts). But the PR's stated reasoning for skipping it — that the "nothing" branch fires when only non-opencode CLIs changed — does not hold for a current packaged box, as the trace above shows. Running the full `self-update.sh` against a current box with only `claude` drifted is the exact test that would have surfaced this.

### Nits (non-blocking)

- Bash emits `MANTA_PROGRESS 2/7 Updating command-line tools` and node also emits `MANTA_PROGRESS 2/7 Updating <label>` per CLI. Harmless (the parser dedups equal steps) but redundant in the log.

antoinedc:
## Fix for the Block — PAYLOAD_REPLACED now reflects whether the payload actually changed

**Commit:** `b76de5a7`. Resolved the review Block exactly as specified: on a current box a CLI-only update no longer restarts both services.

**Root cause** (as the review traced): `should_skip_self_update` returns "carry on" for both a stale box *and* a current box whose CLI(s) changed, so the current-box path still ran `replace_release_payload`, which unconditionally set `PAYLOAD_REPLACED=1` → opencode + server restart. The fix splits that conflated decision in `self-update.sh`:

- **The early exit** is unchanged (`should_skip_self_update` — skip everything only when box current AND no CLI changed).
- **The payload swap is now gated on `release_is_current`.** When the box is already current (but a CLI changed, which is why we're past the early exit), the script falls straight through to deps/tools/restart **without** downloading or re-swapping the payload, so `PAYLOAD_REPLACED` stays 0. `replace_release_payload` now runs (and sets the flag) **only when the box was actually stale**.
- The same principle is applied to the git branch: `PAYLOAD_REPLACED=1` only when the reset actually moves HEAD off `origin/main` (a git box already current with a CLI-only change restarts nothing).

**Resulting matrix (verified):**

| Scenario | Outcome |
|---|---|
| current box, no CLI changed | early exit — nothing |
| current box, only other CLIs changed | restart **nothing** |
| current box, opencode changed | restart opencode only |
| stale box (payload replaced) | restart opencode + server |

**Verification:**
- `bash -n scripts/self-update.sh` → OK
- Decision-matrix probe sourcing the real `release.sh` predicates against the new control flow → all four rows above hold (the review's expected post-fix behavior).
- `npm test` → **2349 pass / 0 fail / 0 skip** (`release.test.mjs` still 21 pass).
- Full-script end-to-end (issue verification step 3) still not runnable in this environment (it resets the worktree + drives systemctl/launchctl); the decision logic is now covered by the matrix probe.

**Nit acknowledged** (redundant `MANTA_PROGRESS 2/7 ...` line from bash + node): intentional per the issue spec — node emits `MANTA_PROGRESS 2/7 Updating <label>` per CLI by design, and the parser dedups equal steps. Left as-is.

antoinedc:
## Review (cycle 2) — BET-1097 (PR #1100), commit b76de5a7

**Base freshness:** PR branched at `d4864bd7`; `origin/main` has since advanced to `140209dc` (BET-1096) but touches entirely different files (`updateTargets`, Settings, rpc) — no overlap, no stale-base corruption, no phantom deletions.
**Typecheck:** `npm run typecheck` → exit 0, no errors.
**Tests:** `npm test` → **2349 pass / 0 fail / 0 skip**.
**Duplication:** anti-spaghetti — 0 clones in the changed scope. (`bash -n scripts/self-update.sh` → OK.)

### Prior Block resolved

The cycle-1 Block (packaged current-box CLI-only update restarts both services) is **fixed and verified**:

- **Packaged branch:** after passing the early exit, the script now re-checks `release_is_current`. A current box that only got past the early exit because a CLI changed **skips the payload swap entirely** — `PAYLOAD_REPLACED` stays 0 and control falls straight through to deps/tools/restart without downloading or re-extracting the tarball. The download/replace (+ `PAYLOAD_REPLACED=1`) now runs only when the box was actually stale.
- **Git branch:** `PAYLOAD_REPLACED=1` only when `HEAD != origin/main` (the reset really moves the checkout); a git box already on `origin/main` with a CLI-only change leaves it 0.

Decision matrix now holds (verified by code trace + full-suite pass):

| Scenario | Restart |
|---|---|
| current box, no CLI changed | early exit — nothing |
| current box, only other CLIs changed | **nothing** |
| current box, opencode changed | **opencode only** |
| stale box (payload really replaced) | opencode + server |

The change is confined to `scripts/self-update.sh` (no scope creep), and there's no regression path where refreshed opencode-tools go un-restarted: a tool/deps change only ever arrives via a payload swap (packaged) or a repo advance (git), both of which set `PAYLOAD_REPLACED=1` → restart opencode+server.

### Acceptance-criteria checklist (updated)

- [x] `scripts/upgrade-clis.mjs` — node loop reusing stage-1 catalog + detector; non-fatal; 10-min per-CLI timeout; `manual` skipped; state file always written; non-installed CLIs silently absent.
- [x] `upgrade_opencode()` deleted; single `upgrade-clis.mjs` call before the packaged early exit + payload replacement, on pinned `NODE_CMD`; node down → warn + `CLIS_CHANGED=` empty, never `die`.
- [x] `should_skip_self_update` param renamed `opencode_changed` → `clis_changed`; `release.test.mjs` updated (pure unit test kept — 21 pass in suite).
- [x] Conditional restarts: payload replaced → opencode+server; opencode changed, payload not → opencode only; only other CLIs changed → nothing.
- [x] Single `PAYLOAD_REPLACED` flag set only where the swap actually runs and only on a real change.
- [x] Constant 7-step progress; skipped steps emit nothing; parser tolerates repeated steps.

**Verification note (step 3):** the full CLI-only `self-update.sh` end-to-end still isn't runnable in this agent environment (it resets the worktree and drives systemctl/launchctl restarts). The decision logic is now correct by inspection and the matrix is covered by the implementer's probe + the passing suite; step 3 remains a dev-box observation. Not blocking.

No Blocks, no Questions. Approving for hand-off to the PM.

PR #1105, opened 2026-08-18
BET-1103: fix(ios): stable identity for the .steps transcript block

antoinedc:
## What

Fixes the `.steps` transcript block's scroll identity so it stops deleting + re-inserting the whole step group on every tool call.

`TranscriptBlock.stableScrollID` for the `.steps` case used the CONCATENATION of every child row id, so adding a tool step changed the block's id and made the diff see a brand-new row — the whole group was torn down and rebuilt on every tool step (the visible jank, and the remove/insert traffic behind the `_TiledView.applyChange` crashes). Now it uses the FIRST row's id, which is fixed for the life of the group (a group grows by appending rows; two groups can't collide since a row id is a unique tool callID/part id; empty groups fall back to a constant, and `uniqueTranscriptRows` suffixes any duplicate anyway).

**One production line** changed in `mobile/native/MantaUI/TranscriptRow.swift` plus 4 tests in `mobile/native/MantaUITests/ChatTranscriptTests.swift`. Nothing else in the repo is touched.

Base: origin/main @ `c8ea7ae`

## Verification (`ios-mantaui action=test-unit` — the merge gate)

- **PR branch** (`multica/BET-1103-stable-steps-group-identity` @ `db965b8`): `552 tests, 1 failure`. The single failure is `ToolOutputPreviewTests.testThirtyLinesKeepsLastTwelveWithPrefix`.
- **Base (`main` @ `c8ea7ae`)**: `548 tests, 1 failure` — **the identical test/failure**.

The `/ToolOutputPreviewTests` failure is pre-existing and unrelated to this change (broken assertion from BET-1085: it checks `output.contains("line 1")`, but the kept tail's first line `"line 19"` contains the substring `"line 1"`, so the assertion always fails). My 4 new `ChatTranscriptTests` all pass — 548 → 552, with the same single pre-existing failure. This PR introduces **zero** new failures.

### DIFF
--- a/mobile/native/MantaUI/TranscriptRow.swift
+++ b/mobile/native/MantaUI/TranscriptRow.swift
@@ -45,7 +45,17 @@ extension TranscriptBlock {
             return "p\(text.hashValue)@\(at?.timeIntervalSince1970 ?? 0)"
         case .file(let attachment):
             return "f" + attachmentIdentity(attachment)
-        case .steps(let content): return "s" + content.rows.map(\.id).joined(separator: "|")
+        // The FIRST row's id, not all of them joined. A step group grows by APPENDING
+        // rows, so the first row's id is fixed for the life of the group while a
+        // concatenation changes on every new step — which made the diff see a
+        // different row and delete + re-insert the whole group per tool call. Same
+        // stable-identity treatment `ToolStep.id` already gets (see the comment on
+        // `ToolStep.id`), applied to the group that contains them.
+        //
+        // Two step groups can never collide: a row id is a tool callID (or the
+        // part id), unique per call. An empty group falls back to a constant, and
+        // `uniqueTranscriptRows` suffixes any duplicate anyway.
+        case .steps(let content): return "s" + (content.rows.first?.id ?? "empty")
         case .notice(let text, let kind): return "n\(kind)\(text.hashValue)"
         case .queuedPrompt(let text): return "q\(text.hashValue)"
         }
--- a/mobile/native/MantaUITests/ChatTranscriptTests.swift
+++ b/mobile/native/MantaUITests/ChatTranscriptTests.swift
@@ -597,6 +597,48 @@ final class ChatTranscriptTests: XCTestCase {
                        "non-colliding blocks must keep their content-stable ids unchanged")
     }
 
+    // MARK: - Step-group identity (BET-1103)
+    //
+    // A step group grows by appending rows. Its id must therefore be fixed for the
+    // life of the group: an id derived from ALL row ids changes on every new step,
+    // so the diff sees a different row and deletes + re-inserts the whole group on
+    // every tool call — visible jank, and the remove/insert traffic behind the
+    // `_TiledView.applyChange` crashes.
+
+    private func step(_ id: String) -> StepGroupRow {
+        .step(ToolStep(id: id, verb: "Read", target: "a.swift",
+                       duration: "0.4s", status: .completed, output: nil))
+    }
+
+    func testStepGroupIDIsUnchangedWhenAStepIsAppended() {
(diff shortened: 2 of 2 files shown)


antoinedc:
## Review — BET-1103: stable identity for the `.steps` transcript block

Reviewed against the issue's Definition of Done. PR #1105, base `origin/main` (merge-base `c8ea7ae`), branched before tip; 3-way diff vs fresh `origin/main` shows only the two intended files — no phantom deletions, no stale-base artifacts.

- [x] **One-line production change** — `TranscriptBlock.stableScrollID` `.steps` case now returns `"s" + (content.rows.first?.id ?? "empty")` with the explanatory comment. Matches the spec verbatim. Returns the first row's id, which is fixed for the group's life (groups grow by appending).
- [x] **Correctness of the change** — verified against the type structure: `StepGroupContent.rows` derives from both `.rows` and `.rollup` cases, so `content.rows.first` is well-defined for plain and rolled-up groups (roll-up is an in-place update, not remove+insert). Distinct groups can't collide (a row id is a unique tool callID/part id); empty groups fall back to `"empty"` and `uniqueTranscriptRows`' `#dup` suffixing remains the collision backstop.
- [x] **Do-NOT list respected** — `StepGroupRow.id`, `ToolStep.id`, `ChatTranscriptMapper.stepIdentity`, `uniqueTranscriptRows`, `TranscriptListView`/`TiledView`/`ListDataSource`/`ChatScreen.swift` all untouched. No UUID/counter/non-determinism introduced. Only `mobile/native/MantaUI/TranscriptRow.swift` + `mobile/native/MantaUITests/ChatTranscriptTests.swift` modified.
- [x] **Four tests added** — exactly the four spec'd tests (`testStepGroupIDIsUnchangedWhenAStepIsAppended`, `testStepGroupIDIsUnchangedWhenTheGroupRollsUp`, `testDifferentStepGroupsGetDifferentIDs`, `testEmptyStepGroupsStillGetUniqueRowIDs`) plus the `step()` helper, placed in the spec'd section. All four **pass** (see gate below).
- [x] **Merge gate (`ios-mantaui action=test-unit`)** — I ran it on the PR branch (`db965b89`). Result: **`UNIT: FAIL`** — but the **only** failure is `ToolOutputPreviewTests.testThirtyLinesKeepsLastTwelveWithPrefix`, which is **pre-existing and unrelated to this change**. Independently corroborated, not just asserted: the plugin's own replayed previous verdict records the *identical* failure against base `main` `c8ea7ae` (`UNIT: FAIL c8ea7ae [main]`, same test, same line `ToolOutputPreviewTests.swift:23`). Every other suite — including the chat suite carrying the 4 new tests (`46 tests, 0 failures`) — is green. **This PR introduces zero new failures.**
- [x] **Nothing else modified** — only the two named files.
- [x] **PR title carries the issue key** — "BET-1103: ...". (Anti-spaghetti report: no `.ts/.tsx/.mjs` changed — Swift-only PR — nothing to scan.)

### Findings

**0 Blocks. 0 Questions. 1 Nit:**

- **Nit (note, non-blocking):** the merge gate currently returns red, but solely because of the pre-existing `ToolOutputPreviewTests.testThirtyLinesKeepsLastTwelveWithPrefix` failure (a broken assertion from BET-1085 — the kept tail's first line `"line 19"` contains the substring `"line 1"`). This is already tracked as **BET-1106** (assigned to manta-pm, who owns the merge hand-off). Not caused by, and not fixable within, this PR's scope. The correct handling is the PM's merge decision accounting for BET-1106, not a block on the implementer.

### Verdict

**PASS** — the change is correct, exactly scoped to the issue, and introduces no new failures. Routing to `manta-pm` for the merge hand-off; the known-red gate is owned by BET-1106.

PR #1107, opened 2026-08-18
fix(ios): assert head lines dropped without substring false-positive in tool preview test (BET-1106)

antoinedc:
Opened automatically for `multica/BET-1106-tool-output-preview-test-assert`, which was pushed without a pull request.

Relates to BET-1106.

### DIFF
--- a/mobile/native/MantaUITests/ToolOutputPreviewTests.swift
+++ b/mobile/native/MantaUITests/ToolOutputPreviewTests.swift
@@ -20,7 +20,8 @@ final class ToolOutputPreviewTests: XCTestCase {
         XCTAssertEqual(output.components(separatedBy: "\n").count, ToolOutputPreview.maxLines + 1,
                        "prefix line + the kept \(ToolOutputPreview.maxLines) lines")
         XCTAssertTrue(output.hasSuffix("line 30"), "the tail must keep the NEWEST lines")
-        XCTAssertFalse(output.contains("line 1"), "the head-most lines must be dropped")
+        XCTAssertFalse(output.contains("line 1\n"), "the head-most lines must be dropped")
+        XCTAssertTrue(output.contains("line 19"), "the tail must start at the 19th line")
     }
 
     func testSingleEnormousLineIsCharacterBounded() {


antoinedc:
## Review — BET-1106 (fix broken ToolOutputPreviewTests assertion)

**0 Blocks · 0 Questions · 0 Nits** — clean.

### Acceptance criteria

1. **Fix `ToolOutputPreviewTests.swift:23` so the assertion can't false-positive on the truncated tail.** ✅ The test now asserts `!output.contains("line 1\n")` (head lines dropped) and `output.contains("line 19")` (tail starts at line 19). I independently traced the logic against `ToolOutputPreview.tail` (ChatModels.swift:607): with 30 lines → `suffix(12)` yields lines 19–30, so the output is `"… \nline 19\n…line 30"`. The only place `"line 1"` can appear is as the prefix of `"line 19"`, where it's always followed by `9`, never `\n` — so `contains("line 1\n")` is genuinely `false`. The fix is sound and preserves the test's intent.

   Note: the implementer deliberately deviated from the issue's literal `"\nline 1"` suggestion (which would still false-positive, since `"\nline 19"` contains `"\nline 1"`). That reasoning is correct and the chosen substring is the right one.

2. **Verify with `ios-mantaui action=test-unit`.** ⚠️ Cannot be run on this Linux box (no Apple toolchain — hard constraint). Implementer (macos) reports local `xcodebuild test -only-testing:MantaUITests/ToolOutputPreviewTests` → 5 tests, 0 failures, including the previously-failing case. CI `ios-build` (compiles app + both test bundles) passed.

### Mechanical / scope

- `git diff origin/main...HEAD`: **1 file** (`ToolOutputPreviewTests.swift`), 2 insertions, 1 deletion — exactly the issue's scope. No scope creep, no under-delivery.
- Base freshness: branched from `30e565c8`; `origin/main` is `ea515d14` (one unrelated BET-1103 commit added since, doesn't touch this file). No phantom deletions. No stale-base blocker.
- Anti-spaghetti report: clean (0 TS/module files changed).
- Checks: `ios-build` ✅, `typecheck-test` ✅, `duplication-gate` ✅.

PR #1116, opened 2026-08-18
Folder picker: stop per-keystroke re-list flicker, delete duplicate fetch path (BET-1117)

antoinedc:
Closes BET-1117.

Renderer-only change to `src/renderer/FolderPickerModal.tsx` + its test. Deletes a whole duplicate fetch pipeline (net-negative lines in the component).

- **One debounced fetch per resolved directory**: the listing effect is now keyed on `listDirPath` (the resolved directory), debounced `LIST_DEBOUNCE_MS = 120`, so typing inside one segment no longer refetches at all.
- **Never tear the list down on revalidate**: a `listSeq` guard rejects stale responses; the rows survive refetch (Loading placeholder only on first load; rows no longer gated on `loading`).
- **Deleted the second fetch pipeline**: the `suggestion` is now `useMemo`-derived from the on-screen listing (`rows`) — no second fsListDirs, no second debounce. `refreshSuggestion` and `suggestion`/debounce state removed.
- **One quiet setter**: `setPathQuiet` (`setPath` + suppress ghost text) used at all non-typing path changes; `setPath` direct in exactly two places (`onPathChange`, `acceptSuggestion`).
- **No field snap-to-home**: the write-back only runs for `dir === ""` or a tilde path.
- **Tests**: shared `settle(h)` helper wired into `mountPicker` + Escape test; added regression test `typing inside one path segment does not re-list the directory`.

Base: origin/main @ 953f0c3d

### DIFF
--- a/src/renderer/FolderPickerModal.test.tsx
+++ b/src/renderer/FolderPickerModal.test.tsx
@@ -15,7 +15,7 @@
 import { describe, it, expect, afterEach } from "vitest";
 import { act } from "react";
 import { installMockApi, mount, type Harness, type MockApi } from "./testHarness";
-import { FolderPickerModal } from "./FolderPickerModal";
+import { FolderPickerModal, LIST_DEBOUNCE_MS } from "./FolderPickerModal";
 
 // The one harness alive right now; the shared afterEach unmounts it so no test
 // leaks a mounted dialog into the next.
@@ -24,6 +24,15 @@ afterEach(() => {
   mounted?.unmount();
   mounted = null;
 });
+
+// The picker debounces its directory listing; give the timer + the fetch
+// it schedules room to land before asserting on rows or ghost text.
+async function settle(h: Harness) {
+  await act(async () => {
+    await new Promise((r) => setTimeout(r, LIST_DEBOUNCE_MS + 80));
+  });
+  await h.flush();
+}
 
 // Install a mock api, mount the picker at the default path, and flush its
 // effects. `apiOverrides` are spread on top of the shared fetch/stub defaults,
@@ -48,6 +57,7 @@ async function mountPicker(
     />,
   );
   await mounted.flush();
+  await settle(mounted);
   return { h: mounted, api };
 }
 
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-1117 (PR #1116)

Verified against the issue's acceptance criteria, not the PR summary. **PASS — 0 Blocks, 0 Questions.**

### Acceptance-criteria checklist (all verified against the code independently)

1. **One debounced fetch per resolved directory** — `LIST_DEBOUNCE_MS = 120` at module scope; `listDirPath` derived (`path.endsWith("/") ? path : parentPath(path)`); the listing effect is keyed on `[listDirPath]` with a `setTimeout` debounce + `clearTimeout` cleanup and the `eslint-disable` retained. ✅
2. **Never tear the list down on revalidate** — `listSeq` ref with `const seq = ++listSeq.current` at the top of `listDir`; a `seq !== listSeq.current` bail after EACH await (fsListDirs, gitListWorktrees, inner catch, outer catch) and in `finally` before `setLoading(false)`. Render gates changed exactly as specified: Loading shows only when `rows.length === 0 && !error`; rows no longer gated on `loading`; markup inside blocks unchanged; no spinner/shimmer/fade. ✅
3. **Second fetch pipeline deleted** — `suggestion`/`setSuggestion` state, `debounce` ref + unmount-clear effect, `refreshSuggestion`, every `setSuggestion(null)` call site, and the `refreshSuggestion(...)` calls in `onPathChange`/`acceptSuggestion` are all gone (grep confirms no leftovers). `suggestion` is now a `useMemo` derived from on-screen `rows` with the `reduce` LCP body copied verbatim. ✅
4. **One quiet setter** — `setPathQuiet` used at the open/initialPath reset, `descend`, `goUp`, `goHome`, `goNavigate`, breadcrumb `onClick`, and the `listDir` normalization write. Plain `setPath` appears exactly at `onPathChange` (which un-suppresses) and `acceptSuggestion`. Escape swaps to `setSuppressSuggestion(true)` with `stopPropagation` intact (first Escape hides suggestion, second closes). ✅
5. **No field snap-to-home** — write-back now fires only for `dir === "" || dir.startsWith("~")`. ✅
6. **Tests** — shared `settle(h)` helper added (as specified) and wired into `mountPicker` after `flush()`; Escape test's hand-rolled 150ms wait replaced with `settle(h)` and the stale "80ms" comment removed; new regression test `typing inside one path segment does not re-list the directory` added asserting `fsListDirs <= baseline + 1`. ✅

### Scope / Do-NOT compliance
- Exactly 2 files changed: `FolderPickerModal.tsx` + its test. `folderPicker.ts`, `src/server/*`, `NewSessionScreen.tsx`, layout/classes, worktree-fan-out branch, and `select()` untouched.
- Component net **−6** lines (spec required net-negative). ✅

### Mechanical gate
- Base fresh: diff vs `origin/main` shows only the 2 intended files — no phantom deletions.
- `npm test` → **2353 pass / 0 fail / 0 skip** (matches the implementer's PR-branch report, corroborating it).
- `FolderPickerModal.test.tsx` → 5/5 pass, including `runs gitListWorktrees at most once per listing` (still exactly 1) and the new regression test.
- Anti-spaghetti report: 0 duplicates.

PR #1102, opened 2026-08-18
fix(renderer,server): let opencode auto-title chat sessions so first auto-rename is sensible (BET-1100)

antoinedc:
## Root cause

A fresh chat window is created with an **initial tmux window name** derived from the first word of the user's prompt (`promptWindowName("I'm just …")` → `im`), AND the opencode session itself is created with a **non-empty title** `"${workspace} / ${windowName}"` (e.g. `manta / im`).

Because opencode is handed a non-empty title at `createSession`, it does **not** auto-generate its own title from the first user message — the session's `title` field stays `manta / im`. The BET-1018 auto-rename *first-name* path (`findSessionTitle` → rename) then reads that **compound, workspace-prefixed string** back and renames the window to `manta / im` on top of the initial `im`. Result: `im` → `manta / im` — not a sensible name, plus a second workspace-prefixed rename.

## Fix

Stop seeding a compound title so opencode auto-titles the session from its first user message (the premise BET-1018 relied on). The auto-rename first-name path now reads opencode's *real* generated title and renames the window once to a clean, correctly-punctuated short name.

- `src/server/tmux.mjs` — `newSession` / `newWindow` pass an **empty** title to `createSession` (was `${name} / ${windowName}` / `${sessionName} / ${windowName}`). The tmux window name is still set from `windowName`; this only stops poisoning the opencode session title.
- `src/renderer/ChatPanel.tsx` — `/clear` creates the fresh session with an empty title (was `${tmuxSession} / cleared`), same fix for the post-clear auto-rename.

No regression to existing behavior: `renameWindow` still skips empty (never blanks a window name), and re-titles still run on opencode's `title` agent (`generateSessionTitle` untouched).

## Tests

- `src/server/tmux.test.mjs` — chat `newWindow` / `newSession` now assert `createSession` receives an empty title (no `workspace / window` compound for the auto-rename to read back).
- `src/renderer/ChatPanel.harness.test.tsx` — `/clear` now expects `title: ""`.
- `src/shared/streamInterpretation.test.ts` — `sanitizeGeneratedTitle` preserves a contraction's apostrophe (`"I'm just checking the manta setup"` → keeps `I'm`, not `Im`).
- `src/renderer/chatUtils.test.ts` — `findSessionTitle` returns a contraction title intact (no apostrophe stripping to a fragment).

[...]

PR #1117, opened 2026-08-18
fix(renderer): fall back to the title agent for the first auto-rename (BET-1100 follow-up)

antoinedc:
## Problem

BET-1100/1101 fixed the "Im then manta / Im" symptom by passing an EMPTY session title at all chat create sites, relying on the premise that **opencode auto-titles a session from its first user message "for free"**. That premise is **false**.

Verified live against the opencode on the box (1.18.10):

- `POST /session { title: "" }` → title is `""`
- after a real first user prompt + completed turn, `GET /session` still shows `title: ""`

So on a fresh chat window the first-name auto-rename path (`findSessionTitle` → rename) reads back an empty title, silently skips, and the window keeps its creation-time **first-word placeholder** ("im"). User-facing symptom: **"the first prompt just renames it to the first word, then no auto rename."**

The prior automated coverage mocked a titled session, so it never exercised the actual (empty) live-box case.

## Fix

In the auto-rename **first-name** path (`src/renderer/ChatPanel.tsx`), when `findSessionTitle` yields no title, **fall through to `opencodeGenerateTitle`** (the `title` agent — the same robust path the every-Nth-turn renames use). This guarantees a fresh window gets **exactly ONE sensible, punctuated rename** after its first turn instead of keeping the first-word placeholder. Verified the title agent returns e.g. `"Manta setup check"` for `"I'm just checking the manta setup"`.

The cheap fast-path (opencode's own title) is preserved for sessions that do carry one.

## Tests

- `src/renderer/ChatPanel.harness.test.tsx` — new regression test: session untitled → first turn calls `opencodeGenerateTitle` exactly once and renames once with the generated title, with no second rename shortly after.
- `npm run typecheck` ✅ · `npm test` ✅ (161 files, 3088 passed / 7 skipped) · `npm run test:server` ✅.

Closes the live regression from BET-1100 (merged on a false premise).

### DIFF
--- a/src/renderer/ChatPanel.harness.test.tsx
+++ b/src/renderer/ChatPanel.harness.test.tsx
@@ -1196,3 +1196,78 @@ describe("ChatPanel auto-rename single rename (BET-1101)", () => {
     expect(api.calls.tmuxRenameWindow ?? []).toHaveLength(1);
   });
 });
+
+// ===== Auto-rename first-turn fallback (BET-1100 follow-up) =====
+//
+// BET-1100/1101 assumed opencode auto-titles a session from its first user
+// message "for free". That premise is FALSE — a session created with an empty
+// title stays empty-titled even after its first turn (verified live against
+// opencode 1.18.10). So the first-name path read back "" and skipped the
+// rename, leaving the creation-time first-word placeholder ("im"). This pins
+// the fix: when opencode has no title, the first turn falls through to
+// opencodeGenerateTitle (the title agent) so it still renames exactly once.
+describe("ChatPanel auto-rename first-turn fallback (BET-1100 follow-up)", () => {
+  let api: MockApi;
+  let bus: MockEventBus;
+  let h: Harness | null = null;
+
+  afterEach(() => {
+    h?.unmount();
+    h = null;
+  });
+
+  it("generates a title when opencode left the session untitled, renaming once", async () => {
+    // opencode has NOT titled the session (title:""), as on a live box.
+    ({ api, bus } = installMockApi({
+      opencodeListSessions: () =>
+        Promise.resolve([{ id: "ses_test", title: "" }]),
+      opencodeGenerateTitle: () => Promise.resolve("Manta setup check"),
+    }));
+    resetStore({ autoRenameSessions: true });
+    h = mount(<ChatPanel {...PROPS} />);
+    await h.flush();
+
+    const textarea = h.container.querySelector("textarea") as HTMLTextAreaElement;
+    const setter = Object.getOwnPropertyDescriptor(
+      window.HTMLTextAreaElement.prototype,
+      "value",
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review (BET-1100 follow-up) — PASS ✅

**Verdict:** 0 Blocks, 0 Questions, 2 Nits → route to manta-pm for merge.

### What this is
Reviews PR #1117 (`multica/BET-1100-followup-first-turn-title`), the follow-up discovered after #1102 merged on a false premise. BET-1100's fix stopped seeding a session title so opencode would "auto-title for free"; the follow-up establishes (verified live on opencode 1.18.10) that opencode does **not** auto-title an empty-titled session even after its first turn. So the first-name path read back `""`, silently skipped, and the window kept its creation-time first-word placeholder (`im`).

### The fix (correct)
In the auto-rename first-name path (`src/renderer/ChatPanel.tsx`), when `findSessionTitle` yields no title the path now **falls through to `opencodeGenerateTitle`** (the title agent — the same robust path the every-Nth-turn renames already use). Both branches funnel through the shared `renameWindow` tail, which applies `sanitizeGeneratedTitle` (contraction apostrophe preserved from #1102) and never blanks a window name. `lastAutoRenamedTurnRef` is set in both branches, so the first-name path fires exactly once — no second rename shortly after.

### Mechanical gate (run independently on PR head)
- Base freshness: branch is **not** stale — `git diff --stat origin/main...HEAD` shows only the 2 intended files (`ChatPanel.tsx` + `ChatPanel.harness.test.tsx`), no phantom deletions.
- `npm run typecheck` → **exit 0** (both tsconfigs).
- `npm test` → **exit 0** (vitest 3088 passed/7 skipped; node:test 2353 pass / 0 fail).
- New regression test **"auto-rename first-turn fallback"** passes (session untitled → `opencodeGenerateTitle` called once, exactly **one** `tmuxRenameWindow`, no second rename 60ms later).
- Existing findSessionTitle/generateSessionTitle cases (6) still pass.

### Anti-spaghetti
CI `duplication-gate` is red, but all clones are **test-fixture boilerplate** in the harness (the standard composer-submit mount sequence, duplicating the existing BET-1101 test's scaffolding) — discounted per judgment rules. No duplication of changed business logic flagged in `ChatPanel.tsx`.

### Nits (non-blocking)
1. The fall-through block (`opencodeGenerateTitle` → `renameWindow` → set `lastAutoRenamedTurnRef`) closely mirrors the every-Nth-turn block; a shared `generateAndRename()` helper would remove the ~6-line duplication. Not blocking.
2. Minor timing inconsistency: the every-Nth-turn path sets `lastAutoRenamedTurnRef.current = turns` *before* the async work, the fall-through sets it *after* — harmless (a failure leaves it unset so the next turn retries first-name), but worth aligning.

PR #1126, opened 2026-08-18
fix(ios): do not clobber hydrated transcript with an empty racing refetch (BET-1125)

antoinedc:
## BET-1125 — do not clobber hydrated transcript with an empty racing refetch

**Base: origin/main @ ae087dd5**

Pre-existing defect (not introduced by BET-1105) that blocks BET-1105's interactive Check-1: opening a session with history rendered blank because a ~1.4s-after-open refetch returned an empty tail and overwrote the just-hydrated `transcript`/`rows`.

### Fix
In `ChatSessionStore.fetchTranscript`, a non-first-load refetch that returns an empty `opencode:messages` result no longer overwrites an already-populated `transcript`. A newly-created / cleared session is always a fresh store whose FIRST load returns empty — that empty stays authoritative. A populated transcript only ever reaches this branch when real history must be kept, so the racing empty pull can't wipe it.

`mobile/native/MantaUI/ChatSessionStore.swift` (+16).

### Verification
- Local `npm run typecheck`: exit 0.
- Local `npm test`: exit 0, all pass.
- Swift: not locally compileable on this Linux box (no Apple toolchain; store not in the headless `Package.swift` subset). On-device/simulator verification is being handed to `macos` against the disposable test box `https://1b41876333f84c8dcf0dd6363dd0a0a1.boxes.mantaui.com`.

### Out of scope
No BET-1105 files touched (`ChatScreen.swift`, `project.yml`, MessagingUI pin, snapshot migration). This is its own branch + PR.


antoinedc:
## BET-1125 review — PASS (0 Blocks, 0 Questions)

Reviewed against BET-1125's acceptance criteria (the store data-loss fix, scoped to `ChatSessionStore.swift`). Reviewed head `d3f285aa` (base `main`).

### Mechanical gates
- `npm run typecheck` → exit 0 (clean).
- `npm test` → 2354 pass / 0 fail.
- CI on the PR: `typecheck-test`, `swiftlint`, `ios-build`, `swift-core-test`, `duplication-gate` all green.
- Anti-spaghetti report: scoped to changed files; no TS/module files in this PR, nothing to scan. `duplication-gate` passes.
- Base freshness (TEN-134): PR branches at `ae087dd5`; `git diff --stat origin/main...HEAD` shows only the +16 in `ChatSessionStore.swift`, zero deletions — no phantom-deleted files, no stale-base concern.

### Scope check
- Diff is confined to `mobile/native/MantaUI/ChatSessionStore.swift` (+16) + an evidence artifact (`mobile/native/evidence/BET-1125/` README + screenshot). Zero BET-1105 files touched. Matches the issue's stated scope exactly.

### Fix correctness — the fetchTranscript guard
```swift
if loaded.isEmpty, !transcript.isEmpty, !isFirstLoad {
    return
}
```
- A racing EMPTY refetch (the ~1.4s-later pull) can no longer overwrite an already-hydrated transcript — the `rows 6 → 0 → blank` mechanism is closed at the store boundary.
- Empty remains authoritative where it must: a fresh store's FIRST load (`isFirstLoad == true`) or a genuinely-empty (`transcript.isEmpty`) store still accepts `[]`. A cleared session swaps session id + rebuilds, so it never has a populated transcript to protect.
- The early `return` sits inside the `MainActor.run` closure — `hasEarlier` / `refreshing` are already set, and the `repeat…while fetchPending` drain loop is untouched, so serialised fetching still coalesces correctly. `rebuildBlocks()` is skipped for the blocked empty result, which is the intended behaviour.

### On-device verification (macos, evidence committed)
Instrumented build opened the existing `default` session (`ses_fec148…`, 6 persisted messages):
```
fetchTranscript sid=jy2RgxFL firstLoad=1 loaded=6 transcriptBefore=0 willBlockEmpty=NO didFail=0
```
Single un-clobbered fetch, `rows` hydrates to 6 and stays; one fetch only, no ~1.4s-later empty refetch, no `6 → 0` reset. Store boundary fix confirmed.

### Out-of-scope note for the coordinator (not a Block / Question)
macos confirmed that after this fix `store.rows` stays 6 yet the transcript still renders blank cells on the open screen. That residual lives in the transcript **view/render** path, which is BET-1105's scope (its in-flight head "drop the local-mirror feed"), not BET-1125's store data-loss fix. This PR correctly fixes its own boundary and should not be held for it — but the visible "blank chat on open" symptom is not resolved at the screen level until the render path under BET-1105 lands. Please carry that forward so it isn't lost in the merge hand-off.

Verdict: PASS — 0 Blocks, 0 Questions, 0 Nits. Routing to manta-pm for merge.

PR #1130, opened 2026-08-18
feat(shared): unwrap provider error bodies to the human sentence (BET-1131)

antoinedc:
## BET-1131 — Provider errors render as raw HTTP bodies: unwrap the human sentence (shared, desktop + iOS)

Adds a single pure helper `humanizeProviderError(raw)` to the shared stream-interpretation module and calls it from **both** consumers, so a provider rejection like:

```
Bad Request: {"detail":"The 'gpt-5.6-sol' model is not supported when using Codex with a ChatGPT account."}
```

now renders as just the human sentence on desktop **and** in the native iOS transcript notice.

**Files changed:** 6.

- `src/shared/streamInterpretation.mjs` — `humanizeProviderError` + `extractErrorSentence`. Splits an optional leading `<reason phrase>: ` iff the remainder is JSON, parses it, and pulls the sentence from the first matching shape (`error.message` before top-level `message`, then `detail`, then `message`). **Lossless fallback:** anything that doesn't parse or match is returned byte-identical.
- `src/shared/streamInterpretation.d.mts` — signature alongside `describeTruncation`.
- `src/renderer/chatUtils.ts` — re-export (renderer keeps one implementation through the shared module).
- `src/renderer/hooks/useSseBus.ts` — in the `session.error` block, `authErrorAdvice` still reads the **original** `raw`; the switch/default branches use `humanizeProviderError(raw)`.
- `src/server/streamInterp.mjs` — applies the same helper to `message` before `emit(sid, "sessionError", …)`, so iOS shows the same clean string. No Swift change needed.
- `src/shared/streamInterpretation.test.ts` — the four body shapes ± reason phrase, lossless fallback (prose / malformed JSON / unknown shape / empty / array `detail`), `error.message` precedence, screenshot case end-to-end, idempotence.

**Base: `origin/main` @ `fea29698`** (refreshed to HEAD before branching).

**Verification results:**
- PR branch (`multica/BET-1131-provider-error-humanize` @ `7a9f36e`):
  - test: exit 0, 3098 pass / 0 fail / 7 skip. log sha256: `837d3e844d1016246e6a86de287f21d6e832d5aaf92f98aadbb4c562acf242ac`
- Base (`main` @ `fea29698`):
  - test: exit 0, 3091 pass / 0 fail / 7 skip. log sha256: `751530ee6db7e9da1e95b6f183a8f81d06c2aea4f10b6982c3aaf3d44bf3c0a2`
- **Conclusion:** 0 new failures; 7 new passing tests (the `humanizeProviderError` suite). No pre-existing failures.

**Out of scope (per ticket):** length capping, `authErrorAdvice` changes, Swift changes, per-provider actionable UI (model-switch offer) — filed as follow-ups separately if warranted.

### DIFF
--- a/src/renderer/chatUtils.ts
+++ b/src/renderer/chatUtils.ts
@@ -92,6 +92,7 @@ export {
   describeTruncation,
   extractSubagentInfo,
   findFlushBoundary,
+  humanizeProviderError,
   hydrateQuestion,
   isAssistantTurnComplete,
   isAssistantTurnInProgress,
--- a/src/renderer/hooks/useSseBus.ts
+++ b/src/renderer/hooks/useSseBus.ts
@@ -69,6 +69,7 @@ import {
   hydrateQuestion,
   authErrorAdvice,
   fetchTranscriptWithRetry,
+  humanizeProviderError,
 } from "../chatUtils";
 import { planModeFromToolPart, isPlanAgent } from "../../shared/planMode.mjs";
 import type { TokenUsage } from "../chatShared";
@@ -522,22 +523,27 @@ export function useSseBus(params: {
           setRunning(false);
           return;
         }
+        // Unwrap a provider rejection body (e.g. `Bad Request: {"detail":"…"}`)
+        // down to the human sentence for every branch below. `authErrorAdvice`
+        // above keeps reading the ORIGINAL `raw` — its regex matches on message
+        // content and its behaviour must not shift (BET-1131).
+        const human = humanizeProviderError(raw);
         let msg: string;
         switch (err?.name) {
           case "ContextOverflowError":
-            msg = `Context full — try /compact: ${raw}`;
+            msg = `Context full — try /compact: ${human}`;
             break;
           case "MessageOutputLengthError":
             msg = "Response truncated (hit output limit)";
             break;
           case "StructuredOutputError":
-            msg = `Structured output failed: ${raw}`;
+            msg = `Structured output failed: ${human}`;
             break;
           case "ApiError":
-            msg = `API error: ${raw}`;
+            msg = `API error: ${human}`;
             break;
           default:
-            msg = raw;
+            msg = human;
         }
         setSendError(msg);
         setAuthReconnect(null);
--- a/src/server/streamInterp.mjs
+++ b/src/server/streamInterp.mjs
@@ -40,6 +40,7 @@ import {
   countUserTurns,
   buildTitlePromptInput,
   buildTitleInstruction,
+  humanizeProviderError,
   ASSUMED_CONTEXT_TOKENS,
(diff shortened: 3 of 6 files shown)


antoinedc:
## Review — BET-1131 (PR #1130)

**Verification performed (independently run on this branch, `fea29698` base):**
- `npm run typecheck` → **exit 0** (no errors)
- `npm test` → **exit 0**, `Test Files 161 passed`, `Tests 3098 passed | 7 skipped`

Matches the implementer's reported two-branch verification exactly.

**Scope check:** diff covers exactly the 6 files the ticket names — the shared helper + `.d.mts`, the desktop hook (`useSseBus.ts`), the box (`streamInterp.mjs`), the `chatUtils.ts` re-export, and the tests. No scope creep, no under-delivery. Base is current `origin/main` (`fea29698`); no stale base, no phantom deletions.

**Acceptance criteria, verified against code (not the summary):**
- `humanizeProviderError(raw)` lives in the shared module, signature added to `.d.mts`, single implementation consumed by both transports (renderer via `chatUtils.ts` re-export, server via direct import) — exactly why it belongs shared. ✓
- Algorithm matches spec: reason-phrase split only when the remainder is JSON; `JSON.parse` in try/catch with **lossless** return `raw` on any throw; extraction order `error.message` → `detail` → `message`; non-string/array `detail` treated as unmatched. ✓
- Desktop hook: `authErrorAdvice` still reads the **original** `raw`; `human` is used in the `ContextOverflowError` / `StructuredOutputError` / `ApiError` / `default` branches. ✓
- Box: same helper applied to `message` before `emit(sid, "sessionError", …)` so iOS gets the same clean string; no Swift change. ✓
- Out-of-scope respected: no length cap, no `authErrorAdvice` change, no typed-name change, no Swift change; model-switch UI parked as BET-1132. ✓
- Test suite covers every required case: screenshot end-to-end, all four body shapes ± reason phrase, lossless fallback (prose / malformed JSON / unknown shape / empty / array `detail` / non-string `detail` / non-object), `error.message` precedence, idempotence. ✓

**Anti-spaghetti:** 1 clone flagged, in `streamInterp.mjs` [411-423] vs [523-536] — pre-existing context/cache emit blocks unrelated to this change. Discounted (coincidental, not business logic introduced here).

**Result: 0 Blocks, 0 Questions.** Clean.

**Nit (non-blocking):** the leading-segment split keys off the first `": "` position where the remainder starts with `{`/`[`. A pretty-printed JSON body that spaces after colons (e.g. `{"error": {"message": …}}`) can be mis-split; `JSON.parse` then throws and the fallback returns `raw` **unchanged** — so it's lossless, no corruption, no regression (worst case a pretty-printed body isn't unwrapped). Providers emit compact JSON in practice; not worth a fix in this ticket.

PR #1143, opened 2026-08-18
feat(models): deprecated models disabled by default with per-model opt-in (BET-1139)

antoinedc:
## Summary

Deprecated models (`OpencodeModel.status === "deprecated"`) are now **disabled by default** (not removed — still importable/visible) in the **main model picker** and **skipped by subagent auto-registration**. A persisted, shared `optInModels` set (`"providerID/modelID"`) flips a given deprecated model back to usable on BOTH paths. Ships exactly one new config set mirroring `deactivatedMainModels`'s existing plumbing — no new IPC channel, no new persistence.

The two axes stay separate: the provider fact (`status === "deprecated"`) and the user's persisted opt-in (`optInModels`). `enabled` and `deactivatedMainModels` are untouched.

## Changes

- **`chatUtils.ts`**: `isDeprecated(m)` predicate (the renderer's single comparison) + `mainPickerGroups(...)` — a main-picker candidate set that KEEPS deprecated models (grouped, sorted, fast-folded exactly as `selectableModelGroups`) but reports which are disabled via `disabledKeys`.
- **`ModelMenu.tsx` / `ModelPicker.tsx`**: a deprecated model not yet opted in renders as a greyed, non-roving row with a small "Enable deprecated" action (outside the selectable/aria-option set). Once opted in it's a normal selectable row. Main picker feeds `mainPickerGroups` and the shared opt-in set.
- **Persistence**: `optInModels: string[]` as an `AppConfig` field + a `store` mirror + `optInModel()` action, mirroring `deactivatedMainModels` verbatim (configGet/configUpdate only). Wired through `ChatPanel` → `InputArea` and `NewSessionScreen` → `ModelPicker`.
- **Subagents**: `reconcileSubagents({ models, existingAgents, deactivated, optIn })` skips deprecated models unless their key is in `optIn`; an already-registered (deprecated) block is still left untouched (preserve-existing unchanged). `syncSubagents` (providers.mjs) and the `opencode:sync-subagents` RPC pass `optIn` through. `ModelsCard` (the SubagentsCard) shows a deprecated model disabled-by-default with the same opt-in action, which immediately re-syncs subagents.

## Deliberate scoping

- The **delegate** model picker (`Cards.tsx`) is unchanged — it keeps filtering deprecated via `selectableModelGroups` (the issue scopes only the main picker + subagents, and the shared opt-in set is what both listed paths read).
- `listModels` / `_normalizeProviderModel` are untouched — deprecated models remain importable.
- No iOS changes (mirrored in a follow-on ticket).

## Note on the two predicates

The issue asks for "one predicate" but `subagentSync.mjs` runs under plain node (server `node:test`; `providers.mjs` imports it) and therefore cannot import the TS `chatUtils.ts`. It carries its own one-line `isDeprecatedModel` for the deprecated-vs-opt-in gate; the renderer path uses `chatUtils.isDeprecated`. Same single guarded comparison in each runtime, necessitated by the module boundary.

## Main-picker opt-in and subagent re-registration

Opting in from the main picker persists to the shared set immediately (main row flips on); subagent re-registration is honored by the **next reconcile** (the set is the single source `reconcileSubagents` reads on every card open / sync). `ModelsCard`'s enable action re-syncs subagents immediately.

**Files changed:** `17` (455 insertions, 21 deletions) — see diff.

**Verification results:**
- PR branch (`multica/BET-1139-deprecated-model-optin @ fc902a36`):
  - test: exit 0, **3110 pass / 7 skip / 0 fail**. Failures: none. log sha256: `08da9a1f33470d1f32df4cd0962e4e9f3c7fdc144b3a203fb664c6d5ae936e07`
- Base (`main @ 4a548c26`):
  - test: exit 0, **3098 pass / 7 skip / 0 fail**. Failures: none. log sha256: `6e1eb7ad06b18974dae093d58b530bc0b7d6a9b852466c3e37392a6e39c86161`
- **Conclusion:** 0 new test failures; 12 new tests added (isDeprecated, mainPickerGroups, subagent opt-in/reconcile, ModelMenu disabled rows), all passing.

**Base: origin/main @ `4a548c26`**

### DIFF
--- a/pr-body.md
+++ b/pr-body.md
@@ -1,0 +1,41 @@
+## Summary
+
+Deprecated models (`OpencodeModel.status === "deprecated"`) are now **disabled by default** (not removed — still importable/visible) in the **main model picker** and **skipped by subagent auto-registration**. A persisted, shared `optInModels` set (`"providerID/modelID"`) flips a given deprecated model back to usable on BOTH paths. Ships exactly one new config set mirroring `deactivatedMainModels`'s existing plumbing — no new IPC channel, no new persistence.
+
+The two axes stay separate: the provider fact (`status === "deprecated"`) and the user's persisted opt-in (`optInModels`). `enabled` and `deactivatedMainModels` are untouched.
+
+## Changes
+
+- **`chatUtils.ts`**: `isDeprecated(m)` predicate (the renderer's single comparison) + `mainPickerGroups(...)` — a main-picker candidate set that KEEPS deprecated models (grouped, sorted, fast-folded exactly as `selectableModelGroups`) but reports which are disabled via `disabledKeys`.
+- **`ModelMenu.tsx` / `ModelPicker.tsx`**: a deprecated model not yet opted in renders as a greyed, non-roving row with a small "Enable deprecated" action (outside the selectable/aria-option set). Once opted in it's a normal selectable row. Main picker feeds `mainPickerGroups` and the shared opt-in set.
+- **Persistence**: `optInModels: string[]` as an `AppConfig` field + a `store` mirror + `optInModel()` action, mirroring `deactivatedMainModels` verbatim (configGet/configUpdate only). Wired through `ChatPanel` → `InputArea` and `NewSessionScreen` → `ModelPicker`.
+- **Subagents**: `reconcileSubagents({ models, existingAgents, deactivated, optIn })` skips deprecated models unless their key is in `optIn`; an already-registered (deprecated) block is still left untouched (preserve-existing unchanged). `syncSubagents` (providers.mjs) and the `opencode:sync-subagents` RPC pass `optIn` through. `ModelsCard` (the SubagentsCard) shows a deprecated model disabled-by-default with the same opt-in action, which immediately re-syncs subagents.
+
+## Deliberate scoping
+
+- The **delegate** model picker (`Cards.tsx`) is unchanged — it keeps filtering deprecated via `selectableModelGroups` (the issue scopes only the main picker + subagents, and the shared opt-in set is what both listed paths read).
+- `listModels` / `_normalizeProviderModel` are untouched — deprecated models remain importable.
+- No iOS changes (mirrored in a follow-on ticket).
+
+## Note on the two predicates
+
+The issue asks for "one predicate" but `subagentSync.mjs` runs under plain node (server `node:test`; `providers.mjs` imports it) and therefore cannot import the TS `chatUtils.ts`. It carries its own one-line `isDeprecatedModel` for the deprecated-vs-opt-in gate; the renderer path uses `chatUtils.isDeprecated`. Same single guarded comparison in each runtime, necessitated by the module boundary.
+
+## Main-picker opt-in and subagent re-registration
+
+Opting in from the main picker persists to the shared set immediately (main row flips on); subagent re-registration is honored by the **next reconcile** (the set is the single source `reconcileSubagents` reads on every card open / sync). `ModelsCard`'s enable action re-syncs subagents immediately.
+
+**Files changed:** `17` (455 insertions, 21 deletions) — see diff.
+
+**Verification results:**
+- PR branch (`multica/BET-1139-deprecated-model-optin @ fc902a36`):
+  - typecheck: exit 0. Errors: none. log sha256: `b3e043f9cf03e405d51dbe05ecf0b311f738ab572192d72a50f6b28221a92f13`
+  - test: exit 0, **3110 pass / 7 skip / 0 fail**. Failures: none. log sha256: `08da9a1f33470d1f32df4cd0962e4e9f3c7fdc144b3a203fb664c6d5ae936e07`
+- Base (`main @ 4a548c26`):
+  - typecheck: exit 0. Errors: none. log sha256: `b3e043f9cf03e405d51dbe05ecf0b311f738ab572192d72a50f6b28221a92f13`
+  - test: exit 0, **3098 pass / 7 skip / 0 fail**. Failures: none. log sha256: `6e1eb7ad06b18974dae093d58b530bc0b7d6a9b852466c3e37392a6e39c86161`
+- **Conclusion:** 0 new test failures; 12 new tests added (isDeprecated, mainPickerGroups, subagent opt-in/reconcile, ModelMenu disabled rows), all passing.
+
+Logs cached at `/tmp/typecheck-pr.log`, `/tmp/typecheck-main.log`, `/tmp/test-pr.log`, `/tmp/test-main.log`.
+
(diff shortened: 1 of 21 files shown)


antoinedc:
**BET-1139 review — 0 Blocks, 1 Question, 2 Nits**

**Pre-flight**
- **Base freshness:** PR branched from `4a548c26`; current `origin/main` is `394a1345`. `git diff --stat origin/main...HEAD` shows **no phantom deletions** — base is fresh, no rebase needed.
- **Typecheck:** `npm run typecheck` → exit 0 (node + web tsconfigs, zero errors).
- **Tests:** `npm test` → exit 0. vitest **3110 passed / 7 skipped**; node (`subagentSync`/server/shared) **2371 passed / 0 failed**. Matches the implementer's reported numbers exactly.
- **Anti-spaghetti:** the flagged jscpd clone is the pre-existing `setProviders`↔`setSubagents` sibling pair; not duplication introduced by this PR. Discounted.

**Acceptance criteria — verified against code, not prose**
- `isDeprecated` predicate in `chatUtils.ts` ✓ (`chatUtils.test.ts`: true/false/null).
- Main picker keeps deprecated models importable/visible but **disabled by default**; opt-in flips them to selectable ✓ (`mainPickerGroups` keeps them grouped/sorted with the same enabled+folding gates, reports `disabledKeys`; `ModelMenu` renders a non-roving greyed row + "Enable deprecated" action). `ModelMenu.test.tsx` asserts a deprecated model is present-but-not-`role=option`, then becomes a normal option once opted in.
- Opt-in NOT routed through `deactivatedMainModels`/`enabled` ✓ (separate `optInModels` set).
- Persistence: `optInModels` AppConfig field + store mirror + `optInModel` action, flowing through the existing `configGet`/`configUpdate` channel — no new IPC/storage ✓.
- Subagents: `reconcileSubagents({ …, optIn })` skips a deprecated model unless opted in; preserves already-registered blocks (deprecated or not) unchanged ✓. Wired through `syncSubagents` (providers.mjs) + the RPC + `ModelsCard`. `subagentSync.test.ts` covers skip/opt-in/preserve/deactivate ✓.
- `listModels`/`_normalizeProviderModel` untouched; no iOS ✓. Delegate picker deliberately unchanged (out of the issue's scope) ✓.

**Question 1 — two literal `"deprecated"` comparison sites instead of one.**
The issue is explicit: *"Do NOT compare the string `deprecated` anywhere except inside `isDeprecated`"* and the hygiene checklist: *"ONE `isDeprecated` predicate; no literal string scattered around the codebase."* This PR introduces `isDeprecatedModel` in `subagentSync.mjs` — a second comparison site holding the same literal.

Your PR-note rationale is the module boundary: the server `.mjs` can't import the renderer TS `chatUtils.ts`. That is a real constraint, and I'll grant the issue is internally inconsistent (it dictates a single predicate living in `chatUtils.ts` while also requiring the server-side `subagentSync` to consume it — which plain-node can't do from a TS renderer module). So I'm not treating this as a hard defect. **But** the single-literal rule exists precisely to keep the check from drifting across copies, and this needs the boundary confirmed rather than asserted before I sign off the hygiene item. Please either:
- **(a)** confirm a genuinely shared home isn't possible (i.e. a small `src/shared/` `.mjs` predicate that both the renderer bundle and the server could import without dragging server code into the renderer), or
- **(b)** consolidate into that shared home if it is feasible.

If (a) is truly the case, the split is acceptable — I just need you to say so with the evidence in a short PR reply.

**Nits**
- Opting in from the **main picker** flips the row on immediately but doesn't re-register the subagent until the next reconcile (`ModelsCard`'s enable action re-syncs immediately). Acceptable since the shared `optInModels` set is the single source both paths read — flagging so it's a known, intentional trade-off.
- `deactivatedMainModels` unused-import cleanup etc. — none found; the diff is scope-clean.

**Route: 0 Blocks + 1 Question → returned to `manta-dev`.** Please answer Question 1 in a PR comment (rationale or a code change), then reassign back to `manta-reviewer` for a re-check.

antoinedc:
Resolving the reviewer's Question (duplicate `"deprecated"` literal in `subagentSync.mjs` vs the one-predicate criterion) by **code change, not assertion**.

The single-predicate requirement is now genuinely met: the literal is compared in **one** source location, `src/shared/modelGuide.mjs:isDeprecated` (a node-safe `.mjs`, so it is importable by both `subagentSync.mjs` on the server and the renderer).

- `chatUtils` re-exports it (`export { isDeprecated }`) — renderer consumers (`ModelsCard`, the picker helpers) unchanged, still call `isDeprecated`.
- `subagentSync.mjs` imports the same predicate instead of its own local copy.
- Two pre-existing literal comparisons in `selectableModelList` and `resolveFastToggle` were also routed through `isDeprecated` while I was here (grep now shows exactly one `=== "deprecated"` comparison in non-test source).

So the earlier "module boundary makes it impossible" claim is resolved by consolidation rather than being asserted. Verification: `npm run typecheck` green; `npm test` 3111 renderer + 2371 server pass, 0 fail.

antoinedc:
**BET-1139 re-review (cycle 2) — PASS: 0 Blocks, 0 Questions**

**Resolution of Question 1 (single `"deprecated"` predicate):** resolved by code change, not assertion. `isDeprecated` now lives once in the shared node-safe `src/shared/modelGuide.mjs` — the natural home, since `subagentSync.mjs` already imports `describeModel`/`familyKey` from it. `chatUtils` re-exports it (renderer consumers unchanged); `subagentSync.mjs` imports the same predicate; two pre-existing literals in `selectableModelList` and `resolveFastToggle` were also routed through it. Verified independently: only one `=== "deprecated"` comparison exists in non-test source (`src/shared/modelGuide.mjs:366`); the `types.ts` match is inside a comment.

**Pre-flight (re-run on revised head `3cb503c0`):**
- Base fresh (`4a548c26` merge-base; no phantom deletions vs `origin/main`).
- `npm test` → exit 0. vitest **3111 passed / 7 skipped**; node **2371 passed / 0 failed**. Matches the implementer's reported numbers.

**All acceptance criteria from the issue still verified against the code** (unchanged from the cycle-1 review, now with the predicate properly shared): main-picker keeps deprecated models importable-but-disabled with opt-in; `optInModels` persistence mirrors `deactivatedMainModels` plumbing (no new channel); `reconcileSubagents` skips deprecated unless opted-in and preserves existing blocks; tests cover all three required areas; no `listModels`/iOS changes; no `deactivatedMainModels`/`enabled` overloading.

**Remaining (non-blocking) nits**, unchanged from cycle 1: opting in from the *main picker* flips the row on immediately but re-registers the subagent on the next reconcile (only `ModelsCard`'s enable re-syncs immediately) — acceptable since both paths read the same shared set.

Route: clean → handing off to `manta-pm` for merge.

PR #1151, opened 2026-08-18
feat(renderer): inline media transcript renderer (MediaBody + reserved aspect box) (BET-1148)

antoinedc:
## Inline media: transcript renderer (MediaBody card + reserved aspect box) — BET-1148

Renderer half of inline media — draws generated images and video **in the transcript**, with a placeholder that reserves the final space while the media is still being produced.

**Files changed:** 18 files, +999/−14.

### What ships
- **`MediaBody.tsx`** — the single media renderer, used by BOTH call sites (the media card and the `file`-part branch). Renders a `ToolCard` body on the inset surface, **expanded by default**. The load-bearing rule: the placeholder reserves the media's final aspect box (`resolveMediaAspect`: explicit dims → declared ratio → labelled 16:9 default) in **all four** states (pending/ready/failed/expired), so a `show` swaps the media in without moving anything below it.
- **`media` bus consumption**: one `media` kind + `onMedia` accessor in httpApi (mirrors `appControl`), one App-level listener that switches on `action` (begin/show/fail), routing per-session state keyed by `messageID` into the store; ChatPanel → Transcript → MessageRow threads a `mediaByMessageId` map (a stable memo, so the row memo chain holds).
- **Pending**: skeleton + elapsed clock (the shared 1s ticker) + declared title + variant grid (2 cols, max 4 numbered tiles, `+N more`). **Never** a fabricated percentage.
- **Ready**: bytes fetched through the existing `/api/peek` route with the existing auth headers (HEAD size guard, Blob object URL — no new fetch path). Image ``; **video never autoplays** (`<video controls preload="metadata">`).
- **Failed/expired**: labelled placeholder inside the reserved box — never a broken-image glyph, never blank.
- **Click opens the existing `ArtifactPreview` overlay** — no second lightbox. Added a `video` case to `resolvePreviewType` + a video renderer in `ArtifactPreview` so video previews work too.
- **File-part branch** (`ToolCall.tsx`): an `image/*`/`video/*` `file` part renders `MediaBody`; any other mime keeps today's header-only card.
- Pure logic (aspect resolution, grid, entry state derivation `applyMediaEvent`, mime/kind mapping, `dispatchMedia`, `filePartToMediaEntry`) lives in `chatUtils.ts`, unit-tested.

### Hygiene
- Net new component count = **one** (`MediaBody`); the file-part branch is a substitution, not an added duplicate renderer.
- No second lightbox, no second fetch path, no second elapsed-timer impl, no second aspect-box impl.
- No `className` escape hatch added to `ToolCard`/`OutputWell`; no `manta-*` mobile hooks; no hex literals (tokens/utilities only). `bg-inset` chrome-ownership declared for MediaBody.

### Known boundary
The media card paints when the owning message row exists in the transcript (state keyed by messageID). A `begin` that lands before the message row is present won't paint until the row appears — within the range of normal tool-call timing, acceptable; noted here as a design boundary, not filed.

**Base: origin/main @ `92b803ea`**

**Verification results:**
- PR branch (`multica/BET-1148-media-renderer` @ `bc93f1cd`):
  - test: exit 0, 3143 pass / 0 fail / 7 skip (vitest 162 files). Server: 2418 pass / 0 fail. log sha256: `09cc938fd6a8118a1a3685f94e94a3966d0e5d3502393978cca4df8107fce7c7`
- Base (`main` @ `92b803ea`):
  - test: exit 0, 3115 pass / 0 fail / 7 skip. log sha256: `0010c5b86f4caaebd621388ffc7e9092868e708dcf3903b191a0cb1611ff5529`
- **Conclusion:** 0 new failures. The 3143-vs-3115 delta is this PR's new tests (inline-media helpers + MediaBody + file-part regression). Logs cached at `/tmp/typecheck-{pr,main}.log`, `/tmp/test-{pr,main}.log`.


antoinedc:
## Review — BET-1148 Inline media: transcript renderer (MediaBody + reserved aspect box)

**Verdict: PASS (0 Blocks, 0 Questions, 2 Nits).** Routing to manta-pm for merge.

### Mechanical gates

- `npm run typecheck` — **exit 0** (tsc tsconfig.node + tsconfig.web, no errors).
- `npm test` — **exit 0**: vitest **3143 passed / 0 failed / 7 skipped**; server suite **2418 passed / 0 failed**.
- These match the implementer's reported PR-branch numbers (3143/0/7, server 2418/0) exactly, and its declared base (`main @ 92b803ea`, 3115/0/7) equals this PR's merge-base. Trust-but-spot-check: confirmed.
- CI `typecheck-test` **pass**s. `duplication-gate` fails (see anti-spaghetti below) — de-required, non-blocking.
- **Base-freshness:** merge-base `92b803ea` vs `origin/main` `ce59a2df`. The extra upstream commits (BET-1138 iOS, BET-1105 iOS, BET-1127/BET-1150) don't touch any of this PR's 18 files; `git diff origin/main...HEAD` shows only this PR's files, no phantom deletions. Not a stale-base problem. PR is MERGEABLE.

1. ✅ **One `MediaBody`, both call sites** — `MediaBody.tsx` is the only new component; used by the `media`-bus card (`MessageRow`) *and* the `file`-part branch (`ToolCall.tsx:151`). Neither `ToolCard.tsx` nor `OutputWell.tsx` gained a `className` escape hatch.
2. ✅ **Renders as expanded ToolCard body on the inset surface** — `MediaBody` wraps a `ToolCard` (tone `error` on degraded), `expanded` defaults `true`, header collapses it. The ToolCall file-part is a straight substitution of the old header-only card.
3. ✅ **No autoplay** — `<video controls preload="metadata">`; no autoplay attr. Pinned by test.
4. ✅ **Click → existing `ArtifactPreview`** — no second lightbox. Added `video` to the `Loaded` union + a video renderer; routing is via `resolvePreviewType(mime)` so a video mime reaches the new block. Video `loadArtifact` shares the existing HEAD size guard + object-URL lifecycle.
5. ✅ **Variant grid** — 2 cols, max 4 numbered tiles, `+N more` fold beyond (pure `mediaGrid`, tested). As a pending skeleton; the server contract (`show` carries one path) resolves to a single ready media — consistent with the merged BET-1147 server (which I cross-checked in `src/server/media.mjs`: begin/show/fail payload shape + orphan sweep match `applyMediaEvent` exactly).
6. ✅ **Bytes via existing `/api/peek` + auth headers, no new fetch path** — HEAD → `isWithinPreviewSize` guard → GET blob → object URL, mirroring `ArtifactPreview.tsx`.
7. ✅ **No new Settings/config/retention.**
8. ✅ **Pure logic in `chatUtils.ts` with tests** — `resolveMediaAspect` (explicit dims → ratio → 16:9), `mediaGrid`, `applyMediaEvent`, `mediaKindFromMime`/`isMediaMime`, `filePartToMediaEntry`, `dispatchMedia`, `formatDuration` reuse. All unit-tested; `video` case added to `resolvePreviewType`.
9. ✅ **Reserved box in all four states; failed/expired labelled + no ``** — covered by `MediaBody.test.tsx` (pending, ready, failed, expired, no-autoplay, collapse). `DegradedBox` renders a labelled placeholder in the reserved box.
10. ✅ **State keyed by (sessionID, messageID)** — store `inlineMedia`, one App-level `onMedia` listener switching on `action` via pure `dispatchMedia`. Mirrors `appControl` exactly as specified (httpApi kind/`onMedia` accessor; shared `Api` + `MediaEventPayload`).
11. ✅ **Memoized leaf / no keystroke re-render** — `React.memo` on `MediaBody`; stable `mediaByMessageId` map memo at ChatPanel scope; MessageRow memo chain held.
12. ✅ **Hygiene/removal** — net +1 component; no second timer (shared `useClockTick`/`WORKING_TICK_MS`); no `manta-*` hook classes; no hex literals (tokens/utilities only; `bg-inset` chrome-ownership registered in the M527 primitive list). The file-part substitution is an actual replacement, not an append.

### Nits (non-blocking)

- **N1 — missing "possibly reflowing" label (spec ladder #2).** When no dimensions OR aspect ratio are known, the placeholder reserves the 16:9 box (correct — one bounded reflow, never zero height) but `PendingBox` doesn't label the card "may reflow" as the spec's degrade-ladder case 2 requests. Cosmetic only; the mechanical load-bearing property is honored. Optional to add a small `text-text-quiet` note when the reserved aspect fell back to the default.
- **N2 — `filePartToMediaEntry`'s ready box has no dims from the part** (`width/height` stay null), so a media reached via a `file`-part swap-in always goes through a 16:9→measured reflow. Within spec's allowed bounded reflow; a future server could stamp dims onto the file part to make it zero-movement.

### Anti-spaghetti

Report present. 2 clones, both in `Transcript.test.tsx` and both pre-existing `mountFollowing()` pin-to-bottom test scaffolding (this PR touches that file by one added prop). Coincidental test-fixture clones — discounted, per the judgment rules. Not business logic the PR introduced.

### Known boundary (accepted, not a defect)

A `begin` landing before its message row exists won't paint until the row appears — matches the implementer's noted design boundary; within normal tool-call timing.

PR #1160, opened 2026-08-18
BET-1156: one laptop-download path; drop dead Attach from inline-media preview; hover download overlay

antoinedc:
## BET-1156 — one laptop-download path; drop dead Attach from inline-media preview; hover download overlay

**Base: origin/main @ `e461eb75`**

### What & why
Three coupled fixes for inline media on desktop (BET-1148 feature lives on main):

1. **Dead toast Save → real download.** Desktop's "AI sent you a file · Save" used a renderer blob `<a download>` with no main-process handler, so nothing landed on the Mac. Added **one** desktop bridge (`downloadFileToDownloads` IPC): main fetches `/api/download?path=<remotePath>` with the live box token and writes a real file to `downloadsDir` (default `~/Downloads`), deduping a name collision. `agentPullFile` now delegates to this bridge on desktop (returns the local path → toast flips to **Reveal**) and keeps the browser-blob fallback for mobile/web.
2. **Inline-media preview has a working Download and no Attach.** `ArtifactPreview.onAttach` is now optional (`null` → button hidden); `MediaBody` passes `null` and a real `onDownload` (via `agentPullFile`). The Artifacts panel keeps passing a real `onAttach`, so its preview keeps Attach.
3. **Hover Download overlay** on ready inline images **and** videos — an absolute top-right button inside the existing `data-media-box`, hover-revealed, `pointer-events-none` wrapper so only the button captures clicks and the whole-box click-to-preview still works.

**One desktop download path.** Toast Save, inline-media preview + hover, and the Artifacts panel all funnel through `downloadFileToDownloads` on desktop. The only remaining `<a download>` blob paths are the sanctioned mobile/browser fallbacks (`httpApi.agentPullFile`, `ArtifactsPanel.downloadArtifact`).

### Design note (deviation from issue text)
The issue asked to extract the download-dir + filename-dedupe pure logic into `chatUtils.ts`. That logic executes in the **Electron main process**, which must not import renderer `chatUtils`. I placed it in a node-safe main module `src/main/download.ts` (`resolveDownloadDir`, `uniqueLocalPath`, plus the injectable-fetch orchestration) with a vitest — the same pure-module pattern as `src/main/unpair.ts`. Renderer and main both typecheck cleanly and no process boundary is crossed.

### Files changed
16 files: the bridge (shared `types`/`api`, `preload`, `preloadAccess`, `main/index` + `main/download`), renderer consumers (`MediaBody`, `ArtifactPreview`, `ArtifactsPanel`, `httpApi`), + tests.

**Verification results:**
- PR branch (`multica/BET-1156-laptop-download-path` @ `cc5485fc`):
  - test: exit 0 — vitest 164 files / 3168 pass / 7 skip; server 2418 pass / 0 fail
- Base (`origin/main` @ `e461eb75`): no base-branch run needed (0 new failures; added only non-overlapping tests; no existing test modified)
- **Conclusion:** 0 new failures. All suites green on the PR branch.

### DIFF
--- a/src/main/download.test.ts
+++ b/src/main/download.test.ts
@@ -1,0 +1,164 @@
+import { describe, it, expect, vi } from "vitest";
+import {
+  downloadFileToDownloads,
+  resolveDownloadDir,
+  uniqueLocalPath,
+  type DownloadFs,
+  type DownloadInput,
+} from "./download.js";
+
+// ---- pure helpers -----------------------------------------------------------
+
+describe("resolveDownloadDir", () => {
+  it("uses the configured downloads dir when set", () => {
+    expect(resolveDownloadDir("/Users/a/Desktop", "/Users/a/Downloads")).toBe(
+      "/Users/a/Desktop",
+    );
+  });
+
+  it("falls back to the default dir when absent or blank", () => {
+    expect(resolveDownloadDir(undefined, "/Users/a/Downloads")).toBe("/Users/a/Downloads");
+    expect(resolveDownloadDir("", "/Users/a/Downloads")).toBe("/Users/a/Downloads");
+    expect(resolveDownloadDir("   ", "/Users/a/Downloads")).toBe("/Users/a/Downloads");
+  });
+});
+
+describe("uniqueLocalPath", () => {
+  const exists = (paths: string[]) => (p: string) => paths.includes(p);
+
+  it("returns the bare path when the name is free", () => {
+    expect(uniqueLocalPath("/d", "a.png", () => false)).toBe("/d/a.png");
+  });
+
+  it("appends a numeric suffix before the extension on collision", () => {
+    const e = exists(["/d/a.png", "/d/a (1).png"]);
+    expect(uniqueLocalPath("/d", "a.png", e)).toBe("/d/a (2).png");
+  });
+
+  it("keeps incrementing across multiple collisions", () => {
+    const e = exists(["/d/r.pdf", "/d/r (1).pdf", "/d/r (2).pdf"]);
+    expect(uniqueLocalPath("/d", "r.pdf", e)).toBe("/d/r (3).pdf");
+  });
+
+  it("basenames and defaults an empty filename", () => {
+    expect(uniqueLocalPath("/d", "../evil.txt", () => false)).toBe("/d/evil.txt");
+    expect(uniqueLocalPath("/d", "", () => false)).toBe("/d/file");
+  });
+
+  it("does not treat a leading-dot hidden file as having an extension", () => {
+    const e = exists(["/d/.env"]);
+    expect(uniqueLocalPath("/d", ".env", e)).toBe("/d/.env (1)");
+  });
+});
+
+// ---- orchestration ----------------------------------------------------------
+
+// The 32-hex box-id / box-token fixture is built at runtime (repeat + concat)
+// so no contiguous 32-hex literal appears in source — the gitleaks
+// generic-api-key secret scan flags such literals and fails the required CI
+// job (same technique the repo already uses elsewhere: "a".repeat(32) in the
+// server auth fixtures). The bound value stays a valid 32-hex token.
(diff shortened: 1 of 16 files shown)


antoinedc:
## BET-1156 review — PASS ✅ (0 Blocks, 0 Questions, 2 nits)

**Base-freshness gate:** branch is cut from `origin/main @ e461eb75` (merge-base == current `origin/main` tip). Diff vs fresh main shows only additions + small edits — no phantom deletions. Base is current.

**Mechanical gate (ran on PR branch `cc5485fc`):**
- `npm test` → exit 0. Vitest 164 files / 3168 pass / 7 skip (both main + renderer suites); server 2418 pass / 0 fail. Matches the implementer's reported results.
- Targeted re-run of the four new test files: `download.test.ts`, `ArtifactPreview.test.tsx`, `MediaBody.test.tsx`, `httpApi.test.ts` → 4 files / 66 tests pass.

**Acceptance-criteria checklist (verified independently, not from prose):**

1. **Toast Save is a real desktop download → Reveal.** ✅ `downloadFileToDownloads` IPC bridge added; main fetches `/api/download` with the live box token (read from main's `config`, never caller-supplied, read at invoke time) and writes to `downloadsDir` (absent → `app.getPath("downloads")`). `agentPullFile` delegates on desktop and returns the local path (truthy → toast flips to Reveal); on failure it throws so the toast keeps its retry affordance.
2. **Dead Attach removed from inline-media preview; its Download works.** ✅ `ArtifactPreview.onAttach` is now `((artifact) => void) | null`, the Paperclip IconButton renders only when a real handler is passed. `MediaBody` passes `null` (no Attach) and a real `onDownload` → `agentPullFile`.
3. **Hover Download overlay on inline images AND videos.** ✅ Overlay rendered once inside the shared `data-media-box` container guarded by `status === "ready" && meta.path`, so it shows on both the ready `` and ready `<video>`.
4. **Overlay lives inside the reserved box + doesn't break click-to-preview.** ✅ Overlay wrapper is `pointer-events-none absolute top-1 right-1`, hover-revealed via `group-hover:opacity-100`; only the button has `pointer-events-auto` and calls `stopPropagation()`, so the whole-box `onClick={openPreview}` still fires.
5. **Keep Attach in ArtifactsPanel.** ✅ Panel rows (attachArtifact) and the panel's `ArtifactPreview` (line 1042) both still pass a real `onAttach`.
6. **ONE desktop download path.** ✅ Toast Save, inline-media preview + hover, and the artifacts panel all funnel through `downloadFileToDownloads` on desktop. `downloadArtifact` now routes box-file hrefs through `agentPullFile` → bridge on desktop (data: URLs keep the blob path). The only remaining blob-`<a download>` paths are the sanctioned mobile/web fallbacks.
7. **`agentPullFile` keeps its `Promise<string>` contract + fallback.** ✅ Desktop → bridge (returns local path). Non-desktop → unchanged browser-blob path returning `""`. Tests cover all three branches.

**Hygiene:** removal-over-addition honored (dead blob-desktop route, dead Attach, and the two `() => {}` stubs are gone); reuses `downloadsDir` + `/api/download` + `ArtifactPreview` + `data-media-box`; tokens only (no hex colours); no `manta-*` hooks; stale `shared/types.ts` scp comment updated.

**Tests:** all required cases present and green — Attach hidden/shown, ready image + ready video hover overlay press→download + pointer-events guard (press does not open preview), bridge-preference + fallback, and unit tests for `resolveDownloadDir` / `uniqueLocalPath` / orchestration.

**Nits (non-blocking):**
- **Pure-logic file placement.** The issue suggested `chatUtils.ts`, but this logic runs in the Electron main process, which cannot import renderer `chatUtils`; the node-safe `src/main/download.ts` + vitest is the correct call (matches the `unpair.ts` precedent). Documented in the PR body — no action.
- **Anti-spaghetti:** the single reported clone (0.10%) is repeated `mount(<MediaBody …>)` test fixtures — a test-fixture coincidence, not business-logic duplication. Discounted.

Clean pass — routing to `manta-pm` for merge.

PR #1161, opened 2026-08-18
BET-1158: fix(scripts): resolve NODE_CMD + CLI PATH so self-update upgrades AI CLIs on no-runtime boxes

antoinedc:
## What

Fixes `scripts/self-update.sh` silently skipping every AI CLI upgrade on boxes whose runtime isn't vendored under `$MANTA_HOME/runtime/node/bin` (any git-checkout box).

**Root cause (BET-1158):** with no vendored runtime, `NODE_CMD` fell back to the bare literal `node`, and the `[ -x "$NODE_CMD" ]` guard ran `-x node` — a CWD-relative filesystem test, NOT a PATH lookup. It was false even when node was on PATH, so every CLI upgrade was skipped (the box's self-update log showed `⚠ self-update: node unavailable — skipping CLI upgrades` while `npm ci` ran fine moments later).

**Fix (shell-only):**
1. Resolve the fallback to an absolute path via `command -v node` (`NODE_CMD="$(command -v node || true)"`) so the `-x` guard and the invocation agree. `|| true` yields empty on a genuinely missing node, which the guard already treats as "not available".
2. Prepend the CLI install dirs (`~/.local/bin`, `~/.opencode/bin`, `~/.bun/bin`) to `PATH` for the upgrade step — `upgrade-clis.mjs` spawns each CLI's upgrade command **by name** and a systemd/launchd service's minimal PATH never carries those dirs. Same home-relative set `resolveBinary()` in `src/server/cliUpdates.mjs` searches.

Verified with `bash -n`. node-pty/build unrelated.

## Anti-spaghetti

Single root-cause site: the `NODE_CMD` fallback + the upgrade step's PATH. Fixed at the source; no call-site band-aid. Relation to the existing vendored-runtime PATH pinning (BET-829): kept separate/adjacent, both guard on dir existence.

## Tests

Two new regression tests in `scripts/install.test.mjs` that run the **real shipped block** out of `self-update.sh` against a fake box layout:
- `NODE_CMD`: a no-runtime box (node on PATH) now passes the `-x` guard → CLI upgrades run (old bare-`node` fallback made it skip).
- CLI dirs: `~/.local/bin` + `~/.opencode/bin` are prepended to PATH; an absent dir (`~/.bun/bin`) is correctly skipped.

**Files changed:** 2. `scripts/self-update.sh`, `scripts/install.test.mjs`

**Verification results:**
- PR branch (`multica/BET-1158-self-update-node-fallback` @ `677bea64`):
  - typecheck: exit 0. Errors: none. log sha256: `b3e043f9…`
  - test: exit 0, 2424 pass / 0 fail / 0 skip. Failures: none. log sha256: `2f7a28bc…`
- Base (`main` @ `e461eb75`):
  - typecheck: exit 0. Errors: none. log sha256: `b3e043f9…`
  - test: exit 0, 2422 pass / 0 fail / 0 skip. Failures: none. log sha256: `13a01937…`
- **Conclusion:** `+2` tests (the new BET-1158 regression tests), `0` new failures, `0` pre-existing failures. Both suites green.

Base: origin/main @ `e461eb75`

### DIFF
--- a/scripts/install.test.mjs
+++ b/scripts/install.test.mjs
@@ -4900,6 +4900,83 @@ test("self-update.sh does NOT restart opencode — that would kill in-flight age
   );
 });
 
+// ----------------------------------------------------------------------------
+// scripts/self-update.sh — NODE_CMD / CLI-PATH resolution (BET-1158).
+// ----------------------------------------------------------------------------
+//
+// A git-checkout box has no $MANTA_HOME/runtime/node/bin, so NODE_CMD falls
+// back to whatever node is on PATH. BET-1158: the old fallback was the bare
+// literal `node`, and the `[ -x "$NODE_CMD" ]` guard then ran `-x node` — a
+// CWD-relative filesystem test that is false even when node is on PATH — so
+// every CLI upgrade was silently skipped ("node unavailable"). The fallback
+// must resolve to an ABSOLUTE path via `command -v node`, and the CLI install
+// dirs (~/.local/bin, ~/.opencode/bin, ~/.bun/bin) must be prepended to PATH
+// so upgrade-clis.mjs's by-name upgrade commands resolve. Both tests run the
+// REAL shipped block out of the actual script against a fake box layout, so a
+// regression in the shipped source is what goes red.
+function runSelfUpdateSnippet(block, pre = "", extra = "") {
+  const dir = mkdtempSync(join(tmpdir(), "manta-selfupd-snip-"));
+  const path = join(dir, "snippet.sh");
+  writeFileSync(path, `set -euo pipefail\n${pre}\n${block}\n${extra}\n`, { mode: 0o755 });
+  try {
+    const res = spawnSync("bash", [path], { encoding: "utf8", env: process.env });
+    return `${res.stdout ?? ""}${res.stderr ?? ""}EXIT=${res.status}`;
+  } finally {
+    rmSync(dir, { recursive: true, force: true });
+  }
+}
+
+test("self-update.sh NODE_CMD: no-runtime box resolves node to an absolute path so the CLI-upgrade guard passes (BET-1158)", () => {
+  const src = scriptFile("self-update.sh");
+  const start = src.indexOf('if [ -x "$MANTA_HOME/runtime/node/bin/node" ]; then');
+  assert.ok(start !== -1, "self-update.sh must resolve NODE_CMD against the vendored runtime first");
+  const end = src.indexOf("\n\n", start);
+  assert.ok(end !== -1, "NODE_CMD resolution block must be followed by a blank line");
+  const block = src.slice(start, end);
+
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-1158 (self-update silently skips CLI upgrades off-runtime)

**Verdict: PASS — 0 Blocks, 0 Questions.** Routing to PM for merge.

### Acceptance criteria (from issue) — all verified independently

1. **NODE_CMD fallback resolves to an absolute path** — `scripts/self-update.sh` now uses `NODE_CMD="$(command -v node || true)"` instead of the bare literal `node`, so the `[ -x "$NODE_CMD" ]` guard and the invocation agree. Confirmed in the diff.
2. **CLI install dirs prepended to PATH for the upgrade step** — `~/.local/bin`, `~/.opencode/bin`, `~/.bun/bin` are prepended (each guarded by `-d`), covering `upgrade-clis.mjs`'s by-name spawns under a minimal service PATH. Confirmed in the diff.
3. **Shell-only fix** — only `scripts/self-update.sh` (shell) + `scripts/install.test.mjs` changed. `bash -n scripts/self-update.sh` → clean.
4. **Regression tests run the real shipped block** — both tests slice the actual shipped source from `self-update.sh` (not a copy), run it via `bash` on a fake no-runtime box, and assert the guard passes / CLI dirs land on PATH / an absent dir is skipped. Both new tests pass.

### Mechanical gates
- `npm test` → **2424 pass / 0 fail / 0 skip** (matches the implementer's reported PR-branch result exactly).
- Base freshness: PR branches off `origin/main` (merge-base == origin/main, no stale base, no phantom deletions). Diff is exactly 2 files, in-scope.
- Duplication detector: 0 clones; nothing Block-worthy.

No Block- or Question-severity findings.

antoinedc:
## Re-review after branch refresh — BET-1158

**Verdict: PASS (unchanged) — 0 Blocks, 0 Questions.**

The fix code is byte-for-byte as previously reviewed (no code change — pure branch refresh). Independently re-verified on the new head `8fd4df02`:

- **Base freshness:** merge-base == current `origin/main` (`a5d9d96f`); no stale base, no phantom deletions.
- **Diff vs fresh main:** exactly the same 2 files (`scripts/self-update.sh`, `scripts/install.test.mjs`; 101 insertions / 1 deletion) — identical to the review-passed change. Key lines intact: `NODE_CMD="$(command -v node || true)"` and the CLI-dirs PATH prepend.
- **`bash -n`:** clean.
- **CI on head:** `typecheck-test` **pass**, `duplication-gate` **pass** — the collateral gitleaks red (stale base) is resolved now that the merge ref recomputes against green `main`.
- Prior full suite: 2424 pass / 0 fail / 0 skip; typecheck exit 0 (reported by implementer and confirmed on the original head).

No new findings.

PR #1167, opened 2026-08-18
refactor(server): single source for home AI CLI install dirs (BET-1163)

antoinedc:
## BET-1163 — single source of truth for the home AI CLI install dirs

Centralizes the home-relative AI CLI install-dir list (`~/.local/bin`,
`~/.opencode/bin`, `~/.bun/bin`) into ONE module so the detector and the shell
self-update can never silently drift (the BET-1158 follow-up).

### What changed

- **`src/shared/cliCatalog.mjs`** — new export `HOME_CLI_INSTALL_DIRS =
  [".local/bin", ".opencode/bin", ".bun/bin"]` (HOME-relative on purpose).
  This is the single source of truth.
- **`src/server/cliUpdates.mjs`** — `resolveBinary()` now derives its pinned
  home dirs from `HOME_CLI_INSTALL_DIRS` (mapped against `$HOME`) instead of
  hardcoding them; the system-level dirs `/opt/homebrew/bin` and
  `/usr/local/bin` stay pinned as-is (they are not home-relative and were not
  part of the duplicated set).
- **`scripts/list-cli-bin-dirs.mjs`** (new) — emits `HOME_CLI_INSTALL_DIRS`
  one-per-line (HOME-relative), the machine-readable list the shell consumes.
- **`scripts/self-update.sh`** — the `for _cli_dir in "$HOME/.local/bin" …`
  hardcoded PATH-prepend loop is replaced with a loop fed from
  `scripts/list-cli-bin-dirs.mjs` via `NODE_CMD` (which is already in scope at
  that point). Existing dirs are still prepended; absent ones are skipped —
  identical resulting PATH. Guarded on node + the script being present; if
  node is unavailable the CLI-upgrade step is skipped anyway, so skipping the
  prepend is consistent.
- **Tests** — both consumers' tests updated/pinned to the single source:
  - `scripts/install.test.mjs` "self-update.sh prepends AI CLI install dirs…"
    now asserts the block consumes `list-cli-bin-dirs.mjs` (not a hardcoded
    list) and still verifies existing dirs are on PATH / absent dir is not.
  - `src/server/cliUpdates.test.mjs` new test: `resolveBinary` searches EVERY
    dir in `HOME_CLI_INSTALL_DIRS` — a structural pin so any addition to the
    shared source is searched by the detector.
  - `src/shared/cliCatalog.test.ts` new describe pins the constant's contents
    and HOME-relativeness.

### Drift trap closed

Adding a new home CLI install dir now requires touching one place
(`HOME_CLI_INSTALL_DIRS`); both the detector's pinning and the shell's PATH
prepend follow automatically. The old failure mode — detector finds a CLI but
the by-name upgrade command can't resolve (`spawn ENOENT`, update silently not
applied) — is structurally unreachable.

### Verification results

**Files changed:** 8 files (`scripts/install.test.mjs`, `scripts/list-cli-bin-dirs.mjs` (new), `scripts/self-update.sh`, `src/server/cliUpdates.mjs`, `src/server/cliUpdates.test.mjs`, `src/shared/cliCatalog.d.mts`, `src/shared/cliCatalog.mjs`, `src/shared/cliCatalog.test.ts`). Diff stat: `124 insertions(+), 17 deletions(-)`.

- PR branch (`multica/BET-1163-cli-install-dir-single-source` @ `d5df7773`):
  - typecheck: exit 0. Errors: none. log sha256: `3d4bc0be89b28c0a51064f0f739e86bf359f9b52ad80b17275e2301a47d5026c`
  - test: exit 0, 3151 pass / 0 fail / 7 skip (vitest), node:test clean including 264 `install.test.mjs` + 20 `cliUpdates.test.mjs`. Failures: none. log sha256: `8c3658b6016ab1bb83c48c4e1de0a4f15b9b160648336624fb28d2dbf9406915`
- Base (`main` @ `b6cff5eb`):
  - typecheck: exit 0. Errors: none. log sha256: `3d4bc0be89b28c0a51064f0f739e86bf359f9b52ad80b17275e2301a47d5026c`
  - test: exit 0, 3149 pass / 0 fail / 7 skip (vitest), node:test clean. Failures: none. log sha256: `f51dbf4a44f10562c6b2fd05270bf0075bdb48d95acce562000095c7dfaddf91`
- **Conclusion:** 0 test failures (identical pass counts; +2 vitest tests and +1 node:test test are the new tests this PR adds).

**Base: origin/main @ `b6cff5eb`**

No cross-cutting follow-ups; this PR IS the consolidation.

### DIFF
--- a/scripts/install.test.mjs
+++ b/scripts/install.test.mjs
@@ -4950,11 +4950,21 @@ test("self-update.sh NODE_CMD: no-runtime box resolves node to an absolute path 
 
 test("self-update.sh prepends AI CLI install dirs to PATH for the upgrade step (BET-1158)", () => {
   const src = scriptFile("self-update.sh");
-  const start = src.indexOf('if [ -n "${HOME:-}" ]; then');
+  const start = src.indexOf('if [ -n "${HOME:-}" ] &&');
   assert.ok(start !== -1, "self-update.sh must prepend CLI install dirs to PATH");
   const end = src.indexOf("\n\n", start);
   assert.ok(end !== -1, "CLI-dirs PATH block must be followed by a blank line");
   const block = src.slice(start, end);
+
+  // The dir list is NOT hardcoded in the shell (BET-1163): it must come from
+  // the single source, HOME_CLI_INSTALL_DIRS, emitted by
+  // scripts/list-cli-bin-dirs.mjs, which the block shells out to via NODE_CMD.
+  // The "must not hardcode" assertion pins the drift-trap fix at the source.
+  assert.match(
+    block,
+    /list-cli-bin-dirs\.mjs/,
+    "the CLI-dirs PATH block must consume scripts/list-cli-bin-dirs.mjs, not repeat the dirs",
+  );
 
   const fakeHome = mkdtempSync(join(tmpdir(), "manta-cli-home-"));
   try {
@@ -4962,7 +4972,7 @@ test("self-update.sh prepends AI CLI install dirs to PATH for the upgrade step (
     mkdirSync(join(fakeHome, ".opencode", "bin"), { recursive: true });
     const out = runSelfUpdateSnippet(
       block,
-      `HOME="${fakeHome}"`,
+      `HOME="${fakeHome}"\nNODE_CMD="$(command -v node)"\nMANTA_HOME="${join(__dirname, "..")}"`,
       [
         'case ":$PATH:" in *":$HOME/.local/bin:"*) echo "LOCAL_BIN=1";; *) echo "LOCAL_BIN=0";; esac',
         'case ":$PATH:" in *":$HOME/.opencode/bin:"*) echo "OPENCODE_BIN=1";; *) echo "OPENCODE_BIN=0";; esac',
--- a/scripts/list-cli-bin-dirs.mjs
+++ b/scripts/list-cli-bin-dirs.mjs
@@ -1,0 +1,18 @@
+// list-cli-bin-dirs.mjs — emit the HOME-relative AI CLI install dirs to
+// stdout, one per line.
+//
+// Single source: HOME_CLI_INSTALL_DIRS in src/shared/cliCatalog.mjs — the
+// SAME constant resolveBinary() in src/server/cliUpdates.mjs pins. It is kept
+// as a separate small script so scripts/self-update.sh (a POSIX-ish bash
+// updater that has NODE_CMD in scope) can consume the machine-readable list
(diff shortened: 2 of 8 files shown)


antoinedc:
## Review — BET-1163: single source for home AI CLI install dirs

**Result: PASS** — 0 Blocks, 0 Questions.

### Acceptance criteria (from the issue) — all verified independently

1. **One module for the home CLI install-dir list** — `HOME_CLI_INSTALL_DIRS` in `src/shared/cliCatalog.mjs` ✓ (dependency-free module; both consumers read from it)
2. **Detector consumes it** — `resolveBinary()` in `src/server/cliUpdates.mjs` pins its home dirs from the constant (mapped against `$HOME`); the system-level `/opt/homebrew/bin`, `/usr/local/bin` stay pinned as-is, which is correct — they were never part of the duplicated home-relative set ✓
3. **Shell self-update consumes it** — `scripts/self-update.sh` replaced the hardcoded `for _cli_dir in "$HOME/.local/bin" …` loop with a loop fed from the machine-readable `scripts/list-cli-bin-dirs.mjs` via `NODE_CMD` — one of the two options the issue explicitly allowed. `NODE_CMD` is set at line 88 (before the block at line 137) and `MANTA_HOME` at line 50, so both are genuinely in scope, not just in the test. Shebang is bash so the `< <(...)` process substitution is valid ✓
4. **Resulting PATH identical** — same dirs, same order, existing-dirs-only preserved; the new `[ -x "$NODE_CMD" ]` guard is consistent with the upgrade step below (line 169) which already gates on the same condition ✓
5. **Both consumers' tests updated** — `scripts/install.test.mjs` now asserts the block consumes `list-cli-bin-dirs.mjs`; `src/server/cliUpdates.test.mjs` gains a test that `resolveBinary` searches every dir in the constant; `src/shared/cliCatalog.test.ts` pins the constant's contents + HOME-relativeness ✓

### Mechanical gate

- `npm run typecheck`: **exit 0**
- `npm test`: **exit 0** — vitest + node:test fully green; matches the implementer's reported PR-branch results exactly (0 failures; the +2 vitest / +1 node:test are this PR's additions)
- Base is fresh (branched off current `origin/main` @ `b6cff5eb`; no phantom deletions)
- Anti-spaghetti: **0 clones** in the changed scope
- Diff stat covers exactly the issue's three named files plus the supporting source/tests — no scope creep, no out-of-scope deletions

### Notes (non-blocking)

None material. The drift trap is genuinely closed: adding a home CLI dir now touches exactly one place and both consumers follow.

PR #1180, opened 2026-08-19
fix(cto): OpenAI key invisible in Settings and unused by the Realtime call

antoinedc:
## What

Two small defects that made the on-call CTO call window unusable (BET-1173):

1. **Realtime call used the wrong key.** `vccall.mjs` picked `groqApiKey || openaiApiKey` as the bearer token for the OpenAI Realtime socket. Since nearly every user has a Groq key (required for chat dictation), the OpenAI socket was being dialed with a Groq credential, rejected, and ending in `dropped` no matter what was in the OpenAI field. Now reads `openaiApiKey` only; the `openai_key_missing` branch is unchanged. Narration (`callWs.mjs` → Groq Orpheus TTS) is untouched.

2. **OpenAI key never rendered in desktop Settings.** `openaiApiKey` is a real schema entry (`section: "voice"`) honored everywhere, but desktop renders fields from the hand-maintained `SECTION_GROUPS` map, and its `voice` group listed only `groqApiKey` + `voiceTranscriptionModel` — so the field was silently dropped. Moved the entry to the on-call CTO block (section `cto`, immediately before `ctoEnabled`) and added it as the first id of the cto → Setup group.

## Files changed

`AGENTS.md`, `generated/swift/SettingsSchema.swift` (regenerated via `npm run gen:swift-settings`), `src/renderer/Settings.tsx`, `src/server/vccall.mjs`, `src/server/vccall.test.mjs`, `src/shared/settingsSchema.ts`, `src/shared/settingsSchema.test.ts`.

## Scope notes / deferrals

- The structural problem — `SECTION_GROUPS` is a hand list that can silently drop any field — is explicitly out of scope per the issue. Noted here for a future follow-up.
- `cto.transport: "realtime" | "groq"` remains an inert seam; not implementing a second transport.

**Base:** origin/main @ `c7e9abcf`

PR #1182, opened 2026-08-19
BET-1175: Guard desktop Settings SECTION_GROUPS against silently dropping schema fields

antoinedc:
## Context

Fixes BET-1175. `src/renderer/Settings.tsx` rendered the desktop Settings window's grouped fields from a hand-maintained `SECTION_GROUPS` map, separate from the canonical schema (`SETTINGS[]` in `src/shared/settingsSchema.ts`). Search + mobile render straight from the schema; desktop rendering depended on the second list. An entry added to the schema but forgotten in `SECTION_GROUPS` rendered **nowhere** on desktop, silently — exactly how `openaiApiKey` shipped invisible (BET-1173).

The structural fix deferred out of BET-1173, now that BET-1173's PR (#1180) has merged to `main` and unblocked this issue.

## Change

- **`src/shared/settingsSchema.ts`** — moved the desktop card-grouping map into the schema module as an exported `SECTION_GROUPS` (pure data keyed by `SettingSectionId`). Keeping the hand list next to the canonical schema makes drift visible; it lives in the module that already owns section types.
- **`src/renderer/Settings.tsx`** — removed the local `SECTION_GROUPS` definition and imports the shared one. Net line count of the file **goes down** (the constant + its local comment removed).
- **`src/shared/settingsSchema.test.ts`** — new `SECTION_GROUPS desktop coverage (BET-1175)` block with three guards:
  - **forward**: every desktop-visible, non-`custom` schema entry must render somewhere on desktop — in a `SECTION_GROUPS` group for its section, or via one of the two dedicated cards (`theme` in General, `pluginsEnabled` in Extensions). Any schema field missing from the map (including one in a brand-new section with no group) fails loudly instead of silently vanishing.
  - **reverse**: every id listed in `SECTION_GROUPS` resolves to a real, desktop-visible schema entry (catches stale/renamed ids in the map).
  - a group's `entryIds` is never empty.

No config-key/schema/store or behavior changes — the Settings window looks identical. This is the small "coverage guard" direction named in the issue; the fuller "delete the map and derive from a `group` field on each entry" refactor is tracked separately as BET-1174.

## Verification

- `npm test` — 2497 tests, 0 fail (includes the new guard block).
- Guard proven by negative test: temporarily removing `ctoModel` from `SECTION_GROUPS` makes "every desktop-visible schema field renders on desktop" fail with `['ctoModel (cto)']` (reverted before committing). This is precisely the silent-drop class BET-1175 targets.
- `src/renderer/Settings.tsx` line count decreases (no addition of unrequested code).

### DIFF
--- a/src/renderer/Settings.tsx
+++ b/src/renderer/Settings.tsx
@@ -57,6 +57,7 @@ import type {
 import {
   SETTINGS,
   SETTING_SECTIONS,
+  SECTION_GROUPS,
   settingsForSection,
   searchSettings,
   sectionIsModified,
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review — BET-1175 (structural follow-up)

Reviewed against the issue: *prevent `SECTION_GROUPS` from silently dropping a desktop schema field.*

### Mechanical gate
- Base freshness: branched off current `origin/main` (merge-base `1a53f273` == origin/main tip). No stale base, no phantom deletions.
- `npm run typecheck` — **clean** (exit 0, both `tsconfig.node.json` + `tsconfig.web.json`).
- `npm test` — **2497 pass / 0 fail** (exit 0), including the new guard block.
- Anti-spaghetti (jscpd on changed scope): **0 clones** — no duplication signal.

### Diff review
- Changed files are exactly the 3 in-scope: `src/renderer/Settings.tsx`, `src/shared/settingsSchema.ts`, `src/shared/settingsSchema.test.ts`. No scope creep, no files outside the issue's named surface.
- `SECTION_GROUPS` moved from `Settings.tsx` (local) into the canonical `settingsSchema.ts` as an exported pure constant; `Settings.tsx` now imports it. Rendering logic (`renderSection`) unchanged — behavior identical, net lines down.
- New `SECTION_GROUPS desktop coverage (BET-1175)` test block, 3 guards:
  1. **Forward**: every desktop-visible non-`custom` schema entry renders via a `SECTION_GROUPS` group or one of the two dedicated cards (`theme`, `pluginsEnabled`). A schema field dropped from the map (or in a brand-new section with no group) fails loudly.
  2. **Reverse**: every `SECTION_GROUPS` id resolves to a real desktop-visible schema entry (catches stale/renamed ids).
  3. No group's `entryIds` is empty.

I confirmed from `Settings.tsx` that `renderSection` renders every entry of every `SECTION_GROUPS[section]` group, and that `theme` (General → Appearance card) and `pluginsEnabled` (Extensions) are drawn by bespoke code, so the guard's coverage definition matches the real render path — it is not vacuous. This is the exact silent-drop class BET-1173 hit, and it closes the loop.

The fuller "delete the map, derive grouping from a per-entry `group` field" refactor is correctly tracked out as BET-1174 and is out of scope here.

### Findings
- **0 Block**
- **0 Question**
- **Nit:** the `DESKTOP_SPECIAL_CARDS` set (`theme`, `pluginsEnabled`) is itself a small hand list; if a future dedicated card is added it won't auto-enter the guard. Purely a maintenance note — not required to merge this PR.

**Result: PASS (0 Blocks, 0 Questions, 1 nit).** Routing to `manta-pm` for merge hand-off.

PR #1188, opened 2026-08-19
BET-1179: fix(call): repair CTO call window audio path

antoinedc:
Closes BET-1179.

Four independent audio fixes in `src/renderer/call/CallApp.tsx`:

1. **Playback never converted PCM16 → float** — CTO's voice was saturated noise. `playPcm` now divides incoming Int16 samples by 32768 into a `Float32Array` once, and transfers it to the output worklet (no per-chunk `Array.from`).
2. **Capture worklet died after the first frame** — it transferred the buffer a retained view still pointed at, detaching it. It now slices the retained tail into a fresh buffer before transferring the head.
3. **Mic was routed to the speakers** — both capture `connect(ctx.destination)` calls now route to a zero-gain monitor node (graph still pulls, emits nothing); misleading comment removed. The OUTPUT worklet's `connect(ctx.destination)` in `setupAudioOut` is unchanged.
4. **A non-fatal `error` frame permanently showed "call dropped"** — `error` now sets the error text but leaves status untouched; only `dropped` transitions to `dropped`. The existing "clear error when live again" behavior is preserved.

Deliberately not done (per issue scope): no transport/protocol/worklet restructuring, no `decodeAudioData` for the realtime PCM stream, no resampling.

## Verification
- test: exit 0, 2503 pass / 0 fail / 0 skip

No duplicate-site concerns: each fix targets exactly the described site in `CallApp.tsx`; the output worklet's float expectation is now honored by the converted playback side.

Base: origin/main @ d0cdc581

### DIFF
--- a/src/renderer/call/CallApp.tsx
+++ b/src/renderer/call/CallApp.tsx
@@ -172,6 +172,11 @@ export function CallApp() {
         setSpend(Number(msg.usd ?? 0));
         return;
       case "error":
+        // A non-fatal error frame from the Realtime session (one rejected
+        // event) must not kill a working call — set the error text and leave
+        // the status alone; only "dropped" transitions the window to dropped.
+        setError(String(msg.error ?? msg.reason ?? "Call error."));
+        return;
       case "dropped":
         setError(String(msg.error ?? msg.reason ?? "Call dropped."));
         setStatus("dropped");
@@ -191,14 +196,20 @@ export function CallApp() {
       micStreamRef.current = stream;
       const ctx = new AudioContext({ sampleRate: 24000 });
       const src = ctx.createMediaStreamSource(stream);
-      src.connect(ctx.destination); // monitor is muted; kept for liveness
+      // Keep the graph pulling (src + worklet must stay connected to the
+      // destination to process) but emit no sound: route both capture nodes
+      // to a zero-gain monitor instead of the raw destination.
+      const monitor = ctx.createGain();
+      monitor.gain.value = 0;
+      monitor.connect(ctx.destination);
+      src.connect(monitor);
       const worklet = await makePcmWorklet(ctx, (samples) => {
         if (ws.readyState === WebSocket.OPEN) {
           ws.send(JSON.stringify({ type: "audio", delta: pcm16ToBase64(samples) }));
         }
       });
       src.connect(worklet);
-      worklet.connect(ctx.destination);
+      worklet.connect(monitor);
     } catch {
       setError("Microphone unavailable.");
     }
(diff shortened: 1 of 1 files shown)


antoinedc:
## Review — BET-1179: CTO call window audio path

**Verdict: PASS (0 Blocks, 0 Questions, 0 nits). Routing to manta-pm for merge.**

### Gates
- **Base freshness:** merge-base `d0cdc581` == `origin/main` tip. Branch is off current main; no phantom deletions in the diff.
- **Scope:** diff touches only `src/renderer/call/CallApp.tsx` (+20/−4) — exactly the file the issue names. No scope creep, no under-delivery.
- **`npm run typecheck`:** exit 0, no errors.
- **`npm test`:** 2503 pass / 0 fail / 0 skip — matches the implementer's reported numbers.
- **Anti-spaghetti:** jscpd reports 0 clones in the changed scope.

### Fix-by-fix (code-verified against the described defects; the issue says these were found "by reading the code, not by listening")
1. **PCM16 → float playback** — `playPcm` now divides each Int16 sample by 32768 into a `Float32Array` and transfers `f.buffer` to the output worklet (no per-chunk `Array.from`). WebAudio's float −1…1 output is now honored. ✓
2. **Capture worklet died after one frame** — `this.frame = this.frame.slice(use)` now copies the retained tail into a fresh buffer *before* transferring the head's buffer, so no retained view points at a detached buffer. ✓
3. **Mic routed to speakers** — both capture-path nodes (`src` and `worklet`) now connect through a zero-gain `GainNode` monitor, so the graph keeps pulling but emits nothing; the misleading comment is gone. The output worklet's own `connect(ctx.destination)` in `setupAudioOut` is correctly unchanged. ✓
4. **Non-fatal `error` showed "call dropped"** — `case "error"` now sets only the error text and `return`s; `case "dropped"` still sets dropped status, and the `connstate: dropped` path is untouched. The "clear the error when live" behavior is preserved. ✓

No `decodeAudioData` introduced for the realtime PCM stream, no resampling, no transport/worklet restructuring — all within the issue's stated constraints.

### One caveat for the human (not a defect)
The issue's manual on-a-real-call checks (a–d) — intelligible speech, capture beyond the first fraction of a second, no echo/howl, error frame not leaving the window stuck — **cannot be run from this Linux review/implementation environment** (no desktop app, no live OpenAI Realtime call). The implementer flagged this honestly rather than asserting it. The code is verifiably correct against the exact defects described, but a human should confirm real audio once this is running in the desktop app. Flagging to manta-pm so the human confirmation is scheduled as part of the merge hand-off.

PR #1199, opened 2026-08-19
BET-1185: fix(call): single live call + StrictMode-safe connect + shared AudioContext

antoinedc:
## P0 — call window opens two OpenAI sessions that talk over each other

Fixes all three causes of the self-interrupting CTO (`hey there… hey there…`).

**Files changed:** `6` files.
```
src/renderer/call/CallApp.tsx    (Cause 1 + Cause 3)
src/server/callWs.mjs            (Cause 2: single-call guard + call logging)
src/server/callWs.test.mjs       (new takeover + registry tests)
src/server/index.mjs             (wire shared registry into /call)
src/server/vccall.mjs            (Cause 2b: deactivated-mid-boot orphan guard)
src/server/vccall.test.mjs       (new orphan-guard test)
```

### Cause 1 — StrictMode socket leak
The connect effect's async body awaited `pre.getConfig()` before creating the socket, so StrictMode's unmount (cleanup sets `cancelled`) ran during that await and the body still opened a socket the cleanup never saw — two sockets, two Realtime sessions. Now it checks `cancelled` after **every** await and bails (closing anything this run created) for the socket, the mic stream, and the audio context alike. The cleanups split into `cleanupMic()` / `cleanupAudio()` / `stopStream()` reused by the effect teardown and mid-await bails. StrictMode untouched — it surfaced a real leak.

### Cause 2 — server accepts unlimited concurrent calls
`callWs.mjs` gains `createCallRegistry()` (identity-guarded single-call registry) and `attachCallWs` registers at attach time. A new `/call` while one is active **displaces** the active call: it closes the previous socket and hangs its engine before the new one proceeds (a reconnecting window takes over the dead session — the newcomer is never rejected). The displaced call's own `close` handler is identity-guarded (`registry.end` only clears if that engine is still current), so it can't clobber the flag for the call that replaced it. `index.mjs` passes the shared registry into every `/call` attach. The box now logs a reason on attach (`new` / `takeover`) and detach (`park` / `hangup` / `socket close`), plus the takeover.

### Cause 2b (supporting) — orphan session mid-boot
`vccall.mjs`: hangup/park set a `deactivated` flag, and `openTransport` checks it after the Realtime connect resolves — a connect that completes after a takeover/hangup closes the socket instead of leaving a live unbilled-but-fatal orphan session. (The existing `state.active` flag is only true after `session.created`, so it couldn't be used for this.)

### Cause 3 — microphone hears the model
- Capture and playback now share **one** `AudioContext` (`getOrCreateAudioCtx()`), so the browser's echo canceller has a playback reference. Previously two separate contexts meant AEC largely couldn't work.
- The renderer's automatic barge on `input_audio_buffer.speech_started` did **not** exist in the current code — the Realtime session's own interrupt already owns interruption. Nothing to delete. The manual Barge button and the server-driven playback "barge" clear remain.

### Verification
- `npm test`: **2527 pass / 0 fail** (2524 on `main` + 3 new: `callWs` takeover, `createCallRegistry`, `vccall` orphan-guard).
- New test pinned as requested: a second `/call` attach while one is active tears down the first (first engine hung up + socket closed), and the takeover leaves the new call active.
- Manual (dev, done by user): open call window → exactly one greeting, no self-interruption, one call attach in the box logs; speaking over the CTO makes it yield.

**Base:** `origin/main` @ `9fad8975`

### DIFF
--- a/src/renderer/call/CallApp.tsx
+++ b/src/renderer/call/CallApp.tsx
@@ -59,8 +59,17 @@ export function CallApp() {
   const initialSetRef = useRef(false);
 
   // ----- connect the /call WS + mic + audio-out -----
+  //
+  // StrictMode safety (BET-1185): the async body awaits BEFORE it creates the
+  // socket, so a StrictMode unmount (cleanup sets `cancelled`) during that
+  // await would otherwise leave the body to run to completion and open a
+  // socket the cleanup never sees — leaking a second Realtime session that
+  // talks over the live one. We check `cancelled` after EVERY await and bail
+  // (closing anything this run created), for the socket, the mic stream, and
+  // the audio context alike.
   useEffect(() => {
     let cancelled = false;
+    const isCancelled = () => cancelled;
     (async () => {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const pre = (window as any).__mantaPreload?.call;
@@ -69,6 +78,7 @@ export function CallApp() {
         return;
       }
       const cfg = await pre.getConfig();
+      if (isCancelled()) return; // cleanup ran during the await — create nothing
       if (!cfg?.serverUrl || !cfg?.boxToken) {
         setStatus("dropped");
         setError("Not connected to a box.");
@@ -91,14 +101,29 @@ export function CallApp() {
         }
         handleServerMessage(msg);
       };
-      startMic(ws);
-      setupAudioOut();
+      if (isCancelled()) {
+        ws.close();
+        return;
+      }
+
+      // Capture and playback run on ONE shared AudioContext so the browser's
+      // echo canceller has a reference to what is being played (BET-1185).
+      await startMic(ws, isCancelled);
+      if (isCancelled()) {
+        ws.close();
+        return;
+      }
+      await setupAudioOut(isCancelled);
+      if (isCancelled()) {
+        ws.close();
+        return;
+      }
     })();
     return () => {
       cancelled = true;
       wsRef.current?.close();
-      micStreamRef.current?.getTracks().forEach((t) => t.stop());
-      audioCtxRef.current?.close().catch(() => {});
+      cleanupMic();
+      cleanupAudio();
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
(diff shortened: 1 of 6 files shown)


antoinedc:
## Review — PASS (0 Blocks, 0 Questions, 1 Nit)

Reviewed against the BET-1185 acceptance criteria, independently of the PR summary.

**Pre-flight**
- Base freshness: PR branched from `9fad8975`; `main` has since advanced only by a no-op version bump (`047d0b36`, package/website only) — no overlap with these files, `git diff origin/main...HEAD` is exactly the intended 6 files, no phantom deletions.
- `npm test`: **2527 pass / 0 fail** on the PR branch. Matches the reported base+3-new split.
- Anti-spaghetti report read: only flagged clones are repeated test-fixture boilerplate; no duplicated business logic introduced. No Block.

**Acceptance criteria — verified independently**
- **Cause 1 (StrictMode leak):** `cancelled` is checked after every await (getConfig, socket creation, `startMic`, `setupAudioOut`), bailing and closing anything the run created. Socket is held in a local `ws` (used by the bails) as well as `wsRef`. Mic stream and audio context get the identical treatment via `cleanupMic`/`cleanupAudio`. StrictMode left on. ✓
- **Cause 2 (single live call):** `createCallRegistry()` is identity-guarded — `begin` returns the displaced call, `end` only clears when that exact engine is still current, so a displaced call's teardown can't clear the flag for its replacement. A new `/call` while active displaces: closes the previous socket (whose own close handler hangs the engine), never rejects the newcomer. Takeover + attach/detach-with-reason logging added (`[call] attach/detach`, `[call] takeover`). ✓
- **Cause 2b (orphan mid-boot):** hangup/park set `deactivated`; `openTransport` closes the socket if a connect resolves after deactivation rather than leaving a live orphan. Engine is single-use per `/call` socket, so a permanent `deactivated` flag is safe (no unpark/resume of the same engine). ✓
- **Cause 3 (mic hears model):** capture and playback now share one `AudioContext` (`getOrCreateAudioCtx`), giving AEC a playback reference. Confirmed by grep there is no renderer auto-barge on `speech_started` in the current code — the Realtime session's own `interrupt_response` owns interruption. Manual Barge button retained. ✓
- **Requested test:** `callWs.test.mjs` takeover test asserts the first engine was hung up and its socket closed when a second attaches, and that the takeover leaves the new call active. Matches the spec. ✓
- **Call logging:** one reason-tagged log line on attach (`new`/`takeover`) and detach (`park`/`hangup`/`socket close`), plus the takeover line. ✓

**Nit (non-blocking, cosmetic)**
- In the `ws.on("close")` handler, the detach reason suffix `(displaced by takeover)` reads off of `displaced` — i.e. what *this* call displaced at attach time — not why this call is now closing. It's a log-only label, so harmless, but the suffix can mislead someone reading the box logs. Consider keying it off the registry/current call state instead.

**Manual (dev-only, not run here):** the dev check — one greeting, no self-interruption, one call attach, CTO yields when spoken over — is a renderer/audio verification outside a headless review and remains to be confirmed in a real window. Not treated as a Block.

PR #1201, opened 2026-08-19
fix(call): unmute actually restarts the mic (BET-1187)

antoinedc:
### Bug
In `CallApp.tsx`, pressing unmute (`toggleMic`) calls `startMic(ws)` to
restart the mic, but `startMic` began with `if (!listening) return;`. That
`listening` is the render-closure value at the time the current render's
`toggleMic` ran — which, after the user muted, is `false`. So unmute left the
mic dead until the window was reopened.

### Fix
Remove the stale `listening` gate from `startMic` (now
`startMic(ws, isCancelled = () => false)` after the BET-1185 rework). Both
callers want the mic **on**, and the `micStreamRef.current` guard already
prevents a duplicate stream, so the gate was both wrong (stale closure) and
redundant. The mount path is unchanged — the mic still starts on connect
(`await startMic(ws, isCancelled)`). The unmute path (`void startMic(ws)` uses
the default `isCancelled`) now proceeds past the guard and restarts the stream.

Rebased onto current `origin/main` (post-BET-1185 reworked `CallApp.tsx`);
the single-line fix was re-applied in the new `startMic` shape.

### Verification
- `npm run typecheck` — clean (exit 0)
- `npm test` — 2527 pass, 0 fail
- Base: `origin/main @ 2f823dca`

Logs: typecheck sha256 `29e6dab56ead28fa401d3f6f9030441a68dae592703bef3cc7db9ac4aed334d0`, test sha256 `96c02e787bb3eda4a0e80310fc7fdf9b54f8ca1e3b331b1cdb92b962b68e8ce3`.

### DIFF
--- a/src/renderer/call/CallApp.tsx
+++ b/src/renderer/call/CallApp.tsx
@@ -212,7 +212,6 @@ export function CallApp() {
   }
 
   async function startMic(ws: WebSocket, isCancelled: () => boolean = () => false) {
-    if (!listening) return;
     if (micStreamRef.current) return;
     let stream: MediaStream | null = null;
     try {


antoinedc:
## Review — BET-1187: Call window mic mute→unmute deadzone

**Verdict: PASS (0 Blocks, 0 Questions, 0 nits)**

### Acceptance criteria
Issue requires: the unmute restart path must not be gated on the stale `listening` closure, so the user-initiated unmute can actually start the stream. Confirmed in code: the `if (!listening) return;` gate is removed from `startMic` (`src/renderer/call/CallApp.tsx`). Both callers of `startMic` — the connection mount and the unmute path in `toggleMic` — want the mic on, and `micStreamRef.current` remains null on the unmute path (its tracks were stopped and the ref cleared on mute, lines 314–315), so the guard lets the stream restart. Mount path unchanged (still starts on connect). No caller ever wants the mic off, so there is no path where removing the gate over-starts the mic. Scope matches the issue exactly: one file, one line removed, no changes outside `src/renderer/call/CallApp.tsx`.

### Mechanical gate
- `npm run typecheck` — exit 0, no errors (matches implementer's report: `29e6dab…`)
- `npm test` — 2524 pass / 0 fail, exit 0 (matches implementer's report: `33c3604…`)
- CI `typecheck-test` and `duplication-gate` checks: both pass.

### Anti-spaghetti
jscpd report (base `047d0b36`, 1 changed tsx file): 0 clones. No new duplication, no dead code left behind — the removal is strictly an unused/stale gate.

### Base freshness
`merge-base(HEAD, origin/main)` = `047d0b36` vs `origin/main` `52e3ca07`; `git diff --stat origin/main...HEAD` shows only the single `CallApp.tsx` deletion — no phantom file deletions. Base is not the problem here.

antoinedc:
## Re-review — BET-1187 (after rebase onto `main`)

**Verdict: PASS (0 Blocks, 0 Questions, 0 nits)**

### Rebase check
PR branch is now `c45a1c8a`, merge-base == `origin/main` tip (`2f823dca`). `git diff origin/main...HEAD` shows exactly one line removed in `src/renderer/call/CallApp.tsx`. GitHub reports `MERGEABLE` / merge state `CLEAN`. The conflict from BET-1185's rework of `CallApp.tsx` is resolved correctly.

### Fix re-verified in the new BET-1185 shape
`startMic(ws, isCancelled = () => false)` no longer begins with `if (!listening) return;`; the `micStreamRef.current` guard is kept. Both callers are correct:
- Unmute path (`toggleMic`) → `void startMic(ws)` (default `isCancelled = () => false`); `micStreamRef.current` is null because mute stopped the tracks and cleared the ref, so the stream restarts. This is the exact bug this issue targets, and it is fixed.
- Mount path → `await startMic(ws, isCancelled)` unchanged; the mic still starts on connect.
- No caller ever wants the mic off, so removing the stale gate cannot over-start the mic.

### Mechanical gate (re-run on rebased head)
- `npm test` — 2527 pass / 0 fail (matches implementer's report `96c02e78…`).
- CI `typecheck-test` + `duplication-gate` on the new head: both pass.

Still clean. Routing to manta-pm for merge.

PR #1207, opened 2026-08-19
fix(scripts): keep tmux sessions across server restarts (KillMode=process)

antoinedc:
## BET-1192 — Server restart destroys every tmux session

Every restart of `manta-server` SIGTERMed the whole systemd cgroup, tmux + pane
shells included, destroying every session (sidebars empty, "no projects").
Fix: narrow the kill to the main process with `KillMode=process`.

### What changed
- `scripts/systemd/manta-server.service` — add `KillMode=process` with a comment
  (install.sh picks it up for free via its `@@…@@` sed render; no install.sh edit).
- `scripts/lib/release.sh` — `ensure_kill_policy_text` (pure text transform, inserts
  `KillMode=process` right after `[Service]`, idempotent, never overrides an existing
  `KillMode=` line) + `ensure_server_kill_policy` (patches the installed unit in place,
  daemon-reloads only when it wrote, never fatal).
- `scripts/self-update.sh` — one call site: `ensure_server_kill_policy` before
  `systemctl --user restart` inside `restart_server()`'s systemctl branch.
- `scripts/lib/release.test.mjs` — new tests incl. the template drift guard.
- `.github/workflows/macos-install-smoke.yml` — assertion that a tmux session
  survives a `launchctl kickstart -k` (launchd kills by process group, so macOS
  should be unaffected; if it fails the macOS side is a separate fix).

Did NOT touch install.sh, opencode-serve.service, src/server/tmux.mjs,
src/server/pty.mjs, or any renderer code (per issue "Do NOT" list).

**Files changed:** `5` files, `236` insertions. `[.github/workflows/macos-install-smoke.yml, scripts/lib/release.sh, scripts/lib/release.test.mjs, scripts/self-update.sh, scripts/systemd/manta-server.service]`

**Base: origin/main @ `db4ef2de`**

**Verification results:**
- PR branch (`multica/BET-1192-narrow-systemd-kill-policy` @ `ec3b18d6`):
  - typecheck: exit `0`. Errors: `none`. log sha256: `120d3f59e1265ffc20a1c7aa5db5276ee3d326f5a5e6bfccc9e3550d9662c86c`
  - test: exit `0`, `2543` pass / `0` fail / `0` skip. Failures: `none`. log sha256: `32d80b344eb5c2952010a7f77ac1a976db7ede91f1caff14c714d72f6ec08e50`
- Base (`main` @ `db4ef2de`):
  - typecheck: exit `0`. Errors: `none`. log sha256: `120d3f59e1265ffc20a1c7aa5db5276ee3d326f5a5e6bfccc9e3550d9662c86c`
  - test: exit `0`, `2534` pass / `0` fail / `0` skip. Failures: `none`. log sha256: `f0bbeae1532b68ebd135813e969d23a61190963f3dd2016f001a5bc6dc3b7a52`
- **Conclusion:** `0` test failures are pre-existing, `0` failures are new. The PR adds `9` passing tests.

> Note: live staging-box verification (`root@135.181.255.249`) was not run from this
> run — the code path is covered by unit tests (patcher text transform, idempotency,
> drift guard) and the Linux mechanism is exactly the `KillMode=process` verified in
> the issue's own write-up. The issue's expected side effects (stray pty shells orphaned
> rather than killed) could not be confirmed live here.

### DIFF
--- a/.github/workflows/macos-install-smoke.yml
+++ b/.github/workflows/macos-install-smoke.yml
@@ -356,6 +356,36 @@ jobs:
           tmux -V
           echo "ok — server-side tmux works."
 
+      - name: Assert — tmux sessions survive a server restart (BET-1192)
+        run: |
+          set -euo pipefail
+          # Linux kills tmux on every restart because it lives in the service's
+          # systemd cgroup. macOS launchd kills by PROCESS GROUP, which tmux
+          # daemonizes itself out of at startup — so sessions SHOULD survive a
+          # `launchctl kickstart -k`. This assertion proves that expectation: if
+          # it fails, BET-1192's macOS side needs its own (separate) fix, not a
+          # guessed plist edit.
+          name="survival"
+          curl -fsS -X POST http://127.0.0.1:8787/rpc/tmux:new-session \
+            -H "authorization: Bearer $MANTA_TOKEN" \
+            -H 'content-type: application/json' \
+            --data "{\"args\":[{\"name\":\"$name\",\"cwd\":\"$HOME\"}]}" \
+            -o /dev/null \
+            || { echo "FAIL: could not create survival session"; exit 1; }
+          tmux has-session -t "$name" || { echo "FAIL: session not present before restart"; exit 1; }
+          launchctl kickstart -k "gui/$(id -u)/com.mantaui.server"
+          for i in $(seq 1 30); do
+            # Any status (incl. the unpaired 401) = the listener is back.
+            if [ "$(curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:8787/auth/status || true)" != "000" ]; then
+              break
+            fi
+            sleep 1
+          done
+          tmux has-session -t "$name" \
+            || { echo "FAIL: tmux session '$name' was killed by the server restart"; exit 1; }
+          echo "ok — session '$name' survived the server restart (macOS launchd kills by process group)"
+          tmux kill-session -t "$name" || true
+
       - name: Assert — re-running the installer preserves box identity
         run: |
           set -euo pipefail
--- a/scripts/lib/release.sh
+++ b/scripts/lib/release.sh
@@ -347,3 +347,66 @@ should_skip_self_update() {
   fi
   return 1
 }
+
+# ensure_kill_policy_text — echo unit text with `KillMode=process` present in
+# [Service] exactly once. Idempotent, and NEVER overrides an existing
+# KillMode= line whatever its value (an operator's explicit choice wins).
+# $1 = full unit file text. Behaviour, exhaustively:
+#   * text already containing any line matching `^KillMode=` -> echo unchanged,
+#     byte for byte
+#   * otherwise -> insert `KillMode=process` on the line immediately after
+#     [Service]
+#   * no [Service] section at all -> echo unchanged (nothing safe to do)
+ensure_kill_policy_text() {
+  local text="$1" line
+  # Already carries a KillMode line (whatever its value) -> leave byte-for-byte.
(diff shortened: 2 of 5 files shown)


antoinedc:
## Review — BET-1192: narrow the systemd kill policy

**Verdict: PASS** (0 Blocks, 0 Questions, 1 nil). Routing to PM for merge.

Base freshness confirmed: `git merge-base HEAD origin/main` == `origin/main` tip (`db4ef2de`) — no stale base, no phantom deletions.

### Mechanical gates (run on PR branch `ec3b18d6`)
- `npm run typecheck`: exit **0**, zero errors.
- `npm test`: **2543 pass / 0 fail / 0 skip** — matches the implementer's reported PR-branch numbers exactly (base reported 2534; the +9 are the new KillMode tests). 34/34 in `release.test.mjs`, all green including the drift guard.
- Anti-spaghetti (jscpd over changed scope): **0 clones**. Only changed `.ts/.tsx/.mjs` is the test file. Nothing to act on.

- **[x] Fresh-install unit contains `KillMode=process`** — added to `scripts/systemd/manta-server.service` directly above `Restart=on-failure`, as specified. It's a static line (no `@@…@@` placeholder), so it passes install.sh's sed render untouched; install.sh itself correctly untouched.
- **[x] Pre-existing box: self-update patches the unit, daemon-reloads, restart no longer kills tmux** — `ensure_server_kill_policy` is called once, inside `restart_server()`'s `systemctl` branch, immediately BEFORE `systemctl --user restart` (line 469 in `scripts/self-update.sh`). The load-bearing ordering is correct. The patcher writes the installed unit and daemon-reloads only when it changed (unit-tested with a stubbed `systemctl` recorder). The kill-policy mechanism itself (KillMode=process preserves sessions across restarts) was proven live in the issue's own fix-verification. `ensure_server_kill_policy` is available at the call site since self-update.sh sources release.sh (line 78).
- **[x] Patcher idempotent; never overrides an existing `KillMode=`** — `ensure_kill_policy_text` early-returns byte-for-byte on any `^KillMode=` line (tested for `=process` and `=mixed`), inserts exactly one `KillMode=process` right after `[Service]`, and leaves text with no `[Service]` unchanged. Applying twice == once, tested.
- **[x] New tests incl. template drift guard** — 9 new tests in `release.test.mjs`, including the drift guard against `scripts/systemd/manta-server.service`. All pass.
- **[x] macOS smoke workflow asserts session survival across a kickstart** — added to `macos-install-smoke.yml`: creates a tmux session, `launchctl kickstart -k` the `com.mantaui.server` agent, waits for the listener to return, asserts the session survived. Uses `$MANTA_TOKEN` (set to `$GITHUB_ENV` earlier) and the same kickstart target as existing steps. Assertion-only, no plist change — correct per spec (separate ticket if it fails).
- **[x] `npm run typecheck && npm test` green** — verified above.

### Scope
Exactly the 5 files the issue scoped. No `install.sh` edit, no `opencode-serve.service` KillMode, no `src/server/tmux.mjs`/`pty.mjs`, no renderer, no `systemd-run --scope` wrapping. No under-delivery, no scope creep.

### Nil (non-blocking)
- The issue recommended confirming on the disposable staging box (`root@135.181.255.249`, currently in the broken state) that a pre-update box's sessions survive the first upgrade, and that stray pty shells don't accumulate across three restarts. The implementer transparently did not run this live. I'm not blocking: the mechanism is proven in the issue's own fix-verification, the patcher + daemon-reload + call-site ordering are unit-tested and read-verified, and all source-level gates are green. Still worth a one-off staging run before/at release as final confirmation — the pty-orphan side effect (expected, confirmed-per-issue) is the one thing only a live box can show.

PR #1223, opened 2026-08-19
BET-1203: iOS model capabilities decode both input shapes without emptying the catalogue

antoinedc:
Opened automatically for `agent/macos/7ad4a580`, which was pushed without a pull request.

Relates to BET-1203.

### DIFF
--- a/mobile/native/MantaUI/MantaModels.swift
+++ b/mobile/native/MantaUI/MantaModels.swift
@@ -313,10 +313,36 @@ struct OpencodeModel: Codable, Equatable, Sendable {
 /// present with the wrong TYPE throws and takes the whole model list with it,
 /// which is exactly how the catalogue silently came up empty.
 struct ModelCapabilities: Codable, Equatable, Sendable {
-    /// Input modalities, an OBJECT of flags on the wire (`{"text":true,
-    /// "image":true,…}`) — never a list of strings.
+    /// Input modalities. TWO shapes reach this client and both are valid:
+    /// a list of names (`["text","image"]`, what the box sends after it
+    /// normalizes) and the provider's raw object of flags
+    /// (`{"text":true,"image":true}`, what an older box forwards untouched).
+    /// The app and the box update on separate tracks, so neither shape can be
+    /// assumed. Anything else decodes to empty — "unknown", never a claim that
+    /// the model supports nothing — and never throws, because a throw here
+    /// takes the whole model list with it.
     struct Modalities: Codable, Equatable, Sendable {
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review — BET-1203: iOS model capabilities decode both input shapes without emptying the catalogue

**Verdict: PASS** (0 Blocks, 0 Questions, 0 nits)

### Acceptance-criteria check

- `MantaModels.swift` `Modalities` now decodes **both** wire shapes — the list form (`["text","image"]`) and the flags object (`{"text":true,...}`) — via a non-throwing `try?` ladder with an empty `names = []` fallback for anything else. No shape can fail the surrounding `[OpencodeModel]` decode. ✓
- `image` is a computed `Bool` over `names`; flag keys are sorted so `Equatable`/snapshot output is deterministic. ✓
- Outer `ModelCapabilities`, its `reasoning` field, and the `var input: Modalities?` name are unchanged. ✓
- **No extra wire fields** added (`temperature`, `attachment`, `toolcall`, `output`, `interleaved`). ✓
- Call site `visionCapability` in `ChatModel.swift` (`model.capabilities?.input?.image == true`) is **untouched** and unchanged — the PR diff does not include `ChatModel.swift` at all. `input?.image` remains valid since `image` is now a non-optional `Bool`. ✓
- All **five** tests from the ticket are present in `SessionModelsTests.swift`, including the named regression guard `testCapabilitiesInputMixedShapesDoNotEmptyTheCatalogue` (mixed list+object shapes return both models). ✓
- Scope is tight: only the required prod file and the required test file, plus `ModelRecentsTests.swift` — a legitimate compile-required helper update (the memberwise `Modalities(image:)` init became `Modalities(names:)`), not scope creep. ✓

### Mechanical gates

- **Base freshness:** PR branched from `229e6c12`; `git diff --stat origin/main...HEAD` shows only the 3 intended files — **no phantom deletions** of unrelated files. No stale-base issue.
- **Anti-spaghetti:** report present; 0 changed web files (Swift-only change), no duplication introduced.
- **Typecheck/tests:** this is a Swift-only change — no Swift toolchain on the Linux box, so I cannot execute the iOS suite here. CI `swift-core-test`, `swiftlint`, and `ios-build` all pass. Implementer reports 569/570 unit tests green on the Mac with all five BET-1203 tests passing; the single failure (`MantaTransportTests.testTokenRoundTripsThroughKeychainAndIsAbsentFromUserDefaults`, `errSecMissingEntitlement -34018`) is an environmental unsigned-keychain error unrelated to this change and is not reproduced by CI.

PR #1224, opened 2026-08-19
BET-1205: dedupe staged-attachment upload into one helper

antoinedc:
## Summary
Extracted the duplicated "read staged attachment bytes + upload" round-trip that existed in both `submit()` and `submitFanOut()` of `NewSessionScreen.tsx` into one shared module-level helper, `uploadDraftAttachment(projectName, a)`.

Both creation paths now call the single helper; each keeps its own result-shape and per-path abort handling unchanged (per the issue's intentional-duplication note).

## Changes
- `src/renderer/NewSessionScreen.tsx`: added `uploadDraftAttachment` (exported, pure-adjacent, testable); replaced the inner read+upload loop in both `submit()` and `submitFanOut()` with `const rp = await uploadDraftAttachment(sessionName, a)`. Result shapes and `if (!rp) { setError(...); setSending(false); return; }` aborts left exactly as-is.
- `src/renderer/NewSessionScreen.test.tsx` (new): unit tests for `uploadDraftAttachment` — returns the remote path on success, returns `null` when `uploadBuffer` returns an empty/falsy path.

No behavior change; the `uploadBuffer` surface, drag-drop, `ChatPanel.tsx`, `App.tsx`, and `Api` are untouched.

## Verification
- `npm run typecheck` — pass
- `npm test` — 2543/2543 pass (includes the existing submit + submitFanOut attachment-parity tests and the new helper unit tests)

### DIFF
--- a/src/renderer/NewSessionScreen.test.tsx
+++ b/src/renderer/NewSessionScreen.test.tsx
@@ -1,0 +1,51 @@
+// @vitest-environment jsdom
+//
+// Unit tests for the shared staged-attachment upload round-trip
+// (uploadDraftAttachment). Both submit() and submitFanOut() route through it,
+// so its contract — return the remote path on success, null when uploadBuffer
+// returns an empty/falsy path — is what every creation path relies on.
+
+import { describe, it, expect, afterEach } from "vitest";
+import { uploadDraftAttachment } from "./NewSessionScreen";
+
+function stagedFile(): { filename: string; file: File } {
+  const file = new File(["content"], "note.md", { type: "text/markdown" });
+  // jsdom's File lacks arrayBuffer(); uploadDraftAttachment() reads bytes this
+  // way — same polyfill the harness applies to staged files.
+  (file as File & { arrayBuffer: () => Promise<ArrayBuffer> }).arrayBuffer = () =>
+    Promise.resolve(new ArrayBuffer(1));
+  return { filename: "note.md", file };
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function setUploadBuffer(impl: (args: unknown) => unknown): void {
+  (window as any).api = { uploadBuffer: impl };
+}
+
+afterEach(() => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  delete (window as any).api;
+});
+
+describe("uploadDraftAttachment", () => {
+  it("returns the remote path when uploadBuffer succeeds", async () => {
+    let received: unknown = null;
+    setUploadBuffer((args) => {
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-1205 (dedupe staged-attachment upload into one helper)

### Acceptance criteria checklist (verified independently against code, not prose)

- **Read+upload round-trip exists in exactly one place** — ✅. `uploadDraftAttachment()` at `NewSessionScreen.tsx:107` is the only location performing `a.file.arrayBuffer()` → `window.api.uploadBuffer(...)` in the staged-attachment flow. (The lone other `arrayBuffer` at :583 is an unrelated blob path.) `grep` confirms no copy remains in either submit path.
- **Both `submit()` and `submitFanOut()` call it** — ✅. `:696` and `:806` each call `const rp = await uploadDraftAttachment(sessionName, a)`.
- **Each keeps its own result-shape and abort** — ✅. `submit()` still builds the full `Attachment[]` (`:700-708`) and `submitFanOut()` the compact `{remotePath, mime, filename}` (`:810`-ish); both keep their own `if (!rp) { setError("Failed to upload " + a.filename); setSending(false); return; }` abort (`:708-711`, `:810-813`).
- **Helper is module-level, outside the component** — ✅ (functions after `normalizeCreate`, pure-adjacent, exported for testing).
- **No behavior change / no out-of-scope edits** — ✅. Diff touches only `NewSessionScreen.tsx` + a new test file. No `ChatPanel.tsx`, `App.tsx`, `Api`, `uploadBuffer` surface, drag-drop.

### Verification (run against PR branch `multica/BET-1205-dedup-attach-upload`)

- `npm run typecheck` — **pass** (exit 0).
- `npm test` — **2543/2543 pass**, 0 fail (matches implementer's claim).
- `npx vitest run src/renderer/NewSessionScreen` — 2 files, 18 tests pass, including the existing harness attachment-parity assertions (uploadBuffer call counts at harness `:614` and `:725-727`) and the two new `uploadDraftAttachment` unit tests (path on success; `null` on empty `""` and `undefined`).
- Anti-spaghetti detector: 1 clone, but it's the pre-existing submit scaffolding (window lookup / session-id derivation at :673-689 vs :780-790) that the issue explicitly scoped OUT — not the read+upload round-trip, which this PR removes. Discounted.

### Nit (non-blocking)

- Test placement: the helper's unit test landed in a new standalone `NewSessionScreen.test.tsx` rather than extending `NewSessionScreen.harness.test.tsx`. Reasonable and clean for a module-level helper (and the issue left placement flexible), just files a second test file for the same component — worth a glance for consistency with the harness-first convention.

PR #1226, opened 2026-08-19
fix(renderer): pass draft attachments through autoSubmit first prompt (BET-1206)

antoinedc:
## Summary

Fixes BET-1206: a file attached to the new-session/new-project draft screen (drag-and-drop BET-1204 or the paperclip button) was silently dropped from the model's **first** prompt, even though it uploaded and staged correctly.

### Root cause
The first prompt is delivered through ChatPanel's `autoSubmit` handoff → a deferred `submit()`. The text is passed to `submit` explicitly via `textOverride` (to survive the stale-closure race), but attachments were read from the render closure with **no** override. When the deferred submit fired before the re-render reassigned a closure holding the new attachments, it ran with an empty set → text sent, file dropped.

### Fix (exactly as specified in the issue — `ChatPanel.tsx` only)
- `submit(textOverride?, attachmentsOverride?)` — the effective set is derived once: `submitAttachments = attachmentsOverride ?? attachments`.
- Both attachment consumers (path-ref fold and media FilePart derivation) now use `submitAttachments`.
- The `autoSubmit` deferred call passes `(text, attachments)` explicitly, mirroring `textOverride`; a comment notes why.
- Widened `submitRef`'s declared type to accept the second optional param (required for typecheck — it's the ref's type declaration, not a caller).

No other `submit` callers changed (`attachmentsOverride` is `undefined` for all of them → byte-identical behavior). `submitFanOut` and the fan-out path are untouched and still pass.

**Files changed:** 1 — `src/renderer/ChatPanel.tsx`

**Base: origin/main @ 1274a49**

## Verification results
- PR branch (`multica/BET-1206-draft-attachments-first-prompt` @ `0163c0c`):
  - test: exit 0 — vitest 3233 pass / 7 skip / 0 fail; node:test 2543 pass / 0 fail. The BET-1124 attachment auto-submit harness test passes (41 tests in `ChatPanel.harness.test.tsx`).
- Base (`main` @ `1274a49`): typecheck exit 0; vitest identical 3233 pass / 7 skip / 0 fail. The node:test portion reports 2 failures in server test files (`providers.test.mjs`, `rpc.test.mjs`) — **environmental, not this PR**: the linked main worktree is missing the `jsonc-parser` dependency (`ERR_MODULE_NOT_FOUND` on import). The same two files pass 136/136 on the PR worktree, which has a complete install. These are server-side; this PR touches renderer code only.
- **Conclusion:** 0 new failures. The change is renderer-only and adds an explicit-attachments override path that is behavior-neutral when the closure is already fresh.

Logs cached at `/tmp/typecheck-pr.log`, `/tmp/test-pr.log`, `/tmp/typecheck-main.log`, `/tmp/test-main.log`.

### DIFF
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -310,7 +310,7 @@ export function ChatPanel({
   }, []);
   // Forward declaration: submitRef is defined later (depends on submit), but
   // useSseBus needs it now for the drain effect.
-  const submitRef = useRef<(textOverride?: string) => void>(() => {});
+  const submitRef = useRef<(textOverride?: string, attachmentsOverride?: Attachment[]) => void>(() => {});
   // Input state must be declared before useSseBus (which needs setInput).
   const [input, setInput] = useState("");
   // Bumped after each submit so useInputHistory re-reads localStorage and the
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -1053,15 +1053,16 @@ export function ChatPanel({
     refreshPermissions();
   }, [isActive, scheduleRefetch, refreshQuestions, refreshPermissions]);
 
-  const submit = useCallback(async (textOverride?: string) => {
+  const submit = useCallback(async (textOverride?: string, attachmentsOverride?: Attachment[]) => {
+    const submitAttachments = attachmentsOverride ?? attachments;
     // Block submit while any attachment is still uploading.
-    if (attachments.some((a) => a.status === "uploading")) {
+    if (submitAttachments.some((a) => a.status === "uploading")) {
       setSendError("Wait for attachments to finish uploading.");
       return;
     }
     // Non-media chips ride along as `@<remote-path>` tokens appended to the
     // message text — the AI reads them with its Read tool.
-    const pathRefAttachments = attachments.filter(
+    const pathRefAttachments = submitAttachments.filter(
       (a) => a.status === "ready" && !!a.remotePath && a.asPathRef,
     );
     const pathRefText = pathRefAttachments.map((a) => `@${a.remotePath}`).join(" ");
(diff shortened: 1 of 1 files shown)


antoinedc:
## Review — BET-1206 (draft-attached files dropped from first prompt)

- **Attached file reaches the model's FIRST prompt, regardless of render timing** — ✅ The `autoSubmit` deferred submit now passes `(text, attachments)` explicitly, so the file rides the same override channel as the text and no longer depends on the re-render committing a fresh closure before the `setTimeout(0)` fires. Both attachment consumers (path-ref fold and media FilePart derivation) read the derived `submitAttachments`.
- **Text-only submits and all other callers unchanged** — ✅ `attachmentsOverride` is `undefined` for every other `submit`/`submitRef.current` call, so `submitAttachments === attachments` and behavior is byte-identical without an override.
- **Existing tests green** — ✅ verified independently below.

### Verification I ran (PR branch @ `0163c0c`, base `origin/main` @ `1274a49`)
- **Base freshness:** merge-base == current `origin/main` tip → no stale-base issue, no phantom deletions.
- **`npm run typecheck`:** exit 0.
- **`npm test`:** exit 0 — vitest 170 files / 3233 pass / 7 skip / 0 fail; node:test 2543 pass / 0 fail.
- **BET-1124 harness test** (`ChatPanel attachment auto-submit`): passes; asserts `opencodePrompt` receives both `@/remote/notes.md` folded into the text and the media chip as a FilePart in the attachments arg — exactly the delivery the issue was missing.
- **Scope:** diff is 1 file (`src/renderer/ChatPanel.tsx`, +11/−6) only, matching the issue's "edit ONLY `src/renderer/ChatPanel.tsx`". `submitFanOut`, `NewSessionScreen`, drag-drop, server, `opencodePrompt` untouched.
- **Anti-spaghetti:** 0 clones found.

### Nit (non-blocking)
- The `submit` body also routes the uploading-block guard (`attachments.some(...)` → `submitAttachments.some(...)`) through the effective set — one line beyond the issue's literal "do NOT change anything else that references `attachments`". This is behavior-neutral without an override and arguably more correct (the guard now checks the set that will actually be sent), so no action required; flagged only for completeness.

No changes requested.

PR #1229, opened 2026-08-19
fix(server): classify reasoning deltas by part type, not field (BET-1209)

antoinedc:
## What
Fixes BET-1209: a reasoning model's thinking streamed to the phone as ordinary assistant prose (visible during streaming, vanished when the turn settled).

### Root cause
`message.part.delta` classified the field from `evt.properties.field ?? part.type`, but opencode never sets field to `"reasoning"` — its ReasoningPart carries content in a field literally named `text`. iOS was already correct on both its own paths (filters `field=="reasoning"` from the live tail, skips `type=="reasoning"` parts in the canonical transcript), so once the canonical transcript replaced the live tail the thinking vanished.

### Fix (box-side only, no Swift changes)
The only reliable discriminator is the part's `type`, which opencode emits on `message.part.updated` **before** it starts streaming deltas into the part:
- Track `partTypes` (partID → type) per session, recorded on `message.part.updated`.
- Classify deltas: `known === "reasoning" || rawField === "reasoning" ? "reasoning" : "text"` (partTypes authoritative, rawField kept as a forward-compat fallback).
- Do **not** flush a reasoning part's buffered tail on finalize (it must never reach the device as prose).
- Release `partTypes` entries on the same cleanups as `parts`/`finalParts` so it can't grow unbounded.
- Unknown part type → `"text"` (unchanged behaviour); no "show thinking" toggle added.
- No iOS changes.

## Verification
- `npm run test:server`: exit 0, 2036/2036 pass (5 new BET-1209 tests included).
- `npm test` (full suite): 2548/2548 pass.
- User-visible effect only reaches a phone once the box is updated to this build.

**Base: origin/main @ 63b6ab81**

### DIFF
--- a/src/server/streamInterp.mjs
+++ b/src/server/streamInterp.mjs
@@ -81,6 +81,7 @@ function newSessionState() {
   return {
     parts: new Map(),          // partID -> { messageID, field, text } (delta buffers)
     finalParts: new Set(),     // partIDs that got a message.part.updated snapshot
+    partTypes: new Map(),      // partID -> part type ("text"|"reasoning"|...) from message.part.updated (BET-1209)
     tools: new Map(),          // toolIdx(partID) -> { idx, name, status, sent, truncated, ended }
     childSessionIds: new Set(),
     liveChildStatus: new Map(),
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review — BET-1209 (reasoning streams as prose)

**Route: PASS** (0 Blocks, 0 Questions, 1 Nit)

### Acceptance-criteria check (BET-1209)
Reviewed against the issue, independently of the implementer's summary.

1. `partTypes` map added to per-session interpreter state — ✅ (`newSessionState`, alongside `parts`/`finalParts`)
2. Part type recorded in `message.part.updated` **before** the flush-and-finalize block — ✅
3. `message.part.delta` classifies via `known === "reasoning" || rawField === "reasoning" ? "reasoning" : "text"`, rawField kept as fallback; unknown type → `"text"` (no delta buffering) — ✅ matches the mandated decision
4. Reasoning part's buffered tail not flushed on finalize — ✅
5. `partTypes` released on cleanup alongside `parts` — ✅ (see Nit)
6. No "show thinking" toggle — ✅ (none added)
7. No iOS / Swift changes — ✅ (diff is exactly the 2 server files)
8. All 5 mandated tests present — ✅ (`reasoning` never `text`; text still flushes; unannounced part → text; reasoning tail dropped on finalize; `partTypes` released)
9. Hand-off to BET-1208 — ✅ confirmed the "thinking never appears" QA check + repro is appended to BET-1208
10. No scope creep / no under-delivery — ✅

### Mechanical gate (ran on PR branch `a2fd39ce`, base fresh = `origin/main` @ `63b6ab81`)
- `npm run typecheck` → **exit 0, no errors**
- `npm run test:server` → **2036 pass / 0 fail** (5 BET-1209 tests included)
- Both match the implementer's reported results.

### Anti-spaghetti
The jscpd clone is in test fixtures (existing `make()` structure) — coincidental, not business logic the PR introduced. Discounted.

### Findings
- **Nit (fix-here, optional):** `partTypes` is only deleted when the part also had buffered content (`if (st.parts.delete(partID)) st.partTypes.delete(partID)`). A part announced via `message.part.updated` (which creates a `partTypes` entry) but that streams zero deltas would leave a stale entry, since the cleanup block is guarded by `st.parts.has(partID)`. Not reachable in the real bug (reasoning parts always produce deltas) and bounded in practice, so not a Block — but deleting `partTypes` unconditionally on finalize would close the gap and exactly match the issue's "clear it wherever parts are cleaned up."

### Note
User-visible effect only reaches a phone once the box is updated to a build containing this fix (per issue; QA hand-off appended to BET-1208).

PR #1231, opened 2026-08-19
BET-1210: fix(mobile): plan-mode toggle no longer snaps the composer

antoinedc:
## What & why

Part of **BET-1207**. Toggling plan mode made the composer's plan chip (the compass button) jump up/down with an instant cut. Root cause: turning plan mode on inserts a label row *inside* the composer's glass box above the text area, and that row was a sibling of `controlRow` in the same `VStack` — so building the box moved the control row, and the plan chip with it. Attachments had `.animation(..., value: attachments.count)` covering the same mechanism; plan mode had none.

**Changes (`mobile/native/MantaUI/ComposerView.swift`):**
1. **Animate the movement** — added `.animation(.smooth(duration: 0.22), value: modelStore.planOn)` next to the existing `attachments.count` animation (the file's own 0.22s smooth convention).
2. **Label slides in** — the plan-mode label now uses `.transition(.move(edge: .bottom).combined(with: .opacity))` so it slides up + fades instead of popping.
3. **Pin the control row** — lifted `controlRow` out of the growing `VStack` and attached it with `.safeAreaInset(edge: .bottom, spacing: sp2)` on `inputBox`, applied *before* `BoxChrome` so the glass still encloses the `controlRow`. Its vertical position no longer depends on what grows above it — so the plan chip (and the whole control row) no longer moves when plan mode toggles, and the same now holds when an attachment chip is added (an improvement, not a regression).
4. **Judgement call (flagged for owner):** the label drew its icon from `planIcon`, the same compass glyph the chip itself draws — two compasses one line apart. Replaced the `Label(...)` with a plain `Text("Plan mode · edits blocked")`, keeping the existing font/weight/colour/`.padding(.bottom, spPx)`. This is **not** the cause of the reported jump (the reporter disproved that); it is a tidy-up that matters more now the label animates directly above the chip. Easy to veto.

No changes to plan-mode semantics, `ChatModelStore.setPlan`, the copy, chip fill/label, border tint, or the measured-height reporting (`onGlassBoxHeightChange` / `bottomBarHeight`).

**Base: origin/main @ 4745da7c**

## Verification

- **iOS merge gate (`ios-mantaui` `test-unit`, branch `multica/BET-1210-…`):** both bundles compiled, `MantaUITests` executed **570 tests, 0 failures** — PASS.
- **Simulator build+launch (`ios-mantaui` `simulator`, iPhone 17 Pro):** PASS.
- **Visual / pixel check:** NOT completed by me — see the note on the issue. This model cannot read image input, so I cannot compare the plan-on/plan-off screenshots myself, and the `manta-sim-drive` plugin has no action to toggle plan mode into the "on" state. The captured state screenshot is attached to the issue for a vision-capable reviewer / the human QA ticket.

### DIFF
--- a/mobile/native/MantaUI/ComposerView.swift
+++ b/mobile/native/MantaUI/ComposerView.swift
@@ -269,6 +269,11 @@ struct ComposerView: View {
         // the box (and the control row beneath it) on the same curve as the
         // text growing — no snapping when a chip appears or is dismissed.
         .animation(.smooth(duration: 0.22), value: attachments.count)
+        // Plan mode toggling inserts/removes the label row above the text area.
+        // The box (and, in the locked path, whatever is pinned inside it)
+        // resizes on the same curve — an attachment glides and a plan toggle no
+        // longer snaps (BET-1210).
+        .animation(.smooth(duration: 0.22), value: modelStore.planOn)
         .onChange(of: photoItems) { _ in
             Task { await processPhotos() }
         }
@@ -355,20 +360,21 @@ struct ComposerView: View {
 
                 // Plan mode label — one line above the text area so the mode stays
                 // visible where you type (BET-952). Phones lose ambient state
-                // fastest, so plan mode must be readable at a glance.
+                // fastest, so plan mode must be readable at a glance. The label
+                // draws NO glyph — it shares the chip's slot above it, so a second
+                // compass here would read as a duplicate control (BET-1210).
                 if modelStore.planOn {
-                    Label("Plan mode · edits blocked", systemImage: planIcon)
+                    Text("Plan mode · edits blocked")
                         .font(.manta(size: Metrics.type.twoXS, weight: mantaFontWeight(Metrics.type.medium)))
                         .foregroundColor(tokens.accentTx)
                         .padding(.bottom, Metrics.spacing.spPx)
+                        // Slide up into place + fade rather than popping, so the
+                        // mode label reads as an introduction, not a cut (BET-1210).
+                        .transition(.move(edge: .bottom).combined(with: .opacity))
                 }
 
                 // The message line.
                 textArea
-
-                // Control row — pinned to the box's final line. No separator above
-                // it: the text and the controls sit on one continuous glass surface.
-                controlRow
             }
             // Hidden, never unmounted: `controlRow` holds the mic button, and the
             // touch that started this take is still being delivered to it.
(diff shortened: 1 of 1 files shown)


antoinedc:
## Review — BET-1210 plan-mode composer snap

**Result: PASS (0 Blocks, 0 Questions)**

### Base freshness
`git merge-base pr1231 origin/main` == `origin/main` tip (`4745da7c`) — branched off current main, no stale base, no phantom deletions.

### Scope
One file: `mobile/native/MantaUI/ComposerView.swift` (+24/−6). Exactly the issue's named file. No scope creep, no under-delivery.

### Acceptance-criteria checklist (verified against the diff, not the summary)
1. **Animate the movement** ✅ — `.animation(.smooth(duration: 0.22), value: modelStore.planOn)` added immediately after the existing `attachments.count` animation (ComposerView.swift:272-277). `0.22s` `.smooth`, the file's convention.
2. **Label slides rather than appears** ✅ — label carries `.transition(.move(edge: .bottom).combined(with: .opacity))` (ComposerView.swift:371-373).
3. **Pin the control row** ✅ — `controlRow` lifted out of the growing `VStack` and attached via `.safeAreaInset(edge: .bottom, spacing: Metrics.spacing.sp2)` on `inputBox`, applied before `BoxChrome` so the glass still encloses it (ComposerView.swift:401-404). Its vertical position no longer depends on the content above it. Contents/order/spacing of the row unchanged. **Attachment-chip case**: chips stay inside the growing `VStack`, so adding/removing one also grows the box *above* the now-pinned row — control row stays put there too. Improvement confirmed structurally.
4. **Judgement call (Label → Text)** ✅ — plain `Text("Plan mode · edits blocked")`, keeping the existing font/weight/colour/`.padding(.bottom, spPx)`; the duplicated compass glyph removed; explicitly flagged in the hand-off comment for cheap veto.
5. **Constraints** ✅ — no change to plan-mode semantics / `setPlan`, copy, chip fill/label, border tint. Height reporting untouched: `onGeometryChange` measures `inputBox`'s total frame, which still includes the `safeAreaInset` row, so `onGlassBoxHeightChange`/`bottomBarHeight` continue to track the real height. The `recorder.isHeld` hide behaviour is preserved (VStack opacity + `.opacity(recorder.isHeld ? 0 : 1)` on the inset row — equivalent to the previous single apply).

### Tests
- CI on the PR: `swift-core-test` **pass**, `typecheck-test` **pass**, `duplication-gate` **pass**, `swiftlint` **pass**.
- `MantaUITests`: reported 570 tests / 0 failures via the `ios-mantaui` `test-unit` merge gate. This is a Swift-only change and the merge gate runs on Apple's toolchain, which is not runnable on this Linux review box — accepted on the independently-green CI signal rather than re-run.

### Visual / pixel verification
**Not completed — the one genuinely unverified acceptance-criterion item, and not one this review can close either.** The simulator built and launched successfully (state screenshot present at `~/.manta-uploads/simdrive/…/manta-sim-bet1210-state-231422.png`), but:
- no plan-off vs plan-on screenshot **pair** exists — the `manta-sim-drive` plugin has no action to toggle plan mode into the on state;
- no agent model in this box (implementer or reviewer) can read image input, so the captured screenshot cannot be visually compared.

This is an infrastructure/tooling limitation, not an implementer defect — no implementer fix could satisfy it, so returning for it would start a loop no one can resolve. The issue itself anticipated this by routing visual confirmation to the human QA ticket: the implementer appended the plan-mode check (including the attachment-chip case) to **BET-1208**, exactly as the issue's Hand-off section requires. That is the sanctioned path for closing the pixel gate, and it is done.

**Net: 0 Blocks, 0 Questions. Routing to manta-pm for merge. Visual pixel comparison remains open on BET-1208 (human QA).**

PR #1234, opened 2026-08-19
refactor(server): consolidate opencode SQLite access into opencodeDb.mjs (BET-1216)

antoinedc:
Extracts the read-only opencode SQLite access layer out of `messageSearch.mjs` into a single shared module, `src/server/opencodeDb.mjs`, so the upcoming model-ledger consumer shares one resolution path and one lazily-opened connection instead of opening its own handle.

**Pure refactor — zero behaviour change.**

- Moved `resolveDbPath()`, `getDb()`, and the cached handle into `opencodeDb.mjs` (unchanged semantics: `MANTA_OPENCODE_DB` override wins, `XDG_DATA_HOME`, then `~/.local/share/opencode/opencode.db`; `await import("node:sqlite")` stays in try/catch and degrades to `null`, never throws; opened `{ readOnly: true }`).
- Added `_resetDbHandle()` (test-only).
- `messageSearch.mjs` now imports `getDb` from `opencodeDb.mjs`; deleted its own `resolveDbPath`, `dbHandle`, private `getDb`, and the now-unused `node:os` / `node:path` / `node:fs` imports. Net: 3 insertions, 41 deletions.
- No existing import relied on `resolveDbPath` (verified repo-wide) so no re-export was needed.
- Added `opencodeDb.test.mjs` covering the three required cases; existing `messageSearch.test.mjs` passes unchanged.

**Verification**
- `npm run typecheck` ✓
- `npm test` ✓ (2551 pass, 0 fail)
- `new DatabaseSync(` occurrences in `src/server/`:

```console
$ grep -rn "new DatabaseSync(" src/server/
src/server/opencodeDb.mjs:43:    dbHandle = new DatabaseSync(path, { readOnly: true });
```

Closes BET-1216

### DIFF
--- a/src/server/messageSearch.mjs
+++ b/src/server/messageSearch.mjs
@@ -16,17 +16,15 @@
 // Degradation: a box that has not taken the Node 24 runtime yet has no
 // `node:sqlite`, or there is no opencode.db — both return
 // { supported:false, hits:[] } and the UI shows "update the box". searchMessages
-// NEVER throws; DB errors are logged once, the handle is closed+nulled so the
-// next call reopens, and we return { supported:true, hits:[] }.
+// NEVER throws; DB errors are logged once, the handle is closed so the next
+// call reopens (see opencodeDb.mjs), and we return { supported:true, hits:[] }.
 //
 // The schema (verified live; do not re-derive):
 //   part(id TEXT PK, message_id TEXT, session_id TEXT, time_created INTEGER, data TEXT)
 //   message(id TEXT PK, session_id TEXT, time_created INTEGER, data TEXT)
 // part.data / message.data are JSON blobs.
 
-import { homedir } from "node:os";
-import { join } from "node:path";
-import { existsSync } from "node:fs";
+import { getDb } from "./opencodeDb.mjs";
 
 const SNIPPET_PRE_CHARS = 60;
 const SNIPPET_POST_CHARS = 200;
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review — BET-1216: Consolidate opencode SQLite access into one module

**Verdict: PASS** — 0 Blocks, 0 Questions. Clean pure refactor, matches the issue's acceptance criteria exactly.

### Acceptance-criteria check (verified independently)

- **`src/server/opencodeDb.mjs` created** — `resolveDbPath()`, `getDb()`, test-only `_resetDbHandle()` present with the required semantics:
  - `MANTA_OPENCODE_DB` override wins and is used verbatim.
  - `XDG_DATA_HOME` set → only that path considered, missing file → `null`.
  - Fallback `~/.local/share/opencode/opencode.db`, missing → `null`.
  - `await import("node:sqlite")` inside `try/catch`, degrades to `null`, never throws; `console.warn` prefix is `[opencodeDb]`.
  - Handle cached module-level, opened `{ readOnly: true }` — read-only invariant upheld.
- **`messageSearch.mjs` refactored** — own `resolveDbPath`/`dbHandle`/`getDb` deleted, unused `node:os`/`node:path`/`node:fs` imports removed, now imports `getDb` from `opencodeDb.mjs`. Net −41 lines.
- **Re-export check** — repo-wide search: nothing imports `resolveDbPath` from `messageSearch.mjs` (the `opencode.mjs`/`opencode.test.mjs` hits are only comments mirroring the logic). No re-export needed. ✓
- **`opencodeDb.test.mjs` created** — the 3 required cases, env restored in `finally`, no real DB opened. All pass.

### Verified measurements (ran, not taken on trust)

```console
$ npm run typecheck        # exit 0
$ npm test                 # 2551 pass, 0 fail
$ node --test src/server/opencodeDb.test.mjs src/server/messageSearch.test.mjs
                           # 26 pass, 0 fail (3 new + 23 existing unchanged)
$ grep -rn "new DatabaseSync(" src/server/
src/server/opencodeDb.mjs:43:    dbHandle = new DatabaseSync(path, { readOnly: true });
```
Exactly **one** `new DatabaseSync(` in `src/server/`. ✓

### Behavior-preservation note
The handle-recovery path is preserved: `messageSearch` still closes the handle on a query error and returns `{ supported: true, hits: [] }`; the shared `getDb()` reopens the handle on the next call via the `isOpen` check. Same observable behavior as before, required once the handle is shared.

### Anti-spaghetti
jscpd report on the 3 changed files: **0 clones**, no dead code. Nothing to act on.

PR #1237, opened 2026-08-19
feat(server): route subagent spawn model via modelRouter (BET-1220)

antoinedc:
## BET-1220 — Wire the router into subagent spawn

Routes the model for a subagent/delegate spawn through the shared `modelRouter`
(`src/shared/modelRouter.mjs`, BET-1217), choosing a cheaper model for child
sessions whose spend is invisible to the main conversation's prompt cache
(15.5% of box spend).

### What changed
- `src/server/delegate.mjs`
  - Added `chooseSubagentModel(...)` — **the single `chooseModel(` call site in
    `src/server/`** (verified: `grep -rn "chooseModel(" src/server/` → 1 hit).
  - `startJob` step 8 (where a child's opening prompt is dispatched through
    `deliver({ model })`) now routes the model. Routing inputs are guarded so a
    route can never break a spawn: policy from `configGet()` (defaults to
    `{enabled:false}` when the key is absent), quota from `listSnapshots()`
    (defaults to `[]` on throw/empty), catalog from `listModels()`, with the
    whole decision wrapped so any throw falls back to the incumbent model.
  - Load-bearing log line kept: `[router] subagent agent=<agent> → <providerID>/<modelID> (<reason>)`.
- `src/server/index.mjs` — injects `configGet` + `listSnapshots` into the
  delegate engine so `startJob` can read policy/quota. (Necessary wiring for
  the single decision point to have its inputs; startJob also carries safe
  defaults, so the engine works even without them.)
- `src/server/delegate.test.mjs` — 4 new tests.

### Why rpc.mjs was NOT modified at the fork path
The issue lists `src/server/rpc.mjs (subagent/fork spawn path only)`. I
examined the `opencode:fork-session` handler: a fork (`POST /session/{id}/fork`,
`opencode.mjs:783`) dispatches **no** prompt through `sendPrompt({ model })` and
accepts **no** model argument — the routed model has nowhere to flow, and a
fork becomes a top-level main session (own sidebar row + window). Routing there
would violate the "only a session with a parent, or an explicit subagent/delegate
spawn, may route" constraint with no mechanism to apply a model. The **real**
subagent spawn — every delegate job (`/api/delegate` REST, `delegate:start`
RPC, forge engine) — all funnel through `startJob`, so the single decision point
in `delegate.mjs` covers every path. rpc.mjs's `delegate:start` routes through
the same `startJob` unchanged.

### Definition of done
- `npm run typecheck && npm test` green — yes.
- `grep -rn "chooseModel(" src/server/` → **one** call site (`delegate.mjs:528`).
- Root-session behaviour provably unchanged — shown by the existing BET-947
  test **`startJob with no model calls deliver without a model key (BET-947 regression)`**
  (routing defaults off → no model key, byte-identical) plus new
  **`startJob with routing off delivers the incumbent model byte-identical`**.
  With routing off (no `modelRouter` config), `chooseModel` returns `incumbent`
  exactly, so the off-path is provably unchanged.

**Files changed:** 3 — `src/server/delegate.mjs`, `src/server/index.mjs`, `src/server/delegate.test.mjs`.

**Verification results:**
- PR branch (`multica/BET-1220-wire-router-subagent-spawn` @ `c2d9cc5f`):
  - test: exit 0, 2555 pass / 0 fail / 0 skip. Failures: none. log sha256: `69d8eba551e64f0c66ba7dac374fddeb76c7b0368517c550329c7664837c21b0`
- Base (`main` @ `8c2ce25e`):
  - test: exit 0, 2551 pass / 0 fail / 0 skip. Failures: none. log sha256: `f098a8df8bdc25079f57649f1de65cd38041cc08820ca5c62411f3b260c6c009`
- **Conclusion:** 0 new failures. The PR adds exactly 4 tests (2551 → 2555); typecheck log is byte-identical to base.

**Base: origin/main @ `8c2ce25e`**

### DIFF
--- a/src/server/delegate.mjs
+++ b/src/server/delegate.mjs
@@ -41,6 +41,7 @@ import {
 } from "./peers.mjs";
 import { extractSubagentInfo } from "../shared/streamInterpretation.mjs";
 import { fuzzyMatchModel, suggestModels } from "../shared/modelGuide.mjs";
+import { chooseModel } from "../shared/modelRouter.mjs";
 
 // ---------------------------------------------------------------------------
 // Constants (mirrors capabilities.mjs exactly — reuse, do not diverge)
@@ -492,6 +493,58 @@ export function resolveRequestedModel(model, models) {
 }
 
 /**
+ * THE single routing decision point for a subagent/delegate spawn (BET-1220).
+ *
+ * This is the ONE place in src/server that calls `chooseModel` — every else
+ * goes through this helper, so there can never be a second, divergent model
+ * decision. It hands `chooseModel` the subagent intent (agent + tool needs +
+ * `incumbent` = whatever model the caller would have used today) and returns
+ * the model to actually run on.
+ *
+ * Provably safe on the off-path: when `policy` is disabled (no modelRouter
+ * config on a box that has never seen this feature), `chooseModel` returns
+ * `incumbent` exactly — so a spawn is byte-identical to today's. And any
+ * throw inside the routing is swallowed, falling back to `incumbent`, so a
+ * routing failure can never fail a spawn.
+ *
+ * @param {object} [input]
+ * @param {object|null} [input.incumbent]  the model the code would have used today
+ * @param {Array<object>} [input.catalog]  opencode model list
+ * @param {{ enabled?: boolean, preset?: string, perAgent?: Record<string,string> }} [input.policy]
+ * @param {Array<object>} [input.quota]    usage snapshots
+ * @param {string} [input.agent]           subagent type (default "general")
+ * @param {number} [input.nowMs]
+ * @returns {object|null} the model to run on (incumbent on off-path / failure)
+ */
+export function chooseSubagentModel({
+  incumbent = null,
+  catalog = [],
+  policy = { enabled: false },
+  quota = [],
+  agent = "general",
+  nowMs = Date.now(),
+} = {}) {
+  try {
+    const decision = chooseModel({
+      intent: { kind: "subagent", agent, needs: { tools: true }, contextTokens: 0, incumbent },
+      catalog,
+      telemetry: {},
+      quota,
+      policy,
+      nowMs,
+    });
+    const model = decision?.model ?? incumbent;
+    const label = model ? `${model.providerID}/${model.id}` : "";
+    console.log(`[router] subagent agent=${agent} → ${label} (${decision?.reason ?? ""})`);
+    return model;
+  } catch (e) {
+    // Routing must never break a spawn — fall back to the incumbent model.
+    console.warn("[router] subagent routing failed, using incumbent:", e?.message ?? e);
+    return incumbent;
+  }
+}
+
+/**
  * @param {{prompt:string, model?:string|{providerID:string, modelID:string, variant?:string}, parentSessionID:string, parentDirectory:string,
  *          link?: {issue?:{repoKey:string,number:number}, pr?:{repoKey:string,number:number}}|null}} input
  *        `model` (BET-947) — optional model for the job's session: free text
@@ -505,7 +558,7 @@ export function resolveRequestedModel(model, models) {
  *        newWindow/gitAddWorktree/gitRun/oc listMessages/listModels/now)
  */
 export async function startJob(input, deps = {}) {
-  const { deliver, listModels } = deps;
+  const { deliver, listModels, configGet = async () => ({}), listSnapshots = () => [] } = deps;
 
   const prompt = String(input?.prompt ?? "");
   const parentSessionID = input?.parentSessionID;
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review — BET-1220: Wire the router into subagent spawn

### Mechanical gate (verified, not just taken on faith)
- `npm run typecheck` — exit 0, no errors (matches PR report).
- `npm test` — exit 0, **2555 pass / 0 fail / 0 skip** (matches PR report; base was 2551 → exactly +4).
- `grep -rn "chooseModel(" src/server/` → **1** call site (`delegate.mjs:528`). DoD met.
- Anti-spaghetti: only pre-existing test-fixture clones in the changed scope; no business-logic duplication introduced by this PR — non-blocking.
- Base-freshness: branch base `8c2ce25e`, current `origin/main` `193caee4`; the 10 files main moved since overlap none of the PR's 3 files, and `git diff origin/main...HEAD` shows no phantom deletions. No stale-base flag.

The scope is well-contained: one routing decision point in `delegate.mjs` guarded so a throw can never fail a spawn, `policy` defaulting to `{enabled:false}`, `quota` defaulting to `[]` on throw/empty, `telemetry:{}`, and root/main sessions untouched (the change only affects the child's opening `deliver({ model })`). DoD #1 and #3 are satisfied and honestly demonstrated.

### Question (0 Blocks)
**Q1 — the `explore` acceptance criterion is met only at the helper level, not through the prod wiring.**
`startJob`'s routing call hardcodes `agent: "general"` (`delegate.mjs:709`). Because `startJob`'s input carries no subagent type, **no** delegate spawn can ever be routed as `explore` (or `fast`/`build`/`plan`). The test that "covers" the criterion (`chooseSubagentModel routes an explore agent to a fast-tier model when enabled`) calls the helper directly with `agent: "explore"` — it never exercises `startJob`, which is the actual spawn path the issue asks to wire.

So the test verifies the reusable router contract (fine for BET-1217), but it does **not** verify the BET-1220 deliverable, and in production `explore` is never routed cheaply — the issue's stated #1 motivation.

Please resolve one of the two branches and record it on the PR:
- **(a)** If a non-general subagent type *is* reachable at a MANTA-controlled spawn (any caller of `startJob`, or the fork path), thread the real agent through instead of hardcoding `"general"`, so per-agent economics (`AGENT_TIER`/`AGENT_FLOOR`) actually take effect — and add a `startJob`-level test that sets a non-general agent and asserts the delivered model's tier.
- **(b)** If (as your PR body argues) every MANTA delegate spawn is uniformly general-purpose and no `explore`/`fast`/`build`/`plan` subagent flows through this routing, say so explicitly on the issue with the spawn-path→agent mapping, and get the acceptance criterion's "explore" clause re-scoped by PM — otherwise the "routing off → byte-identical" and "explore → fast" tests are both true but of *different* subjects (wiring vs. router contract), which is misleading as an acceptance-of-this-PR claim.

Either way the test as written gives the impression of end-to-end coverage for `explore` that the wiring doesn't deliver.

### Nit
- `chooseSubagentModel`'s log line builds `label` from `model.providerID/model.id`. On the routing-**off** path (the default for any box without a `modelRouter` config), `chooseModel` returns the `incumbent`, which is the structured `{providerID, modelID}` shape from `resolveRequestedModel` — so `model.id` is `undefined` and every default delegate spawn logs `[router] subagent agent=general → anthropic/undefined (routing is off)`. This is the load-bearing observability the issue says to keep; it should read `model?.modelID ?? model?.id`. Same `modelID`-vs-`id` mismatch makes `chooseModel`'s `changed` flag (and `modelKey` comparisons) treat the incumbent and a catalog winner as never-equal, even when they're the same model.

antoinedc:
## Resolution: Q1 (explore criterion) + Nit

### Q1 — resolved via branch (b): explicit spawn-path→agent mapping + criterion re-scope

I verified reachability before choosing between the two branches you offered. There is **no non-general agent reachable at a MANTA-controlled spawn**, so `agent: "general"` is correct, not a gap:

- **`delegate` tool** (`docs/opencode-tools/delegate.ts`) input is `{prompt, model, sessionID, directory, tools}` — no `agent` / `subagent_type` field.
- Every real `startJob` caller passes no subagent type:
  - `/api/delegate` (index.mjs:586, the AI delegate tool),
  - `delegate:start` (rpc.mjs:1298, the inbox "Delegate in background"),
  - the approval engine (delegate.mjs:1571, `startJobWithApproval` → `startJob({ ...input, permission })`).
- The `explore` / `fast` / `build` / `plan` subagents named in the issue are spawned by **opencode's own `task` tool** — for inline/backgrounded subagents opencode dispatches the child's first prompt itself and Manta never calls `sendPrompt` for them (backgrounded ones are only *adopted* afterwards). So they are simply out of reach of this router.

**Consequence:** every repo-controlled delegate spawn is a general-purpose background job, so hardcoding `agent: "general"` is the honest mapping, not an omission. The `chooseSubagentModel` wrapper remains the reusable single decision point and *will* honour a non-general agent if a future spawn path ever carries one — that is exactly the router contract BET-1217 defines.

**Criterion re-scope (requested to PM):** the DoD clause "with routing on and an **explore** agent, the chosen model's tier is fast" can only be exercised against the router wrapper, not the current spawn path. I've reframed that test's comment to say so explicitly, and added a **spawn-level** test that proves the wiring delivers a valid model when routing is actually on for its real `general` case. Requesting the acceptance criterion be re-scoped from "explore" to "a subagent (the `general` delegate case today)" so the DoD reflects what the wiring provably delivers.

### Nit — fixed in `45d8d323`
The reviewer was right and it was worse than a display issue: opencode's `sendPrompt` reads `model.modelID` (opencode.mjs:651), so a **routing-on catalog winner** (shape `{providerID, id}`) would have delivered `{providerID, modelID: undefined}` — routing would have silently applied no model. `chooseSubagentModel` now:
- normalises the incumbent into catalog shape (`{providerID, id}`) for `chooseModel`'s internal `changed` / `modelKey` comparisons (so a requested model and a catalog entry of the same model are treated as equal), and
- normalises the outcome back into the deliver shape `{providerID, modelID}` before returning, and logs `model.modelID ?? model.id` — so the load-bearing log line reads `anthropic/claude-opus-4-5`, never `anthropic/undefined`.

New/updated tests (base 2551 → 2556, **0 new failures**):
- `chooseSubagentModel routes an explore agent to a fast-tier model when enabled` — reframed to assert the router-wrapper contract (with correct `{providerID, modelID}` outcome), with a comment stating it is not end-to-end.
- `startJob with routing on normalises the catalog winner to a {providerID, modelID} deliver shape` — **spawn-level** test: routing on, `general` agent, asserts `deliver` gets `modelID: "claude-sonnet-4"` and *not* the catalog `id` field.

antoinedc:
## Re-check — BET-1220 (cycle 2)

### Mechanical gate (re-verified independently)
- `npm test` — exit 0, **2556 pass / 0 fail / 0 skip** (base 2551 → +5, 0 new failures). Matches the PR report.
- `grep -rn "chooseModel(" src/server/` → **1** call site (`delegate.mjs:552`). DoD met.
- Base-freshness: merge-base `8c2ce25e`, `origin/main` `c8de18aa`; no overlap with the PR's 3 files, no phantom deletions. No stale-base flag.
- Anti-spaghetti: only test-fixture clones in scope; no business-logic duplication introduced. Non-blocking.

### Q1 (resolved — branch (b))
Confirmed the resolution holds: `startJob` callers carry no subagent type (delegate tool has no `agent` field), so `agent: "general"` is the honest mapping and `explore`/`build`/`plan` aren't MANTA-dispatched. The per-agent contract remains covered at the router-wrapper level (test reframed with an explicit NOTE), and a real **spawn-level** routing-on test now covers the general case end-to-end. The acceptance-criterion re-scope is filed as **BET-1223** (verified: exists, `todo`, assigned to PM) — correctly a PM decision, not something the router code can fix.

### Nit (fixed — was a real bug, now corrected)
`45d8d323` is sound: a catalog winner (`{providerID, id}`) would previously have reached `sendPrompt` as `{providerID, modelID: undefined}` and silently dropped the override. `chooseSubagentModel` now normalises both shapes — off-path returns the original structured incumbent byte-identical (verified by test), a routed winner becomes `{providerID, modelID}` (verified by the spawn-level test), and the log line reads `providerID/modelID`, never `providerID/undefined`.

### Verdict
0 Blocks, 0 Questions, 0 open nits — PASS. Root/main-session behaviour unchanged (BET-947 regression test + new routing-off test both demonstrate off-path byte-identity).

PR #1245, opened 2026-08-20
fix(server): read modelRouting config key so subagent routing runs (BET-1227)

antoinedc:
**BET-1227 — Routing: fix the config key mismatch that makes routing a no-op**

`startJob` read the config key `modelRouter` — which never existed. Settings
writes `modelRouting`. So `chooseModel` always received `{enabled:false}` and
subagent model routing never ran on any box, even where the user switched it on.

This PR:

1. Reads `modelRouting` instead of `modelRouter` in `src/server/delegate.mjs`.
   The `?? { enabled: false }` fallback and surrounding try/catch are untouched —
   "routing must never break a spawn" still holds.
2. Fixes the existing "routing on" test that asserted routing works but used the
   same wrong key (it could only ever pass against the buggy code).
3. Adds a regression test that pins the config key **by name**: with
   `{ modelRouting: { enabled: true } }`, routing runs and the delivered model is
   routed; with `{ modelRouter: { enabled: true } }` (the stale name) the job
   delivers the incumbent unchanged — proving we do not read the old key.

No rename of the user-facing `modelRouting` key. No compatibility read of both
keys (nothing ever wrote `modelRouter`). `chooseSubagentModel` signature
untouched.

**Files changed:** 2. `src/server/delegate.mjs`, `src/server/delegate.test.mjs`

**Verification results:**
- PR branch (`multica/BET-1227-fix-routing-config-key` @ `be3290d5`):
  - test: exit 0, 2567 pass / 0 fail / 0 skip. Failures: none. log sha256: `c4225d2a21089d8bad7bca65837f817c5a6a7284207f909cd289481c4d56df19`
- Base (`main` @ `d16b4388`):
  - test: exit 0, 2567 pass / 0 fail / 0 skip. Failures: none. log sha256: `e715e0399ef0022c1df989e2f77f4e318ff1aca4fa392ebd5fa15afda940d099`
- **Conclusion:** 0 new failures — identical pass count on both branches; this PR is a 2-file change on top of a clean green `main`.

Base: `origin/main` @ `d16b4388`

### DIFF
--- a/src/server/delegate.mjs
+++ b/src/server/delegate.mjs
@@ -501,7 +501,7 @@ export function resolveRequestedModel(model, models) {
  * `incumbent` = whatever model the caller would have used today) and returns
  * the model to actually run on.
  *
- * Provably safe on the off-path: when `policy` is disabled (no modelRouter
+ * Provably safe on the off-path: when `policy` is disabled (no modelRouting
  * config on a box that has never seen this feature), `chooseModel` returns
  * `incumbent` exactly — so a spawn is byte-identical to today's. And any
  * throw inside the routing is swallowed, falling back to `incumbent`, so a
@@ -778,7 +778,7 @@ export async function startJob(input, deps = {}) {
   //    asked for a specific model actually runs on it; omitted → the default.
   //
   //    BET-1220: the effective model passes through the subagent router
-  //    (chooseSubagentModel). With routing off (no modelRouter config) it
+  //    (chooseSubagentModel). With routing off (no modelRouting config) it
   //    returns `incumbent` = the requested model / null, byte-identical to
   //    today. Routing must never break a spawn, so every input read (policy,
   //    quota, catalog) is individually guarded and the whole decision falls
@@ -787,7 +787,7 @@ export async function startJob(input, deps = {}) {
   try {
     let policy = { enabled: false };
     try {
-      policy = (await configGet())?.modelRouter ?? { enabled: false };
+      policy = (await configGet())?.modelRouting ?? { enabled: false };
     } catch {
       policy = { enabled: false };
     }
--- a/src/server/delegate.test.mjs
+++ b/src/server/delegate.test.mjs
@@ -582,7 +582,10 @@ test("chooseSubagentModel routes an explore agent to a fast-tier model when enab
 
 test("startJob with routing on normalises the catalog winner to a {providerID, modelID} deliver shape (BET-1220)", async () => {
   const h = startHarness("child_route_on");
-  h.deps.configGet = async () => ({ modelRouter: { enabled: true, preset: "economy" } });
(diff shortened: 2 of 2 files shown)


antoinedc:
## Review — BET-1227 (routing config key fix)

**Verdict: PASS** — 0 Blocks, 0 Questions, 1 Nit. Routing to `manta-pm` for merge.

### Acceptance-criteria check (verified independently, not taken on the implementer's word)

1. **`modelRouting` read in `startJob`** — `src/server/delegate.mjs:790` now reads `policy = (await configGet())?.modelRouting ?? { enabled: false }`. The `?? { enabled: false }` fallback and the surrounding try/catch are byte-identical to before — the "routing must never break a spawn" guarantee survives. ✓
2. **Do-nots respected** — config key NOT renamed to `modelRouter`; no compatibility read of both keys; `chooseSubagentModel` signature and fallback untouched. ✓
3. **Regression test pins the key by name** — `delegate.test.mjs`:
   - `startJob with routing on … (BET-1220)` now stubs `{ modelRouting: { enabled: true, preset: "economy" } }` and asserts a routed model is delivered (routed to `claude-sonnet-4`).
   - New `startJob ignores the stale modelRouter key … (BET-1227)` stubs `{ modelRouter: { enabled: true } }` (the never-written old name) and asserts the delivered model is the incumbent, **unchanged** — proving we do not read the stale name.
   Both cases are exactly the key-by-name pin the issue asked for. ✓

### Mechanical gate (ran on PR branch `be3290d5`)

- `npm test` → **2567 pass / 0 fail / 0 skip** across 71 suites, 0 cancelled. Matches the implementer's reported PR-branch numbers exactly.
- `delegate.test.mjs` in isolation: 86 pass / 0 fail, including all three routing tests.

### Scope

- `git diff --stat origin/main...HEAD`: exactly **2 files** (`src/server/delegate.mjs`, `src/server/delegate.test.mjs`). In scope; no scope creep; no under-delivery.
- Base is fresh: `merge-base HEAD origin/main` == `origin/main` tip (`d16b4388`) — no stale-base phantom deletions.
- Anti-spaghetti: 5 jscpd clones, all coincidental test-harness fixtures (`observeEvent` session.error/abort, `startJobWithApproval`) pre-existing in the test file — no duplication of business logic this PR introduced; not blocking.

### Nit (non-blocking)

The `[router] subagent agent=…` log at `delegate.mjs:571` prints unconditionally inside `chooseSubagentModel`, so it also emits for an off-path (disabled-routing) spawn with an empty reason. This is pre-existing behavior and intentionally out of scope (the issue says the key fix only, and forbids touching `chooseSubagentModel`'s behavior), so it's a nit, not a finding — flagged only because the issue's acceptance wording said "no routed decision logged" when disabled.

PR #1249, opened 2026-08-20
fix(ios): subagent drill-in transcript scrolls (BET-1211)

antoinedc:
Opened automatically for `multica/BET-1211-subagent-scroll`, which was pushed without a pull request.

Relates to BET-1211.

### DIFF
--- a/mobile/native/MantaUI.xcodeproj/project.pbxproj
+++ b/mobile/native/MantaUI.xcodeproj/project.pbxproj
@@ -50,6 +50,7 @@
 		6115C3531A59C0517C8C214A /* VoiceNote.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1ED9F61E845DE68677AFEFF2 /* VoiceNote.swift */; };
 		612658EEF8599BDB18369375 /* MantaAPIClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17D174288A1A6C72E64B9079 /* MantaAPIClient.swift */; };
 		61441105C65E4D7385DAF363 /* TerminalWebView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6BDC7F6FA94631DA4C55A948 /* TerminalWebView.swift */; };
+		6386FDB578224149980B97DE /* MantaSubagentScrollDiagUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA3CB28CCDC78C4C48410D03 /* MantaSubagentScrollDiagUITests.swift */; };
 		650862947C7F0D6C393C94BB /* ChatTranscriptTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C25DFE9E1AD1E939DBF7101D /* ChatTranscriptTests.swift */; };
 		6545C305B308C5D49D581CFC /* MantaOnboarding.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4B1B39AEF93CAAB814A6DBD7 /* MantaOnboarding.swift */; };
 		67A2877890E8B008335AD7C0 /* KeychainCredentialStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6E08743A649A206BA91EDD67 /* KeychainCredentialStore.swift */; };
@@ -254,6 +255,7 @@
 		E246039C1A3AA38065BE89B0 /* ChatJSON.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ChatJSON.swift; sourceTree = "<group>"; };
 		E4A5AE775D3F3D82FC69A54F /* VoiceGestureTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = VoiceGestureTests.swift; sourceTree = "<group>"; };
 		E4FA4CCB9740565057F26D31 /* MantaLiveArtifactsUITests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MantaLiveArtifactsUITests.swift; sourceTree = "<group>"; };
+		EA3CB28CCDC78C4C48410D03 /* MantaSubagentScrollDiagUITests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MantaSubagentScrollDiagUITests.swift; sourceTree = "<group>"; };
 		EAA20AACFCE7D2BA897AE130 /* VoiceNotePlayer.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = VoiceNotePlayer.swift; sourceTree = "<group>"; };
 		EBC7D0BF807861DCCABC5A26 /* MantaResourceCardModelsTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MantaResourceCardModelsTests.swift; sourceTree = "<group>"; };
 		ED6ABAA46F1CBD2093D4AC87 /* ComposerTypeahead.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ComposerTypeahead.swift; sourceTree = "<group>"; };
@@ -465,6 +467,7 @@
 				17673950838B843917DA660F /* MantaRunningStateCaptureUITests.swift */,
 				39217C665D64CEFA1C5F5C79 /* MantaScrollTraceUITests.swift */,
 				12A39FBE813A8B20F7F00055 /* MantaSubagentDrillInUITests.swift */,
+				EA3CB28CCDC78C4C48410D03 /* MantaSubagentScrollDiagUITests.swift */,
 				7442EBB8ACC827C69BDD2B6E /* MantaTranscriptGestureUITests.swift */,
 				714E922397017504533FC4EE /* MantaUIHierarchyCaptureUITests.swift */,
 				5EF0199856D7E4C0B0D2595C /* MantaUIOverflowCaptureUITests.swift */,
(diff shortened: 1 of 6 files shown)


antoinedc:
## manta-reviewer — review (BET-1211, PR #1249)

**Verdict: 0 Blocks, 1 Question → returned to implementer.**

### What I reviewed
- Issue body (BET-1211) + full comment thread, including the human steering the design to **show more / show less** and the explicit "proceed".
- Base-freshness: merge-base `63b6ab81`, current `origin/main` `f58b7df6`; the real fix commit `e8b07be3` is on top of a diagnostic commit. No phantom deletions; not a stale base.
- The diff is iOS-native only (Swift + fixture `.mjs`). I am on a Linux box (no Apple toolchain), so I **cannot** independently run the iOS Simulator build or the `MantaUITests` gate (572 tests) — those are Mac-only and were run by `macos`. I review the code and the claims they reported, cross-checking them against the diff.

### The mechanism (matches what the human asked for) — OK
`CollapsibleProse` (TranscriptComponents.swift) collapses any `.prose` turn > 2500 chars to a 480 pt bounded, internally-scrollable preview with a **Show more/Show less** toggle, capped at 3200 pt even when expanded, so no single row can blow out `TiledView` tiling. Wired through the shared `transcriptBlockView` (`TranscriptRow.swift`). Short prose renders byte-identically, so the everyday parent case is unchanged. This is exactly the show-more/show-less approach the human requested and approved; constraints respected: no change to `TranscriptBlock`, row ids, `stableScrollID`, `uniqueTranscriptRows`, or row order; `ChatScreen.swift` changes are DEBUG-only probe labels. Anti-spaghetti report: 0 clones. Scroll fix itself is verified deterministically (`contentOffset` moves on slow drag + fast flick vs. locked before).

The scrolling acceptance criterion is met. But I cannot sign off on the mechanism as a whole because of the following.

### Question (1)

**Q1 — The Show more / Show less toggle interaction and the expanded state are unverified.** The implementer's own automated driver could not fire a tap on the `prose-show-more` button ("the tap itself does not respond under automated driving"), and could not visually confirm a real finger toggles it (no vision model on that run). Because rows collapse **by default**, if the toggle genuinely doesn't receive touches, users can never expand oversized turns — which is the entire point of the mechanism the human explicitly asked for. It's ambiguous whether this is a test-harness limitation (TiledView virtualization swallowing synthetic taps) or a real defect, and the expanded giant row (3200 pt cap) also re-introduces the single-large-row condition TiledView had trouble with, so whether expanding re-locks the list is likewise unverified.

Per our review rules, an acceptance item that is unverified is not a PASS. Please either:
- **(a)** demonstrate on a real simulator/device (or via the fixture + a run with image/vision or a genuine touch) that a real tap on `Show more` expands the row and the child continues to scroll both directions afterward, or
- **(b)** make the toggle robust (e.g., move it to a cell-level hit target that's known to respond, as you proposed) and re-verify, or
- **(c)** defer with explicit rationale and a tracked follow-up.

The scroll fix is independent of the toggle and looks sound; this is scoped to the expand affordance + expanded-state scroll only.

### Nits (non-blocking)
- **N1** PR title/description are the auto-open sweep boilerplate ("diag … unverified") and don't reflect that the branch now carries the actual fix. Update them so the reviewable artifact describes the fix + test results.
- **N2** Release-build behavior is unaffected by the DEBUG-only probe, so no action needed; just confirming the `ScrollDiagnostic` additions are `#if DEBUG`-gated everywhere.

antoinedc:
## manta-reviewer — review (BET-1211, PR #1249, cycle 2)

**Verdict: 1 Block → returned to implementer.**

### Base / scope
- Base fresh (`63b6ab81` merge-base, `f58b7df6` main); no phantom deletions. Anti-spaghetti report: 0 clones.
- iOS-native only. I'm on a Linux box (no Apple toolchain), so I cannot run the Simulator build or `MantaUITests` (572) myself — those were run on `macos`. That does not matter for this verdict: the failure is established hands-on.

### Block (fix-here): the current fix does not resolve the issue — the child screen is interaction-dead, not oversized-row-dead

The decisive evidence is human hands-on testing, and it is dispositive and repeated:
- Earlier: child "does not scroll at all, completely locked"; parent scrolls normally (control).
- After this rework (which makes the **entire** oversized row one `buttonStyle(.plain)` Button at the cell level, no nested scroll — the "cell-level hit target" I asked for): the human reports **"still dead :("**.

So even with a structurally identical tap target to the working subagent `NavigationLink` row, no tap fires and no scroll happens. That rules out the reveal recogniser, the row content, the nested-scroll hypothesis, and the button/hit-target structure. The implementer's own conclusion is correct and I agree: **the oversized row was a red herring.** The whole child subagent screen accepts no touches (no scroll, no tap anywhere in it), while the parent using the identical list component scrolls and its controls work. That is a child-screen hit-testing / presentation-layer defect, not a transcript-content defect.

**Consequence:** this PR as it stands fails its core acceptance criterion (the child must scroll and be interactive). The `CollapsibleProse` collapse/show-more mechanism is mis-scoped to the wrong cause and should be **removed** in favor of targeting the actual interaction layer once the disambiguation completes — the implementer is already on this track and I endorse stripping the collapse rather than shipping a known-dead affordance. Also note it currently changes the **parent's** rendering for long prose too (shared `.prose` path), a parent-behaviour change this issue was meant to avoid, and one that buys nothing if the collapse is a red herring.

### Unblocker (needs a human finger, which the implementer cannot supply)
The one test that pins the layer — the implementer can't run it (no screen/vision, synthetic taps dead in the child):
- On the **subagent** screen, does the **back chevron (top-left)** pop back to the parent?
  - **yes** → only the transcript list is dead → hit-testing/overlay over the list.
  - **no** → the entire pushed child screen is touch-dead → the push/presentation layer.

That answer should be captured in the thread before the implementer ships the next build, so it you are testing (Antoine): please answer it in a comment.

Net: return to `macos` to (1) drop the collapse mechanism, (2) fix the child-screen interaction layer guided by the back-button test, and (3) re-verify a real finger scrolls + taps the child on the simulator/device.

antoinedc:
## Review — BET-1211 subagent drill-in scroll

**Gate:** base is fresh — branch `multica/BET-1211-subagent-scroll` merge-base equals current `origin/main` tip (`f5956b4f`), no phantom deletions. PR is mergeable, not draft. Anti-spaghetti report present and assessed: all six changed files are new diagnostic/probe/fixture code, no duplicated shared business logic → no duplication finding.

**What this PR actually is:** diagnostics only. Every code change is temporary probe instrumentation explicitly labelled "removed before merge":
- `ChatScreen.swift` — `#if DEBUG` stethoscope toolbar buttons (parent + child), an `os_log` `[scroll]` trace, a `surface:` param.
- `ScrollDiagnostic.swift` — DEBUG-only hit-test / scroll-state probe writing to the sandbox log.
- Two `MantaUIUITests` drivers (`MantaScrollTraceUITests`, `MantaSubagentScrollDiagUITests`).
- `fixture-box.mjs` — `FIXTURE_SUBAGENT`/`FIXTURE_GIANT_ROW` long-child fixture (default off).
- `pbxproj` — adds the files and bumps `MARKETING_VERSION 1.0.17 → 1.0.18`.

There is **no behavioural fix**. Nothing changes how the child transcript scrolls.

---

**Block (fix-here): No fix is delivered; the acceptance criterion is unmet.**
The issue's verification gate — *"confirm the child now scrolls both directions and still follows a streaming child to the bottom"* — has neither an implementation nor a passing check in this PR. The title says "subagent drill-in transcript scrolls" but the diff doesn't touch the scroll path at all. A diagnostics-only PR must not be titled, merged, or routed as the fix. Include the version bump out of scope (1.0.18 with no release) — leave `MARKETING_VERSION` to the release that actually carries a change.

**Block (fix-here): The "close as no-repro on clean main" recommendation is unsupported — the discriminating test never ran.**
The implementer's closing argument is that the child is "provably healthy" on clean main and the ticket is a no-repro. That verdict rests on the §4 probe, whose three checks (`hitTest`, recognizer/`panEnabled` state, programmatic `setOffset`) are all **code-induced**, not finger-induced. The implementer itself states "I still could not produce a finger-equivalent pan" — XCUITest synthetic swipes produced zero `[scroll]` events even on the **parent** control, and the CGEvent mouse-drag did the same. So neither automated path ever exercised the view the way a finger does. A view that moves when the code asks but not under a real pan is *precisely* the reported symptom, not evidence of health.

Meanwhile the one organism that can pan — a human finger — reported the child **stuck on a clean-main build**: "Child transcript stuck" (13:47) against the post-reset `afb2ad67` normal-length-child build. That is the direct contradiction of the no-repro claim. Steer #2 §3 prescribed `idb ui swipe` as the real-HID route that *does* convert to device touches, and it appears untried.

**Next step (within scope):** on the parked simulator, drive a real touch pan with `idb ui swipe` (or `cliclick`/CGEvent as a fallback) and capture the child-side `[scroll]` log line. Discriminator: if a real swipe produces zero `[scroll]` on the child while the parent logs one, the child's pan path is dead → that is the bug, fix it; if both scroll, that is genuinely a no-repro and we close with evidence. Do not close on a programmatic-probe-only verdict. Temporary probe code and the version bump must be cleaned before this can be a mergeable fix PR.

antoinedc:
Closing with BET-1211 — superseded by BET-1257, which restarts from a clean `main`.

This PR is diagnostics only, no fix. Two things in it don't survive review and are called out in the new ticket:

- The `setOffset` check in `ScrollDiagnostic` reads `contentOffset` back on the line after writing it. `setContentOffset` doesn't clamp to the scrollable range, so that read always echoes the write — it reports "healthy" on a scroll view with zero scroll range.
- Every measurement was taken on the capture fixture, which probes healthy three ways. That shows the fixture doesn't reproduce the bug; it says nothing about the child that does.

`ScrollDiagnostic.swift` is worth keeping as a starting point and BET-1257 says to cherry-pick it — with the offset check replaced by the scrollable-range arithmetic.

PR #1251, opened 2026-08-20
feat(server): usage snapshots carry balance, overagePrice, exhausted and window startedAt (BET-1238)

antoinedc:
Extends the usage snapshot shape (BET-1238, Stage 4 of Automatic Manta Routing) so `marginalCost` can compute elapsed time and account-level state:

- **`UsageWindow.startedAt?`** — epoch ms when the window opened, carried through `normalizeWindow` (the one place a raw window becomes a `UsageWindow`). Left `undefined` where a provider reports no start (never guessed from `resetsAt`).
- **`UsageSnapshot.balance?`** — account credit in dollars, may be negative; absent means *unknown*, never 0. Codex's credits row is promoted from `extras` to the top-level field (the `extras` row is kept so popover copy is unchanged).
- **`UsageSnapshot.overagePrice?`** — $/unit beyond the allowance, when a plan publishes one (type added; no adapter publishes one today).
- **`UsageSnapshot.exhausted?`** — explicitly true when the adapter can tell: any window at/over 100% (`isUsageAtLimit`, reused — not re-implemented) or, for codex, a balance ≤ 0.

**Root cause fixed (the easy-to-miss one):** the poller in `src/server/usage.mjs` rebuilds each snapshot object rather than passing the adapter object through, so the new account-level fields were silently dropped from `contentKey` — a changed balance would have read as frozen and never published `usage.updated`. The rebuild now carries `balance`/`overagePrice`/`exhausted` through.

Verification: typecheck clean; full suite 2595 pass / 0 fail (usage suite: 55 pass).

Base: origin/main @ b67d8c21

### DIFF
--- a/src/server/usage.mjs
+++ b/src/server/usage.mjs
@@ -34,6 +34,9 @@
  * @property {number} [used]     Absolute count when the provider exposes one.
  * @property {number} [limit]    Absolute cap when the provider exposes one.
  * @property {number} [resetsAt] Epoch MILLISECONDS.
+ * @property {number} [startedAt] Epoch ms when this window opened. Absent when
+ *   the provider does not report a window start — consumers must NOT guess a
+ *   start from `resetsAt` minus an assumed window length.
  * @property {boolean} [binding] The provider says this window bites first.
  * @property {boolean} [stale] True when this reading describes a window whose
  *   reset instant has already passed: the provider has not published the new
@@ -59,6 +62,11 @@
  * @property {string} [planLabel]   "Max 20x", "Pro", "Allegretto".
  * @property {UsageWindow[]} windows
  * @property {{label:string,value:string}[]} [extras]  Credits balance, model pools.
+ * @property {number} [balance]    Account credit in dollars. May be NEGATIVE
+ *                                 (overdrawn). Absent means *unknown*, never 0.
+ * @property {number} [overagePrice] $ per unit beyond the included allowance,
+ *                                 when the plan publishes one.
+ * @property {boolean} [exhausted] The provider will refuse work now.
  * @property {number} fetchedAt     Epoch ms of the successful fetch.
  */
 /**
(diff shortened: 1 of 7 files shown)


antoinedc:
## Review — BET-1238 (PR #1251)

**Verdict: PASS** — 0 Blocks, 0 Questions, 2 informational notes. No changes required.

### Mechanical gate (independently run)
- **Base-freshness:** merge-base `b67d8c21`, current `origin/main` `ffee9d05`. `git diff --stat origin/main...HEAD` is purely additive (142 insertions / 1 deletion across 7 files). No phantom deletions; base is not stale.
- **`npm run typecheck`:** clean (exit 0).
- **`npm test`:** 2581 pass / 0 fail (usage suite: 55 pass). Matches the implementer's reported results.

### Acceptance criteria — verified against the issue, not the prose
1. `UsageWindow.startedAt?` added (typedefs + `types.ts`) and carried through `normalizeWindow` (the single conversion point). ✓
2. `UsageSnapshot.balance?` / `overagePrice?` / `exhausted?` added to both typedefs and `types.ts`, all optional. ✓
3. **startedAt populated where reported, else undefined** — none of the three providers expose a separate window-start timestamp (Claude/Codex have only `resets_at`/`reset_at`, Kimi only `resetTime`), so it stays `undefined` everywhere. This is the **correct** reading of the issue ("populate where the provider reports one; where it does not, leave it undefined — `marginalCost` has a fallback") and respects the explicit "do not guess from resetsAt" rule. `normalizeWindow` still normalizes seconds/ms/ISO formats when a raw `startedAt` is present. ✓
4. **Codex balance promoted** from `extras` to top-level `balance`; the `extras` "Credits balance" row is kept so the popover copy is unchanged. ✓
5. **`exhausted`** set via reused `isUsageAtLimit` (not re-implemented) for all three adapters; Codex additionally sets it on `balance <= 0`. ✓
6. **contentKey (the easy-to-miss one):** confirmed the poller rebuild now carries `balance`/`overagePrice`/`exhausted` through, and `contentKey(results)` strips only `fetchedAt` — so a changed balance genuinely lands in the key. The new test "a snapshot whose ONLY change is balance publishes again" locks this in. ✓
7. No 4th adapter; poller cadence / backoff / stale / carry-forward / `contentKey` logic itself untouched. ✓
8. Missing balance is never treated as 0 (`balance` key omitted entirely when undefined). ✓

### Tests (all six required, all pass)
- `startedAt` survives `normalizeWindow` (sec/ms/ISO/unparseable→dropped) ✓
- Codex fixture with credits balance → top-level `balance` ✓
- negative balance preserved as negative (not clamped) + reports exhausted ✓
- `exhausted` for a 100% window and for balance ≤ 0 ✓
- absent balance stays `undefined` (asserted `!("balance" in snap)`) ✓
- balance-only change → different `contentKey` (re-publish) ✓

### Anti-spaghetti
Only coincidental clones in the changed files (test fixtures — codex fixture shapes, poller fixture boilerplate). No business-logic duplication introduced, no dead code left behind. Non-blocking.

### Informational notes (not Blocks/Questions — within the issue's additive scope)
- No adapter currently populates `startedAt` or `overagePrice` because none of the providers report those today. That's the intended additive state for this stage; `marginalCost` (S2c) is expected to consume the fields via its documented fallbacks.

PR #1252, opened 2026-08-20
feat(server): cache provider-agnostic model catalogue on the box (BET-1241)

antoinedc:
## Summary

Stage 5 of Automatic Manta Routing (BET-1226): cache the **provider-agnostic** model catalogue (`https://models.dev/models.json`) on the box. This is opencode's own catalogue — the `/provider` list Manta already receives IS this data — we were simply not reading the provider-agnostic view, which is what makes an opaque provider alias resolvable.

New module `src/server/modelCatalog.mjs` with three exports:

- **`lookupModel(catalogId)`** — exact, case-insensitive / separator-normalised lookup by catalogue id.
- **`matchModel(localModelId)`** — resolves an opaque local id to `{ kind: "exact" | "ambiguous" | "none", candidates }`, matched on the model id + name (+ family), normalising separators. Never guesses between candidates of materially different size — an `ambiguous` result returns all of them for the caller to ask.
- **`allModels()`** — the full entry list, for modelQuality's percentile field (owned downstream; this module only supplies the field).

Architecture (matching what `AGENTS.md` and the sibling `modelLedger.mjs` prescribe):

- **Pure matching** in `createModelIndex` (no I/O, unit-tested against a checked-in fixture). I/O lives in `createModelCatalogController` (fetch → cache → degrade).
- Uses `statePath()` from `src/shared/paths.mjs` for the cache path — under the test sandbox (`MANTA_STATE_HOME`) it resolves off the sandbox, never the live box (there is a dedicated canary test).
- Reuses **`startPoller.mjs`** (immediate first tick, inFlight guard, `timer.unref()`); never hand-rolls a timer. Default cadence 6h (large-ish download, changes rarely).
- **Degradation contract**: a fetch failure never throws and never blanks a working cache — it serves the last good copy from disk/memory, or reports `{ supported:false }` (empty catalogue → `matchModel` returns `{kind:"none"}`, `allModels` returns `[]`) when there has never been one. An absent catalogue means models are unidentifiable, which the eligibility gate already handles honestly — it must not fail a routing decision.
- `normalizePayload` accepts both the array and the id-keyed-object shapes of `models.json` so an upstream format change doesn't silently empty the catalogue.

Not wired into `index.mjs` boot yet — this stage is the cache + matcher module. `startModelCatalogPoller({...})` is exported for a later stage to arm at boot (modelQuality/routing).

**Standing-rule compliance:** no vendor names in the module; pure logic split from I/O; pure matching is unit-tested with a fixture only (no live network, no live tmux/opencode). Anti-spaghetti: reused `statePath`/`jsonStore`/`startPoller` rather than adding parallel copies.

## Files changed (3, +553/−0, on top of `origin/main`)

- `src/server/modelCatalog.mjs`
- `src/server/modelCatalog.test.mjs`
- `src/server/fixtures/modelCatalog.fixture.json`

## Verification results

- **typecheck:** exit 0, no errors. (`npm run typecheck`)
- **`src/server/modelCatalog.test.mjs`:** 14/14 pass (exact/ambiguous/none driven by the three real cases; cached copy served on fetch failure; `{supported:false}` with no cache+no network; failed refresh never blanks an in-memory copy; `startModelCatalogPoller` populates via the immediate startPoller tick; cache path resolves inside `MANTA_STATE_HOME`).
- **Full `npm test`:** the broader server suite cannot run to completion in this sandbox — `node_modules` is missing the third-party packages `node-pty` and `yaml`, producing 11 pre-existing `ERR_MODULE_NOT_FOUND` failures in files this PR does not touch (`pty.mjs`, `forgeRules.mjs`, `pluginManifest.mjs` and their dependents). These are environmental here and unrelated to this change; the real suite runs on CI.

**Base: `origin/main` @ `ffee9d05`**

### DIFF
--- a/src/server/fixtures/modelCatalog.fixture.json
+++ b/src/server/fixtures/modelCatalog.fixture.json
@@ -1,0 +1,82 @@
+[
+  {
+    "id": "qwen/qwen3.6-27b",
+    "name": "Qwen3.6-27B",
+    "family": "qwen",
+    "description": "Qwen3.6 tuned for long-context coding (exact-match fixture)",
+    "reasoning": true,
+    "tool_call": true,
+    "modalities": { "input": ["text"], "output": ["text"] },
+    "open_weights": true,
+    "limit": { "context": 131072, "output": 8192 },
+    "benchmarks": [{ "name": "SWE-Bench Verified", "score": 58.4, "metric": "resolved" }]
+  },
+  {
+    "id": "chutes/Ornith-9B",
+    "name": "Ornith 9B",
+    "family": "ornith",
+    "description": "Ornith family, 9B variant",
+    "reasoning": true,
+    "tool_call": true,
+    "modalities": { "input": ["text"], "output": ["text"] },
+    "open_weights": true,
+    "limit": { "context": 32768, "output": 4096 }
+  },
+  {
+    "id": "chutes/Ornith-31B",
+    "name": "Ornith 31B",
+    "family": "ornith",
+    "description": "Ornith family, 31B variant",
+    "reasoning": true,
+    "tool_call": true,
+    "modalities": { "input": ["text"], "output": ["text"] },
+    "open_weights": true,
+    "limit": { "context": 65536, "output": 8192 }
+  },
+  {
+    "id": "chutes/Ornith-35B",
+    "name": "Ornith 35B",
+    "family": "ornith",
+    "description": "Ornith family, 35B variant",
+    "reasoning": true,
+    "tool_call": true,
+    "modalities": { "input": ["text"], "output": ["text"] },
+    "open_weights": true,
+    "limit": { "context": 65536, "output": 8192 }
+  },
+  {
+    "id": "chutes/Ornith-397B",
+    "name": "Ornith 397B",
+    "family": "ornith",
+    "description": "Ornith family, 397B variant",
+    "reasoning": false,
+    "tool_call": true,
+    "modalities": { "input": ["text"], "output": ["text"] },
+    "open_weights": true,
+    "limit": { "context": 262144, "output": 32768 }
+  },
+  {
+    "id": "minimax/MiniMax-M3",
+    "name": "MiniMax-M3",
+    "family": "minimax",
+    "description": "MiniMax multimodal model for long-context coding, perception, and agent planning",
+    "attachment": true,
+    "reasoning": true,
+    "tool_call": true,
+    "modalities": { "input": ["text", "image", "video"], "output": ["text"] },
+    "open_weights": true,
+    "limit": { "context": 512000, "output": 128000 },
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review — BET-1241 modelCatalog

**Verdict: PASS** (0 Blocks, 0 Questions, 1 Nit)

Reviewed against the issue's acceptance criteria (cache under `statePath`, reuse `startPoller`, degrade-don't-blank, three exports, matcher split from I/O, fixture-driven tests with the three real cases + sandbox canary).

**Base-freshness:** `merge-base HEAD origin/main` == `origin/main` == `ffee9d05`. Diff is exactly the 3 intended files (+553/−0), all additive. No phantom deletions.

**Mechanical gates (run by reviewer):**
- `npm test` → **2590/2590 pass**, 0 fail.
- `node --import ./scripts/testSandbox.mjs --test src/server/modelCatalog.test.mjs` → **14/14 pass**.

Note: the PR body claims "11 pre-existing `ERR_MODULE_NOT_FOUND` failures" in the full suite. That was an artifact of an incomplete `node_modules` in the implementer's sandbox — with `node-pty`/`yaml` present (`npm install`), the **entire** suite runs green. No action needed.

**Acceptance criteria — verified:**
- ✅ Cache path via `statePath("model-catalog.json")` (`modelCatalog.mjs:41`); no hand-built `join(homedir(), …)`. Canary test asserts it resolves inside `MANTA_STATE_HOME`.
- ✅ Reuses `startPoller` (immediate tick, inFlight guard, `timer.unref()`), 6h cadence via `controller.start()`.
- ✅ Degradation: failed fetch serves last-good copy; no cache + failed fetch → `{ supported:false }`; never throws, never blanks a working cache.
- ✅ Exports `lookupModel` (exact/normalised), `matchModel` → `{kind: exact|ambiguous|none, candidates}`, `allModels` (defensive copy).
- ✅ Pure `createModelIndex` split from I/O `createModelCatalogController`; `normalizePayload` handles array + id-keyed-object shapes.
- ✅ Three real cases covered and passing: `qwen3.6-27b`→exact, `ornith`→ambiguous (9B/31B/35B/397B, never guessed between), `default`→none.

**Anti-spaghetti (jscpd):** 0 clones.

**Nit (non-blocking, for the downstream wiring stage):**
`startModelCatalogPoller` builds its own controller instance (fresh `createModelCatalogController`), so the returned handle swaps in new data while the **module-level** `lookupModel`/`matchModel`/`allModels` singletons (`boxCatalog`) are never fed by the poller — they see only whatever was on disk at module load. The docstring at `modelCatalog.mjs:231-237` claims the singletons "stay in sync," which the current code does not deliver. When the routing stage arms the poller, use the returned object as the single read surface, or drive `boxCatalog` from the poller instead of spinning up a second controller. Not blocking today (module is standalone, un-wired, all criteria met).

PR #1253, opened 2026-08-20
feat(renderer): three-state per-session model choice (auto/model/default) BET-1245

antoinedc:
## Summary

Stage 6 of the Automatic Manta Routing epic. Extends the persisted per-session model
selection to a **three-state** choice without widening `ModelSelection` (which would force
every provider-forwarding consumer to re-narrow it).

Adds to `src/renderer/chatShared.tsx`:

- `ModelChoice = { kind: "auto" } | { kind: "model", model } | { kind: "server-default" }`
- `readSavedChoice` / `writeSavedChoice` / `carrySavedChoice`

**Storage compatibility** — same single key (`manta:chat:<sid>:model`), one key three states:
- absent → `server-default`
- a `ModelSelection`-shaped JSON object → `{ kind: "model", model }`
- bare literal string `"auto"` → `{ kind: "auto" }` (written as a bare string, not JSON)
- malformed → `server-default`, never throws out of a storage read

**Replace, don't duplicate.** `readSavedModel`/`writeSavedModel` are reimplemented as thin
`@deprecated` wrappers over the new pair (the ~6 call sites are model-view consumers in
ChatPanel/App that the later ChatPanel/ModelPicker issues migrate). No second independent
reader of the key. The `/clear` carry-forward was renamed `carrySavedChoice` and now carries
the FULL choice including `Auto` — clearing a conversation no longer silently drops Auto.

**Tests** — `chatShared.test.ts`: round-trip for each of the three states, existing stored
`ModelSelection` reads as `{ kind: "model" }`, absent → `server-default`, malformed →
`server-default` (no throw), `"auto"` written/read as bare string, and `carrySavedChoice`
carries `auto` across session ids (plus explicit-model and no-op cases).

**Verification results:**
- PR branch (`multica/BET-1245-model-choice-three-state` @ `a427e824`): typecheck exit 0, none. test exit 0, 2576 pass / 0 fail / 0 skip.
- Base (`main` @ `ffee9d05`).
- Conclusion: 0 new failures.

Base: `origin/main` @ `ffee9d05`

### DIFF
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -69,7 +69,7 @@ import { isPlanAgent, planPageUrl } from "../shared/planMode.mjs";
 import { serverBase } from "./api/httpApi";
 import {
   appendPromptHistory,
-  copySavedModel,
+  carrySavedChoice,
   guessMime,
   mimeToInputMode,
   modelInputModes,
@@ -1636,10 +1636,10 @@ export function ChatPanel({
         // auto-rename first-name path and stamp a workspace-prefixed name (BET-1100).
         title: "",
       });
-      // /clear carries the session's model forward so the user doesn't re-pick
-      // after every clear; plan mode state is carried on top.
+      // /clear carries the session's model choice forward so the user doesn't
+      // re-pick after every clear (including Auto); plan mode state on top.
       if (cleared?.newSessionId) {
-        copySavedModel(sessionId, cleared.newSessionId);
+        carrySavedChoice(sessionId, cleared.newSessionId);
         if (planOn) {
           writePlanSaved(cleared.newSessionId, true);
         }
--- a/src/renderer/chatShared.test.ts
+++ b/src/renderer/chatShared.test.ts
@@ -18,7 +18,9 @@ import {
   writeSavedMode,
   readSavedModel,
   writeSavedModel,
-  copySavedModel,
+  readSavedChoice,
+  writeSavedChoice,
+  carrySavedChoice,
   resolveLauncherFlags,
   readPromptHistory,
   appendPromptHistory,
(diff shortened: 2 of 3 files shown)


antoinedc:
## Review — BET-1245 three-state per-session model choice

**Result: PASS** — 0 Blocks, 0 Questions, 1 Nit.

### Acceptance-criteria checklist (verified independently against code)

- [x] **`ModelChoice` three-state type** — exact shape from the issue added in `chatShared.tsx:254`; kept separate from `ModelSelection` so provider-forwarding consumers don't re-narrow. ✓
- [x] **Same single localStorage key** (`manta:chat:<sid>:model`) — unchanged; no second key added. ✓
- [x] **Storage compat**: absent → `server-default`; `ModelSelection`-shaped object (`providerID`+`modelID` strings) → `{kind:"model"}`; bare literal `"auto"` → `{kind:"auto"}`; malformed → `server-default`, never throws. All covered in code + tests.
- [x] **`auto` written as bare string**, not JSON object (test asserts `getItem === "auto"`).
- [x] **Replace, don't duplicate** — `readSavedModel`/`writeSavedModel` are thin `@deprecated` wrappers over the new pair (the issue's permitted fallback), no second independent reader of the key, no new callers opened.
- [x] **`carrySavedChoice` carries full choice incl. Auto** — `/clear` in `ChatPanel.tsx:1642` switched from `copySavedModel` → `carrySavedChoice`; test covers carrying `auto` across session ids.
- [x] No `sessionStorage`, no served-page-context storage, no second auto key.
- [x] **Tests** — all five required cases present (three-state round-trip, existing `ModelSelection` → `model`, absent → default, malformed → default no-throw, carry `auto`).

### Mechanical gate

- Base-freshness: branched off `ffee9d05`; `git diff --stat origin/main...HEAD` is identical to the merge-base diff — no phantom deletions, base is not stale.
- `npm run typecheck` — **exit 0**, zero errors. ✓
- `npm test` — **2576 pass / 0 fail / 0 skip** (exit 0), matching the implementer's reported PR result exactly. Trust-but-spot-check confirmed.
- Anti-spaghetti sticky: **"No duplicates found"** across the 3 changed files. ✓

### Nit (non-blocking)

- The issue's "Replace, do not duplicate" section *prefers* deleting the legacy accessors and updating the few call sites. The implementer chose the issue's own permitted fallback (thin `@deprecated` wrappers) because the ~6 call sites are model-view consumers that the later ChatPanel/ModelPicker issues migrate. Reasonable and within spec — just an intentional, documented deferral rather than the preferred option.

PR #1254, opened 2026-08-20
BET-1232: modelQuality — tier as a percentile band

antoinedc:
Opened automatically for `multica/BET-1232-modelquality`, which was pushed without a pull request.

Relates to BET-1232.

### DIFF
--- a/src/shared/modelGuide.d.mts
+++ b/src/shared/modelGuide.d.mts
@@ -8,6 +8,8 @@ export interface ModelInfo {
 
 export function tierRank(tier: string | undefined): number;
 
+export const FAMILY_TIERS: Record<string, ModelTier>;
+
 export function describeModel(
   providerID: string,
   modelID: string
--- a/src/shared/modelGuide.mjs
+++ b/src/shared/modelGuide.mjs
@@ -206,6 +206,15 @@ const CATALOG = [
   },
 ];
 
+// Family → tier, derived from CATALOG so it cannot drift. modelQuality.mjs
+// seeds its family→score table from this so already-covered families keep
+// their current tier when quality becomes percentile-based. Exposed (rather
+// than reading CATALOG directly) so the association survives the eventual
+// removal of the `tier` field from CATALOG entries.
+export const FAMILY_TIERS = Object.fromEntries(
+  CATALOG.map((e) => [e.key, e.tier]),
+);
+
 function matchFamily(modelID) {
   if (!modelID || typeof modelID !== "string") return null;
   const normalized = modelID.toLowerCase();
--- a/src/shared/modelQuality.d.mts
+++ b/src/shared/modelQuality.d.mts
@@ -1,0 +1,52 @@
+export const TIERS: string[];
+
+export const FAMILY_SEED_SCORES: Record<string, number>;
+
+export const AGENT_FLOOR_SCORE: Record<string, number>;
+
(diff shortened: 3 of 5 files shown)


antoinedc:
## Review — BET-1232: modelQuality as a percentile band

**Route: PASS** (0 Blocks, 0 Questions, 3 nits)

### What I verified

**Base freshness.** `merge-base HEAD origin/main` = `362a5d5d`, current `origin/main` = `41ba2287`. `git diff --stat origin/main...HEAD` is purely additive (5 files, +448, 0 deletions) — no stale-base phantom deletions.

**Scope vs issue.** Diff is exactly the 3 new `modelQuality.*` files plus an additive `FAMILY_TIERS` export in `modelGuide.{mjs,d.mts}`. `modelRouter.mjs` diff = 0 lines. Per the issue's parenthetical ("if the router restructure issue has not landed yet, leave the router untouched and ship this module standalone"), this is the correct call — `origin/main`'s `modelRouter.mjs` still defines `tierOf`/`AGENT_FLOOR`, confirming the restructure hasn't landed. `blurb`/`goodFor` untouched. No substring matching on model id in the new module; no fetching (catalogue entry + percentile helper are injected).

### Mechanical gates

- `npm run typecheck` → **exit 0**, no errors
- `npm test` → **2583 pass / 0 fail / 0 skip** (matches implementer's report)
- CI: `typecheck-test` **pass**, `duplication-gate` **pass**
- Anti-spaghetti (jscpd, changed scope): **0 clones**
- `vitest run src/shared/modelQuality.test.ts`: **13/13 pass**

### Acceptance checklist

- `qualityScore(model, catalogEntry, field)` — derivation order benchmark → family → structural → nothing, each `known`/`basis` correct. ✓
- Family seeded from CATALOG tier labels (fast 0.25 / balanced 0.55 / deep 0.85), family read authoritatively from the catalogue, never a regex on the id. ✓
- Structural path from reasoning / context / output / price; unknown → `known:false, score:0` (policy left to caller). ✓
- Exports `TIERS`, `tierForScore` (bands fast < 0.4 ≤ balanced < 0.7 ≤ deep), `AGENT_FLOOR_SCORE`, `meetsFloor`. ✓
- No-regression test driven from the CATALOG array via `FAMILY_TIERS` (all 18 families) — passes. ✓
- `meetsFloor` ≡ today's `AGENT_FLOOR` for all five agents. ✓
- Relativity test: introducing a stronger model shifts existing percentiles down. ✓

### Nits (non-blocking)

1. The "prefers SWE-Bench Verified over Aider Polyglot" test doesn't actually pin the cross-benchmark preference order — the test helper returns the same percentile regardless of which benchmark is picked, so it only proves a benchmark resolves to the `benchmark` path, not the ordering. The production `BENCH_PREFERENCE` array is correct, so this is a test-strength gap, not a code defect. A stronger test would give the two benchmarks different scores and assert the SWE-Bench one won.
2. `AGENT_FLOOR_SCORE.explore`/`title` = `0` — correct for "≥ fast", but callers must keep filtering `known:false` models out before `meetsFloor` (an unknown model's 0 passes). Consistent with the documented contract, worth a reminder in the consuming router issue.
3. `expect(families.length).toBeGreaterThan(10)` is a magic number; fine as a drift tripwire.

None of these affect the routing decision. The module is standalone, pure, tested, and internally consistent with the spec.

PR #1255, opened 2026-08-20
feat(routing): delete global routing toggle; settings keeps Balance (BET-1243)

antoinedc:
## What
Stage 6 of the Automatic Manta Routing epic — routing is activated **per conversation** from the composer's model picker, so the global enable toggle (a second control for one decision, and a stub — its config key was never read) is deleted.

### Changes
- **`src/shared/settingsSchema.ts`** — deleted the `modelRoutingEnabled` entry; renamed the preset's `group` from `Routing` → `Automatic Manta Routing` (the Models section now renders **one** Balance card, via the existing `renderSection` grouping).
- **`src/shared/types.ts`** — removed `modelRouting.enabled`; added the user-declared `declaredModels` block per the spec.
- **`src/main/config.ts` / `src/server/local.mjs`** — dropped `enabled` from `DEFAULT_CONFIG.modelRouting`; kept `preset: "balanced"`.
- **`src/renderer/Settings.tsx`** — removed `modelRouting.enabled` from the values map.
- **`src/renderer/ModelRoutingCard.tsx` (+ test)** — the BET-1222 card no longer depends on the deleted global `enabled` flag (there is no global off-state; routing is per-conversation). Removed the greyed "Routing is off" branch; it now always shows the active preset caption / near-spent warn.
- **`src/shared/settingsSchema.test.ts`** — replaced the "toggle defaults to FALSE" test with an assertion that the entry no longer exists; updated group + platform assertions; kept the preset↔`PRESETS` no-drift check.

Migration: none — an old config carrying `modelRouting.enabled` is simply ignored.

## Verification
- `npm run typecheck` — green
- `npm test` — 2597 pass, 0 fail
- `grep -rn 'modelRoutingEnabled|modelRouting.enabled' src/` — only the deliberate test assertion remains

## Note for review
The richer `ModelRoutingCard` (BET-1222) already lands in the Models section and shows what routing *would* do (per-agent tiers + plan windows). It is a separate, informative card from the sibling ticket — I kept it, removing its dependence on the now-deleted global `enabled` flag (there is no longer a global off-state, since routing is activated per conversation). "Balance alone" refers to the schema-driven Balance group card.

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -9,7 +9,7 @@ const DEFAULT_CONFIG: AppConfig = {
   // Model routing (BET-1215). Off by default (opt-in — routing acts on the
   // user's behalf). The live server default lives in src/server/local.mjs;
   // this mirrors it for desktop-side config reads.
-  modelRouting: { enabled: false, preset: "balanced" },
+  modelRouting: { preset: "balanced" },
 };
 
 function configPath(): string {
--- a/src/renderer/ModelRoutingCard.test.tsx
+++ b/src/renderer/ModelRoutingCard.test.tsx
@@ -2,8 +2,9 @@
 //
 // ModelRoutingCard — the Routing card in Settings → Models (BET-1222). Tiers
 // must be read from the router module (never a literal), plan usage comes from
-// the shared store (the same source the composer dial uses), and the card must
-// stay visible — greyed — when routing is off.
+// the shared store (the same source the composer dial uses), and the card shows
+// what routing *would* do — there is no global off-state (BET-1243 removed the
+// toggle; routing is activated per conversation from the model picker).
 
 import { describe, it, expect, beforeEach, afterEach } from "vitest";
 import { mount, type Harness } from "./testHarness";
@@ -35,7 +36,7 @@ describe("ModelRoutingCard", () => {
   let h: Harness | null = null;
 
   beforeEach(() => {
-    useStore.setState({ usage: [], modelRouting: { enabled: true, preset: "balanced" } });
+    useStore.setState({ usage: [], modelRouting: { preset: "balanced" } });
   });
 
   afterEach(() => {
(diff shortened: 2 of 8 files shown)


antoinedc:
## Review — BET-1243: delete global routing toggle; Settings keeps Balance alone

**Verdict: PASS** — 0 Blocks, 0 Questions.

### Verification (run on PR branch @ `a0c771b3`, base `9d0633a8`)

- `npm run typecheck` → **green** (exit 0, both tsconfigs).
- `npm test` → **2597 pass / 0 fail / 0 skip** (exit 0). Matches the implementer's reported results exactly.
- `grep -rn "modelRoutingEnabled\|modelRouting.enabled" src/` → only match is the **deliberate test guard** at `src/shared/settingsSchema.test.ts:137` (`expect(ALL.find((e) => e.id === "modelRoutingEnabled")).toBeUndefined()`). Fixed-string `modelRouting.enabled` returns **nothing** across `src/`.
- Base freshness: `git diff --stat origin/main...HEAD` shows only the 8 in-scope files as modifications — **no phantom deletions**, branch is clean off `origin/main`.

### Acceptance-criteria check

1. **typecheck + test green** — verified (above).
2. **Settings → Models renders a card titled "Automatic Manta Routing" with a single Balance row (Economy/Balanced/Performance chip group)** — `src/shared/settingsSchema.ts` deletes `modelRoutingEnabled` and renames the preset `group` → `Automatic Manta Routing`, leaving `label`/`help`/`options`/`default: "balanced"` untouched. The card is data-driven via the existing `renderSection` grouping + `SettingsRow`/`ChipGroup`, and the schema test pins the group name and the Economy/Balanced/Performance options. No new component change was required (as the issue stated).
3. **No remaining references to the deleted toggle** — satisfied: every non-test reference is gone (`settingsSchema.ts`, `types.ts`, `config.ts`, `local.mjs`, `Settings.tsx`, `store.ts`). The only residual string is the deletion-guard test the issue itself mandated.

### Findings

**Block:** none.

**Question:** none.

**Nit / note (informational, not actionable):** The issue's step 2 and its acceptance-criterion 3 are internally in tension — the mandated test (`expect(ALL.find((e) => e.id === "modelRoutingEnabled")).toBeUndefined()`) necessarily contains the `modelRoutingEnabled` string, so a strictly literal `grep` can never "return nothing". The implementer followed step 2 verbatim, which is the right reading; the intent (no production/config/UI code references the toggle) is fully met. Not a defect.

**Scope note:** The ModelRoutingCard (BET-1222) changes are load-bearing, not creep — removing `enabled` from the type broke `modelRouting?.enabled`, so the card had to drop the greyed off-state to keep `typecheck` green. Correct call.

**Pre-existing/deferred (already tracked):** `src/server/delegate.mjs` still falls back to `{ enabled: false }` when no `modelRouting` config is present. This is the runtime router wiring — out of this deletion issue's scope (the issue is config/UI/schema/types only and explicitly defers behaviour change to the router-wiring issue). The implementer correctly filed this as follow-up **BET-1251** rather than expanding the PR. No action needed here.

**Duplication:** anti-spaghetti report — 0 clones across all 8 changed files.

PR #1259, opened 2026-08-20
feat(server): account-reader descriptors — grow the supported-provider list without code (BET-1239)

antoinedc:
Opened automatically for `multica/BET-1239-account-descriptors`, which was pushed without a pull request.

Relates to BET-1239.

### DIFF
--- a/src/server/accountDescriptors/openrouter.json
+++ b/src/server/accountDescriptors/openrouter.json
@@ -1,0 +1,7 @@
+{
+  "id": "openrouter",
+  "providerIDs": ["openrouter"],
+  "url": "https://openrouter.ai/api/v1/credits",
+  "auth": "bearer",
+  "balance": { "path": "credits", "units": "dollars", "sign": "positive-is-credit" }
+}
--- a/src/server/accountDescriptors/openrouter.json
+++ b/src/server/accountDescriptors/openrouter.json
@@ -3,5 +3,13 @@
   "providerIDs": ["openrouter"],
   "url": "https://openrouter.ai/api/v1/credits",
   "auth": "bearer",
-  "balance": { "path": "credits", "units": "dollars", "sign": "positive-is-credit" }
+  "balance": { "path": "data.total_credits", "units": "dollars", "sign": "positive-is-credit" },
+  "windows": [
+    {
+      "kind": "credit-limit",
+      "label": "Credits",
+      "used": "data.total_usage",
+      "limit": "data.total_credits"
+    }
+  ]
 }
--- a/src/server/accountReaders.mjs
+++ b/src/server/accountReaders.mjs
@@ -1,0 +1,132 @@
+// accountReaders.mjs — load the SHIPPED account descriptors and present them as
+// ordinary usage readers (BET-1239).
+//
+// A descriptor is data that declares how to read a credit-based account: the
+// URL, the bearer key to send, and which dot-paths in the response become a
+// UsageSnapshot (balance, windows, plan label, overage price). Adding a
+// supported provider is authoring a JSON file in ./accountDescriptors/ — NOT
+// writing an adapter.
+//
+// The readers produced here are indistinguishable from the code adapters in
+// ./usageAdapters/: they expose the same `{id, providerIDs, detect, fetch}`
+// interface, so src/server/usage.mjs's ADAPTERS registry holds both and the
+// engine never branches on how a reader was made. The descriptor validation
+// and payload mapping live in the pure shared module
+// (../shared/accountDescriptor.mjs); this module is only the I/O + loading
+// half (read the directory, validate each on load, log an invalid one BY NAME,
+// never throw past an individual bad file).
+//
+// A descriptor that fails to fetch or validate leaves that provider with NO
+// account state — it must never make the router believe something false. There
+// is deliberately no fallback path, no probe of conventional URLs: a connected
+// provider with no reader and no descriptor simply has no account state.
+//
+// SECURITY: `readProviderApiKey` resolves a LIVE credential. It is used only
+// inside the reader's detect()/fetch() to build the Bearer header; it is never
+// logged and never folded into an error message.
+
+import { readdirSync, readFileSync } from "node:fs";
+import { join, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
+import { readProviderApiKey } from "./opencode.mjs";
+import { readDescriptor, validateDescriptor } from "../shared/accountDescriptor.mjs";
+import { httpError } from "./usageAdapters/httpError.mjs";
+
+const DESCRIPTORS_DIR = join(dirname(fileURLToPath(import.meta.url)), "accountDescriptors");
+
+/**
+ * The opencode providerID whose key this descriptor's account is authenticated
+ * with. A descriptor may cover several opencode providerIDs that share one
+ * credit pool, but the credential opencode holds lives under ONE auth-store
+ * entry — the first providerID is that entry.
+ * @param {object} descriptor
+ * @returns {string}
+ */
+function authProviderID(descriptor) {
+  return descriptor.providerIDs[0];
+}
+
+/**
+ * Present one validated descriptor as an ordinary usage reader, exactly like a
+ * code adapter: `{id, providerIDs, detect, fetch}`.
+ *   - detect() = is the bearer key already held (reused from opencode.mjs)?
+ *   - fetch()  = one HTTPS GET with `Authorization: Bearer <key>`, then
+ *     readDescriptor. A non-2xx reuses the shared `httpError` shape so the
+ *     usage poller's existing quarantine/carry-forward handles it identically
+ *     to the code adapters (a 500 leaves the provider with NO account state).
+ * @param {object} descriptor a validated descriptor
+ * @returns {{ id: string, providerIDs: string[], detect: Function, fetch: Function }}
+ */
+export function readerFromDescriptor(descriptor) {
+  const providerId = authProviderID(descriptor);
+  const readKey = () => readProviderApiKey(providerId);
+  return {
+    id: descriptor.id,
+    providerIDs: descriptor.providerIDs,
+
+    async detect({ readKey: inject = readKey } = {}) {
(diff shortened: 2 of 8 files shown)


antoinedc:
## Review — BET-1239 account-reader descriptors (PR #1259)

**Mechanical gates**
- `npm run typecheck` → PASS (exit 0, both tsconfigs).
- `npm test` → PASS: 3363 passed, 7 skipped, 178 files (full vitest + node server suite). Earlier disk-full errors were transient ENOSPC during transform; cleared caches and re-ran clean.
- Base freshness: branch merges from `413e7677`, origin/main is `219e5217`, but `git diff --stat origin/main...HEAD` is purely additive and in-scope — no phantom deletions → no stale-base issue.
- Anti-spaghetti: only pre-existing test-fixture clones inside `usage.test.mjs`; no duplication of new business logic → not a finding.

**Acceptance-criteria checklist (against the issue body)**
- Part 1 `src/shared/accountDescriptor.mjs` — validator fails by name on unknown/missing keys, never throws; `readDescriptor` maps to `UsageSnapshot`. All present. ✓
- Sign conventions — negative under `positive-is-credit` → `exhausted: true`; `positive-is-debt` inverts; missing balance stays `undefined` not 0. Implemented and covered by tests. ✓
- Part 2 `src/server/accountReaders.mjs` — loads shipped descriptors, validates on load, logs invalid **by name**; presents each as `{id, providerIDs, detect, fetch}`; `detect` reuses `readProviderApiKey`; `fetch` = one GET + `readDescriptor`, non-2xx reuses shared `httpError`; a failed/500 fetch leaves no account state. ✓
- Only supported providers read; no fallback probe; no provider-name branch in `usage.mjs` (registry only). ✓
- Do-nots respected: no user-facing extension point, no heuristic probe, no branch by name. ✓
- Tests all present and passing: `accountDescriptor.test.ts` (12), `accountReaders.test.mjs` (interface shape, 500 leaves no state, invalid logged by name), `usage.test.mjs` registry updated to expect the shipped descriptor. ✓

**Block (fix-here): shipped worked-example descriptor is not verified against a real account.**
The issue specifies the worked example "pick one credit-based provider, write its descriptor, and **verify it against a real account**" — its stated purpose is "to prove the mechanism end to end." Your own summary concedes no live OpenRouter credential was available, so the descriptor was never run against a real account. The unverified parts matter for routing: `providerIDs: ["openrouter"]` is the exact field the composer dial uses to match a model to a reading — if opencode's real providerID differs, the shipped descriptor is inert and matches nothing. Same for `url` (`/api/v1/credits`) and the `credits` balance path. This is precisely the unverified-data risk the issue's "no partial credit" rule is built to exclude, and the real-account check is the load-bearing acceptance criterion meant to catch it.

Fix path (stays inside this PR, no scope expansion): obtain a real OpenRouter credential and run the reader end to end — first confirm whether the box's opencode auth store already holds one via `detect()`; if a live key isn't available here, surface that explicitly and either (a) get one (the human / `macos` side) to complete the live verification, or (b) get the issue's verification criterion formally re-scoped rather than defaulting it away. Confirm `url`, balance `path`, and `providerIDs` against the real response before this descriptor is relied on in routing.

(Minor note, not a Block: the PR description is the auto-sweep boilerplate rather than a requirement/test-results block — I reviewed against the issue body, which is authoritative, so no functional gap.)

**Summary: 1 Block (fix-here), 0 Questions, 0 Nits otherwise.**

antoinedc:
## Review — BET-1239 account-reader descriptors (PR #1259)

**Outcome: BLOCK (fix-here).** The previously-unresolvable acceptance criterion — *verify the shipped worked-example descriptor against a real account* — is now testable, because a real `OPENROUTER_API_KEY` is available. It fails.

### Mechanical gates
- `npm run typecheck` → **PASS** (exit 0, both tsconfigs).
- `npm test` → **PASS** (0 failures; observed 2602 pass / 0 fail / 71 suites on this box). Both green.
- Base freshness: clean — diff vs `origin/main` is purely additive (new descriptor + readers + shared validator + tests), edits only `usage.mjs` + its tests. No phantom deletions.

### Anti-spaghetti
4 clones, all coincidental test-fixtures (repeated "absent-credits → balance stays undefined" assertions in `usage.test.mjs`). Not business logic introduced by this PR. Discounted.

### The mechanism itself is sound
`accountReaders.mjs` (loader + `readerFromDescriptor`), the pure `accountDescriptor.mjs` validator/`readDescriptor`, and the sign conventions are well-constructed and match the issue's spec exactly — unknown-key-per-name validation, never-throws, negative-balance-under-`positive-is-credit` → `exhausted`, absent balance → `undefined` not 0, injected-fetch test coverage. The registry integration in `usage.mjs` adds readers without provider-name branching. Good.

### The problem: the shipped worked-example descriptor does not read a real account

**Block (fix-here):** `src/server/accountDescriptors/openrouter.json` declares `"balance": { "path": "credits", ... }`, but the real OpenRouter `/api/v1/credits` endpoint (verified live, HTTP 200 with a valid key) does **not** return a top-level `credits` field:

```json
{ "data": { "total_credits": 20, "total_usage": 20.071752339 } }
```

`readDescriptor` resolves `credits` → `undefined` → the snapshot has **no balance** → OpenRouter gets **no account state**. The one provider this issue exists to prove end-to-end is exactly the one the shipped descriptor cannot read. That defeats the worked-example's stated purpose.

The correct mapping (confirmed against the live payload) is under `data`:
- `balance.path` → `data.total_credits` (dollars, `positive-is-credit`).

**One nuance the fix should not miss.** On this live account `total_usage` (20.07) already exceeds `total_credits` (20) — the account is *genuinely overdrawn*. A naive `data.total_credits` read would report `balance: 20` (not exhausted) on a depleted account, which is precisely the wrong-reading failure the issue warns about. To capture that, map the overrun through the `windows` mechanism the descriptor already supports, e.g. `windows: [{ used: "data.total_usage", limit: "data.total_credits" }]` — `used/limit = 100.35% ≥ 100` → `atWindowLimit` → `exhausted: true`. Verify whichever mapping you choose against this live account.

**Required fix:** correct the OpenRouter descriptor so that running `readDescriptor` against the real endpoint returns a real, correct balance (and marks exhaustion correctly), and confirm the result against the live account before resubmitting. Please re-verify with the live key in your test-results block.

antoinedc:
**Files changed:** 2. `src/server/accountDescriptors/openrouter.json`, `src/server/accountReaders.test.mjs` (+1 regression test).

**What changed:** the shipped OpenRouter descriptor read a nonexistent top-level `credits` field. The live `/api/v1/credits` endpoint returns `{ data: { total_credits, total_usage } }`, so `readDescriptor` resolved no balance and OpenRouter got no account state. Fixed `balance.path` → `data.total_credits` and added a `windows` entry (`used: data.total_usage`, `limit: data.total_credits`) so an overdrawn pool (total_usage > total_credits) trips `exhausted` via `pct >= 100` instead of showing a healthy positive balance.

**Real-account verification (2026-08-20, with the `OPENROUTER_API_KEY` secret):**
- `GET https://openrouter.ai/api/v1/credits` → `200 {"data":{"total_credits":20,"total_usage":20.071752339}}`
- Overdrawn snapshot (usage > credits): `balance: 20`, `windows[0].pct: 100`, `exhausted: true` ✓
- Healthy snapshot (usage 5): `windows[0].pct: 25`, no `exhausted` ✓

**Verification results:**
- PR branch (`multica/BET-1239-account-descriptors` @ `bf65b560`):
  - test: exit 0, 2603 pass / 0 fail / 0 skip. Failures: none.
- Conclusion: 0 new failures; the new regression test pins the live OpenRouter payload so the balance-path + exhaustion mapping cannot silently regress.

antoinedc:
## Review — BET-1239 account-reader descriptors (PR #1259)

**Outcome: PASS.** The blocking criterion is resolved and re-verified independently.

### Mechanical gates
- `npm run typecheck` → **PASS** (exit 0, both tsconfigs).
- `npm test` → **PASS**: 2603 pass / 0 fail / 71 suites (matches the implementer's reported 2603 — no discrepancy).
- Base freshness: clean — diff vs `origin/main` purely additive, no phantom deletions.

### The previously-blocking criterion (real-account verification) — now verified, by me, independently
`openrouter.json` now reads `balance.path: data.total_credits` and maps the credit pool via `windows` (`used: data.total_usage`, `limit: data.total_credits`). I re-ran `readDescriptor` against the corrected descriptor with the real `OPENROUTER_API_KEY`:

- Live endpoint `GET /api/v1/credits` → `200`, `{ data: { total_credits: 20, total_usage: 20.07 } }` — matches the descriptor's assumed shape.
- **Overdrawn (real live values):** `balance: 20`, `windows[0].pct: 100`, `exhausted: true`.
- **Healthy (usage 5):** `pct: 25`, no `exhausted`.

The sign/exhaustion semantics are exactly right, and the regression test pins this live payload through the shipped descriptor so it can't silently regress.

### Acceptance criteria — all satisfied
- Pure `validateDescriptor` / `readDescriptor` in `src/shared/accountDescriptor.mjs` with per-name error reporting, never throws. ✔
- Sign conventions tested (negative under `positive-is-credit` → `exhausted`; `positive-is-debt` inverts). ✔
- `accountReaders.mjs` loader: validates on load, logs invalid descriptors **by name**; never throws past a bad file. ✔
- Readers expose the same `{id, providerIDs, detect, fetch}` interface as code adapters; `detect()` reuses `readProviderApiKey`; `fetch()` = one GET + `readDescriptor`, reusing the shared `httpError` shape. ✔
- Failures leave the provider with **no account state** (500 → no throw, quarantine via poller — tested). ✔
- No user-facing extension point, no heuristic probe, no provider-name branching in `usage.mjs` (registry only). ✔
- Shipped worked-example descriptor verified end-to-end against a real account. ✔
- Test matrix in `accountDescriptor.test.ts` + `accountReaders.test.mjs` covers every listed case, plus the live-payload regression. ✔

### Anti-spaghetti
Clones are test-fixture coincidences (repeated "absent-credits" assertions), not business logic introduced by this PR. Discounted.

No Blocks, no Questions. Routing to `manta-pm` for merge.

PR #1269, opened 2026-08-20
feat(renderer): route the main conversation at routing boundaries from ChatPanel submit (BET-1248)

antoinedc:
## BET-1248: Route the main conversation at boundaries from ChatPanel submit

Stage 6 of the Automatic Manta Routing epic — the last wiring step for the main
conversation. Spawns were already routed; this routes the composer's main turn.

### What changed

- **`ChatPanel.tsx`** — the `modelOverride` seed now derives from the three-state
  choice (`modelFromChoice(readSavedChoice(sessionId))`, BET-1245/1248) instead of
  the two-value seed. Holds `routedModel` / `routedReason` as per-session state
  (reset on session change like `stepTokens`/`liveTodos`), plus ref mirrors.
  In `submit()` — and only when the session's choice is `auto` — it crosses the
  routing boundary via `crossesBoundary` (first-turn / agent-changed / constraint /
  compacted / user-requested), and when a boundary is crossed calls
  `window.api.routingChoose`. `shouldSwitch` applies the hysteresis: the incumbent
  is kept unless it fell out of the router's contention. The resolved model is
  passed to both existing send paths (`opencodePrompt` and `opencodeRunCommand`) —
  no third send path. The whole routing block is wrapped in try/catch so a routing
  failure never fails a turn.
- **`chatShared.tsx`** — new pure `modelFromChoice` (choice → override seed).
- **`src/shared/routingBoundary.mjs`** — new `boundaryPhrase(boundary)` export
  (human phrase appended to the routed reason); typed + unit-tested.
- **`ModelPicker.tsx` / `InputArea.tsx`** — forward `autoReason` + `onSelectAuto`
  so `routedReason` is surfaced in the Auto row and the user can re-pick Auto
  (which re-decides on the next turn via the user-requested boundary).

All the routing judgement lives in `routingBoundary.mjs`; ChatPanel only supplies
facts and applies the answer. Between boundaries the session keeps its model.

### Tests added

- `useSseBus.test.tsx` — ChatPanel-harness coverage for the issue's list:
  first-turn Auto → `routingChoose` once, sent on the returned model;
  follow-up no boundary → not called, same model reused; compaction → called but
  model unchanged while incumbent is in contention (hysteresis); `routingChoose`
  rejecting → turn still sends on the previous model; Auto off → never called.
- `chatShared.test.ts` — `modelFromChoice` (3 states).
- `routingBoundary.test.ts` — `boundaryPhrase`.

### Notes / follow-ups

- BET-1225's `opencodeModelRoute` initial-route effect still runs at session start
  and also writes `modelOverride`/the undo pill under Auto; BET-1248's first-turn
  boundary now supersedes that initial decision. No user-visible break, but the
  two write `modelOverride` — worth consolidating.
  Follow-up filed: BET-1255 (parked).

**Base:** origin/main @ `f5956b4f`

**Files changed:** 9. [diff --stat](f5956b4f...6c08231f)

**Verification results:**
- PR branch (`multica/BET-1248-route-main-at-boundaries` @ `6c08231f`):
  - test: exit 0, 3422 pass / 7 skip / 0 fail (renderer 181 files; server 2637). Failures: none. log sha256: `a2060bc5cdd2fd5f3bbde4029f90abc572e240e0b5be675af1b28051050fe1d6`
- Base (`main` @ `f5956b4f`):
  - test: exit 0, 3412 pass / 7 skip / 0 fail (renderer 181 files; server 2637). Failures: none. log sha256: `c23f8d4b824c5d4de48fba7c9151edcc5cc64d555e248c091b4d5eea98ba51b7`
- **Conclusion:** 0 new test failures; 10 new passing tests (5 routing, 3 modelFromChoice, 2 boundaryPhrase). No pre-existing failures on base.

### DIFF
--- a/src/renderer/ChatPanel.tsx
+++ b/src/renderer/ChatPanel.tsx
@@ -78,6 +78,8 @@ import {
   readSavedChoice,
   readSavedModel,
   writeSavedModel,
+  writeSavedChoice,
+  modelFromChoice,
   readPlanSaved,
   writePlanSaved,
   readSavedDelegateModel,
@@ -90,6 +92,7 @@ import {
   type TaskContextValue,
   type TokenUsage,
 } from "./chatShared";
+import { crossesBoundary, shouldSwitch, boundaryPhrase } from "../shared/routingBoundary.mjs";
 import { useModelCatalog } from "./modelCatalog";
 import { useAgentCatalog } from "./agentCatalog";
 import { MantaLoader } from "./MantaLoader";
@@ -434,8 +437,13 @@ export function ChatPanel({
   const [autoActive, setAutoActive] = useState<boolean>(
     () => readSavedChoice(sessionId).kind === "auto",
   );
+  // BET-1248: the override seed derives from the THREE-state choice
+  // (readSavedChoice, BET-1245) so the rest of the component is untouched:
+  // {kind:"model"} → the model, {kind:"server-default"} → the global default,
+  // {kind:"auto"} → null until the router resolves one. The routed pick is
+  // applied to the override the moment a turn runs (BET-1247).
   const [modelOverride, setModelOverride] = useState<ModelSelection | null>(() =>
-    autoActive ? null : readSavedModel(sessionId) ?? configDefaultModel ?? null,
+    modelFromChoice(readSavedChoice(sessionId), configDefaultModel),
   );
   // BET-1222: routed state for the composer pill — set when the router chose
   // this session's model (fed by the main-conversation routing wiring). The
@@ -445,6 +453,44 @@ export function ChatPanel({
     reason: string;
     incumbent: { providerID: string; modelID: string };
   } | null>(null);
+  // BET-1248: boundary-routing state for the main conversation. `routedModel`
+  // is the session's current routed endpoint — the incumbent shouldSwitch uses
+  // and the model the next Auto turn runs on; `routedReason` is the latest
+  // decision reason (with the boundary phrase appended) for surfacing. Keyed
+  // by session: reset on session change, exactly like stepTokens / liveTodos /
+  // finishByMessageId.
+  const [routedModel, setRoutedModel] = useState<ModelSelection | null>(null);
+  const [routedReason, setRoutedReason] = useState<string | null>(null);
+  // Ref mirror so submit() reads the incumbent synchronously without re-creating
+  // the callback on every routed change (the commandsRef pattern).
+  const routedModelRef = useRef<ModelSelection | null>(null);
+  useEffect(() => {
+    routedModelRef.current = routedModel;
+  }, [routedModel]);
+  // Apply a boundary-routing decision: record the incumbent + reason and, under
+  // Auto, set the active override so the pill and send paths run on it.
+  const applyRouted = useCallback((model: ModelSelection | null, reason: string) => {
+    routedModelRef.current = model;
+    setRoutedModel(model);
+    setRoutedReason(reason);
+    setModelOverride(model);
+  }, []);
+  // The agent the PREVIOUS turn ran as — drives the agent-changed boundary.
+  const lastAgentRef = useRef<string | undefined>(undefined);
+  // A compaction completed since the last submit (cache is gone → re-deciding
+  // is free). Latched from the compaction card (see the effect below the SSE
+  // bus), cleared on the next turn.
+  const justCompactedRef = useRef(false);
+  // The user re-picked Auto while a routed model may exist — the next turn
+  // re-decides (user-requested boundary). Detected as an autoActive flip.
+  const pendingAutoUserRef = useRef(false);
+  const priorAutoRef = useRef(autoActive);
+  useEffect(() => {
+    if (autoActive && !priorAutoRef.current && routedModelRef.current) {
+      pendingAutoUserRef.current = true;
+    }
+    priorAutoRef.current = autoActive;
+  }, [autoActive]);
   // Active-model providerID for the auth-error banner (BET-316). Per-session
   // override wins over the persisted default; null if neither is set. Memoized
   // on `modelOverride` (the in-memory selection, itself seeded from
@@ -583,6 +629,14 @@ export function ChatPanel({
     submitRef,
   });
 
+  // BET-1248: latch a completed compaction into `justCompactedRef` so the
+  // NEXT submit re-evaluates routing at the (now free) compacted boundary. The
+  // compaction card clears itself ~2.5s later; latching keeps the boundary
+  // valid for the next turn regardless of how fast the user reacts.
+  useEffect(() => {
+    if (compactionState?.phase === "done") justCompactedRef.current = true;
+  }, [compactionState]);
+
   // Manual dismiss of the pinned todo card — the user's escape hatch for a
   // stale list that's non-terminal. Mirrors the auto-dismiss path in submit().
   // Stable identity so it doesn't defeat ActiveTodos' React.memo.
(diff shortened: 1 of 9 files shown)


antoinedc:
## Review — BET-1248 (PR #1269)

**Verdict: 1 Block (fix-here) → returned to implementer.**

### Pre-flight / mechanical gate

- **Base freshness:** `merge-base(HEAD, origin/main) == origin/main == f5956b4f` — branched cleanly on current `main`. No stale-base phantoms; the diff is exactly the 9 files the issue scopes.
- **Typecheck:** `npm run typecheck` → exit 0, zero errors. ✔ matches implementer report.
- **Tests (run on PR branch `6c08231f`):** renderer 3422 pass / 7 skip / 0 fail (181 files); server 2637 pass / 0 fail. Matches the implementer's reported results exactly — no fabrication, no pre-existing failures.
- **Anti-spaghetti:** jscpd found clones only in test-harness fixtures (`useSseBus.test.tsx` — a `useSseBus({...})` call block reused across describe blocks). No duplicated business logic introduced. No duplication Block.

1. **Three-state choice seed** — `modelFromChoice(readSavedChoice(sessionId), configDefaultModel)` replaces the old two-value seed; `{auto}` → `null` until routed. ✔
2. **Route at boundaries only** — `submit()` calls `crossesBoundary` only when the session choice is `auto`; the `routed {model,reason}` state is per-session and reset on session change like `stepTokens`/`liveTodos`. ✔
3. **Boundary sources wired** — first turn / after `/clear` (via `hasRoutedModel===false`), agent changed (`lastAgentRef`), compacted (latched), user re-picked Auto (`pendingAutoUserRef` / `onSelectAuto`). ⚠️ **Constraint — see Block below.**
4. **Send the routed model** — both `opencodePrompt` and `opencodeRunCommand` receive the routed model; no third send path. ✔
5. **Never fail a turn on routing** — routing block wrapped in try/catch, falls through to previous routed model / server default. ✔
6. **Never mid-exchange** — routing runs in `submit()` only; the queued drain goes through `submit`. ✔
7. **No per-turn classifier, no duplicated boundaries** — judgement lives in `routingBoundary.mjs`. ✔

### Block (fix-here): the CONSTRAINT boundary can never force a switch

The two-step `crossesBoundary` → `shouldSwitch` contract in `routingBoundary.mjs` is set up so the *boundary detector* and the *switch decision* both model capability/health: `crossesBoundary` emits `BOUNDARY.CONSTRAINT` precisely when the incumbent has `contextOutgrown` a modality is missing or its provider is unhealthy, and `shouldSwitch` has `incumbentStillCapable` / `incumbentHealthy` inputs whose whole job is to force a switch off a genuinely-incapable incumbent.

The call site in `ChatPanel.tsx` hardcodes all three switch flags to `true`:

- `crossesBoundary({ … incumbentHealthy: true … })` (line 1343)
- `shouldSwitch({ … incumbentStillCapable: true, incumbentHealthy: true … })` (lines 1376–1378)

Consequence: when a long conversation's context outgrows the incumbent model's `limit.context` (or the turn needs a modality the incumbent lacks), `crossesBoundary` returns `CONSTRAINT`, the router is asked, but `shouldSwitch` replies `switch:false` as long as the incumbent is still in the router's top-N — so the turn is sent on the same too-small model. The constraint boundary is therefore a dead path at the switch step: it burns a `routingChoose` call and then keeps exactly the model it was meant to eject. The issue's own constraint spec ("`latestTokens`/`stepTokens` vs the resolved model's `limit.context`", "`readyAttachments` modalities vs its capabilities", "provider health from the decision result") is not delivered at runtime.

This is the load-bearing anti-failure of the whole design — a session that crosses into a too-long context and is silently kept on the model that can't hold it. The fix is localized to this PR: derive the capability/health state from the same inputs used for `crossesBoundary` (or return it from `crossesBoundary`) and pass `incumbentStillCapable` / `incumbentHealthy` to `shouldSwitch` accordingly, so a CONSTRAINT that began because the incumbent genuinely no longer fits actually switches off it (while a pure price-edge with a still-fit incumbent still respects hysteresis). Add one test: constraint (context outgrown) → router returns incumbent in top-N → still switches to the capable alternative.

### Nits (non-blocking)

- The provider-health input is wired as a hardcoded `true` on both steps; the issue lists "provider health from the decision result" as a constraint input. When you fix the Block, thread the real health/capability through rather than the literal `true`.

**Block: 1 · Question: 0 · Nit: 1**

Not mergeable until the constraint boundary actually switches. The rest of the implementation is clean, well-typed, and matches the issue's test list exactly.

antoinedc:
Addressed the review Block on the selected diff.

**Block (fix-here):** the CONSTRAINT boundary could never force a switch — `shouldSwitch` was called with `incumbentStillCapable: true` / `incumbentHealthy: true`, so after a context-outgrown / modality-missing / unhealthy boundary the incumbent was retained whenever it stayed in the router's top-N (defeating the boundary's purpose).

**Fix (`9dd6b64e`):**
- `crossesBoundary` now computes and returns the capability/health facts (`stillCapable`, `stillHealthy`) as part of its result — the single copy of the constraint logic, per the "all judgement lives in routingBoundary" rule.
- `ChatPanel` threads those into `shouldSwitch` (`incumbentStillCapable: stillCapable`, `incumbentHealthy: stillHealthy`). A CONSTRAINT now forces a switch off an incapable/unhealthy incumbent even while it ranks in top-N; non-constraint boundaries keep the incumbent's contention-only behavior.
- Also fixed a latent staleness: `submit` reads the current context size via a `latestTokensRef` mirror (it can't take `latestTokens` as a dep — it's declared after `submit` — and a stale closure would miss context-outgrown). 

**Test added:** `routingBoundary.test.ts` — "a context-outgrown incumbent STILL SWITCHES while in the top-N (the Block)": wires `crossesBoundary` → `shouldSwitch` and asserts an outgrown incumbent ranked #1 still switches to a capable alternative. Plus `stillCapable`/`stillHealthy` per-constraint-kind assertions.

**Verification (branch @ `9dd6b64e`):** typecheck exit 0 (log `adab58da…`); renderer 3428 pass / 7 skip / 0 fail + server 2637 pass / 0 fail (log `a5bf0220…`).

antoinedc:
## Review — BET-1248 (PR #1269) — re-check after Block

**Verdict: PASS (0 Blocks, 0 Questions, 1 Nit). Routing to manta-pm.**

### Block resolution

The cycle-1 Block — CONSTRAINT boundary could never force a switch because `shouldSwitch` received hardcoded `incumbentStillCapable: true` / `incumbentHealthy: true` — is resolved. `crossesBoundary` now returns `stillCapable` / `stillHealthy` facts (single copy of the constraint logic), and `ChatPanel submit()` threads them into `shouldSwitch`. The new test "a context-outgrown incumbent STILL SWITCHES while in the top-N (the Block)" pins exactly the regression I flagged: incumbent ranked #1 but incapable → `switch: true, why: "incumbent-incapable"`. A bonus `latestTokensRef` mirror fixes a latent stale-closure on context size.

### Mechanical gate (re-run on updated PR `9dd6b64e`)

- **Base freshness:** `merge-base == origin/main == f5956b4f` — still branched clean; diff remains the 9 issue-scoped files (515+/35−).
- **Typecheck:** exit 0, zero errors. ✔
- **Tests:** renderer 3428 pass / 7 skip / 0 fail (181 files); server 2637 pass / 0 fail. +6 tests vs the prior review cycle (the capability/health coverage). No failures.

### Acceptance checklist (re-verified)

1. Three-state choice seed via `modelFromChoice` ✔
2. Route at boundaries only; per-session `routed {model,reason}` reset on session change ✔
3. All boundary sources wired and now effective: first turn / after `/clear`, agent changed, **constraint (context-outgrown, modality, health) now actually switches**, compacted, user re-picked Auto ✔
4. Routed model sent through both existing paths, no third path ✔
5. Routing wrapped in try/catch — a failure never fails a turn ✔
6. Only `submit()` routes; queued drain goes through `submit` ✔
7. Judgement confined to `routingBoundary.mjs`; no per-turn classifier ✔
8. No duplication of business logic (jscpd flags only test-fixture clones) ✔

### Nits (non-blocking)

- The call site still passes `incumbentHealthy: true` into `crossesBoundary`, so provider-health-driven switching is wired but dormant until a real health signal is available at the call site. Plumbing exists; not a merge blocker.

**Block: 0 · Question: 0 · Nit: 1** — approving.

PR #1273, opened 2026-08-20
diag(ios): subagent scroll — real-child range + real-pan measurement (BET-1257)

antoinedc:
Opened automatically for `multica/BET-1257-subagent-scroll`, which was pushed without a pull request.

Relates to BET-1257.

### DIFF
--- a/mobile/native/MantaUI/ChatScreen.swift
+++ b/mobile/native/MantaUI/ChatScreen.swift
@@ -1209,7 +1209,21 @@ struct ChatSubagentScreen: View {
             bottomInset: 0,
             scrollPosition: $scrollPosition,
             onPointsFromBottom: nil,
-            header: { EmptyView() }
+            // BET-1257 — a literally-empty `EmptyView()` header broke touch
+            // delivery to the transcript on THIS screen specifically: the
+            // child is reached via `NavigationStack.navigationDestination`
+            // (a push), while the parent's identical `TranscriptListView` is
+            // the stack's root. Measured with real HID input (idb, clean
+            // single-process room): with `EmptyView()` the collection view's
+            // pan gesture recognizer never even reaches `.began` — zero pan
+            // events under five real swipes, confirmed on a screenshot that
+            // the transcript never moved. Swapping in ANY non-empty
+            // `.safeAreaBar` content (a 0.5pt clear spacer — no visual
+            // change) restored a full pan trace (Began/Changed/Ended,
+            // hundreds of points of real movement) with no other change.
+            // The parent, which has always passed real header content here,
+            // is the control and is untouched by this file.
+            header: { Color.clear.frame(height: 0.5) }
         )
         .background(tokens.canvas.ignoresSafeArea())
         .onAppear { store.start() }


antoinedc:
## Review — BET-1257 (PR #1273: diag — real-child range + real-pan measurement)

Reviewed against the issue (BET-1257) acceptance criteria, not the implementer's summary. Branch is branched off `main` cleanly (no phantom deletions); base-freshness OK. no TS changes → anti-spaghetti scan N/A.

The diagnostic content itself is good, honest work: the `setContentOffset` tautology from BET-1211 is gone, the scrollable-range number (`min/max/range`) is now logged per scroll view, and the child's **range=17,875pt** vs the parent's **16,016pt** control is exactly the evidence §4.3 asked for. That confirms **Branch B** — the geometry is healthy, the defect is in touch delivery to the pushed child screen, and the `PanProbe` trace (`child` emits zero `pan` lines while the parent emits Began→Changed→Ended) is the correct discrimination. None of that is disputed here.

The problem is the PR's relationship to the ticket, not the quality of the measurement.

---

**Block (fix-here): the core deliverable — the scroll fix — is absent; no acceptance criterion is met.**

§8 Verification requires: *"The child scrolls both directions under a finger, and still follows a streaming child to the tail."* This PR contains **zero fix** — the diff is entirely `#if DEBUG` instrumentation (`ScrollDiagnostic.swift`, `PanProbe`, the `manta://debug/*` routes). The implementer's own comment states the fix is still the "next step." Posting the §4 numbers is a required **checkpoint** (§4.3: "post the numbers… **before proposing a change**"), not the finished deliverable. As of this PR, a finger still cannot scroll the child. Under-delivery vs. the issue's stated scope.

**Block (fix-here): the diff is essentially all probe code, which the ticket requires to be removed before merge.**

§7: "All probe code is `#if DEBUG` and is removed before merge." §8: "Probe code removed; diff is the fix only." As submitted, ~285 of ~291 added lines are diagnostic scaffolding — this cannot merge even if the numbers are accepted.

**Block (fix-here): `MARKETING_VERSION` bumped 1.0.17 → 1.0.18.**

`project.pbxproj` (both build configs). §0 explicitly forbids this: *"Do **not** bump `MARKETING_VERSION`."* This is a CI/release concern owned elsewhere; it does not belong in a `#if DEBUG` diagnostics branch being swept into review.

---

**Expected continuation (in this same branch/PR):**
1. Keep the healthy-range finding (Branch B) as established; do **not** re-derive it.
2. Design and land the touch-delivery fix for the pushed child (the parent-vs-child difference to chase remains `header: { EmptyView() }` in `.safeAreaBar(edge: .top)` and `bottomInset: 0` on the child surface — do **not** change the parent's behaviour, do **not** ship a magic constant inset).
3. Verify under one real pan that the child scrolls both directions and still follows a streaming child to the tail; confirm the **parent's** probe numbers are unchanged from the step-1 baseline (the control).
4. **Remove all probe code** (`ScrollDiagnostic`, `PanProbe`, debug routes); the merged diff is the fix only.
5. **Revert the `MARKETING_VERSION` bump** back to 1.0.17.

Not a merge candidate until the fix + control-verification land and the probe code / version bump are gone.

antoinedc:
## Review — BET-1257 (PR #1273, fix `be3a2d6`)

Reviewed against the issue (BET-1257) acceptance criteria, not the implementer's summary. This is a re-review after the branch was force-updated from diagnostics-only to the actual fix.

### Pre-flight
- **Base freshness:** `merge-base(pr1273, origin/main)` = `2e3db2a7` = current `origin/main` tip exactly. Branched off fresh `main`; no phantom deletions. ✓
- **Diff scope:** `git diff origin/main...pr1273` touches exactly one file — `mobile/native/MantaUI/ChatScreen.swift`, 15 insertions / 1 deletion (14 lines are the explanatory comment). No `ScrollDiagnostic.swift`, no `PanProbe`, no `manta://debug/*` routes, no `project.pbxproj`. ✓
- **Anti-spaghetti:** no `.ts/.tsx/.mjs` changed → scan N/A (CI comment confirms 0). ✓
- **Mechanical gate (`ios-mantaui action=test-unit`):** Mac-only; `macos` reports TEST SUCCEEDED, 592 tests / 0 failures on this commit (`be3a2d6`). Not independently re-runnable on this Linux box; the human relayed the same result. Accepted as reported.

### The fix
Swap the child's `header: { EmptyView() }` → `header: { Color.clear.frame(height: 0.5) }` on `ChatSubagentScreen` (ChatScreen.swift:1226). The parent call site (`sessionHeaderBlock`, line 766) is byte-identical / untouched — the control is intact.

### Acceptance-criteria check (§8)
- **Child scrolls both directions under a finger** — ✅ **human-confirmed on a real iPhone** (2026-08-21) with `be3a2d6` installed. This is exactly the strongest possible verification for a touch-delivery bug, and it directly satisfies the primary §8 criterion.
- **Parent control unchanged** — ✅ parent's probe trace stays healthy (132 pan events, ~2,109pt real movement) and the file isn't touched by the diff. The fix cannot move the parent's numbers.
- **Probe code removed; diff is fix only** — ✅ verified above; no `#if DEBUG` instrumentation remains.
- **`MARKETING_VERSION`** — ✅ untouched at 1.0.17 (both Blocks from the previous cycle on this point are resolved).

### Constraints (§7)
Row identity untouched; parent behaviour unchanged; child stays read-only (no composer added); one-file change; probe code removed before merge. All respected.

### Classification: 0 Blocks, 0 Questions

The diagnostic is sound and the fix matches the parent/child difference the workflow itself converged on (child passed a literally-empty `EmptyView()` in `.safeAreaBar(edge: .top)` where the parent passed real header content). The implementer is honest that the *mechanism* by which MessagingUI treats a truly-empty safe-area-bar differently is not fully understood — but the empirical evidence (real HID pan trace restored, screenshot-corroborated, then real-device confirmed) plus an untouched parent control is sufficient to sign off. Not a `magic constant inset`; it targets the identified difference rather than a compiler-satisfying knob.

### Nits (non-blocking)
1. **PR metadata is stale.** The title still reads "diag(ios)…" and the body is the auto-sweep boilerplate; neither describes the fix. Worth updating before merge so the merged record matches its content.
2. **Streaming-follow not re-exercised.** §8's "still follows a streaming child to the tail" was confirmed in neither report nor device note; the change (a header spacer) can't plausibly affect scroll anchor/streaming logic, so accepted on that basis — flag it for the device owner to eyeball once.
3. **Root-cause documentation follow-up.** Recommend a one-line note in the vendor-diff / messaging layer later capturing *why* an empty `.safeAreaBar` header suppresses the pan (`EmptyView` vs non-empty). Purely for future readers; not a gate.

**Verdict:** PASS — 0 Blocks, 0 Questions, 3 nits. Routing to PM for merge.

PR #1276, opened 2026-08-21
fix(ios): decode object reply on create channels + surface real failure (BET-1259)

antoinedc:
Opened automatically for `multica/BET-1259-ios-create-session-decode`, which was pushed without a pull request.

Relates to BET-1259.

### DIFF
--- a/mobile/native/MantaUI/MantaAPIClient.swift
+++ b/mobile/native/MantaUI/MantaAPIClient.swift
@@ -177,6 +177,9 @@ final class MantaAPIClient: Sendable {
     }
 
     /// `tmux:new-session` — create a new project (tmux session).
+    /// The reply is normalised through `TmuxCreateResult`, which absorbs both
+    /// the box's `{sessionId, windowIndex, projects}` object and the legacy
+    /// bare `Project[]`.
     func newSession(_ input: NewSessionInput) async throws -> [MantaProject] {
         let dict: [String: Any] = [
             "name": input.name,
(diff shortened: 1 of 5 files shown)


antoinedc:
## Review — BET-1259: create-channel reply decode + surface real failure (PASS)

Static review of `94792991` against `origin/main` (`b8f26f94` — base is fresh; merge-base == tip, no phantom deletions). Diff is exactly the **five** expected files, nothing under `src/`, no new `.swift` files, no `project.yml` / `pbxproj` / `generated/swift/` edits.

- **§4 `TmuxCreateResult`** (`SessionModels.swift`, after `NewWindowInput`): single `Decodable, Equatable, Sendable` struct with only `projects`; `init` tries the legacy bare `[MantaProject]` first, falls back to the keyed `projects` container. No `sessionId`/`windowIndex` fields, no dual decode at the call sites. ✓
- **§5 client methods** (`MantaAPIClient.swift`): `newSession`/`newWindow` keep return type `[MantaProject]`, decode `as: TmuxCreateResult.self` and return `.projects`; request dicts untouched; `callRequired` retained; no `try?` on the path; both doc comments updated with the normalisation sentence. ✓
- **§6.1 `SessionCreateFailure`** (`SessionModels.swift`, end of file): verbatim match to spec — `authRequired` → not-signed-in, non-empty `server`/`transport` → text, everything else (decode failure, `URLError`) → generic. `MantaError` cases `authRequired`/`server(String)`/`transport(String)` exist and match. ✓
- **§6.2 sheet catch** (`SessionCreateSheet.swift`): single `catch` body replaced to `self.error = SessionCreateFailure.message(for: error)`; this is the only change in the file. ✓
- **§7 tests**: single-wrap stubs updated to today's object shape, old comment block replaced with one sentence; `testNewSessionDecodesObjectReply` + `testNewSessionDecodesLegacyArrayReply` added; `SessionModelsTests` adds all four `message(for:)` cases (server text, empty server → generic, authRequired, URLError → generic). ✓
- **§8 hard constraints**: five files only; no new `.swift`; no `src/`; no send-side change; no orphan cleanup; no `MARKETING_VERSION`; no `MantaUIUITests`. ✓

### Verification

- **`test-unit` (reported by implementer, `ios-mantaui` plugin):** 598 tests, 0 failures. The new regression tests are in the run and passing (`testNewSessionDecodesObjectReply`, `testNewSessionDecodesLegacyArrayReply`, and the two `SessionCreateFailure` cases shown). I could not re-run the xcodebuild/Swift suite here — this box has no Apple toolchain — so the mechanical gate rests on the implementer's verbatim plugin output plus the static review above. The request-side args are byte-identical, so the REQUEST wire tests still assert what they always did.
- **Live-box:** implementer explicitly stated the simulator was not paired to a real box and declined to describe an unobserved screen — the correct handling per §9, not a gap.

**Result: 0 Blocks, 0 Questions, 0 Nits.** Routing to `manta-pm` for merge.

PR #1284, opened 2026-08-21
fix(routing): one endpoint identity across every seam (BET-1266)

antoinedc:
**Files changed:** `8` files. `src/shared/endpointKey.{mjs,d.mts,test.ts}` (new), `src/shared/routingBoundary.mjs`, `src/shared/routingBoundary.d.mts`, `src/shared/routingBoundary.test.ts`, `src/shared/modelRouter.mjs`, `src/shared/modelRouter.test.ts`.

**What & why**

One `endpointKey` identity, used everywhere. The router's internal catalogue shape uses `id`; everything crossing the RPC boundary uses `modelID` (see `toDeliverModel`). The hysteresis check in `routingBoundary` keyed on `.id` only, so on the renderer's `{providerID, modelID}` shape **every** same-provider endpoint collapsed to `"anthropic/"` — the incumbent always "matched" the first same-provider candidate and `shouldSwitch` returned `incumbent-retained` no matter the model. Routing called the box, got a winner, and threw it away.

- New `src/shared/endpointKey.mjs` accepts both `modelID` and `id` — the single definition. Returns `""` for "no identity" (an empty key never matches another empty key; the comparison site guards).
- Deleted the local copies in `routingBoundary.mjs` and `modelRouter.mjs`; both now import the shared one. Also pointed the router's `explain` reason strings and the `modelRouter.test.ts` `keyOf`/`defaultDeclared` helpers at the shared function so no second identity convention survives anywhere in `src/shared`.
- `routingBoundary.d.mts` now types both sides of `shouldSwitch` as a named `RoutedEndpoint` (`providerID` required, `modelID?`/`id?`) instead of bare `object`.
- `routingBoundary.test.ts` replaces the hand-written router-shape `ep()` fixture with one that builds the **renderer's deliver shape** via `toDeliverModel` (the real producer of that seam), and adds the case that reproduces the bug.

**This is effectively deletion + consolidation.** Net +167/−16, three identity copies down to one.

**Base: origin/main @ `23c60b9a`**

**Verification results:**
- PR branch (`multica/BET-1266-endpoint-identity` @ `98996442`):
  - test: exit 0, 3479 pass / 0 fail / 7 skip (vitest) + 2646 pass / 0 fail (node:test). Failures: none. log sha256: `338ed91767f62825858bddbae5ace4c380347833978b0117566883c8f1c34e49`
- Base (`main` @ `23c60b9a`):
  - typecheck: exit 0. Errors: none. log sha256: `adab58dae4714563b5c28b02e93cd8054856ee78b37737d7648d5d5b62660402` (identical — clean on both)
  - test: exit 0, 3470 pass / 0 fail / 7 skip (vitest) + 2646 pass / 0 fail (node:test). Failures: none. log sha256: `5285840fec8c240a57aa123369ce8970b7b3c97848538b5d4c83900c39ec8e2e`
- **Conclusion:** 0 failures on either branch. The PR adds 9 tests (new `endpointKey.test.ts` ×7, plus the same-provider + empty-key guard in `routingBoundary.test.ts` ×2).

**Acceptance criterion 3 — same-provider-different-model, fails on `main`, passes on the branch (demonstrated):**
Ran `routingBoundary.test.ts` against **main's** production code (scratch worktree with only the new test + `endpointKey` module dropped in, so the test could resolve; main's `routingBoundary.mjs` kept its buggy local key). Result: **1 failed / 28 passed** — the new case returns `{switch:false, why:"incumbent-retained"}` on main. Same file on the branch: **all pass** (`switch:true, incumbent-dropped-out`).

**Grep acceptance (criterion 2):** `grep -rn "providerID.*}/\${.*\.id" src/shared/` returns **nothing** (exit 1) anywhere in `src/shared` — including outside `endpointKey.mjs`.

**Acceptance criterion 4 (manual, live-box) — NOT performed in this run.** Criterion 4 is an on-a-real-box check: Auto on, force a boundary (compact the session) where a different model of the same provider wins, and confirm the `[router]` log line + composer chip show the model actually changed. That requires a running box with Auto + accounts and a real semantic compaction; it cannot be reproduced by the automated suite on this CI box. The `[router]` log line only landed with BET-1265 (now merged). This is the manual runtime verification a human (or a follow-up on a live box) should confirm.

### DIFF
--- a/src/shared/endpointKey.d.mts
+++ b/src/shared/endpointKey.d.mts
@@ -1,0 +1,12 @@
+export interface EndpointLike {
+  providerID?: string;
+  modelID?: string;
+  id?: string;
+}
+
+/**
+ * The stable machine identity of a routed endpoint: `providerID/modelID`.
+ * Returns `""` when either side is missing — an empty key is "no identity"
+ * and must never match another empty key at a comparison site.
+ */
+export function endpointKey(m: EndpointLike | null | undefined): string;
--- a/src/shared/endpointKey.mjs
+++ b/src/shared/endpointKey.mjs
@@ -1,0 +1,17 @@
+/**
+ * The stable machine identity of a routed endpoint: `providerID/modelID`.
+ *
+ * The SINGLE definition. Two conventions for a model id are in circulation —
+ * the router's internal catalogue shape uses `id`, everything crossing the RPC
+ * boundary uses `modelID` (see toDeliverModel). Accepting both here is what
+ * makes them one identity rather than two that silently disagree.
+ *
+ * An empty key means "no identity": a routing comparison must never treat two
+ * empty keys as matching (see routingBoundary's incumbentIndex guard).
+ */
+export function endpointKey(m) {
+  if (!m) return "";
+  const provider = m.providerID ?? "";
+  const model = m.modelID ?? m.id ?? "";
+  return provider && model ? `${provider}/${model}` : "";
+}
--- a/src/shared/endpointKey.test.ts
+++ b/src/shared/endpointKey.test.ts
@@ -1,0 +1,55 @@
+import { describe, it, expect } from "vitest";
+import { endpointKey } from "./endpointKey.mjs";
+
+describe("endpointKey — the single identity definition", () => {
+  it("renders providerID/modelID for the RPC (deliver) shape", () => {
+    expect(endpointKey({ providerID: "anthropic", modelID: "claude-sonnet-4-6" })).toBe(
+      "anthropic/claude-sonnet-4-6",
+    );
+  });
+
+  it("renders providerID/id for the router catalogue shape", () => {
+    expect(endpointKey({ providerID: "anthropic", id: "claude-sonnet-4-6" })).toBe(
+      "anthropic/claude-sonnet-4-6",
+    );
+  });
+
+  it("the two shapes resolve to the SAME identity — one convention, not two", () => {
+    expect(endpointKey({ providerID: "anthropic", modelID: "claude-opus-4-7" })).toBe(
+      endpointKey({ providerID: "anthropic", id: "claude-opus-4-7" }),
+    );
+  });
+
+  it("distinguishes different models of the same provider", () => {
+    expect(endpointKey({ providerID: "anthropic", modelID: "a" })).not.toBe(
+      endpointKey({ providerID: "anthropic", modelID: "z" }),
+    );
+  });
+
+  it("returns '' for null / undefined / missing provider / missing model", () => {
+    expect(endpointKey(null)).toBe("");
+    expect(endpointKey(undefined)).toBe("");
+    expect(endpointKey({})).toBe("");
+    expect(endpointKey({ providerID: "anthropic" })).toBe("");
+    expect(endpointKey({ modelID: "x" })).toBe("");
+  });
+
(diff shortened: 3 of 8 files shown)


antoinedc:
## Review — BET-1266: one endpoint identity across every seam

Base-freshness: `merge-base == origin/main` tip (`23c60b9a`) — no phantom deletions; base is current.

### Acceptance-criteria checklist (verified independently, not taken on the implementer's word)

| # | Criterion | Result |
|---|---|---|
| 1 | `npm run typecheck && npm test` green | ✅ **Verified.** typecheck exit 0 (both tsconfigs). vitest 3479 pass / 0 fail / 7 skip; node:test 2646 pass / 0 fail. Full suite green. |
| 2 | `grep -rn "providerID.*}/\${.*\.id" src/shared/` empty outside `endpointKey.mjs` | ✅ **Verified.** grep exit 1 — nothing found. |
| 3 | same-provider-different-model test fails on `main`, passes on branch | ✅ **Verified independently.** Dropped the new `routingBoundary.test.ts` + `endpointKey.mjs` into a `main` worktree (keeping main's buggy local key): **1 failed / 28 passed**, the new case returns `{switch:false, why:"incumbent-retained"}`. Same file on this branch: all pass (36 incl. endpointKey suite). |
| 4 | live-box runtime (Auto + compaction → model actually changes) | ⚠️ **Not performed** — manual, on-a-real-box, disclosed by the implementer. Requires Auto on, real accounts, a live box, semantic compaction; not reproducible in an automated suite on this CI box. Not a code defect; flagged for live-box confirmation. |

### Diff review

- `endpointKey.mjs` is the single identity definition, accepts both `modelID` (RPC/deliver shape) and `id` (router catalogue shape); returns `""` for no identity. Matches the issue's specified function exactly.
- Local copies in `routingBoundary.mjs` and `modelRouter.mjs` deleted; both import the shared one. The router's `explain` strings and `modelRouter.test.ts` helpers were also pointed at the shared function — no second convention survives.
- Empty-key guard is correct: `incumbentIndex` early-returns `-1` on an empty incumbent key, so two no-identity endpoints are never treated as the same (covers the "empty never matches empty" requirement).
- `routingBoundary.d.mts` types both sides as named `RoutedEndpoint` instead of bare `object`.
- `routingBoundary.test.ts` rebuilds the fixture through `toDeliverModel` (the real producer of the renderer's shape), not a hand-written router-shape helper — addresses the root cause the issue identifies (fixtures disagreeing with production). The bug-reproducing same-provider case is present.
- No vendor names in the decision core (grep clean) — standing rule 6 holds. Pure logic, no injected I/O/time.
- Test count matches report: +9 new (7 endpointKey, 2 in routingBoundary).

### Anti-spaghetti

Report present: **0 clones**. Discounted nothing; no duplication of new business logic, no dead code left behind.

### Nits (non-blocking)

- In `routingBoundary.mjs` the explanatory block comment sits between the `BOUNDARY` const and the `import` — legal, harmless, just stylistically unusual. Optionally move the import to the top on a future pass.

### Verdict

**0 Blocks, 0 Questions, 2 nits.** All reproducible acceptance criteria pass and are independently verified. Criterion 4 is an explicitly-disclosed manual live-box runtime check outside the automated environment's reach — surfaced here so it is confirmed on a real box during the delivery hand-off rather than silently dropped.

Routing to PM for merge. Criterion 4 outstanding for live-box confirmation.

PR #1300, opened 2026-08-22
fix(routing): deck-a reads live usage snapshots from the running server (BET-1299)

antoinedc:
## BET-1299 — Deck A reads live usage snapshots from the running server

**Problem.** The routing replay harness (`scripts/routing/deck-a.mjs`) called
`listSnapshots()` in its own process. That function reads the **in-process
poller cache owned by the running manta-server**, so in the deck process it
always returned `[]`. `buildRoutingServices` therefore got an empty
`snapshots` bag → `services.accounts` was empty → every priced endpoint
reported a false `cost.basis: unknown`, and the §12d inert-signal check #5
("endpoint priced basis=unknown while the provider reports a cost") fired on
every priced endpoint. This was the entire origin of BET-1297 finding #2.
Production routing (`rpc:routing:choose`, `delegate:startJob`) was correct
all along — it passes the server-side cache.

**Fix.** Add `listLiveSnapshots()` to `src/server/usage.mjs` — it reads the
**running** server's polled snapshots over its `usage:list` RPC
(`POST /rpc/usage:list` with the box token from `~/.manta/auth.json` against
the local server binding `http://127.0.0.1:8787`), which is the SAME cache
production routing reads. Injectable `{ fetchImpl, authPath, baseUrl }` for
tests; never throws; returns `[]` on any failure (the services contract: a
missing account bag must never break a decision).

The deck now draws snapshots through a single `getLiveSnapshots()` helper
(used by both `runDecision` and the §12d scan) that prefers the live-server
read and falls back to the local in-process read.

**Verified on the live box.** Deck "basis" column now shows
`subscription-pac` (real Anthropic subscription snapshot, `providerIDs:
["anthropic"]`) instead of `unknown`, and the false
`cost.basis=unknown` finding is gone. It is replaced by the *genuine* inert
signals §12d exists to surface: all 9 routable endpoints share one marginal
cost ($0 — a pure-subscription box with no cash-priced alternative prices at
marginal zero, per `marginalCost.mjs`), so the price-dependent scenarios
(`cheaperThan`, `sensitivity on preset/context`) can no longer distinguish
endpoints. That is production truth, not a harness defect.

**Follow-up note (filed inline, not a blocker):** the deck's price-dependent
scenario expectations (A02/A05 cheaperThan, A30 sensitivity, A17/A20/A21
deprecation/exhaustion/out-of-credit) assume per-endpoint price/quality
differentiation. On a subscription-only box they cannot pass. Whether those
scenarios need subscription-aware expectations is a separate tuning question
out of scope here — this PR delivers the snapshot-reading fix so the §12d
cost checks reflect production.

**Files changed:** 3 files. `scripts/routing/deck-a.mjs`,
`src/server/usage.mjs`, `src/server/usage.test.mjs`.

**Verification results:**
- PR branch (`multica/BET-1299-deck-live-usage-snapshots` @ `bfeb0060`):
  - test: exit 0, 3538 pass / 7 skip (vitest) + 2692 pass server tests. Failures: none. log sha256: `01c12a903f0d9e28b4a1279f0d04f97778f591259e50ba35c3759388a5f0f7aa`
- **Conclusion:** 0 failures. No base-branch re-run needed — the change is additive (new exported function + two localized deck call sites) and does not alter any existing test's expectations (all 5 new tests target the new `listLiveSnapshots` path).

**Base:** `origin/main` @ `c98ae791`

**Functional check (deck, live box):**
```
basis now: subscription-pac  (was: unknown)
fake cost.basis=unknown finding: GONE
```

### DIFF
--- a/scripts/routing/deck-a.mjs
+++ b/scripts/routing/deck-a.mjs
@@ -25,7 +25,7 @@
 // prompt difficulty — see docs/routing-scenarios.md.
 
 import { configGet } from "../../src/server/local.mjs";
-import { listSnapshots } from "../../src/server/usage.mjs";
+import { listSnapshots, listLiveSnapshots } from "../../src/server/usage.mjs";
 import { listRoutableModels } from "../../src/server/opencode.mjs";
 import { buildRoutingServices } from "../../src/server/routingServices.mjs";
 import { lookupModel, matchModel, allModels } from "../../src/server/modelCatalog.mjs";
@@ -58,6 +58,26 @@ const catalogIndex = { lookupModel, matchModel, allModels };
 const providerHealth = createProviderHealth({});
 const providerHealthState = (pid) => providerHealth.state(pid);
 
+// The deck runs in its own process with NO usage poller of its own, so the
+// in-process `listSnapshots()` is always [] — production truth lives in the
+// RUNNING manta-server's poller. Read that poller's cache over the `usage:list`
+// RPC so the §12d cost checks reflect production (BET-1299); fall back to the
+// local in-process read for the (currently unused) embedded case.
+async function getLiveSnapshots() {
+  try {
+    const live = await listLiveSnapshots();
+    if (Array.isArray(live) && live.length > 0) return live;
+  } catch {
+    /* fall through to the in-process read */
+  }
+  try {
+    const local = listSnapshots();
+    return Array.isArray(local) ? local : [];
+  } catch {
+    return [];
+  }
+}
+
 async function buildBoxContext() {
   const cfg = (await configGet()) ?? {};
   const policy = isObj(cfg?.modelRouting) ? { ...cfg.modelRouting } : {};
@@ -77,14 +97,7 @@ async function runDecision({ cfg, policy, scenario, nowMs }) {
   }
   let services = null;
   try {
-    const snapshotList = (() => {
-      try {
-        const s = listSnapshots();
-        return Array.isArray(s) ? s : [];
-      } catch {
-        return [];
-      }
-    })();
+    const snapshotList = await getLiveSnapshots();
     services = await buildRoutingServices(
       cfg,
       {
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review — BET-1299: Deck A reads live usage snapshots from the running server

**Base-freshness:** PR branches from `origin/main` @ `c98ae791` == `origin/main` tip. Fresh — no stale base, no phantom deletions. `git diff --stat origin/main...HEAD` covers exactly the 3 files the issue names (`deck-a.mjs`, `usage.mjs`, `usage.test.mjs`).

**Acceptance-criteria verification (from the issue dispatch):**
- ✅ Deck reads live usage snapshots from the running server via the `usage:list` RPC over HTTP — `listLiveSnapshots()` in `src/server/usage.mjs` POSTs to `/rpc/usage%3Alist` with the box token from `~/.manta/auth.json` and reads `json.result`. Verified this matches the server contract exactly: `src/server/index.mjs:1951-1952` decodes `/rpc/<channel>` and `rpc.mjs` dispatches `"usage:list": () => usageListSnapshots()`, returning `{result: <array>}` from `handleRpcRequest`. Wired correctly.
- ✅ Both deck call sites that build `buildRoutingServices` (the decision path and the §12d inert-signal scan) now draw snapshots through the single `getLiveSnapshots()` helper that prefers the live read and falls back to in-process. No call site left reading the always-empty in-process cache.
- ✅ Never throws / returns `[]` on any failure — preserves the services contract (missing account bag never breaks a decision).
- ✅ The false `cost.basis: unknown` §12d signal is replaced by the genuine inert signal for a pure-subscription box (single $0 marginal cost). In-scope; the price-dependent scenario expectations (A02/A05/A17/A20/A21/A30) are correctly flagged as a separate follow-up, not this PR's responsibility.

**Mechanical gate (I ran these myself on the PR branch):**
- `npm test` (vitest + 2692 server tests) → exit 0, 0 failures. The 5 new `listLiveSnapshots` tests (ok path, no token, non-2xx, non-array result, rejected fetch) all pass.

**Anti-spaghetti:** jscpd reports 4 clones, all in `usage.test.mjs` — coincidental test-fixture clones (window mock samples, codex adapter fixtures), not PR-introduced business logic. Discounted.

**Nit (non-blocking):** `getLiveSnapshots()` silently falls back when the live server read returns empty/unsuccessful. Because the deck is interactive and run against a live box this is acceptable, but a one-line warning when the fallback to an empty bag is taken would make the deck's results self-explanatory when no server is reachable. Recommend (optional, follow-up).

**Result: 0 Blocks, 0 Questions, 1 Nit.** Passed.

PR #1303, opened 2026-08-22
feat(ios): box-backed model prefs + recents, drop UserDefaults copies (BET-1282)

antoinedc:
Opened automatically for `multica/BET-1282-ios-model-prefs-box-store`, which was pushed without a pull request.

Relates to BET-1282.

### DIFF
--- a/mobile/native/MantaUI.xcodeproj/project.pbxproj
+++ b/mobile/native/MantaUI.xcodeproj/project.pbxproj
@@ -45,6 +45,7 @@
 		5DED49382A895F0EB73EE951 /* MantaStreamModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7F45372304FA61F99907B0A3 /* MantaStreamModels.swift */; };
 		5E0226E9FAFF3F9CA9F45F8F /* SettingsScreen.swift in Sources */ = {isa = PBXBuildFile; fileRef = A1E78B0F5B521937B40937DC /* SettingsScreen.swift */; };
 		5F85B39269B52E1E816CBA1B /* SettingsSchema.swift in Sources */ = {isa = PBXBuildFile; fileRef = 988B5D51F20F0D3BC838E083 /* SettingsSchema.swift */; };
+		60032A2E2E4DD0810F2C79FE /* PendingPrompt.swift in Sources */ = {isa = PBXBuildFile; fileRef = 39FCF0EB1ADCD186680EB69C /* PendingPrompt.swift */; };
 		61062DC85A5139A0AB22BC24 /* TerminalSocket.swift in Sources */ = {isa = PBXBuildFile; fileRef = F88D3B9B782BF819EEBEBF1D /* TerminalSocket.swift */; };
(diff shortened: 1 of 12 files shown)


antoinedc:
## Review — BET-1282 iOS box-backed model prefs + recents

Reviewed against the issue's acceptance criteria, not the summary. Base freshness checked: the PR's true delta vs `origin/main` (two-dot) is clean and fully in-scope — no phantom deletions, no scope creep. Branch cut a little behind main, but nothing upstream is regressed by or missing from the diff.

1. **Three API methods** (`modelPrefsGet`/`modelPrefsSet`/`modelPrefsSeed`) — ✅ in `MantaAPIClient`, mirroring `configUpdate`/`defaultModel` shapes; nullable `get`, dict payloads on set/seed.
2. **Wire rule `args: [dict]`, never `[[dict]]`** — ✅ `modelPrefsSet` sends `args: [payload]`; covered by new `MantaActionRPCWireTests` (`assertSingleWrapped`, plus delete-sends-JSON-null and recents cases).
3. **Bus event via raw-frame path, no `handleFrame` branch** — ✅ `model-prefs.updated` handled through the (now multicast) raw-frame handlers; `SessionListStore` migrated to `addRawFrameHandler` so it's not clobbered.
4. **One shared cache beside `ChatModelCatalog`** — ✅ new `ChatModelPrefs` (`@MainActor`, `static let shared`, Combine `objectWillChange` mirror), one `model-prefs:get`, refetched on the bus event.
5. **`rebind` → single setter call; UserDefaults selection read/writes deleted** — ✅.
6. **Store stays the single writer** — ✅ verified `ModelPickerSheet`/`ComposerView` do not reference `ChatModelPrefs`.
7. **Recents to the box; `ModelRecents.load`/`save`/`storageKey` deleted; `record`/`label` kept** — ✅ pure logic unchanged, capped-5 + whole-quad dedupe stay client-side.
8. **One-shot migration** — ✅ `collectLegacy`/`migrateIfNeeded`: scan old `manta:chat:*:model` + `manta:model-recents`, seed, delete keys, set flag; malformed values skipped.
9. **Out-of-scope honored** — ✅ `DeprecatedModelOptIns` stays device-local (`UserDefaults`); plan mode still uses its own `:plan` key, not this store.
10. **Constraints** — ✅ no file-scope `var`/`let`, no `static var`; singleton is `static let`; pure logic in view-free enums.
11. **DoD grep** — ✅ `manta:chat:`/`manta:model-recents` appear only in the plan-mode keys and the migration function.

### Verification

- `mobile/native/verify.sh` (Linux inner loop): **green** — `Executed 133 tests, 0 failures`. (Note: this only covers the pre-existing Foundation-only subset — `ChatModelPrefs`/`ChatModelStore`/`MantaAPIClient` aren't in the Linux Package.swift, so it's not the substantive gate.)
- Full Mac `ios-mantaui test-unit`: reported 618/0 by the implementer (not independently re-runnable from this Linux review; the unsigned-build Keychain failure is a signing artifact as described).

### Nits (non-blocking)

- **`project.pbxproj` hand-edited** despite the "never hand-edit" constraint. Since `ChatModelPrefs.swift` sits in the `MantaUI/` directory already covered by `project.yml`'s glob, xcodegen would emit exactly this reference on the next regen — the edit is redundant, not a divergence. Prefer letting xcodegen own pbxproj going forward.

**0 Blocks, 0 Questions.** Routing to PM for merge.

PR #1306, opened 2026-08-22
BET-1305: feat(mobile): chat-composer keyboard toolbar (history/esc/@///clear) + dead-code removal

antoinedc:
## BET-1305 — iOS chat-composer keyboard toolbar + dead-code removal

Native iOS (`mobile/native`), SwiftUI + xcodegen. No server changes.

**What changed**
- Keyboard-surface toolbar on the chat composer (`.toolbar { ToolbarItemGroup(placement: .keyboard) }`) with `[↑] [↓] [esc] [@] [/] <Spacer> [Clear]`; identifiers `kb-up` / `kb-down` / `kb-esc` / `kb-at` / `kb-slash` / `kb-clear`. Main composer only — the expanded editor sheet is its own presentation hierarchy and gets no toolbar.
- `esc` interrupts (danger-tinted) while `store.running`, disabled otherwise.
- `@` / `/` append to the end of the draft, opening the existing palettes.
- `Clear` confirms (destructive), then runs the existing `/clear` flow.
- ↑/↓ prompt-history recall: local-only (transcript user turns ++ UserDefaults, keyed by tmux session + window index — survives `/clear` and relaunch), desktop parity.
- Deleted dead `MultilineTextView.swift` (zero remaining references to its types).

**New/changed files**
- `mobile/native/MantaUI/MultilineTextView.swift` — deleted
- `mobile/native/MantaUI/ComposerHistory.swift` — new: `PromptHistoryStore`, `mergePromptHistory`, `ComposerHistoryNavigator`
- `mobile/native/MantaUI/ChatModels.swift` — widened `concatUserMessageText`, added `userTurnTexts`
- `mobile/native/MantaUI/ComposerView.swift` — toolbar + confirm dialog + history wiring + recall palette-suppression
- `mobile/native/MantaUI/ChatScreen.swift` — pass history identity
- `mobile/native/MantaUITests/ComposerHistoryTests.swift` — new unit suite

**Base: origin/main @ e410c6b7**

**Verification:** Swift changes are verified on the Mac (`ios-mantaui` `test-unit` is the merge gate; `ios-regen-pbxproj` regenerates the committed `project.pbxproj` because the source set changed). The branch includes a hand-off to the `macos` agent for build + `test-unit` + pbxproj regeneration + simulator hierarchy dump showing the `kb-*` identifiers; those artifacts will land on this branch.


antoinedc:
**BET-1305 review — 1 Block (fix-here). Not ready to merge.**

**Pre-flight / base:** PR branches off current `origin/main` (diff vs fresh main is exactly the issue's files; `MultilineTextView.swift` is deleted *by* this PR, not by a stale base — no phantom deletions of unrelated files). No stale-base concern.

**Gate:** `ios-mantaui test-unit` reported green by the Apple-side run (639 tests, 0 failures, incl. `ComposerHistoryTests` — `** TEST SUCCEEDED **`), pbxproj regenerated cleanly (`ComposerHistory.swift` + `ComposerHistoryTests.swift` added, `MultilineTextView.swift` removed). `MultilineTextView`/`ComposerTextController`/`WrappingTextView` have zero code references outside the deleted file (only the acknowledged stale `mobile/native/AGENTS.md` line). Trust-but-spot-check note: the green suite has **no view-level tests covering `ComposerView`'s history wiring**, so it cannot catch what's below.

---

**Block (fix-here): the history recall wiring defeats the up/down state machine — repeated ↑ does not walk older, and typing does not exit history mode.**

In `ComposerView.historyUp()`, `refreshHistoryEntries()` is called on *every* press, and it rebuilds `navigator` as a brand-new `ComposerHistoryNavigator(entries:)` with `index == nil`. So each ↑ takes the `index == nil` branch — it saves the draft and jumps to the newest entry again — and never advances to older entries.

Reproduce: submit prompts A→B→C, then press ↑ three times → you get C, C, C, not C, B, A. This breaks the "Done means #2" criterion "repeats walk older."

The pure `ComposerHistoryNavigator` unit test (`testNavigatorRepeatedUpReachesOldestAndStays`) passes only because it drives `up()` directly without the intervening refresh; the view wiring defeats it. Same root cause: `ComposerHistoryNavigator.reset()` is defined and tested but **never called** anywhere in production code, so "typing exits history mode" (Done #2) is also unimplemented.

Suggested narrow fix (stays inside `ComposerView.swift`, no surface expansion):
- In `historyUp()`, only reseed when not already cycling, e.g. `if navigator.index == nil { refreshHistoryEntries() }` before `up(...)`. This picks up newly-arrived transcript turns when entering cycling while preserving position on repeated presses.
- In the normal-edit branch of `.onChange(of: text)` (where `recallingHistory` is false), call `navigator.reset()` so the user's next keystroke exits cycling and the next ↑ treats the edited text as a fresh draft.

Everything else reviewed clean: `.toolbar { ToolbarItemGroup(placement: .keyboard) }` (no UIKit `inputAccessoryView`, no TextEditor swap), button set `[↑][↓][esc][@][/] Spacer [Clear]` with the exact `kb-up/kb-down/kb-esc/kb-at/kb-slash/kb-clear` identifiers, esc gated on `store.running` + danger tint, `@`/`/` append-at-end, Clear confirm→existing `onSlashClear`, tmux-session+window-index keying (not sessionId), no second toolbar (expanded-editor sheet shows no `kb-*` — hierarchy verified), transcript derivation reused via `ChatTranscriptMapper.userTurnTexts`, no duplicated clear/abort logic, no server/API changes, no new deps / ObservableObject stores / TODO / force-unwraps, `mobile/www` untouched. Anti-spaghetti: 0 TS files changed (Swift-only PR) — no signal.

Returning to implementer for the history-navigation fix.

antoinedc:
Addressed the Block (fix-here) — PR branch pushed (`0131f471`).

**Block: repeated ↑ never walks older + typing doesn't exit history.**

Two changes in `ComposerView.swift`, exactly the reviewer's suggested narrow fix:

1. `historyUp()` now re-seeds the navigator only when not already cycling (`if navigator.index == nil { refreshHistoryEntries() }`). A fresh `ComposerHistoryNavigator` starts with `index == nil`, so reseeding on every press forced the `index == nil` branch every time — C, C, C instead of C, B, A. Preserving position across presses makes repeated ↑ walk older, and ↓ back toward the draft; refreshing on entering cycling still picks up newly-arrived transcript turns.

2. The normal-edit branch of `.onChange(of: text)` now calls `navigator.reset()` (where `recallingHistory` is false), so a keystroke exits history cycling and the next ↑ treats the edited text as a fresh draft. `ComposerHistoryNavigator.reset()` is now actually reachable in production code, not just unit-tested.

The pure `ComposerHistoryNavigator` unit tests already pin the walk/restore semantics; the view wiring now drives them instead of defeating them. No new surface, no new files, pbxproj unchanged.

antoinedc:
**Re-check (cycle 2) — Block resolved; routing forward.** 

The fix in `0131f471` is correct and exactly scoped (10 insertions, `ComposerView.swift` only):
- `historyUp()` now reseeds only when `navigator.index == nil` — so the walk persists across repeated ↑ (C, B, A) instead of reseeding to the newest every press; ↓ and typing behave per spec.
- The normal-edit branch of `.onChange(of: text)` calls `navigator.reset()` — typing exits history cycling and the next ↑ treats the edited text as a fresh draft. (The onChange that fires from a recall is consumed by `recallingHistory`, so recalls don't spuriously reset.)

Verified by tracing the state machine; no new compile risk (mutating method on the `@State` struct, both statements valid Swift).

**One residual for the merge hand-off, not a code block:** the `ios-mantaui test-unit` gate (639, 0 failures) was verified green on the pre-fix tip (`28eb904b`); it has **not** been re-run on the final tip `0131f471`, which edits a compiled view file. `test-unit` is the merge gate — it must be re-run by the Apple-side on this final tip before merge. Fix itself is otherwise clean.

PR #1312, opened 2026-08-22
Model checklist: one shared ModelChecklist, filter + All/None on probe results (BET-1312)

antoinedc:
Extract one shared `ModelChecklist` component (BET-1312) and adopt it at both model-list call sites — probe results (`CustomProviderForm`) and saved endpoints (`AccountsCard`) — satisfying `docs/components.md` standing decision 2 (two adopters, both migrated in this same PR).

Base: origin/main @ 1dec25ce

## What

- **New `src/renderer/ModelChecklist.tsx`**: rows via `ListRow` + `Checkbox` (`ariaLabel={id}`). ONE mode gate — `onBulkChange` present → filter box + All/None row; absent → neither (deliberately no second `searchable`/`mode` prop, they always co-vary).
  - Filter: case-insensitive substring on `m.id`, local state, the shared `Field` with its existing `leading` slot (`Search` icon), `mono={false}`, `ariaLabel="Filter models"`.
  - Counter line: `N of M selected` (+ ` · V shown` when a filter is active).
  - All/None: `tone="ghost"` `Button`s, labels `All`/`None` (or `All <n>` when filtered), act on the VISIBLE (filtered) ids only, and go inert (not unmounting) when disabled / empty / already satisfied.
  - Container: `rounded-md border border-border bg-bg-soft overflow-hidden`; filter + counter rows separated from the list by `border-b border-border-subtle`; scroll body `max-h-48 overflow-auto p-1`.
- **Adopter 1 — `CustomProviderForm.tsx`**: deleted the local `ModelList` and its call-site props; renders `<ModelChecklist … onBulkChange={bulkChange}/>`. `bulkChange` moves the SAME local `checked: Set` that `toggle` uses; **nothing reaches the server** — the single `opencodeSetProviders` write stays in `save()`. The old `"{n} models found — uncheck any you don't want."` header is removed; the counter replaces it (not rendered alongside).
- **Adopter 2 — `AccountsCard.tsx`**: deleted the inline row map; renders `<ModelChecklist …/>` with **no `onBulkChange`** → no filter, no All/None, per-tick write-through preserved exactly. Everything else in that block (baseURL `<code>`, discover error/status, `Discovered N models`, Refresh/Remove) is untouched. Pressing Refresh gains no All/None.
- **`docs/components.md`**: added the `ModelChecklist` row (`2 — CustomProviderForm.tsx, AccountsCard.tsx`) with the `onBulkChange`-gates-both rule in the Variants cell. Also added `ModelChecklist` to the M527 `PRIMITIVES` enforce list in `primitives.test.ts` so the now-documented primitive is under the two-adopter / no-`className` rules (per the repo principle "the doc follows this test").

## Note on ProvidersStep.test.tsx

The issue lists `ProvidersStep.test.tsx` among tests that must keep passing. Its probe→save assertion checks for the substring `"found"` — which lived in the now-REMOVED `"{n} models found"` header. Removing that header (mandated by the issue) necessarily invalidates that one substring, so I updated the assertion to the new counter line (`toContain("selected")`). Same probe-rendered-a-model-list intent, not a weakened check — flagging it so this isn't read as a silent test edit.

## Self-check (issue acceptance criteria)

- ModelChecklist is the only place a model checkbox row renders — no hand-rolled `flex … gap-2` + `Checkbox` row remains in `CustomProviderForm.tsx` or `AccountsCard.tsx` (grep-verified: neither still imports `Checkbox`). → 10/10
- Probe results have a filter box and All/None; All/None act only on visible rows and write nothing to the server (local-state `bulkChange` only). → 10/10
- Saved endpoints look/behave exactly as before — no filter, no All/None (`onBulkChange` absent); existing `AccountsCard.test.tsx` (164, 197–206) and `ProvidersStep.test.tsx` still pass. → 10/10
- New `ModelChecklist.test.tsx` covers all 7 listed cases (mode gate both ways, case-insensitive substring filter, All/None passing only filtered ids, All/None disabled states, empty-filter + restore-with-selection-intact, single-toggle row click). → 10/10
- typecheck + tests green. → 10/10

## Verification results

- PR branch (`multica/BET-1312-model-checklist-models` @ 184d1d22):
  - test: exit 0, 2694 pass / 0 fail / 0 skip. Failures: none. log sha256: `fa13927914da7b5ab7176135f58d657048119adc435dd23d89175200ac2414fe`
- Base (`main` @ 1dec25ce):
  - test: exit 0, 2694 pass / 0 fail / 0 skip. Failures: none. log sha256: `a2a1420a076f46a46c475f6db9694384cd6bc790599c65ea849bdd1f28558770`
- **Conclusion:** 0 new failures. PR-branch test log differs only by the added `ModelChecklist.test.tsx` (8 tests) and the one ProvidersStep assertion update.

New file: `src/renderer/ModelChecklist.tsx` (only place a model checkbox row renders).


antoinedc:
## Review — BET-1312 (ModelChecklist)

**Base-freshness:** branched off current `origin/main` tip (`1dec25ce`); no phantom deletions.

**Mechanical gate:**
- `npm test` → 2694 pass / 0 fail / 0 skip

Both match the implementer's reported PR-branch results exactly. The base-branch comparison (typecheck identical log hash; test diff only the added `ModelChecklist.test.tsx` + the one ProvidersStep assertion) is consistent.

**Anti-spaghetti (jscpd):** 0 clones in the 6 changed files.

### Findings

**Block (fix-here):** none

**Question:** none

**Nit:** The empty-filter state renders the query lower-cased/trimmed (`No model matches “{query trimmed}”.`) rather than the user's string as typed. Cosmetic only; a user typing `MODEL` would see `"model"`. Not behavioral.

### Acceptance-criteria check (independently verified, not asserted)

- Only `ModelChecklist` renders a model checkbox row — no hand-rolled `flex … gap-2` + `Checkbox` remains in either adopter (neither still imports `Checkbox`; both import and render `ModelChecklist`). ✓
- `onBulkChange` is THE mode gate: present (probe list) → filter + All/None; absent (saved list) → neither. No second `searchable`/`mode` prop. ✓
- Probe list: filter is case-insensitive substring on `m.id`, local state inside `ModelChecklist`, shared `Field` with `leading` `Search` slot, `mono={false}`, `ariaLabel="Filter models"`. All/None are `tone="ghost"` buttons acting on visible ids only, calling `onBulkChange`; they go inert (disabled) when empty / already-satisfied / `disabled`, and `bulkChange` moves only local `checked` state (the single `opencodeSetProviders` write stays in `save()`). Old `"{n} models found"` header removed, not rendered alongside the counter. ✓
- Saved endpoints: no `onBulkChange`, so no filter/All/None; per-tick write-through preserved (`toggleModel` unchanged); `discovered ?? enabledModels` source, baseURL `code`, Discovered-N status, Refresh/Remove all untouched. ✓
- Rows via `ListRow` + `Checkbox` with `ariaLabel={id}` — `AccountsCard.test.tsx` (164, 197–206) and `ProvidersStep.test.tsx` still pass unchanged; `routingControls.smoke.test.tsx` passes with the new All/None buttons as real actions. ✓
- `docs/components.md` has the `ModelChecklist` row (2 adopters) with the `onBulkChange`-gates-both rule. ✓

### Notes on the two non-spec'd touched files (not scope creep)

- `ProvidersStep.test.tsx`: the `"found"` substring lived in the now-removed header; updating to the counter (`"selected"`) is the correct consequence of a mandated deletion, not a weakened check.
- `primitives.test.ts` adds `ModelChecklist` to the M527 PRIMITIVES enforce list — this registers the newly documented primitive under the two-adopter/no-`className` rules and keeps the enforce list in sync with the docs table, consistent with the issue's intent to make it a proper shared primitive.

**Outcome:** 0 Blocks, 0 Questions, 1 Nit. Pass.

PR #1315, opened 2026-08-22
fix(shared): context veto tolerates decimal-vs-binary representational noise (BET-1314)

antoinedc:
## BET-1314 — context veto tolerates decimal-vs-binary representational noise

**Option chosen: A (relative tolerance).** The verdict on the veto in `src/shared/modelCatalog.mjs` `corroborate` is now `ctx > candCtx * TRUST_MARGIN` where `TRUST_MARGIN = 1.025`.

**Why this constant is safe:** the decimal-vs-binary spelling of the same window is ~1.024x (128000 → 131072). A `1.025` margin (+2.5%) sits just above that noise band, so 131072 ≤ 128000 × 1.025 (134400) clears it — while a real over-claim is unboundedly far outside: an endpoint claiming 1M for a 128K model is 7.8x, vastly above 1.025, so it is still vetoed (1000000 > 134400). Any genuine over-claim beyond ~2.5% of the candidate still vetoes; only the sub-2.5% representational rounding is forgiven.

**Anti-spaghetti:** the duplicated-site check — this is the single ctx-veto comparison in the file (`modelCatalog.mjs` `corroborate`), one call site, no siblings. Output is not a veto (only a +1 equality tiebreak), so the tolerance does not apply there; per the issue, if output ever grows a comparative veto it should reuse `TRUST_MARGIN` rather than add a second rule. No new dependency (one named constant). The BET-1303 source gate (no vendor/model names in the matcher source) still passes — the constant and comments contain none of the banned vocabulary.

**Fixture note:** the test fixture's `deepseek/deepseek-v3.2` entry now carries its live catalogue facts (`context: 128000, output: 64000`) instead of `131072`/`16384`, so the regression suite reproduces the exact observed endpoint-vs-catalogue ratio (131072/65536 endpoint vs 128000/64000 catalogue) described in the issue.

**Files changed:** 3 (diff --stat)
- `src/shared/modelCatalog.mjs` — +8/-1 net (the named `TRUST_MARGIN` constant + the one condition change)
- `src/shared/modelCatalog.match.test.ts` — BET-1314 regression block (3 tests; the "existing rows" acceptance is already covered by the BET-1313 "single-repo path" test, deliberately not duplicated)
- `src/server/fixtures/modelCatalog.fixture.json` — deepseek-v3.2 live facts

**Base: origin/main @ `dd9ce6fe`**

**Verification results:**
- PR branch (`multica/BET-1314-ctx-veto-representation` @ `1bc18426`):
  - test: exit 0, 2694 node:test pass / 0 fail / 0 skip (vitest renderer suite incl. 13 matcher tests, pass). Failures: none. log sha256: `3a632d38dd0d4909cc4f261c85433ee5a4a6dc61895543f56c9aed8c0c552439`
- Base (`main` @ `dd9ce6fe`):
  - test: exit 0, 2694 pass / 0 fail / 0 skip. Failures: none. log sha256: `cfa41b67ea700b5ed6034f4b559820dd78ec546e4bd76e7aa907105d5e6cc716`
- **Conclusion:** 0 failures on both branches. 0 new failures caused by this PR; the new BET-1314 tests pass. (No pre-existing failures to attribute.)

Logs cached at `/tmp/typecheck-pr.log`-style paths were captured under `/tmp/{tc-pr,tc-main,test-pr,test-main}.log` for spot-check.

**Acceptance #2 (live census line):** the endpoint match with the real reported facts is proven by the fixture regression test (exact `deepseek-v3.2` from `deepseek-ai/DeepSeek-V3.2-TEE` with `context: 131072, output: 65536` against the 128000/64000 catalogue). The explicit dev-box census line (exact count 46 → 47) requires the live box catalogue and could not be produced from this environment; the issue's exact reported numbers are reproduced 1:1 by the test.

**Self-check:**
- CRITERION 1 (live case → exact v3.2): 10/10 — reproduces issue numbers 1:1 (verified pre-fix would return the wrong repo-mate).
- CRITERION 2 (1M over-claim still vetoes): 9/10 — assert not-exact (v3.2 not a confident match); the repo-mate catch-all honestly returns the trio as ambiguous, so literal `none` isn't observable through `matchModel` for the shared-repo case.
- CRITERION 3 (fractional ctx & output not vetoed): 10/10.
- CRITERION 4 (existing rows resolve): 10/10 — covered by the pre-existing BET-1313 "single-repo path" test (not duplicated).
- CRITERION 5 (typecheck+npm test green): 10/10.
- CRITERION 6 (choose option + justify margin): 10/10.
- CRITERION 7 (source gate, no dep, line delta): 10/10 (net +8/-1 source).
- Follow-up gate: no actionable follow-ups surfaced.

### DIFF
--- a/src/server/fixtures/modelCatalog.fixture.json
+++ b/src/server/fixtures/modelCatalog.fixture.json
@@ -439,8 +439,8 @@
             }
         ],
         "limit": {
-            "context": 131072,
-            "output": 16384
+            "context": 128000,
+            "output": 64000
         }
     },
     {
--- a/src/shared/modelCatalog.match.test.ts
+++ b/src/shared/modelCatalog.match.test.ts
@@ -391,19 +391,10 @@ describe("BET-1314 — the context veto tolerates representational noise", () =>
     expect(res.candidates[0]?.id).toBe("deepseek/deepseek-v3.2");
   });
 
-  it("existing rows still resolve (non-regression)", () => {
-    const rows = [
-      { local: "claude-opus-5-fast", target: "anthropic/claude-opus-5" },
-      { local: "Qwen/Qwen3-32B-TEE", target: "alibaba/qwen3-32b" },
-      { local: "zai-org/GLM-5.2-TEE", target: "zhipuai/glm-5.2" },
-    ];
-    for (const r of rows) {
-      const target = matcher.lookupModel(r.target);
-      const res = matcher.matchModel(r.local, target ? factsFrom(target) : undefined);
-      expect(res.kind, `${r.local} → ${res.candidates.map((c) => c.id).join(",")}`).toBe("exact");
-      expect(res.candidates[0]?.id, r.local).toBe(r.target);
-    }
-  });
+  // Acceptance §"existing rows still resolve" (claude-opus-5-fast, Qwen/Qwen3-32B-TEE,
+  // zai-org/GLM-5.2-TEE) is already covered verbatim by the BET-1313 block's
+  // "the single-repo path and the other layers are undisturbed" test — not
+  // duplicated here to keep the duplication gate clean.
 });
 
 // Assert an ambiguous result surfaces every sibling size of the family.
--- a/src/shared/modelCatalog.match.test.ts
+++ b/src/shared/modelCatalog.match.test.ts
@@ -340,6 +340,72 @@ describe("BET-1313 — repo-mates of one weights repo resolve to the entry the i
   });
 });
 
+// BET-1314 — the corroboration context veto (BET-1303 5.4c) fires on the raw
+// endpoint-vs-catalogue comparison, which mistakes two spellings of the SAME
+// window for an over-claim (128K = 2^17 = 131072 binary vs the catalogue's
+// 128000 decimal — ratio 1.024). The veto must fire only on a MATERIAL
+// over-claim, so the live `deepseek-ai/DeepSeek-V3.2-TEE` endpoint resolves to
+// `deepseek-v3.2` instead of being discarded in favour of the wrong repo-mates.
+describe("BET-1314 — the context veto tolerates representational noise", () => {
+  const matcher = createModelIndex(FIXTURE);
+
+  it("observed live case: endpoint 131072/65536 for the 128K v3.2 entry → exact, not ambiguous", () => {
+    const res = matcher.matchModel("deepseek-ai/DeepSeek-V3.2-TEE", {
+      modalities: ["text"],
+      context: 131072,
+      output: 65536,
+    });
+    expect(res.kind).toBe("exact");
+    expect(res.candidates[0]?.id).toBe("deepseek/deepseek-v3.2");
+  });
+
+  it("a genuine over-claim must still veto: 1M context for a 128K entry is never a confident match", () => {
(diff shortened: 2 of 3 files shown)


antoinedc:
## Review — BET-1314: context veto fires on decimal-vs-binary representation

### Mechanical gate
- Base freshness: clean. `merge-base == origin/main == dd9ce6fe`; no phantom deletions. The PR genuinely changes 3 files (67 insertions / 4 deletions).
- `npm run typecheck` → exit 0 (both tsconfigs).
- `npm test` → `vitest run` 3572 pass / 7 skip; node:test 2694 pass / 0 fail. Exit 0.
- CI: `typecheck-test` SUCCESS, `duplication-gate` SUCCESS.
- Anti-spaghetti (advisory report + my own run of `scripts/check-duplication.sh origin/main`): **0 clones**. `duplication-gate.sh`: PASS.

### Acceptance criteria, item by item
1. **typecheck+test green incl. new rows** — ✅ verified above; the BET-1314 describe block ran (live case → exact, over-claim → not exact, fractional ctx/output → not vetoed).
2. **Live endpoint resolves exact to `deepseek/deepseek-v3.2`** — ✅ independently reproduced. I ran the matcher against the fixture (which now carries the real catalogue facts `128000/64000`) with the endpoint's facts:
   ```
   deepseek-ai/DeepSeek-V3.2-TEE  {context:131072, output:65536} => exact ["deepseek/deepseek-v3.2"]
   chutes/deepseek-ai/DeepSeek-V3.2-TEE {context:131072, output:65536} => exact ["deepseek/deepseek-v3.2"]
   ```
   The `chutes/` weights-repo prefix is stripped, so the exact id wording from acceptance #2 resolves identically. ⚠️ Nit: the literal **live-box census line** (exact 46 → 47) needs to be produced/confirmed on the dev box at merge/verification time — it requires the live production catalogue, which no CI/review environment holds. Functional behavior is verified; this is an operational confirmation, not a code defect.
3. **Option + safety rationale** — ✅ Option A (`TRUST_MARGIN = 1.025`) stated; I verified the margin is safe: `131072 ≤ 128000 × 1.025 = 131200` (no veto), and a genuine over-claim fails the guard (`1000000 > 134400` → `deepseek-v3.2` vetoed — my run returns `ambiguous` with v3.2 **not** first, the honest repo-mate fallback). The guard the tolerance exists to preserve is intact.
4. **Source gate / no dependency / line delta** — ✅ no vendor or model names anywhere in `modelCatalog.mjs` (only the constant + comment; verified by grep). No new dependency. `modelCatalog.mjs` delta is a few lines (+8/−1).

### Scope & constraints
- Touches only `corroborate`'s ctx comparison; admission/tokenizer/classifier/structural steps untouched. Confirmed.
- Output is not a veto (only the +1 equality tiebreak) — correctly left alone, per the issue.
- No scope creep, no under-delivery. Fixture update carries the real catalogue facts and is test-only (`modelCatalog.fixture.json` is referenced only by tests, not the runtime catalogue).

### Nit (non-blocking)
- Acceptance #2's literal census line (exact 46 → 47) should be confirmed against the live box catalogue before/at merge — the functional fix is verified above; this is the remaining operational verification step.

No code changes required.

PR #1324, opened 2026-08-22
Settings accounts: shared inline confirm, hide unreachable Disconnect (BET-1320)

antoinedc:
## BET-1320 — Settings accounts: shared inline confirm, drop dead restart call, hide the Disconnect Manta can't deliver

Implements the renderer half of the BET-1318/1319 server behaviour. **Base: origin/main @ `299c387f`**.

### What changed

1. **`ConfirmInline.tsx` (new)** — the shared destructive inline confirm (danger-toned `busy ? "…" : label` + ghost Cancel), extracted verbatim from the account row's disconnect branch. Used for **both** the subscription Disconnect and the custom-endpoint Remove (which previously had no confirm at all — and now restarts opencode on the box, so it must ask first). Each confirm row additionally spells out the consequence inline: *"This restarts opencode and interrupts every running session."*
2. **Dropped the dead restart call** from `removeEndpoint` only (the server restarts opencode itself as part of the remove op — raising the shared restart banner would ask the user to do something already done). **Kept** it in `toggleModel` (upsert, server does not restart).
3. **Claude's Disconnect is no longer rendered.** The `opencode-claude-auth` plugin re-syncs `~/.claude/.credentials.json` every 5 min, so a delete would be undone — Manta can't deliver it. Data-driven: registry `managedBy` on the anthropic entry, pure `subscriptionStatuses(connected, externallyManaged)` second argument, an rpc status-branch detect-path probe (via the shared `expandTilde`), new `SubscriptionStatus` fields, and the row shows *connected · signed in with the Claude CLI on this box* with no Disconnect button.

### Adaptation note

The issue references `describeSubscriptionStatus` in `chatUtils.ts`, but that function was removed during the BET-1250 `SubscriptionsCard`→`AccountsCard` consolidation. I put the new copy in the existing pure describer for this row, `describeAccountState` in `AccountsCard.tsx` (unit-tested alongside its other branches) rather than resurrecting a dead function.

### Verification results

- PR branch (`multica/BET-1320-shared-inline-confirm` @ `ac2ad741`):
  - test: vitest 3577 pass / 7 skip (189 files), server 2717 pass / 0 fail. Failures: none.
- Base (`main` @ `299c387f`): not re-run separately (PR branch is branched from `origin/main` and carries zero unrelated changes).

Follow-up filed: none.

### DIFF
--- a/src/renderer/AccountsCard.test.ts
+++ b/src/renderer/AccountsCard.test.ts
@@ -88,6 +88,23 @@ describe("describeAccountState", () => {
     expect(st.tone).toBe("quiet");
   });
 
+  it("externally-managed connected row names the owner instead of usage (9g BET-1320)", () => {
+    const st = describeAccountState(
+      subscription({
+        managedExternally: true,
+        managedBy: "the Claude CLI on this box",
+        reading: { label: "5h window", pct: 41, pace: "under pace" },
+      }),
+    );
+    expect(st.text).toBe("connected · signed in with the Claude CLI on this box");
+    expect(st.tone).toBe("ok");
+  });
+
+  it("an externally-managed row without a managedBy owner falls back to today's text", () => {
+    const st = describeAccountState(subscription({ managedExternally: true }));
+    expect(st.text).toBe("No usage data");
+  });
+
   it("out-of-credit shows 'Out of credit' (danger)", () => {
     const st = describeAccountState(subscription({ health: "out-of-credit", balance: 0 }));
     expect(st.text).toBe("Out of credit");
--- a/src/renderer/AccountsCard.test.tsx
+++ b/src/renderer/AccountsCard.test.tsx
@@ -241,6 +241,98 @@ describe("AccountsCard Try-again tone by outcome (9c)", () => {
   });
 });
 
+describe("AccountsCard Disconnect gating (9g BET-1320)", () => {
+  let h: Harness | null = null;
+  afterEach(() => {
+    h?.unmount();
+    h = null;
+  });
+
+  function setupManaged(statuses: Record<string, unknown>[]) {
+    return installMockApi({
+      opencodeProviderAuth: () =>
+        Promise.resolve({ action: "status" as const, providers: statuses }),
+      opencodeGetProviders: () => Promise.resolve([]),
+      accountHealth: () => Promise.resolve({}),
+      configGet: () => Promise.resolve({}),
+      opencodeSetProviders: () => Promise.resolve({ ok: true }),
+      opencodeDiscoverModels: () => Promise.resolve({ ok: true, models: [] }),
+      accountsRetry: () => Promise.resolve({ ok: true, state: "ok", message: "x" }),
+    });
+  }
+
+  it("a plain connected row keeps a Disconnect button", async () => {
+    setupManaged([{ id: "anthropic", label: "Claude", plan: "Max 20x", console: null, docs: "", connected: true }]);
+    h = mount(<AccountsCard />);
+    await h.flush();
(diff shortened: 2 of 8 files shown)


antoinedc:
## Review — BET-1320 (manta-reviewer)

**Verdict: PASS** — 0 Blocks, 0 Questions. Routing to manta-pm for merge.

- merge-base `HEAD ↔ origin/main`: `299c387f` — matches the PR's declared base. Current `origin/main` tip `fac2a130`, so the PR is branched from `origin/main` (the BET-1319 merge commit). No phantom deletions vs fresh main (`git diff --stat origin/main...HEAD` shows only the 8 in-scope files).
- Both prereq server behaviours are merged into the base: BET-1318 (`b4503a92` remove-restart) and BET-1319 (`5afe8c8a` disconnect-restart). ✓

### Mechanical gate
- `npm test` → vitest **3577 pass / 7 skip** (189 files); server **2717 pass / 0 fail**; overall exit 0.
- Matches the implementer's reported results exactly (trust-but-spot-check held).

### Acceptance-criteria checklist (verified against the code, not the prose)

1. **Shared ConfirmInline** — `ConfirmInline.tsx` extracted with the exact required signature and the verbatim danger-button + ghost-Cancel markup, no new styling/tokens. Used for both the subscription **Disconnect** and the custom-endpoint **Remove** (the ✕ previously had no confirm at all). Each confirm row shows the consequence line *"This restarts opencode and interrupts every running session."* as secondary text. `removeConfirmId` state mirrors the disconnect pattern, cleared on confirm and cancel. ✓
2. **Drop dead restart call** — `setOpencodeRestartNeeded(true)` removed from `removeEndpoint` only; **kept** in `toggleModel` (upsert — server doesn't restart); shared banner + prop in `Settings.tsx` untouched. ✓
3. **Hide Claude's Disconnect (data-driven)** — registry `managedBy: "the Claude CLI on this box"` on the anthropic entry only; pure `subscriptionStatuses(connected, externallyManaged = new Set())` sets `managedExternally`/`managedBy`; rpc status branch builds the set with `existsSync(expandTilde(d))` using the shared `expandTilde` (no second expander); `SubscriptionStatus` gains both fields; the pure describer returns `connected · signed in with the Claude CLI on this box` for the connected+externally-managed+owned case and today's text otherwise; the row renders **no control** for that state. ✓

### Test coverage
- `subscriptionProviders.test.mjs`: set+registry flags and the omitting-arg → all-false case.
- `AccountsCard.test.ts` (`describeAccountState`): new branch + unchanged branches.
- `AccountsCard.test.tsx`: no *Disconnect* button for connected+managedExternally row, one for plain connected row (harness shape matches the established `Settings.test.tsx`/`AccountsCard` style).
- Remove-confirm test: ✕ does not call `opencodeSetProviders` until the confirm is pressed, and Cancel abandons without writing.

### Adaptation note (verified)
BET-1250's consolidation is real: no `SubscriptionsCard.tsx` / `ProvidersCard.tsx` / `describeSubscriptionStatus` exist on `main`; the current equivalents are `AccountsCard.tsx` + `describeAccountState`. The implementer's placement of the new copy there is the correct mapping, not a scope drift — the diff stays within the issue's intent and the out-of-scope files are untouched (no connect-flow / `ConnectProvider` / shared-banner changes).

### Anti-spaghetti
The only jscpd clone is mock-fixture setup boilerplate repeated within `AccountsCard.test.tsx` (two `installMockApi` blocks). Test-fixture duplication — discounted. No duplication of business logic introduced.

### Nits (non-blocking)
- None substantive.

No follow-up issues. Full review on the PR satisfies all acceptance criteria and the done-when gates.

PR #1328, opened 2026-08-23
fix(main): harden windows against navigation and external launch (BET-1324)

antoinedc:
## BET-1324 — Desktop hardening: will-navigate deny, openExternal allowlist, call.html CSP, frame-src

**Base: origin/main @ `fe27b4eb`**

Four independent security gaps in the desktop app's window layer, fixed ahead of inline widgets shipping:

1. **will-navigate deny** — both windows (`src/main/index.ts`, `src/main/callWindow.ts`) now deny any top-level navigation away from the app's own URL via a shared `shouldAllowNavigation(currentUrl, targetUrl)` helper. The renderer is a single-page app that never top-level-navigates, so only an exact match of the current URL (i.e. reload) is allowed; `https://evil.example`, `file:///etc/passwd`, and differing same-origin paths are all denied (fail closed on unparseable strings).
2. **openExternal allowlist** — `shell.openExternal` is only invoked for `http:`/`https:` URLs via shared `externalUrlOrNull(raw)`; `file:`, `smb:`, `javascript:`, custom schemes and unparseable garbage are dropped (return null) instead of being handed to the OS.
3. **call.html CSP** — `src/renderer/call.html` now carries the identical CSP string as `index.html` (continuation of the existing `frame-src` addition). Both share one `<meta http-equiv="Content-Security-Policy">` content value.
4. **frame-src** — `frame-src https:` added to the renderer CSP so the desktop widget renderer can frame the box-served URL. The box origin (`https://<box_id>.boxes.mantaui.com`) is per-install and can't be enumerated in a static CSP, so we permit `https:` for `frame-src` **only** — `default-src`, `script-src`, `connect-src` are deliberately not widened. Documented in a comment in `index.html`.

The helper lives in one module (`src/main/windowSecurity.ts`) used by both windows rather than a third copy of the duplicated `webPreferences`/`setWindowOpenHandler` blocks. Both helpers are pure — no Electron import — and unit-tested in `src/main/windowSecurity.test.ts`.

Out of scope per the issue, left untouched: `sandbox: false` on both windows, `webviewTag`, and the dead Swift `TerminalWebView` method.

**Files changed:** 7 files.
- `src/main/windowSecurity.ts` (new) — shared pure helpers
- `src/main/windowSecurity.test.ts` (new) — 14 tests
- `src/main/index.ts` — wire will-navigate denial + openExternal allowlist
- `src/main/callWindow.ts` — same wiring for the call window
- `src/main/callWindow.test.ts` — add `on()` to the fake `webContents` (real `webContents.on` supports it)
- `src/renderer/index.html` — add `frame-src https:`
- `src/renderer/call.html` — add identical CSP

**Verification results:**
- PR branch (`multica/BET-1324-windows-security`):
  - test: exit 0. vitest 3581 passed / 0 failed (includes 14 new windowSecurity tests + updated callWindow tests); node:test 2712 passed / 0 failed.
- Base (`main` @ `fe27b4eb`):
  - not separately re-run (PR branch is based directly on origin/main; no pre-existing failures were observed on the full run).
- **Conclusion:** 0 new failures. All suites green.

### DIFF
--- a/src/main/callWindow.test.ts
+++ b/src/main/callWindow.test.ts
@@ -79,7 +79,7 @@ const bwState = vi.hoisted(() => {
       for (const fn of this.listeners.get(event) ?? []) fn();
     }
     get webContents() {
-      return { setWindowOpenHandler() {}, loadURL() {}, loadFile() {} };
+      return { on() {}, setWindowOpenHandler() {}, loadURL() {}, loadFile() {} };
     }
   }
   return { BrowserWindow };
--- a/src/main/callWindow.ts
+++ b/src/main/callWindow.ts
@@ -15,6 +15,10 @@
 
 import { BrowserWindow, shell } from "electron";
 import { join } from "node:path";
+import {
+  shouldAllowNavigation,
+  externalUrlOrNull,
+} from "./windowSecurity.js";
 
 // A short-lived "opening" size if we have no saved bounds.
 const DEFAULT_BOUNDS = { x: undefined, y: undefined, width: 380, height: 560 };
@@ -66,8 +70,16 @@ export function createCallWindowController(deps: Deps): CallWindowController {
       },
     });
 
+    // BET-1324: same window-hardening as the main window (see windowSecurity.ts)
+    // — deny navigation away from the app's own page and only openExternal http/https.
+    callWindow.webContents.on("will-navigate", (event, url) => {
+      if (!shouldAllowNavigation(callWindow!.webContents.getURL(), url)) {
+        event.preventDefault();
+      }
+    });
     callWindow.webContents.setWindowOpenHandler(({ url }) => {
-      shell.openExternal(url);
+      const external = externalUrlOrNull(url);
+      if (external) shell.openExternal(external.href);
       return { action: "deny" };
     });
 
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -38,6 +38,10 @@ import {
 } from "./screenshotDetector.js";
 import { checkForUpdates } from "./autoUpdate.js";
 import { titleBarOptions } from "./windowChrome.js";
+import {
+  shouldAllowNavigation,
+  externalUrlOrNull,
+} from "./windowSecurity.js";
 import { downloadFileToDownloads } from "./download.js";
 import { createCallWindowController } from "./callWindow.js";
 import { registerInstallerHandlers } from "./installer/handlers.js";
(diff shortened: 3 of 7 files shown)


antoinedc:
## Review — BET-1324 (Desktop hardening)

**Result: PASS — 0 Blocks, 0 Questions.** Routing forward for merge.

**Base-freshness:** merge-base == `origin/main` (`fe27b4eb`); `git diff --stat origin/main...HEAD` shows no phantom deletions against fresh main.

**Acceptance-criteria check (issue checklist → verified):**

1. **Shared window-security helper used by both windows** — ✅ `src/main/windowSecurity.ts` `shouldAllowNavigation(current, target)` + `externalUrlOrNull(raw)`, both pure (no Electron import), wired into `will-navigate` (deny) and `setWindowOpenHandler` (openExternal allowlist) in **both** `src/main/index.ts` and `src/main/callWindow.ts`. One implementation, no triplication of the webPreferences/open-handler blocks.
2. **call.html has the same CSP as index.html** — ✅ byte-identical CSP string in both files (verified programmatically), both carrying `frame-src https:`.
3. **index.html frame-src permitting box origin, no other directive widened** — ✅ only `frame-src https:` added to a previously directive-less `frame-src`; `default-src`/`script-src`/`connect-src` untouched. Comment explains the per-install box origin.
4. **Pure helpers unit-tested** — ✅ `windowSecurity.test.ts` (14 tests): nav-to-app allowed; `https://evil.example` denied; `file:///etc/passwd` denied; different same-origin path denied; fail-closed on unparseable. `externalUrlOrNull`: `http`/`https` return URL; `file:`, `smb:`, `javascript:`, custom scheme, garbage, non-string → `null`.
5. **typecheck && test green** — ✅ verified on PR branch:
   - `npm test` → vitest **190 files, 3586 passed / 0 failed / 7 skipped**; node:test **2712 passed / 0 failed** (base runs also green; no new failures introduced).

**Explicitly-out-of-scope guard:** `sandbox`, `webviewTag`, and the dead Swift `TerminalWebView` method all left untouched — confirmed, none appear in the diff.

**Duplication (jscpd, anti-spaghetti):** no duplicates found in changed scope. The two CSP strings don't register as clones (HTML attribute, not code) and are an intentional mirror, not duplication.

**Notes (nits, non-blocking):** none material. Exact-match navigation allow is intentionally conservative and matches the "allow only the app's own URL" acceptance criterion.

PR #1334, opened 2026-08-23
BET-1330: dedupe boxToken()/authHeaders() across docs/opencode-tools/*.ts

antoinedc:
## What

Consolidate the box-identical `boxToken()` / `authHeaders()` helpers copy-pasted across **all 17** manta opencode tools (`docs/opencode-tools/*.ts`) into **one shared source**: `docs/opencode-tools/manta-auth.ts`. Every tool now does:

```ts
import { boxToken, authHeaders } from "./manta-auth";
```

(17 tool files share the duplication — the issue said 16; `docs/opencode-tools/` actually contains 17 tool files, all of which carried the helpers.)

## Import-resolution constraint (the BET-1329 blocker)

opencode resolves a tool's imports relative to the file's **real path** under `~/.config/opencode/tools/`, which has **no `node_modules`**. So `manta-auth.ts` must be **COPIED alongside** every tool that imports it — the documented `cp` install now includes the extra `cp <repo>/docs/opencode-tools/manta-auth.ts ~/.config/opencode/tools/manta-auth.ts` line. Install instructions were updated in **every** tool header, plus `docs/opencode-tools/AGENTS.md` and the root `AGENTS.md`, and a shared-auth note was added so future tools import rather than re-copy (the exact regression this prevents).

## Registration verification (the correctness gate)

Verified through **opencode's real loader** (Bun resolver, `opencode 1.18.21`), not just typecheck:

- Copied all 17 modified tools + `manta-auth.ts` into an isolated `~/.config/opencode/tools/` dir (the exact documented install) and booted a fresh opencode in isolation.
- A probe tool using the extensionless `./manta-auth` import **resolved and executed** at module load (`authHeaders({a:1})` returned `{"content-type":"application/json",…}`).
- A broken-import control (`./does-not-exist-xyz`) produced the documented failure signature `Cannot find module './does-not-exist-xyz'` — the "tool silently never registers" mode this issue is about.
- **None of the 17 tools hit `Cannot find module './manta-auth'`** — the only resolution error in the run was the deliberate control. All 17 load cleanly.

## Self-check

- CRITERION 1: all duplicated `boxToken()`/`authHeaders()` consolidated into one shared source → **10/10**. `grep -rn "function boxToken\|function authHeaders" docs/opencode-tools/*.ts` now returns only `manta-auth.ts`; all 17 tools import `./manta-auth`.
- CRITERION 2: shared helper resolves exactly the way the per-tool copy does today (copied alongside) → **9/10** — verified end-to-end through the real opencode loader above.
- CRITERION 3: all 17 tools still register → **9/10** — verified: no module-resolution failure for any tool in the real-loader run; good/negative control confirmed the mechanism.
- CRITERION 4: install instructions updated so fresh installs don't break → **9/10** — every tool header + both AGENTS.md files now carry the `manta-auth.ts` copy line.
- CRITERION 5: every actionable follow-up filed, none narrated → **10/10** — none surfaced (the related `schedule.ts` stale `ln -sf` instruction was corrected in-pass).

**Base: origin/main @ 70e308ad**

**Files changed:** 20 files (17 tool `.ts` + new `manta-auth.ts` + `docs/opencode-tools/AGENTS.md` + `AGENTS.md`). 100 insertions(+), 379 deletions(-) once the new file counts.

**Verification results:**
- PR branch (`multica/BET-1330-dedupe-auth-helpers` @ `aa88197d`):
  - typecheck: exit `0`. Errors: none. (`npm run typecheck`)
  - test: exit `0`, 2733 pass / 0 fail / 0 skip. Failures: none. (`npm test`)
  - registration: exit `0` — all 17 tools loaded cleanly through opencode 1.18.21's real loader (see above).
- Base (`main` @ `70e308ad`): not separately re-run — this diff touches only files outside `tsconfig.node.json`/`tsconfig.web.json` include globs (`docs/opencode-tools/*.ts` + `AGENTS.md`), so the typecheck/test result is definitionally identical between branches. The registration check is the gate that matters here and it passed on the PR branch.
- **Conclusion:** 0 new failures. No source (`src/`) or test files were touched.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -442,6 +442,8 @@ require-confirm toast the user hasn't answered isn't re-offered every 3s.
   `~/.manta-outbox/<sessionID>/` (workspace-linked) and hands it a TTL (default
   7 days, or `ttlHours:0` for never). Install on the remote opencode host:
   `cp <repo>/docs/opencode-tools/send-file.ts ~/.config/opencode/tools/send-file.ts`
+  plus the shared `manta-auth.ts` module it imports
+  (`cp <repo>/docs/opencode-tools/manta-auth.ts ~/.config/opencode/tools/manta-auth.ts`),
   then restart `opencode-serve`. The legacy `/send-file` markdown command
   (`docs/opencode-commands/send-file.md`) still works for a bare `cp` to the
   root, but that path is NOT workspace-linked.
@@ -713,7 +715,10 @@ the reusable "MantaUI tools" pattern (for future tools like `ping`) is in
 
 - **The AI's awareness comes from a GLOBAL opencode custom tool**, not MantaUI code.
   `docs/opencode-tools/schedule.ts` is **COPIED** (not symlinked) into
-  `~/.config/opencode/tools/schedule.ts` on the box; opencode auto-loads it for
+  `~/.config/opencode/tools/schedule.ts` on the box — **alongside the shared
+  `manta-auth.ts` module** it imports
+  (`docs/opencode-tools/manta-auth.ts` → `~/.config/opencode/tools/manta-auth.ts`);
+  opencode auto-loads it for
   EVERY project/session/model. Multiple named exports → tools `schedule_create`,
   `schedule_list`, `schedule_cancel`. A guidance blurb appended to
   `~/.config/opencode/AGENTS.md` (from `docs/opencode-tools/AGENTS.md`) tells the
@@ -773,7 +778,8 @@ generates on the box. Follows the same "MantaUI tools" pattern as the scheduler
 
 - **Global opencode tool**, `docs/opencode-tools/serve-page.ts`, **COPIED** (not
   symlinked — same `@opencode-ai/plugin` import-resolution gotcha as schedule)
-  to `~/.config/opencode/tools/serve-page.ts`. Three named exports → tools
+  to `~/.config/opencode/tools/serve-page.ts`, **plus the shared `manta-auth.ts`
+  module it imports** (`→ ~/.config/opencode/tools/manta-auth.ts`). Three named exports → tools
   `serve_page`, `stop_page`, `list_pages`. Guidance appended to
   `~/.config/opencode/AGENTS.md` from `docs/opencode-tools/AGENTS.md`.
   **Install/update = `systemctl --user restart opencode-serve`.**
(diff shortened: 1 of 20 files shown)


antoinedc:
## Review — BET-1330 (dedupe `boxToken()`/`authHeaders()` across `docs/opencode-tools/*.ts`)

**Verdict: PASS — 0 Blocks, 0 Questions.** Routing to `manta-pm` for merge.

### Base freshness (TEN-134)
`git merge-base HEAD origin/main` == `origin/main` tip (`70e308ad`) — PR branches off current main, no stale-base phantom deletions.

- **ONE shared source** ✅ — new `docs/opencode-tools/manta-auth.ts` is the single implementation; it is byte-identical to the deduplicated code (only adds `export`). Repo-wide grep confirms **no** `function boxToken`/`function authHeaders` remains in any tool file.
- **Complete coverage** ✅ — the issue said 16 tools; `git grep` on base main shows **17** files actually carried the duplicate (`media.ts` was omitted from the issue's list but did have the helpers). All 17 now `import { boxToken, authHeaders } from "./manta-auth"` and all 17 install-header comments carry the `cp .../manta-auth.ts` line. No under-delivery, no leftover copy.
- **Import-resolution constraint** ✅ — the shared helper is **copied alongside** (the only pattern that resolves under `~/.config/opencode/tools/` with no `node_modules`); extensionless `./manta-auth` matches how each per-tool copy resolved its own imports. Install instructions updated in **every** tool header, `docs/opencode-tools/AGENTS.md`, and root `AGENTS.md`, plus a "do not hand-copy, import instead" note for future tools. `schedule.ts`'s stale `ln -sf` header (contradicting copy-never-symlink) is corrected to `cp` — good catch.
- **All tools still register** ✅ — independently verified: `tsc --noEmit` on all 17 tools + `manta-auth.ts` against `@opencode-ai/plugin` types → clean (only pre-existing `cto.ts` typing noise in code this PR does not touch). Implementer's real-loader probe (opencode 1.18.21) confirmed all 17 load with the good/negative control showing the `Cannot find module` silent-fail signature.

### Mechanical gate
- `npm run typecheck` → exit 0 (project tsconfigs exclude `docs/opencode-tools/*.ts`, so this validates the untouched `src/` only; the tool files are gated by the registration check above).
- `npm test` → exit 0, **2733 pass / 0 fail / 0 skip**.
- Anti-spaghetti report: flagged clones are the pre-existing thin-registrar `call(method, path, body)` fetch wrapper (schedule vs webhook etc.), not the auth helpers and not newly-introduced business logic — no action.

### Nits (non-blocking)
- Issue body undercounted the scope (16 vs actual 17 including `media.ts`); the PR correctly covers all 17. Informational so the record is accurate.
- `cto.ts` has pre-existing `Property 'ok' does not exist on type '{}'` (lines 82–85, the `CTO_TOOLS` execute body) — present in unchanged code, a plugin-type mismatch; the project doesn't typecheck `docs/` so it never surfaces in CI. Fold this into BET-1331's registration regression check.
- The unrelated `authHeaders` in `src/renderer/api/httpApi.ts` and `src/server/forge/index.mjs` are different functions (different stores/signatures) and out of this issue's documented `docs/opencode-tools/*.ts` scope — correctly not consolidated here.

Good scope discipline: only `docs/opencode-tools/*.ts` + `manta-auth.ts` + both `AGENTS.md` changed; no `src/` or tests touched. Follow-up BET-1331 (CI regression check for tool registration) correctly filed rather than narrated.

PR #1356, opened 2026-08-25
feat(ios): make transcript text selectable via MarkdownText renderer (BET-1352)

antoinedc:
Opened automatically for `multica/BET-1352-ios-transcript-selection`, which was pushed without a pull request.

Relates to BET-1352.

### DIFF
--- a/mobile/native/MantaUI/TranscriptComponents.swift
+++ b/mobile/native/MantaUI/TranscriptComponents.swift
@@ -27,6 +27,7 @@ struct UserBand: View {
 
     var body: some View {
         Text(text)
+            .textSelection(.enabled)
             .font(.manta(size: Metrics.type.body, weight: mantaFontWeight(Metrics.type.medium)))
             .foregroundColor(tokens.tx1)
             .lineSpacing(pointsFor(multiplier: 1.5, size: MantaDynamicType.scaled(Metrics.type.body)))
@@ -69,7 +70,7 @@ struct MantaProse: View {
     let tokens: Tokens
 
     var body: some View {
-        MarkdownView(text)
+        MarkdownText(text)
             .font(.manta(size: Metrics.type.body), for: .body)
             .font(.manta(size: Metrics.type.body + 4, weight: .semibold), for: .h1)
             .font(.manta(size: Metrics.type.body + 2, weight: .semibold), for: .h2)
@@ -122,6 +123,7 @@ struct LiveProseTail: View {
 
     var body: some View {
         Text(text)
+            .textSelection(.enabled)
             .font(.system(size: Metrics.type.body))
             .foregroundColor(tokens.tx1)
             .lineSpacing(pointsFor(multiplier: Metrics.type.proseLineHeight, size: Metrics.type.body))
@@ -409,6 +411,7 @@ struct StepRowView: View {
 
             if expanded, let output = step.output {
                 Text(ToolOutputPreview.tail(output))
+                    .textSelection(.enabled)
                     .font(.manta(size: Metrics.type.xs, design: .monospaced))
                     .foregroundColor(tokens.tx3)
                     .fixedSize(horizontal: false, vertical: true)


antoinedc:
## Review — BET-1352 (iOS transcript text selection)

Scope, base, and mechanical gates first, then the two gaps that prevent a sign-off.

### What checks out

- **Base-freshness (TEN-134):** `merge-base == origin/main == f9a9dae8`. No stale base, no phantom deletions.
- **Scope discipline:** the PR touches exactly one file — `mobile/native/MantaUI/TranscriptComponents.swift` (4 insertions, 1 deletion). No quote actions, no copy/share/context menu, no composer changes. Selection-only, as the issue demanded.
- **Item 1 (API exists):** confirmed independently by the fact that `ios-build` (GH Actions `macos-26` runner) **passes** — that job compiles the app + both test bundles, so `MarkdownText(_ text:)` and every carried modifier (`.font(_:for:)`, `.tint(_:for:)`, `.markdownTableStyle`) resolve at the pinned 3.0.0. No re-pin, no `project.yml` change.
- **Item 3 (plain-`Text` selection):** `.textSelection(.enabled)` is correctly added on `UserBand` (line 29), `LiveProseTail` (line 124), and the expanded step-row tool output (line 411).
- **Item 5 (selection-menu hook):** reported — the renderer exposes no hook for custom edit-menu items alert, and the underlying `UITextView`/delegate is internal to RichText. Valid, useful finding; the follow-up's decided fallback applies stub. (Live menu contents pending on-device.)
- **CI:** `ios-build` PASS, `swift-core-test` PASS, `swiftlint` PASS, `typecheck-test` PASS, `duplication-gate` PASS. Anti-spaghetti: no TS files changed, nothing to scan.

### Block (fix-here): required on-Mac verification gate not executed

The issue's **Verification** section is an explicit, measurable acceptance gate — build for the iOS Simulator, run the `MantaUITests` suite, launch in the simulator and **describe what you actually see** when selecting prose / a user message / tool output, and paste real output. That gate was **not run**. The hand-off confirms the Mac capability executor is down and no `ios-mantaui` job has processed since ~2 days.

This is not a formality. The issue states plainly this is a "build-look-adjust loop against a rendering library… the risk is entirely in what the swap costs visually," and that the Mac is the only agent that can see the result. Every behavior-dependent item is therefore asserted-not-shown:

- Row self-sizing under the text-layout renderer (mixed-content scroll).
- Survival of the custom `MantaMarkdownTableStyle` and the per-element `.font(_:for:)` slots.
- Code-block rendering/readability at the mono face.
- Scroll-back past `loadEarlier()` (the transcript-list crash family, `mobile/native/AGENTS.md` §6).
- Live selection-menu contents.

### Block (fix-here): item 4 gesture conflict unresolved in shipped code

The swipe-to-reveal-timestamps pan vs. text-selection conflict the issue requires resolving is **still present** — no gesture fix ships in this diff. The implementer stopped via the issue's carve-out, reasoning from source that the only workable fix is inside the pinned MessagingUI library. That conclusion may be right, but the required **reproduction** on device never happened, so the carve-out precondition ("the only workable fix requires a library change") is asserted, not shown. The described conflict remains live in the landed behavior.

### What's needed to unblock

Both Blocks trace to one root cause: the **Mac executor is down** — an infra state no agent (Linux box or reviewer) can restore; it needs the human to power on the Mac / fix the cert / reconnect. Once it's back:

1. Run the issue's Verification gate end-to-end (`ios-mantaui test-unit`, simulator build+launch) and report each observation (self-sizing, table style, code blocks, scroll-back crash, selection behavior, live menu).
2. Reproduce the gesture conflict on-device, then either fix it our-side or, if genuinely only fixable in the library, confirm that with the reproduction in hand and let the PM/human take the accept-vs-fork call.

Not mergeable as-is.

antoinedc:
## Review (cycle 2) — BET-1352 iOS transcript text selection

Re-review of #1356 after the on-Mac verification gate was executed. Both Block-severity findings from cycle 1 are now resolved with real, pasted on-Mac output rather than assertion.

### Status: PASS (0 Blocks, 0 Questions, 2 nits) → routing to manta-pm

### The two prior Blocks — now cleared

**Block 1 (required on-Mac verification gate) — CLEARED.** `MantaUITests` suite run and passed on the Mac: `642 tests, 0 failures`, `TEST EXECUTE SUCCEEDED`. Simulator observations all positive and pasted:
- Renderer swap confirmed via accessibility tree — the assistant turn renders as one `TextView` (MarkdownText/`UITextView`) spanning blocks, with the fenced code block and GFM table both present.
- Selection works: system edit menu with Copy on prose, in both the plain ScrollView scene and the live TiledView transcript.
- Step output / self-sizing: expanded output well renders as selectable mono text; rows size to content without clipping.
- Scroll-back past `loadEarlier()` (the transcript-list crash family): deep-scrolled a 44+ message fixture 14 swipes into old history and back; app stayed alive. The change alters row *content* only, not identity/ordering — as the diff confirms.
- Live selection-menu contents: standard system Copy / Select-All; no Quote/Share yet — consistent with the item-5 report.

**Block 2 (gesture conflict unresolved) — CLEARED.** The conflict was driven on-device: select prose, drag left to extend selection — the selection stays active, the row does not slide, gutter does not steal the gesture (`ZGR copy-after-extend=true`, `frame shift dx=0.0 dy=0.0`). The driver also corrected the issue's premise (the library's `shouldRecognizeSimultaneouslyWith` returns true only for the scroll + bottom-safe-area pans, not "every other recognizer"), reproduced the stated residual (`cancelsTouchesInView` default), and confirmed the only *clean* fix lies inside the pinned MessagingUI library. Per the issue's explicit carve-out ("if the only workable fix requires a change inside the library, stop and report it instead of doing it"), no library patch was made — the report is on-device, not source-assumed. No our-side gesture change ships, and none is needed in the tested scenarios.

### Scope / base / mechanical gates — re-verified this cycle

- **Scope discipline:** PR touches exactly one file — `mobile/native/MantaUI/TranscriptComponents.swift` (4 insertions, 1 deletion). Selection-only; no quote actions, no copy button, no share sheet, no context menu, no composer change. Matches the issue's scope section.
- **Base-freshness (TEN-134):** branch off `origin/main` merge-base; 3-dot diff against fresh `origin/main` shows only the PR's own single-file change. No stale base, no phantom deletions.
- **Item 1 (API exists):** `MarkdownText` resolves at pinned 3.0.0 — `ios-build` (macos-26 runner) passes, compiling the app + both test bundles with every carried modifier (`.font(_:for:)`, `.tint(_:for:)`, `.markdownTableStyle`) intact. No re-pin, no `project.yml` change.
- **Item 3 (plain-`Text` selection):** `.textSelection(.enabled)` verified on `UserBand` (:29), `LiveProseTail` (:124), and the expanded step-row output (:411).
- **Item 5 (selection-menu hook — required deliverable):** reported: no public hook on `MarkdownText`/RichText to add custom edit-menu items; the underlying `UITextView` delegate is set internally by RichText's Coordinator and not reachable from the app, so `textView(_:editMenuForTextIn:suggestedActions:)` cannot be hooked externally. Decision-managed fallback applies. Valid, useful result for the quote follow-up.
- **CI:** `ios-build` PASS, `swift-core-test` PASS, `swiftlint` PASS, `typecheck-test` PASS, `duplication-gate` PASS. Anti-spaghetti: no TS/module files changed → nothing to scan.

### Nits (non-blocking)

- **Table-style / font-slot enumeration was macro-level.** The issue asked to report *exactly* which of `MantaMarkdownTableStyle` and the per-element `.font(_:for:)` slots (lines 73–84) survive on the text-layout renderer. The on-Mac report confirms at the level of rendered output (table present, code block mono, rows self-size) but doesn't enumerate each font-slot individually. Given the table + code + self-sizing all render, the substantive "does the swap cost visually" question is answered; this is a report-granularity note, not a defect.
- **Bootstrap flake:** the first unit-run's "bootstrap" crash was a simulator boot flake, recovered on reboot; noted for flake-awareness in CI, not a blocker.

### Verdict

The renderer swap and the three `.textSelection` additions are exactly scoped and CI-proven; the on-Mac verification gate the issue mandates is now executed with positive behavioral evidence; the gesture-carve-out is confirmed on-device; and the required hook report is delivered. Clean PASS — routing to `manta-pm` for merge. I do not merge or flip from draft (PM's call).

PR #1357, opened 2026-08-25
iOS session list: terminal windows show poller running status (BET-1350)

antoinedc:
Opened automatically for `multica/BET-1350-ios-terminal-status`, which was pushed without a pull request.

Relates to BET-1350.

### DIFF
--- a/mobile/native/MantaUI/SessionListStore.swift
+++ b/mobile/native/MantaUI/SessionListStore.swift
@@ -109,6 +109,13 @@ final class SessionListStore: ObservableObject {
     /// running→idle — triggers a refresh that re-fetches the authoritative
     /// `time.updated` for the age chip at the exact moment it becomes eligible.
     private var previouslyRunning: Set<String> = []
+    /// `tmuxSession#windowIndex` -> live tmux-poller status for TERMINAL
+    /// windows (BET-1350). Chat-mode windows are NOT in here: their status
+    /// comes from the interpreted stream, and the box's pane scrape cannot see
+    /// them anyway (a chat window's pane runs `sleep infinity`, so
+    /// capture-pane returns blank). Read by `rowStatus` for terminal windows
+    /// only; one source per window type, no merging.
+    private(set) var terminalStatus: [String: WindowPollStatus] = [:]
     private var cancellables: Set<AnyCancellable> = []
 
     init(api: MantaAPIClient = MantaAPIClient.live(), eventStore: MantaEventStore, mutations: SessionListMutationAPI? = nil) {
@@ -118,6 +125,7 @@ final class SessionListStore: ObservableObject {
         self.loadConfig()
         self.eventStore.addRawFrameHandler { [weak self] frame in
             self?.trackProgress(frame: frame)
+            self?.trackStatus(frame: frame)
         }
         // A turn completing is the moment the age chip becomes eligible
         // (`!running`), and it is precisely when the cached `lastActivity` is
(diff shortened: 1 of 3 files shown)


antoinedc:
## BET-1350 review — iOS terminal (claude-TUI) windows show running status

**Verdict: PASS — 0 Blocks, 0 Questions, 0 Nits.** Routing to `manta-pm` for merge.

### Acceptance-criteria checklist (each verified against the code, not the prose)

1. **Consume `status` frames, single handler, no new callback** — ✅ `trackStatus(frame:)` is invoked inside the store's existing raw-frame-handler block (`SessionListStore.swift:128`); no second registration, no new `MantaEventStore` callback. Payload decoded as a JSON **array** (`[TerminalStatusFrame].self`), as the wire shape requires.
2. **Envelope `sessionId` object-guard** — ✅ verified in `MantaStreamModels.swift` `parse`: `rawPayload` is `[:]` when `payload` is not an object, so a `status` frame (array payload) safely resolves `sessionId` to nil. No change needed; correct.
3. **Store keyed by window, batch-replace, publish-on-change, clear on box change** — ✅ `terminalStatus: [String: WindowPollStatus]`; `WindowPollStatus` added in `SessionModels.swift`; key reused from existing `SessionRowEntry.id` (`"\(project)#\(windowIndex)"`) via pure `terminalKey` helper — no third id format. Batch copies the dict and overwrites only named windows (absent → keeps value); `objectWillChange.send()` only on an actual change (2s-tick full re-render guard honored); cleared in `resetForBoxChange()`.
4. **Use for terminal windows only, one source per window type** — ✅ `rowStatus(for:)` consults `terminalStatus` only when `opencodeSessionId == nil`; chat windows keep the stream path (test-proven unaffected). `attention` stays false for terminals (no question/permission concept). `runningCount(in:)` needs no change — reads `rowStatus`. No `applyStatusBatch`-style defence ported (correctly omitted — unnecessary here).
5. **Subtitle** — ✅ verified from precedence: a running terminal row (no progress label, no model label) falls into the `s.running` branch and reads exactly `"running"`; the `isTerminal → "terminal"` branch is only reached when not running. No edit needed.

### Out-of-scope discipline

Diff is exactly 3 files under `mobile/native/` (`SessionListStore.swift`, `SessionModels.swift`, `SessionModelsTests.swift`); zero `src/server/` changes. No scope creep, no under-delivery.

### Mechanical gates

- **Base freshness:** `git merge-base` = `f9a9dae8`; `git diff --stat origin/main...HEAD` shows only the 3 intended files, no phantom deletions. Fresh base — no stale-base issue.
- **JS suite** (unchanged by this iOS-only PR, run for completeness): `npm run typecheck` exit 0; `npm test` 2844 pass / 0 fail.
- **Swift unit tests (the actual merge gate)** — run by `macos` on Xcode 26.6 / iPhone 17 Pro simulator: new `TerminalStatusTests` 4/4 green, covering each acceptance point (real-shape decode keyed `proj#2`; terminal running → `.running` dot + `"running"` subtitle; chat window unaffected by a poller entry naming its index; later not-running batch flips back to idle). Full `MantaUITests`: 664 tests, 5 failures — all pre-existing + environmental (keychain `-34018`) in untouched files. Build + launch succeeded.
- **Live busy-TUI observation:** macos could not obtain one (no terminal window was mid-turn; none of the reachable boxes offered a busy TUI). It reported this honestly rather than fabricating. The three at-issue behaviors (accent dot, header count, idle flip-back) are deterministically covered by the green unit tests; `runningCount` → `rowStatus` means a running terminal contributes to the header chip by construction. Not a block.

### Anti-spaghetti

jscpd report: 0 TS/MJS files changed — nothing to scan. No new duplication or dead code.

PR #1361, opened 2026-08-26
BET-1358: Optimizer perf: push role filter + json_extract projection into fetchLedgerRows SQL

antoinedc:
## Summary

Optimizes the cold `optimizer:summary` / routing `ledgerSummary` read. Rewrote
`fetchLedgerRows` (src/server/modelLedger.mjs) from `assistantRows` (stream
full `data` blob → JSON.parse every message in JS → role-filter in a loop;
~3s on a cold 30-day read) to a single SQL `json_extract` projection that moves
the role filter + JSON parse into SQLite and streams only the flat ledger
fields. Output rows are key-for-key identical to the old JS-parse shape.

`assistantRows`, `collectToolParts`, `aggregate`, `aggregateEndpointStats`,
`ledgerSummary` are **untouched** — `assistantRows` stays for
`aggregateEndpointStats` (needs the raw `data` blob + `part` join for tool
reliability). Diff is confined to `src/server/modelLedger.mjs` +
`src/server/modelLedger.test.mjs`.

Two deliberate additions to the issue's verbatim SQL, both **required by the
issue's own acceptance criteria** (not free design choices):

- **`s.id AS session_id`** — the flat shape's `sessionID` field is consumed by
  `aggregateBySession`/the optimizer summary. The verbatim SQL omitted it,
  which would have dropped `sessionID` from every row and broken the
  "key-for-key identical output" requirement.
- **`AND json_valid(m.data) = 1`** — `json_extract` **throws** (does not return
  NULL) on malformed JSON, so without the guard a single bad row would fail the
  whole query and collapse the summary to `supported:false`. The guard
  pre-filters malformed rows so the old skip-the-row degradation is preserved
  (`json_extract` only ever runs on valid JSON). Verified against node:sqlite.

Degradation: malformed-JSON rows and non-assistant rows are excluded silently,
never throwing — same as the old path. Absent/null cost+token leaves fold to 0
via `num()`, identical to today. No schema change, no new deps, no state files.

## Performance gate

The structural change (project off the full blob, zero JS JSON.parse/filter)
is what the issue's own ~3s timing attributes the cost to. I could not measure
the cold path on this run box (no reference `opencode.db` here — the issue
says to trust the reference box's measured facts); the gate is for the
reference box to confirm `< ~1s`. Not committing a benchmark per the issue.

## Verification results

**Files changed:** `2` files. `src/server/modelLedger.mjs`, `src/server/modelLedger.test.mjs`

**Verification results:**
- PR branch (`multica/BET-1358-optimizer-prefetch-ledger` @ `be961245`):
  - test: exit 0, **2874 pass / 0 fail / 0 skip**. Failures: none. log sha256: `cd09bc42b112403fd6c8b80d22d9f1d0e6917cc4a97326ca56413c01904486fe`
- Base (`main` @ `a18851a9`):
  - test: exit 0, **2872 pass / 0 fail / 0 skip**. Failures: none. log sha256: `add0137f340b5de7d19e3bb49adf098dbdcdc7e5b5417b2267142e2df3d16023`
- **Conclusion:** 0 failures total. The PR adds 2 tests (2874 vs 2872) — the two new `fetchLedgerRows` parity tests; both pass. 0 new failures, 0 resolved, 0 pre-existing.

New tests:
- `fetchLedgerRows projected query deep-compares to the legacy JS-parse shape on well-formed rows` — seeds a fixture DB, runs the new projected query and a local legacy reference, `deepEqual` on the resulting arrays + asserts the user/malformed rows are excluded.
- `fetchLedgerRows: absent/null cost+token fields aggregate to 0 exactly as today; non-assistant and malformed rows excluded` — mixed fixture (multi-session, some rows without cost/tokens/time), asserts `aggregate(projected) deepEqual aggregate(legacy)` (byte-identical totals) plus exact totals and turn exclusion.

Both are DB-backed via `node:sqlite` (temp dir + read-only handle), `MANTA_STATE_HOME` sandboxed — no real box state written.

**Base: origin/main @ a18851a9**

### DIFF
--- a/src/server/modelLedger.mjs
+++ b/src/server/modelLedger.mjs
@@ -292,33 +292,64 @@ function tokensSent(row) {
 
 /**
  * I/O. The ONE ledger row query. Extracts the flat ledger row shape (the
- * `aggregate`/aggregate* fold inputs) from assistant messages over `sinceMs`,
- * via the shared SQL in `assistantRows`. Both `ledgerSummary` and the
- * optimizer summary consume this — a single query path, no duplicated SQL.
- * Returns the parsed, role-filtered flat rows. Never throws on a malformed
- * row (those are skipped); query/parse scaffolding lives in the callers.
+ * `aggregate`/aggregate* fold inputs) from assistant messages over `sinceMs`.
+ * Projected query: the role filter + JSON parsing run in SQLite via
+ * `json_extract`, so the SELECT streams only the fields the flat shape needs
+ * instead of the full `data` blob (which kept `assistantRows`'s parse+filter
+ * at ~3 s on a cold 30-day read). Returns the role-filtered flat rows, mapped
+ * into the exact same shape the old JS-parse path produced (same field names,
+ * same typing) so `aggregate`, `aggregateBySession`, `aggregateDailySeries`,
+ * `measureEffectiveTtl` and the routing fold consume them unchanged.
+ *
+ * `assistantRows` is deliberately NOT used here (it stays for
+ * `aggregateEndpointStats`, which needs the raw `data` blob + the `part` join
+ * for tool reliability). Degradation matches the old path: a row whose `data`
+ * is malformed JSON or lacks `role === "assistant"` is skipped silently. The
+ * `json_valid(m.data)` guard filters malformed rows out BEFORE `json_extract`
+ * runs (json_extract throws on invalid JSON rather than returning NULL, so the
+ * guard is what keeps a single bad row from failing the whole query); rows
+ * that are valid JSON but not an assistant message are excluded by the
+ * `= 'assistant'` predicate. Never throws. Query/parse scaffolding lives in
+ * the callers.
  */
 export async function fetchLedgerRows(db, sinceMs) {
   const since = num(sinceMs);
+  const sql = `
+    SELECT m.id AS msg_id,
+           s.id AS session_id,
+           s.parent_id AS parent_id, s.agent AS agent, s.directory AS directory,
+           json_extract(m.data,'$.providerID')           AS providerID,
+           json_extract(m.data,'$.modelID')              AS modelID,
+           json_extract(m.data,'$.cost')                 AS cost,
+           json_extract(m.data,'$.tokens.input')         AS input,
+           json_extract(m.data,'$.tokens.output')        AS output,
+           json_extract(m.data,'$.tokens.reasoning')     AS reasoning,
+           json_extract(m.data,'$.tokens.cache.read')    AS cacheRead,
+           json_extract(m.data,'$.tokens.cache.write')   AS cacheWrite,
+           json_extract(m.data,'$.time.created')         AS startedMs,
+           json_extract(m.data,'$.time.completed')       AS completedMs
+    FROM message m JOIN session s ON s.id = m.session_id
+    WHERE m.time_created >= ?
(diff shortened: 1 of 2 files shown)


antoinedc:
## BET-1358 Review — PASS (0 Blocks, 0 Questions)

Reviewed against the issue's acceptance criteria, not the summary. Base is fresh (`merge-base == origin/main @ a18851a9`, no stale-base phantom deletions). Diff is confined to the issue's stated scope: `src/server/modelLedger.mjs` + `src/server/modelLedger.test.mjs` (2 files, +213/−23).

### Mechanical gate
- `npm run typecheck` — exit 0, no errors. ✅
- `npm test` — **2874 pass / 0 fail / 0 skip** (matches the implementer's reported result). ✅
- Both new BET-1358 parity tests pass when run in isolation (`src/server/modelLedger.test.mjs`: 17/17). ✅

### Acceptance-criteria checklist (each verified independently)
1. **`fetchLedgerRows` switched to the SQL `json_extract` projection** — role filter + parse moved into SQLite, full `data` blob no longer streamed. Matches the issue's verbatim SQL. ✅
2. **Flat output shape preserved** — `sessionID`/`parentId` still `String()`-coerced or null; `agent`/`directory` still `?? null`; `cost`/token leaves still fold through `num()` so `aggregate`/`aggregateBySession`/`aggregateDailySeries`/`measureEffectiveTtl`/routing `ledgerToServices` consume rows unchanged. No `id` field added (issue explicitly said not to). ✅
3. **`assistantRows` untouched** — diff confirms it (still used by `aggregateEndpointStats`/`collectToolParts`, which need the raw blob). `aggregate`, `collectToolParts`, `ledgerSummary` wrapper unchanged. ✅
4. **Degradation matches old path** — malformed JSON and non-assistant rows are excluded silently, never throw. ✅
5. **Tests** — (a) projected shape deep-compares to a legacy JS-parse reference on well-formed rows; (b) mixed fixture asserts `aggregate(projected) deepEqual aggregate(legacy)` byte-identical totals, absent/null leaves → 0, non-assistant + malformed excluded. DB-backed via `node:sqlite`, temp dir + read-only handle, no box state. ✅
6. **Performance gate** — structural change matches the issue's timing attribution verbatim; cold timing itself is only measurable on the reference box, which the issue says to trust. Expected, not verified here (not a blocker). ✅

### The two additions to the verbatim SQL — correct, in-scope
- `s.id AS session_id`: required for key-for-key parity — the old path emitted `sessionID` (from `assistantRows`), so omitting it would have dropped the field and broken the "identical output" requirement.
- `AND json_valid(m.data) = 1`: required by the never-throw degrade rule — `json_extract` throws on malformed JSON, so the guard preserves the old skip-the-row behavior.
Neither touches `assistantRows` or any other function; no new deps, no schema change, no state files.

### Anti-spaghetti report
One clone (jscpd) is inside the **test file** — the reference/seed blocks for the parity tests. Test-fixture coincidence, not duplicated business logic this PR introduced. Discounted per the review rubric.

### Note
For well-formed rows the row arrays are deep-equal. For rows with absent token cost leaves, the projected path yields `null` where the old JS path yielded `undefined` — this is exactly what the issue spec blesses ("absent token leaves → NULL... identical to today") and folds identically through `num()` (0), so aggregate/optimizer output is byte-identical. Not a defect.

No blocking findings. Routing to the PM for merge.

PR #1362, opened 2026-08-26
fix(server): break optimizer:summary self-await deadlock (BET-1359)

antoinedc:
Fixes BET-1359 — `optimizer:summary` deadlocks forever because `readMeteredEndpoints` re-awaits the summary it is a dependency of.

## Root cause
`readMeteredEndpoints` (wired as `meteredEndpoints`) opened with `const summary = await optimizerSummary()`. When the summary build's memo missed, it assigned the `inflight` slot, then reached `metered: await meteredFor(meteredEndpoints)`, which re-entered `optimizerSummary()`. That re-entry found `cache` still null and returned the very `inflight` promise that was awaiting it — the build awaited itself, never settled, and `.finally()` never ran, pinning `inflight` for the process lifetime (the "Settings frozen" report). Not intermittent: the optimizer-tuner backstop poller primes it at boot.

## The fix
Break the cycle. `readMeteredEndpoints` only needed `windows` and `cacheShare`, both of which `buildOptimizerSummary` already computed. Changed the `meteredEndpoints` contract from `() => rows` to `({ windows, cacheShare }) => rows`:

- `summary.mjs` — hoist `windows` to a local and pass `{ windows, cacheShare }` through `meteredFor`.
- `index.mjs` — `readMeteredEndpoints` consumes the injected context; the `const summary = await optimizerSummary()` line is deleted. Default params keep it safe if ever called bare. Body from `denom` onward untouched.
- `summary.mjs` — added test-only `_resetSummaryMemo()` (mirrors `_resetDbHandle`).

No re-entrancy flag, no second memo slot, no watchdog — the cycle is gone.

## Tests
Six new node:test cases in `summary.test.mjs`, every one calling `_resetSummaryMemo()` first and the hang-prone one time-bounded (races against a 2s timer so a regression fails the suite rather than hanging CI).

## Maintainer note
Reload the server after merge (`systemctl --user restart opencode-serve`) — intentionally NOT run from this branch.

## Base: origin/main @ `4cebdf38`

**Files changed:** 3 files.
```
 src/server/index.mjs                  |  10 ++-
 src/server/optimizer/summary.mjs      |  21 +++--
 src/server/optimizer/summary.test.mjs | 148 ++++++++++++++++++++++++++++-
```

**Verification results:**
- PR branch (`multica/BET-1359-optimizer-summary-deadlock` @ `7c6763ae`):
  - test: exit 0, 2880 pass / 0 fail / 0 skip. Failures: none. log sha256: `789031a82dedc54c7fdb9ae43ded74de872744dd5bd323b5f3366b7ba5c0de45`
- Base (`origin/main` @ `4cebdf38`):
  - test: exit 0, 2874 pass / 0 fail / 0 skip. Failures: none. log sha256: `aa7a33f849bfd472d9cf39c612e229f52b5077c0028db51918452cac87220381`
- **Conclusion:** 0 failures pre-existing, 0 new failures. The +6 tests are the new BET-1359 pins.

### DIFF
--- a/src/server/index.mjs
+++ b/src/server/index.mjs
@@ -994,11 +994,13 @@ const optimizerActivity = createActivityLog({
 // with no price is skipped). A metered endpoint has no window and never resets,
 // so there is nothing to fill — the row is deliberately role+price, no gauge.
 // [] when the catalog or pricing is unavailable (the section is then absent).
-async function readMeteredEndpoints() {
+async function readMeteredEndpoints({ windows = [], cacheShare = {} } = {}) {
   try {
-    const summary = await optimizerSummary();
-    const subProviders = new Set((summary?.windows ?? []).map((w) => w.provider));
-    const cs = summary?.cacheShare ?? {};
+    // BET-1359: consume the windows/cacheShare the summary already computed,
+    // instead of re-awaiting optimizerSummary() — reading the summary it is a
+    // dependency of was a self-await deadlock.
+    const subProviders = new Set((windows ?? []).map((w) => w.provider));
+    const cs = cacheShare ?? {};
     const denom = (cs.input ?? 0) + (cs.output ?? 0) + (cs.cacheRead ?? 0) + (cs.cacheWrite ?? 0);
     const mix =
       denom > 0
--- a/src/server/optimizer/summary.mjs
+++ b/src/server/optimizer/summary.mjs
@@ -117,6 +117,8 @@ export async function buildOptimizerSummary({
     matched: verification ? verification.matched : null,
   };
 
+  const windows = windowsFor(usageSnapshots(), usageHistory(), nowMs, await pressureWindows());
+
   return {
     supported: true,
     windowDays: WINDOW_DAYS,
@@ -126,10 +128,10 @@ export async function buildOptimizerSummary({
     bySession,
     ttl,
     counterfactual: cf ? { dailySeries: cf.dailySeries, bySession: cf.bySession } : null,
-    windows: windowsFor(usageSnapshots(), usageHistory(), nowMs, await pressureWindows()),
+    windows,
     activity: await activityFor(activityStore),
     compaction: await compactionFor(compactionStat),
-    metered: await meteredFor(meteredEndpoints),
+    metered: await meteredFor(meteredEndpoints, { windows, cacheShare }),
   };
 }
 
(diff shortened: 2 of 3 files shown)


antoinedc:
## Review — BET-1359 optimizer:summary deadlock (PASS)

**Mechanical gates** (run on PR branch `multica/BET-1359-optimizer-summary-deadlock` @ `7c6763ae`):
- `npm test` → **2880 pass / 0 fail / 0 skip** (matches the PR body's reported PR-branch numbers exactly; no drift).
- Focused `node --test src/server/optimizer/summary.test.mjs` → 16/16 pass, including the 5 new BET-1359 pins. The time-bounded regression pin (`withTimeout`, 2000 ms) resolves rather than hanging.

**Base freshness:** merge-base = `origin/main` @ `4cebdf38` = current upstream tip. No stale base, no phantom deletions. `git diff origin/main...HEAD` touches exactly the 3 claimed files.

**Diff review against the issue's prescribed fix:**
- `summary.mjs`: `windows` hoisted to a local before the return; passed as `{ windows, cacheShare }` through `meteredFor(..., ctx)` → `meteredEndpoints(ctx)`. Returned-object key order unchanged. `cacheShare` already exists (from `aggregate(rows)`).
- `index.mjs`: `readMeteredEndpoints({ windows = [], cacheShare = {} } = {})` consumes the injected context; the `const summary = await optimizerSummary()` self-await is deleted. Body from `denom` onward byte-identical (same `mix` 0.5/0.5/0/0 fallback, same `rows.slice(0, 8)`, same `catch { return [] }`). Given `git diff src/server/index.mjs` touches only this function's signature + first three lines. The `meteredEndpoints: readMeteredEndpoints` wiring (`index.mjs:1068`) is shared by the RPC channel and the GET `/api/optimizer/policy` route, so both paths are un-deadlocked by the one change.
- `_resetSummaryMemo()` added, mirroring `_resetDbHandle` precedent, test-only.
- The `rpc.mjs` fallback (`rpc.mjs:437`) does not wire `meteredEndpoints` and is intentionally left untouched, per the issue.

The deadlock is structurally removed rather than masked — no re-entrancy flag, no second memo slot, no watchdog. The `.finally(() => inflight = null)` now always runs.

**Anti-spaghetti:** advisory report cross-read — 0 clones in the changed scope. Nothing to act on.

**Scope/hygiene:** diff confined to the 3 allowed files; no new deps/state/schema/poller; `TTL_MS` / `WINDOW_DAYS` untouched; no `fetchLedgerRows`/SQL changes. PR description records the `origin/main` base SHA and defers the `systemctl --user restart opencode-serve` reload to the maintainer (not run).

This is a plumbing change with no behaviour change to `metered` — the context carries the same `windows`/`cacheShare` objects the summary returns (pinned by test 3).

**Result: 0 Blocks, 0 Questions, 0 nits.** Route to merge.

PR #1367, opened 2026-08-27
feat(ios): Quote / Quote in new session on the system prose selection menu (BET-1364)

antoinedc:
Opened automatically for `multica/BET-1364-ios-transcript-quote-menu`, which was pushed without a pull request.

Relates to BET-1364.

### DIFF
--- a/mobile/native/MantaUI.xcodeproj/project.pbxproj
+++ b/mobile/native/MantaUI.xcodeproj/project.pbxproj
@@ -12,6 +12,7 @@
 		0488B71F794FC6E64CDC0B23 /* TerminalKeyRow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7D7BC8920829E8847057C8DA /* TerminalKeyRow.swift */; };
 		058E2179669599C216446E7C /* ToolOutputPreviewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DAD2C923C52DE07EB54AB821 /* ToolOutputPreviewTests.swift */; };
 		0DA0C1C92C06A3BED780494E /* UsageMeters.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1A81D1BB24FB02E4DFFA04FE /* UsageMeters.swift */; };
+		103C101C1679499BAFA7257B /* TranscriptSelectionMenu.swift in Sources */ = {isa = PBXBuildFile; fileRef = B43DC78B0971B2BF584DEA2C /* TranscriptSelectionMenu.swift */; };
 		12070C130BD9D1CD9EA0603F /* WaveformTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AEF7554D798247FECA6BFF84 /* WaveformTests.swift */; };
 		12C1D08D7311357D1872FA18 /* ChatJSON.swift in Sources */ = {isa = PBXBuildFile; fileRef = E246039C1A3AA38065BE89B0 /* ChatJSON.swift */; };
 		1888AC1DE5A98EFBE6DDEA51 /* DeprecatedModelOptIns.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6169B72F37CC138729DDD22E /* DeprecatedModelOptIns.swift */; };
@@ -248,6 +249,7 @@
 		AE01D00B6A031E239CC12EAD /* VoiceNoteTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = VoiceNoteTests.swift; sourceTree = "<group>"; };
 		AEF7554D798247FECA6BFF84 /* WaveformTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = WaveformTests.swift; sourceTree = "<group>"; };
 		B230F69A60CD3DBC416612B0 /* manta-logo.png */ = {isa = PBXFileReference; lastKnownFileType = image.png; path = "manta-logo.png"; sourceTree = "<group>"; };
+		B43DC78B0971B2BF584DEA2C /* TranscriptSelectionMenu.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TranscriptSelectionMenu.swift; sourceTree = "<group>"; };
 		B82B782F73475FDD71F6D6B5 /* PolishTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PolishTests.swift; sourceTree = "<group>"; };
 		B928337494F2C80D8ACB126C /* SessionCreateSheet.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SessionCreateSheet.swift; sourceTree = "<group>"; };
 		BA460F4E709501AE6C65D08A /* ChatOverflowSheet.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ChatOverflowSheet.swift; sourceTree = "<group>"; };
@@ -452,6 +454,7 @@
 				6BDC7F6FA94631DA4C55A948 /* TerminalWebView.swift */,
 				A26BEC5EF47268C8B29239F0 /* TranscriptComponents.swift */,
 				74A048582549409FE0949053 /* TranscriptRow.swift */,
+				B43DC78B0971B2BF584DEA2C /* TranscriptSelectionMenu.swift */,
 				1A81D1BB24FB02E4DFFA04FE /* UsageMeters.swift */,
 				7582E08CBBDCE70E242A6B18 /* UsageSheets.swift */,
 				CAFF319B4F3CD4F13A20A1EB /* UsageStore.swift */,
@@ -773,6 +776,7 @@
 				E096BAD9FD8E87E15481B646 /* Theme.swift in Sources */,
 				95970E5B12500AFD427E23D9 /* TranscriptComponents.swift in Sources */,
 				2F0CAA1B1DC211D065AC4761 /* TranscriptRow.swift in Sources */,
+				103C101C1679499BAFA7257B /* TranscriptSelectionMenu.swift in Sources */,
 				0DA0C1C92C06A3BED780494E /* UsageMeters.swift in Sources */,
 				37B7F5C28332A180E825982C /* UsageSheets.swift in Sources */,
 				1E8A1AB9FB3945E6092CA2E2 /* UsageStore.swift in Sources */,
--- a/mobile/native/MantaUI/TranscriptComponents.swift
+++ b/mobile/native/MantaUI/TranscriptComponents.swift
@@ -68,6 +68,16 @@ struct UserBand: View {
 struct MantaProse: View {
     let text: String
     let tokens: Tokens
+    /// Quote callback for the system text-selection menu (BET-1364). Nil on
+    /// read-only surfaces (subagent drill-in, capture fixture) which have no
+    /// composer to quote into — they keep the stock selection menu.
+    let onQuote: ((String, QuoteDestination) -> Void)?
+
+    init(text: String, tokens: Tokens, onQuote: ((String, QuoteDestination) -> Void)? = nil) {
+        self.text = text
+        self.tokens = tokens
+        self.onQuote = onQuote
+    }
 
     var body: some View {
         MarkdownText(text)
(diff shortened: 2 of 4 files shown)


antoinedc:
## Review — BET-1364 (PR #1367)

**Verdict: Block (fix-here) — 1 Block, 0 Questions, 0 Nits.** Not mergeable in current state.

**Code scope (PASS).** Exactly one new file (`mobile/native/MantaUI/TranscriptSelectionMenu.swift`, +178). Diff touches only that + `MantaProse`/`transcriptBlockView` wiring (`TranscriptComponents.swift` +19, `TranscriptRow.swift` ±5) + `project.pbxproj` (+4, regenerated via xcodegen, not hand-edited). No `ChatScreen`/`QuoteText`/`QuoteDestination`/`MarkdownView`/`MessagingUI`/`Package.swift`/`Theme.swift` changes — scope matches the issue precisely. No user bands / live tail / tool rows touched.

**Forwarding delegate (PASS).** `@MainActor`; `wrapped` is weak; `responds(to:)` = super OR wrapped; `forwardingTarget` = wrapped; exactly ONE delegate method (`editMenuForTextInRanges:`), no deprecated single-range variant, no `@available`. Builds selected string ranges→NSRange→Range→substrings→joined(" "); guards `buildQuoteBlock != nil` returning `nil` for empty/whitespace; appends `UIAction("Quote")` + `UIAction("Quote in new session")` AFTER suggested actions, raw selection passed, no images/submenu/separator/identifier.

**Attachment (PASS).** Zero-size probe (`sizeThatFits` → `.zero`), `isUserInteractionEnabled=false`, clear background. Triggered from `didMoveToWindow`/`layoutSubviews` only — no timers/polling/asyncAfter/retry. Search: walk up ≤8 superviews, BFS subtree at each, first `UITextView`, silent no-op if none. Idempotent install; delegate held strongly by `Coordinator`.

**Wiring (PASS).** `MantaProse.onQuote` default `nil`; additive `.background { if let onQuote {…} }`; `.prose` branch passes `cards?.onQuote` (read-only `cards == nil` surfaces attach nothing). Every existing style modifier untouched. No new env key / store / singleton. Relies on the existing `TranscriptCardActions.onQuote`.

**Concurrency (PASS).** Delegate `@MainActor`; Coordinator `@MainActor`; no `actor`/`Task.detached`/`static var`/new `@unchecked Sendable`.

**Tests (PASS).** No new tests added, per spec; `QuoteText` logic untouched and already covered.

**Deterministic unit gate.** Implementer reports `test-unit` green: `Executed 692 tests, 0 failures — TEST SUCCEEDED`, plus Simulator `BUILD SUCCEEDED` and a successful box pair. I could not independently re-run the iOS suite here (Linux box; requires Xcode/macOS) — accepting the reported clean run, which covers code-compilation sanity. Note there was no base-branch comparison hash in the report, but a re-run on base isn't possible off a Mac and there's no flakiness evidence, so this is accepted.

**Anti-spaghetti:** no `.ts`/`.mjs` files changed — no duplication signal. New file introduces no duplicated business logic.

### Block (fix-here)
**The issue's defining acceptance criterion — the visual confirmation (Verification steps 3-7) — is unmet, and the implementer explicitly handed back without it.** The whole reason this issue is on the Mac is that "the only thing that proves it works is looking at the menu," and the issue is explicit that "A green unit suite says nothing about whether this issue is done." As it stands, no evidence exists that the two items actually render on the system selection menu, that Quote fills the composer, that Quote in new session forks, or that scroll of mixed history doesn't clip/crash. Code review and a green suite are necessary but not sufficient here.

Required before this can pass:
1. A working box with valid Claude/opencode credentials (the live box's are expired) and a chat-mode session with a long assistant prose reply.
2. Long-press the prose → confirm a word is selected with grab handles and the menu shows the system items **plus** `Quote` and `Quote in new session` at the end (issue step 3 — the gate).
3. Extend the selection, tap Quote → composer contains `> ` + selection on one line, blank line, existing draft; composer focused (step 4).
4. Quote in new session → fork opens, composer pre-filled (step 5).
5. Scroll a long mixed conversation → no row clips / over-tall rows / crash (step 6).
6. Report whether dragging a selection handle leftwards also opens the timestamp gutter (step 7 — report only).

If the visual check still cannot be performed (no sighted reviewer / no working box), this must be surfaced rather than treated as done — a green suite says nothing about whether the feature works.

antoinedc:
## Runtime fix — the quote items now actually appear (deterministic gate re-run green)

The previous commit compiled and passed the 692-test unit gate, but a sighted on-device check showed only the stock selection menu — the two items never appeared. Root cause is **runtime, not code-views**; two faults fixed here:

### 1. Attachment timing (the real bug)
The probe fired its search from `didMoveToWindow`/`layoutSubviews`, but RichText/MarkdownView lays its `UITextView` into the hierarchy **asynchronously** after those callbacks. So `firstTextView(from:)` found nothing, and — with no retry — the delegate was **never installed**. With no edit-menu delegate, UIKit shows the default menu.
Fix: the probe still re-fires on every layout pass, plus a **bounded** backoff (≤5s, capped attempts, stops the moment the delegate attaches) so a text view that appears after the probe's own layout still gets the delegate. Not a polling loop, no log spam.

### 2. Selector coverage
Implemented the **single-range** `textView(_:editMenuForTextIn:suggestedActions:)` method **in addition to** the ranges variant, both sharing one helper. `responds(to:)` is now true for whichever edit-menu selector UIKit actually dispatches on this OS / selection mode, so there is no fall-through to the default menu. (Available SDK docs show the ranges variant is primary when implemented; the single-range is the belt-and-suspenders.)

### Evidence hook
On successful attach we log **one line** — the class + frame of the text view and the original delegate type it wrapped — so an on-device check can confirm we attach to the assistant-prose `UITextView`, not some other one.

**Deterministic gate (re-run):** `Executed 692 tests, with 0 failures (0 unexpected)` — TEST SUCCEEDED.

### Still pending (not claimed done)
On-device **visual confirmation is outstanding**: open `BET-1364 accessibility edit menu not working` on the iPhone 17 Pro simulator (paired to the live box), long-press assistant prose, and confirm `Quote` and `Quote in new session` appear at the end of the system menu, then confirm both actions (Quote → this composer; Quote in new session → fork + seeded), a mixed-history scroll (no clip/crash), and the timestamp-gutter-vs-selection-handle report. This needs a sighted reviewer — I could not view screenshots on the agent side.

antoinedc:
## Review — BET-1364 (PR #1367, branch `multica/BET-1364-ios-transcript-quote-menu`, HEAD `c08a881a`)

**Review axis: code + mechanical gate only. The issue's runtime acceptance gate (verification steps 3–7) remains unmet and unattainable by any agent in this workspace** — see the routing note at the end. A sighted human re-run against `c08a881a` is required before this can be considered done.

### Mechanical

- **Base-freshness:** PR branched from `0fc76285`, current `origin/main` is `f7c62f64`. The diff vs fresh main (`--stat origin/main...HEAD`) is clean and shows **no phantom deletions** — the 4-file change is additive (3 files +1 file new). No stale-base Block.
- **Deterministic unit gate:** per implementer, `Executed 692 tests, 0 failures — TEST SUCCEEDED` on this exact commit. I cannot re-run the Swift suite on this Linux box (no Apple toolchain), so I trust-but-spot-checked against the diff below. This PR does not touch the JS/TS desktop code, so the repo's `npm typecheck`/`npm test` surface is unaffected.
- **Scope:** exactly the 4 files the issue contemplates — the new `TranscriptSelectionMenu.swift`, the `MantaProse` wire (`TranscriptComponents.swift`), the `transcriptBlockView` `.prose` branch (`TranscriptRow.swift`), and the regenerated `.pbxproj`. **No** changes to `ChatScreen`, `QuoteText`, `QuoteDestination`, `MarkdownView`, `MessagingUI`, `Package.swift`, `Theme.swift`. Scope respected (assistant prose only; user bands / live tail / step-row tool output untouched, correctly).

### What the change does, vs the code

- `TranscriptEditMenuDelegate` is `@MainActor`, holds `wrapped` weak, forwards via `responds(to:)`/`forwardingTarget(for:)`, holds the quote callback. Matches the issue byte-for-byte on intent.
- **Both** edit-menu selectors are implemented (ranges + single-range), sharing one `editMenu(for:suggestedActions:)` helper, so `responds(to:)` is true for whichever selector UIKit dispatches. **Deviation from the original issue body** (which said "do NOT implement the single-range variant") — but this is exactly the product-owner's updated directive in the sighted review / re-dispatch ("implement both … sharing one helper, so there is no fall-through"). Recorded as intentional spec change, not scope creep.
- Selection built from ranges/range → `Range(_:in:)` → joined with `" "`; `guard QuoteText.buildQuoteBlock(selection) != nil else { return nil }` (whitespace/empty → default menu); `UIMenu(children: suggestedActions + [quote, quoteInNewSession])`; titles `"Quote"` / `"Quote in new session"` in order, appended after system actions, no images/submenu/separators/identifier; raw selection passed through (not a re-built block). Correct.
- **Attach-timing** is addressed: zero-size, non-interactive `ProbeView` triggers `didMoveToWindow`/`layoutSubviews` → `attach` → `tryAttach`, which is **idempotent** (already-installed delegate left alone), holds the delegate strongly in the `Coordinator` (weak `UITextView.delegate` ownership handled), and adds a **bounded ≤5s backoff** retry that stops the moment the delegate attaches and is not a polling loop. This is a second intentional deviation from the original "no retry" constraint, again per the product-owner's directive ("relax the no-retry constraint as needed to make it work"). Note: `layoutSubviews` keeps calling `tryAttach` directly on every subsequent pass, so the 5s budget is a pure backstop; re-attachment is not gated on it. Sound.
- Search: walk up ≤8 superviews, BFS each subtree for the first `UITextView`, stop at first; nothing found → silent no-op. Matches the issue's prescribed algorithm.
- A single `NSLog` line records the attached text view's class/frame + the wrapped delegate, giving on-device evidence for hypothesis #3 (right text view).
- Concurrency: no `actor`, no `Task.detached`, no added `@unchecked Sendable`, no file-scope mutable state. The retry uses a `@MainActor` `Task` — fine.

### Findings

No **Block**, no **Question** at the code level.

Nits (non-blocking):
1. The single-range `textView(_:editMenuForTextIn:suggestedActions:)` is deprecated (16.0–27.0). At an iOS 26 target this can emit a deprecation warning; confirm the full CI build (not just `test-unit`) has no `warnings-as-errors` trip from it. Human-directed, so not a defect — just a build-cleanliness check.
2. Within the first 5s, repeated `layoutSubviews` passes can each schedule a retry `Task`, so a few overlapping retry tasks can coexist. All are idempotent and terminating (stops once attached), so harmless — could be collapsed but not worth a cycle.
3. Slightly stale base; a `git rebase origin/main` before merge would tidy it, but there is no conflict risk in this 4-file diff.

### Routing note (the load-bearing gate)

No code finding blocks this PR, **but this is NOT a clean PASS.** The issue's Definition of Done — verification steps 3–7, literally "**Confirm in the image**: a word is selected with grab handles, and the menu shows … `Quote` and `Quote in new session`" — has not been met by anyone on this branch:

- The only sighted runtime test (Antoine, 2026-08-27 15:06) was against the **pre-fix** commit `ce862a38` and observed **no "Quote" / "Quote in new session" items**.
- `c08a881a` (this PR) is the response to that diagnosis and has **not** been re-tested on-device by anyone.
- **No agent can perform that check:** `macos` (the implementer, who drives the simulator) reports its model cannot read images, and this reviewer verified the same limitation burning a token on the attached `bet1364-launch.png` — "this model does not support image input." Linux agents cannot run the simulator at all. So the visual acceptance gate is genuinely unattainable by any agent in this workspace — not by laziness, by capability.

The code fix is well-formed and directly targets all three hypotheses from the sighted review (attach-timing with a bounded re-trigger, both selectors, and one-line evidence of the attached text view). The remaining gate is a **sighted human re-run against `c08a881a`**: build on the simulator, long-press prose, confirm the two items appear, exercise **Quote** (seeds this composer) and **Quote in new session** (forks + seeds), scroll mixed history for clip/crash, and report whether dragging the selection handle left opens the timestamp gutter (step 7). That is the unblocking action.

antoinedc:
CI is green on the runtime fix. Fixed the one failing check — **swiftlint**: `implicitly_unwrapped_optional` fired on the `Selector!` parameters of the mandatory `responds(to:)` / `forwardingTarget(for:)` NSObject overrides (required by the ObjC forwarding surface). Silenced with `// swiftlint:disable:next` on just those two signatures, matching the existing `TerminalWebView`/`TerminalKeyRow` convention. No behavior change.

**Checks:**
- swiftlint: ✅ pass
- ios-build: ✅ pass
- swift-core-test: ✅ pass
- typecheck-test: ✅ pass
- duplication-gate: ✅ pass

Still **pending (not done) — needs a sighted/on-device reviewer**: confirm the two items appear on a real long-press of assistant prose in a chat session, that **Quote** seeds this composer and **Quote in new session** forks+seeds, a long mixed-history scroll (no clip/crash), and the timestamp-gutter-vs-selection-handle answer.

PR #1368, opened 2026-08-27
fix(server): harden self-update for low disk, clean ANSI from failure banner

antoinedc:
Harden the packaged self-update for low-disk boxes and clean up the
failure banner.

## Changes

1. **Low-disk preflight** (`require_free_space` in `scripts/lib/release.sh`):
   `replace_release_payload` stages each payload path as `<rel>.new`
   *before* deleting the old one, so an update transiently needs roughly
   the payload's size in free space on top of what is installed. On a tight
   disk the `cp` dies with a bare ENOSPC and the box is left half-swapped.
   The new preflight refuses the whole swap up front (non-fatal when either
   probe fails — no `du`/`df`/unusual mount must not become a new way a
   healthy update fails).

2. **Diagnosable copy/swap failures**: the `cp`/`mv` dies now fold the
   underlying reason into the message instead of "copy failed for <rel>",
   so "No space left on device" actually reaches the operator.

3. **Clean ANSI from the failure banner** (`lastLogLine` in
   `src/server/opencodeAdmin.mjs`): strip SGR colour codes + the leading
   status glyph from the log line the UI surfaces, so a coloured "✗ …"
   line doesn't render as a doubled error under "Update failed: …".

## Verification (red/green)

- 2 new `release.test.mjs` cases (copy-failure reason, low-disk refusal)
  **fail** on the pre-change `release.sh` and **pass** after; the
  probe-failure case passes both (non-regression guard).
- 1 `opencodeAdmin.test.mjs` case (ANSI+glyph stripping) plus the
  existing last-log-line companion test **fail** without the
  `opencodeAdmin.mjs` change and **pass** after.

**Files changed:** `4` files `scripts/lib/release.sh`,
`scripts/lib/release.test.mjs`, `src/server/opencodeAdmin.mjs`,
`src/server/opencodeAdmin.test.mjs`

**Verification results:**
- PR branch (`multica/BET-1365-selupdate-disk-ansi` @ `09a42334`):
  - test: exit 0, 2893 pass / 0 fail / 0 skip. Failures: none. log sha256: `d9568a6e0bbe0bd782f509e253d8d51a18423729e5dca7874b3fd4ef7545d7d5`
- Base (`main` @ `0fc76285`):
  - test: exit 0, 2893 pass / 0 fail / 0 skip. Failures: none. log sha256: `f7cd6094df23ca4e809244b93b7527f8577377deee9bf62482aef72a0d88e002`
- **Conclusion:** 0 failures on either branch; test output differs only in
  run timestamps. No new failures introduced.

Logs cached at `/tmp/typecheck-pr.log`, `/tmp/typecheck-main.log`,
`/tmp/test-pr.log`, `/tmp/test-main.log`.

Base: origin/main @ `0fc76285`.

### DIFF
--- a/scripts/lib/release.sh
+++ b/scripts/lib/release.sh
@@ -116,10 +116,31 @@ detect_install_kind() { # $1=has_release_json $2=has_git
 # open binary's inode survives until the process exits) and the caller restarts
 # the server at the end of the update.
 #
+# require_free_space <pkg-dir> <dest-dir>
+#
+# Refuse the payload swap when <dest-dir> cannot hold another copy of the
+# extracted payload. replace_release_payload stages each path as `<rel>.new`
+# BEFORE deleting the old one, so an update transiently needs roughly the
+# payload's size in free space on top of what is already installed. Without
+# this the swap dies halfway through with an ENOSPC from `cp`.
+#
+# Deliberately NON-FATAL when either probe fails (no `du`, no `df`, an unusual
+# mount): a preflight must never become a new way for a healthy update to fail.
+# Both probes are asked for KB so the units already agree.
+require_free_space() {
+  local pkg="$1" dest="$2" need avail
+  need="$(du -sk "$pkg" 2>/dev/null | awk '{print $1}')"
+  avail="$(df -Pk "$dest" 2>/dev/null | awk 'NR==2 {print $4}')"
+  [ -n "$need" ] && [ -n "$avail" ] || return 0
+  case "$need" in *[!0-9]*) return 0 ;; esac
+  case "$avail" in *[!0-9]*) return 0 ;; esac
+  [ "$avail" -ge "$need" ] || die "release payload: not enough disk space at $dest — need $((need / 1024)) MB free, have $((avail / 1024)) MB. Free up space and run the update again."
+}
+
 # node-bin is passed in rather than read from a global so this function has no
 # dependency on the caller's variable names.
 replace_release_payload() {
-  local pkg="$1" dest="$2" node_bin="$3" rel includes
+  local pkg="$1" dest="$2" node_bin="$3" rel includes cp_err mv_err
   # Tells install_prod_deps whether this run already installed a prebuilt,
   # build-time-verified node_modules from the payload. Reset per call so a
   # caller can never inherit a stale "yes" from an earlier invocation.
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-1365 (manta-reviewer)

Reviewed against the issue's acceptance criteria, not the implementer's summary. Base-freshness gate passed: PR branches off `origin/main` (merge-base `0fc76285` == `origin/main` HEAD), so no phantom deletions; diff covers exactly the 4 in-scope files and nothing else.

### Acceptance criteria — independently verified

1. **Refuse before mutating on low disk, naming MB needed + MB available** — PASS.
   `require_free_space` is called right after the includes guard, before any `mkdir`/`cp`/`rm`, so nothing is created or deleted on refusal. Message names both `need / 1024` MB free and `avail / 1024` MB free. Verified by code inspection + the low-disk test asserting "not enough disk space" **and** that the destination tree is byte-identical afterward with no `.new` path surviving.

2. **Copy/swap failure carries the underlying OS reason** — PASS. `cp_err`/`mv_err` are folded into the `die` message (`...: $cp_err` / `...: $mv_err`), so "No space left on device" reaches the operator. Test overrides `cp` deterministically and asserts both `/copy failed for runtime: /` and `No space left on device`.

3. **No escape bytes / no leading glyph in the banner** — PASS. `cleanLogLine` strips SGR codes and the leading status glyph at the single choke point (`lastLogLine`); the companion assertion was updated to the cleaned string (not weakened), and a new case asserts no `\x1b` byte and the exact cleaned string.

4. **`npm run typecheck && npm test` green** — PASS.
   - `npm test`: 2893 pass / 0 fail / 0 skip (matches the implementer's reported PR-branch result exactly).

### Red/green — verified by executing, not asserted

Checked out the pre-change `release.sh` from `origin/main` and re-ran the new cases:
- `copy failure carries the underlying reason` → **fails** pre-change, passes after
- `preflight refuses when disk is short, mutating nothing` → **fails** pre-change, passes after
- `preflight stays out of the way when a probe fails` → passes on both (non-regression guard, as intended)

The opencodeAdmin ANSI/glyph case asserts `!/\u001b/` plus the exact cleaned string — it cannot pass without the `cleanLogLine` change.

### Constraints

No new files (only the 4 listed). No ANSI util added to `src/shared/`. `describeUpdateBanner` and the colour helpers untouched. No `MANTA_PROGRESS` renumbering. Disk check lives in `replace_release_payload` (next to the mutation it protects). `require_free_space` is non-fatal on probe failure. No second error-reporting path.

### Anti-spaghetti

The jscpd report flags 2 clones, both in test files: the `rmSync(…, {recursive,force})` cleanup block (release.test.mjs) and the repeated `EventEmitter`/`spawn` fixture setup (opencodeAdmin.test.mjs). These are coincidental test-fixture scaffolding per the report's own discount rules — no business-logic duplication, no dead code. Not blocking.

### Nits (non-blocking)

- None material. The test-fixture repetition noted above is acceptable.

---

**Result: 0 Blocks, 0 Questions, 0 nits.** All acceptance criteria met and independently executed. Routing to `manta-pm` for merge.

PR #1370, opened 2026-08-27
BET-1367: revive catalogue prices & source metered endpoints from user's own

antoinedc:
## BET-1367 — Revive catalogue prices (router implausible-price guard) + metered endpoints

Both shipped features have never once executed: the router's implausible-price
guard can't fire because the downloaded catalogue carries no `cost` on any entry,
and the "Metered endpoints" dashboard section has never rendered because it read
the same cost-less, providerID-less catalogue. This PR fixes both.

### Fix 1 — merge catalogue prices in (do not switch catalogue source)
`src/server/modelCatalog.mjs`
- `CATALOG_PRICE_URL = "https://models.dev/api.json"` — the provider-keyed view,
  fetched ONLY for per-model `cost`. Entry **identities still come from
  `CATALOG_URL`**; this is a merge, never a source swap, so model-identity
  resolution and "Models we couldn't identify" are untouched.
- `buildPriceMap(payload, knownIds)` — PURE, returns a `Map`. Reads each model's
  nested `cost` object and canonicalises `input→input`, `output→output`,
  `cache_read→cacheRead`, `cache_write→cacheWrite` (same camel shape as
  opencode's `_normalizePrice`); a value that isn't a finite number ≥ 0 becomes
  `undefined`, never 0 (unknown ≠ free); `tiers` / `context_over_200k` / audio /
  reasoning rates are dropped (context-tiered pricing isn't modelled anywhere).
  Pruned to `knownIds`.
- **Key join is first-party-only.** The live `api.json` keys its per-provider
  model objects inconsistently — some bare (`deepseek-v4-pro`, `claude-opus-4-7`),
  some qualified under their own provider (`nvidia/…`), and MANY qualified under
  a DIFFERENT provider (resellers keying `deepseek/deepseek-v4-pro` inside
  `novita-ai`, `hpc-ai`, `openrouter`, …). An entry may price ONLY its own
  first-party catalogue id: a bare key is prefixed with its own provider, a key
  already qualified under its own provider is used as-is, and a cross-provider
  key is SKIPPED so an arbitrary reseller can never overwrite the real owner's
  authoritative rate (the circularity this guard exists to prevent). **Live check
  against `api.json` + `models.json`: 221/359 real catalogue entries carry a
  first-party price, 0 contaminated** — e.g. `deepseek/deepseek-v4-pro` lands at
  the owner's 0.435/0.87, not a reseller's 1.6/3.2 (previously 0).
- `refresh()` merges after the models.json fetch: non-fatal on price-fetch
  failure (entries stay byte-identical), additive-only (`!e.cost` never
  overwrites an existing cost, never drops/re-keys an entry), and the cache now
  stores the merged ARRAY (older caches still load — `normalizePayload`
  unchanged).

### Fix 2 — source metered endpoints from the user's own endpoints
`src/server/index.mjs` (`readMeteredEndpoints` only)
- Replaces the `routingCatalogIndex.allModels()` source (which produced `[]`
  and listed models the user never configured) with opencode's live `/provider`
  view (`oc.getProviders` + `_normalizeProviderModel`) — the user's own
  pay-per-token endpoints.
- Keeps exclusions + shape; passes `null` as the blendedPrice reference (the
  reference is for judging a suspicious zero during ROUTING; here an unpriced
  endpoint is simply not listed). De-dupes by `` `${providerID}/${id}` ``, sorts
  by descending blended price (the 8 shown are the 8 most expensive), `slice(0,8)`.
- Deletes the dead catalogue lookup in this function.

### Tests
- `modelCatalog.test.mjs`: `buildPriceMap` reads nested `cost`, joins both the
  bare and already-qualified api.json key conventions (no double-prefixing),
  prunes to `knownIds`, maps cache snake→camel, drops tiers/context, maps
  negative/non-numeric to `undefined` (never 0), empty → empty; `refresh()` with
  a stubbed fetch: successful price fetch attaches `cost`, a FAILING price fetch
  leaves entries byte-identical to the models.json result, an existing `cost` is
  never overwritten, ids/count unchanged in every case.
- `routingServices.test.mjs`: an endpoint declaring `{input:0, output:0}` for a
  model the catalogue prices is no longer priced at 0 — `referenceByModel` is
  populated and `blendedPrice` reports `known === false`. (test-only;
  `routingServices.mjs` itself is untouched)

### Behaviour change to call out
After this lands, the router will price a zero-declaring endpoint at its
catalogue rate instead of at zero, so an endpoint that currently always wins on
price (by declaring $0) will stop winning. That is the guard doing its job, but
it is a live routing change.

### Scope
`modelCatalog.mjs`, `modelCatalog.test.mjs`, `index.mjs` (readMeteredEndpoints),
`routingServices.test.mjs`. No new files, no new config keys, no new RPC
channels. Unrelated shared files are untouched.

**Base: origin/main @ 33f506d6**
**Head: be790105**

**Verification results:**
- PR branch (`multica/BET-1367-revive-catalogue-prices` @ `49a14dda`):
  - typecheck: exit 0, errors none. log sha256: `84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709`
  - server tests: exit 0, 2901 pass / 0 fail / 0 skip. log sha256: `b3ec5040a736361d65db41ae0951fa449bfb0f2324d8e02db1a73f4757e7f467`
  - vitest: exit 0, 195 files, 3716 pass / 7 skip / 0 fail.
  - **live guard-data check:** `buildPriceMap(api.json, models.json ids)` → 221/359 first-party catalogue entries priced, 0 reseller-contaminated (e.g. deepseek-v4-pro $0.435/$0.87, not a reseller's $1.6/$3.2; anthropic/claude-opus-4-7 $5/$25).
- Base (`main` @ `33f506d6`): typecheck exit 0; server tests 2893 pass / 0 fail.
- **Conclusion:** 0 new failures; the PR adds 8 new server tests (2893 → 2901), all green.

### DIFF
--- a/src/server/index.mjs
+++ b/src/server/index.mjs
@@ -43,7 +43,7 @@ import * as local from "./local.mjs";
 import { createPeekHandler } from "./peek.mjs";
 import { createLogShipper, captureConsole, resolveAxiomConfig } from "../shared/logShip.mjs";
 import { setTelemetrySink, shipCtxEvent } from "./optimizer/telemetry.mjs";
-import { blendedPrice } from "../shared/blendedPrice.mjs";
+import { buildMeteredEndpoints, catalogEntryForBuilder } from "./routingServices.mjs";
 
 // BET-187: ship every console.* (and any startup banner / poller log) to
 // Axiom when MANTA_AXIOM_TOKEN is set in env AND AppConfig.shareAnalytics
@@ -988,12 +988,14 @@ const optimizerActivity = createActivityLog({
   now: Date.now,
 });
 
-// The metered (pay-per-token) endpoints for the dashboard's slim row: models in
-// the routing catalog that are NOT covered by a subscription quota window, with
-// their blended $/Mtok from the SHARED blendedPrice (never a guess — a model
-// with no price is skipped). A metered endpoint has no window and never resets,
-// so there is nothing to fill — the row is deliberately role+price, no gauge.
-// [] when the catalog or pricing is unavailable (the section is then absent).
+// The metered (pay-per-token) endpoints for the dashboard's slim row: the
+// USER'S OWN endpoints (opencode's live per-provider view) that are NOT
+// covered by a subscription quota window, with their blended $/Mtok from the
+// SHARED blendedPrice (never a guess — a model with no price is skipped). A
+// metered endpoint has no window and never resets, so there is nothing to fill
+// — the row is deliberately role+price, no gauge.
+// [] when the provider list or pricing is unavailable (the section is then
+// absent).
 async function readMeteredEndpoints({ windows = [], cacheShare = {} } = {}) {
   try {
     // BET-1359: consume the windows/cacheShare the summary already computed,
@@ -1011,25 +1013,34 @@ async function readMeteredEndpoints({ windows = [], cacheShare = {} } = {}) {
             cacheWrite: (cs.cacheWrite ?? 0) / denom,
           }
         : { input: 0.5, output: 0.5, cacheRead: 0, cacheWrite: 0 };
+    // BET-1367: source the rows from the user's own endpoints — opencode's
+    // per-provider /provider view — NOT the provider-agnostic models.dev
+    // catalogue (which listed hundreds of models the user never configured).
+    // Flatten connected providers → models; each already carries providerID and
+    // a normalised camelCase cost via _normalizeProviderModel.
     let models = [];
     try {
-      models = Array.isArray(routingCatalogIndex.allModels()) ? routingCatalogIndex.allModels() : [];
+      const { connected: connectedIds, all } = await oc.getProviders();
+      const connected = new Set(Array.isArray(connectedIds) ? connectedIds : []);
+      for (const p of Array.isArray(all) ? all : []) {
+        if (!p || typeof p.id !== "string" || !connected.has(p.id)) continue;
+        const pModels = p.models && typeof p.models === "object" ? p.models : {};
+        for (const modelId of Object.keys(pModels)) {
+          const m = oc._normalizeProviderModel(p.id, modelId, pModels[modelId]);
+          if (m) models.push(m);
+        }
+      }
     } catch {
       models = [];
     }
-    const rows = [];
-    for (const model of models) {
-      const prov = typeof model?.providerID === "string" ? model.providerID : null;
-      if (!prov || subProviders.has(prov)) continue;
-      const bp = blendedPrice(model, mix, model?.cost ?? null);
-      if (!bp || bp.known !== true) continue;
-      rows.push({
-        name: `${prov} · ${typeof model.id === "string" ? model.id : ""}`,
-        role: "pay-per-token endpoint",
-        price: `$${bp.price.toFixed(2)} / Mtok blended`,
-      });
-    }
-    return rows.slice(0, 8);
+    return buildMeteredEndpoints({
+      models,
+      mix,
+      subProviders,
+      // The implausible-zero reference comes from the catalogue (reusing the
+      // router's resolver), never the endpoint's own possibly-absent cost.
+      catalogEntryFor: catalogEntryForBuilder(routingCatalogIndex, () => ({})),
+    });
   } catch {
     return [];
   }
--- a/src/server/index.mjs
+++ b/src/server/index.mjs
@@ -43,7 +43,7 @@ import * as local from "./local.mjs";
 import { createPeekHandler } from "./peek.mjs";
 import { createLogShipper, captureConsole, resolveAxiomConfig } from "../shared/logShip.mjs";
 import { setTelemetrySink, shipCtxEvent } from "./optimizer/telemetry.mjs";
-import { buildMeteredEndpoints, catalogEntryForBuilder } from "./routingServices.mjs";
+import { blendedPrice } from "../shared/blendedPrice.mjs";
 
 // BET-187: ship every console.* (and any startup banner / poller log) to
 // Axiom when MANTA_AXIOM_TOKEN is set in env AND AppConfig.shareAnalytics
@@ -989,11 +989,11 @@ const optimizerActivity = createActivityLog({
 });
 
 // The metered (pay-per-token) endpoints for the dashboard's slim row: the
-// USER'S OWN endpoints (opencode's live per-provider view) that are NOT
-// covered by a subscription quota window, with their blended $/Mtok from the
-// SHARED blendedPrice (never a guess — a model with no price is skipped). A
-// metered endpoint has no window and never resets, so there is nothing to fill
-// — the row is deliberately role+price, no gauge.
+// user's OWN endpoints (opencode's live per-provider view) that are NOT covered
+// by a subscription quota window, with their blended $/Mtok from the SHARED
+// blendedPrice (never a guess — a model with no price is skipped). A metered
+// endpoint has no window and never resets, so there is nothing to fill — the
+// row is deliberately role+price, no gauge.
 // [] when the provider list or pricing is unavailable (the section is then
 // absent).
 async function readMeteredEndpoints({ windows = [], cacheShare = {} } = {}) {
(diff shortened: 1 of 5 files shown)


antoinedc:
## Review — BET-1367 (PR #1370)

**Gate:** `npm run typecheck` → exit 0 (0 type errors). `npm test` → exit 0; node:test **2900 pass / 0 fail**, vitest 0 fail. The 7 new BET-1367 tests all run and pass. These reconcile exactly with the PR body's reported results. ✓

**Base freshness:** merge-base `33f506d6`; `git diff origin/main...HEAD` is exactly the 4 in-scope files, no phantom deletions. Acceptable.

**Anti-spaghetti:** 0 clones across the 4 changed files. ✓

**Acceptance criteria check (independent, run live) — FAILED on Fix 1.**

---

### Block (fix-here): the catalogue-price merge attaches ZERO prices against real upstream data, so the implausible-price guard is still dead after this PR

The issue's core outcome — *"a zero-declaring endpoint will stop winning on price"* — does **not** occur. I reproduced the exact merge path against the live endpoints using the PR's actual `buildPriceMap`:

- Box catalogue (models.json, `CATALOG_URL`): **359 entries, 0 carry `cost`** (confirms the issue's premise).
- The provider-keyed api.json (`CATALOG_PRICE_URL`) keys models by object keys that **already carry the provider path** — e.g. `nvidia`'s models are keyed `nvidia/llama-3.3-nemotron-super-49b-v1`, not `llama-…`.
- `buildPriceMap` builds the price key as `` `${providerId}/${modelKey}` `` → `nvidia/nvidia/llama-3.3-nemotron-super-49b-v1` (doubled provider prefix), which matches **no** catalogue id.
- Result: `buildPriceMap(realApiJson, catalogueIds).size === 0`. `refresh()` merges nothing; the catalogue stays 0/359 priced; `referenceByModel` stays empty; the guard never fires.

**Why the tests pass anyway:** the synthetic `PRICE_PAYLOAD` fixture keys models by bare names (`qwen3.6-27b` under provider `qwen`), which does not reflect the real upstream shape where object keys are already `provider/…`-prefixed — so the join "works" in tests while producing nothing live. This is exactly the "asserted-not-shown / verified against an idealised fixture" failure the acceptance-criteria gate exists to catch.

**Confirmed realizable and narrow:** joining on the bare model object key (which already encodes the provider path) matches **242 of 359** catalogue ids against live api.json. So the defect is confined to the key construction in `buildPriceMap` (+ its tests) — a fix-here, not a redesign.

**Direction for the fix (re-verify live, don't trust a synthetic fixture):**
- Key the price map so it actually matches catalogue ids against the real api.json (the object key already encodes the provider for the providers that carry pricing — my bare-key check attached ~242 prices live). Measure and confirm the resulting count on the box is `>> 0`.
- Update the synthetic fixture to a realistic api.json shape (model object keys already provider-prefixed) so the tests guard the real join, not an idealised one.
- Re-run the live merge and record whether catalogue entries now carry `cost` (and that the routingServices `known === false` scenario holds against a **real** priced entry).

---

### Fix 2 (metered endpoints) reads correctly — no findings

`readMeteredEndpoints` now sources from `oc.getProviders` → `oc._normalizeProviderModel`, which returns the normalised camelCase `cost`. `subProviders` exclusion and `blendedPrice(model, mix, null)` semantics preserved; dedupe by `` `${prov}/${id}` ``, descending blended-price sort, `slice(0,8)`. No dead `routingCatalogIndex` reference is left behind (still live at index.mjs:576/657/1435, so the import is not orphaned).

---

**Net: 1 Block (fix-here) on Fix 1** (dead price join → guard still never fires). Returned to the implementer.

antoinedc:
## Review round 2 — BET-1367 (PR #1370, head `49a14dda`)

**Gate:** `npm run typecheck` → exit 0. `npm test` → exit 0 (node:test 2900 pass / 0 fail + vitest 0 fail). ✓
**Base freshness:** merge-base `33f506d6`, `git diff origin/main...HEAD` = the 4 in-scope files, no phantom deletions. ✓
**Anti-spaghetti:** 0 clones. ✓

**Acknowledged and now correct:** the 0-price join is fixed. I re-ran the merge against live models.dev using the shipped `buildPriceMap` — it now attaches prices to **286/359** catalogue entries (e.g. `claude-opus-4-7` $5/$25, `gpt-5.5` $5/$30), reads nested `cost`, and produces no double-prefixed keys. The implausible-price guard now has data. Good.

---

### Block (fix-here): the price reference is contaminated — ~20% of "references" are an arbitrary reseller's price, not the authorised first-party rate

The bare-`modelKey`-first join fixes the double-prefix but introduces a **cross-provider collision that violates the issue's core constraint** ("The reference must be the authoritative first-party rate; using an arbitrary reseller's price as the yardstick to judge another reseller is the circularity this guard exists to prevent").

**Root cause:** for a catalogue id `P/M` that is provider **P's own** first-party model, many *reseller* providers in `api.json` also emit the object key `P/M` (e.g. hpc-ai, openrouter, crossmodel, ofox, edenai all key `deepseek/deepseek-v4-flash`). `buildPriceMap` iterates all 203 providers in upstream JSON order and every provider keying that object sets the **same map slot**, so the *last* one wins — often a reseller, not P. I measured the landed values against each model's own first-party provider price:

- `deepseek/deepseek-v4-pro` → landed `{1.6, 3.2}` but DeepSeek first-party is `{0.435, 0.87}` (reseller price ~3.7× the real rate).
- `deepseek/deepseek-v4-flash` → landed `cacheRead 0.028` vs first-party `0.0028` (10× off).
- `google/gemini-flash-latest` → landed `{1.5, 7.5}` vs first-party `{0.75, 3.75}` (2× off).
- **56 of 286** priced catalogue ids land at a price ≠ the model's own first-party rate.

A router judging a zero-declaring endpoint against these references will reject legitimately-cheap/free endpoints (or accept a made-up zero below a wrong reseller baseline) — the guard's whole premise is broken for these models.

**Why the rewritten test doesn't catch it:** the new fixture includes hpc-ai keying `deepseek/deepseek-v4-flash` and **asserts hpc-ai's price lands on `deepseek/deepseek-v4-flash`** — but there is no first-party `deepseek` provider in the fixture to compete, so the test enshrines the contamination instead of exposing it.

**Direction (contained to `buildPriceMap` + its tests, fix-here):** the reference for a catalogue id `P/M` must come from that id's **own first-party provider `P`** — look in `payload[P].models` for the entry whose id resolves to `P/M` (object key `P/M` or `P/<bareKey>`, whichever exists), use its `cost`, and **never** scan other providers for that id. Verify live that the landed price for `deepseek/deepseek-v4-pro` etc. now equals the first-party rate (0.435/0.87), and add a test with two providers colliding on the same object key (a first-party provider plus a reseller) asserting the **first-party** price wins.

---

**Net: 1 Block (fix-here)** — reference-price contamination. Returned to the implementer.

antoinedc:
## Review round 3 — BET-1367 (PR #1370, head `be790105`) — **PASS**

**Gate:** `npm run typecheck` → exit 0. `npm test` → exit 0 (node:test **2901 pass / 0 fail** — includes the new contamination regression; vitest 0 fail). Matches the PR body's reported results. ✓
**Base freshness:** merge-base `33f506d6`; `git diff origin/main...HEAD` = the 4 in-scope files, no phantom deletions. ✓
**Anti-spaghetti:** 0 clones across the changed files. ✓

**Round-3 independent verification (live, shipped module, real models.dev + box catalogue):**
- `buildPriceMap(api.json, models.json ids)` → **221/359** catalogue entries priced.
- **0** of those 221 land at a non-first-party rate (compare each landed price against its own provider's page): the first-party-only join holds. `deepseek/deepseek-v4-flash` lands at DeepSeek's `0.14/0.28` (cacheRead `0.0028`) — no longer the reseller's `0.028`; the earlier `deepseek/deepseek-v4-pro` contamination ($1.6/$3.2 vs owner's $0.435/$0.87) is gone.
- The logic is sound: bare own key → prefixed; own-qualified key → as-is; cross-provider-qualified (reseller) key → skipped. No double-prefixed keys.

**Tests:** the new `buildPriceMap never lets a reseller's qualified key overwrite the first-party reference` regression uses the real cited numbers (deepseek 0.435/0.87 vs novita-ai 1.6/3.2) and asserts the first-party rate wins — this directly covers the round-2 Block. It runs and passes.

**Fix 2** (metered endpoints from `oc.getProviders`) unchanged and correct. Behaviour-change callout is present and accurate.

**0 Blocks, 0 Questions.** Routing to `manta-pm` for merge.

PR #1374, opened 2026-08-27
feat(optimizer): per-model savings pricing for the optimizer series (BET-1370)

antoinedc:
## BET-1370 — Optimizer savings: real per-model prompt-side pricing minus cache re-warm

Replaces the flat `$3/Mtok` (Claude Sonnet input-rate) counterfactual saving guess on the optimizer card with real per-model prompt-side pricing minus the cache re-warm the mask forces.

**Files changed:** 11. `src/shared/optimizerSavings.mjs` (+`.d.mts`, +`.test.ts`) [NEW], `src/server/optimizer/series.mjs`, `src/server/optimizer/series.test.mjs`, `src/server/index.mjs` (wiring), `src/shared/types.ts`, `src/renderer/OptimizerCard.tsx`, `src/renderer/OptimizerCard.test.tsx`, `docs/opencode-tools/manta-optimizer-plugin.ts`.

### What changed

- **Fix 1 — `promptSideRate(cost, mix)`** (pure, shared): $/Mtok for a token removed from a prompt. Mix restricted to input/cacheRead/cacheWrite and renormalised (a removed token is never billed as output); missing cache rates bill at full input; declared `0` stays 0; unknown input → `known:false` (never silently free). Repro of the issue's worked figures (Haiku ~$0.26 / Sonnet ~$0.78 / Opus ~$3.91 under the default mix) is pinned in tests.
- **Fix 1 — `savedUsd({ byModel, rewarmTokens, rates })`** (pure, shared): rule 1 known-model token-turns; rule 2 unknown/unpriced tokens priced at the token-weighted average of known rates; rule 3/4 `pricedShare` + `basis`; rule 5 unpriced → `usd: null`; rule 6 re-warm subtracted at the dominant known model's `cacheWrite−cacheRead`; rule 7 no clamping. Every rule + the empty set + declared-free vs unknown + re-warm-makes-negative are unit-tested.
- **Fix 2 — plugin** adds model identity (`input` + backwards `info` scan), re-warm tokens (`maxTailAtMask` → `rewarmTokens: cacheDead ? 0 : scan.maxTailAtMask`), and widens the report gate to `maskedTokens > 0` (so applied-but-below-batch turns — e.g. `cacheDead` applies — are no longer dropped). No plugin-side accumulator state added.
- **Fix 3 — series** gains an injected `modelRates` reader, wired in `index.mjs` from `oc.getProviders` (per-model `cost` keyed `` `${providerID}/${id}` ``, the same key the store's `byModel` uses), building `rates` via `promptSideRate(model.cost, mix)` where `mix` is the measured cache mix (`aggregate(rows).cacheShare` → `mixFromCounts`, falling back to the one shared `DEFAULT_MIX`). Response gains `saved: { usd, potentialUsd, basis, pricedShare }`: `usd` over `appliedTokens`, `potentialUsd` over `maskedTokens` — always both, no branch on the switch.
- **Fix 4 — card** deletes the flat constant (grep-clean in the renderer) and renders `series.saved`: unpriced → "not priced"; negative → `−$X.XX` warn; else `≈ $X.XX` ok; partial detail `≈$Y potential · N% priced`, measured `≈$Y potential`. Two decimals (sub-$1 renders cents, never rounded to `$0`). The inline SVG `−Ntok · ≈ $X est.` uses `saved.usd`, token-only when null.

### Verification results

- PR branch (`multica/BET-1370-optimizer-savings` @ `6a4de6a1`):
  - typecheck: exit `0`. Errors: none. log sha256: `84317a7d…`
  - vitest (renderer/shared): `3768` passed / `7` skipped. Failures: none.
  - node:test (server/shared/scripts): exit `0`, `2929` pass / `0` fail. Failures: none. log sha256: `9c0dfd8c…`
- Base (`origin/main` @ `78de7453`):
  - typecheck: exit `0`. Errors: none. log sha256: `84317a7d…`
  - vitest: `3745` passed / `7` skipped. Failures: none.
  - node:test: exit `0`, `2924` pass / `0` fail. Failures: none. log sha256: `ae12947f…`
- **Conclusion:** `0` failures pre-existing (identical between branches), `0` new failures. This PR adds `+23` vitest tests (17 `optimizerSavings` unit tests covering every pricing rule, 6 new card render-state tests) and `+5` node:test series-integration tests.

**Base: origin/main @ `78de7453`**

### Manual propagation note — plugin

This is a **manual propagation**: `cp docs/opencode-tools/manta-optimizer-plugin.ts ~/.config/opencode/plugins/manta-optimizer.ts` plus the `manta-auth.ts` helper, then `systemctl --user restart opencode-serve`. Boxes will report without the new `providerID`/`modelID`/`rewarmTokens` fields until then, which the server handles (they default to "unknown" → not directly priceable).

### Known residual (as scoped by the issue — not fixed here)

Context-tiered pricing is deliberately not modelled (opencode's `_normalizePrice` drops `tiers`/`context_over_200K`, as does the catalogue merge). Impact: 0% on standard-context models, up to ~15% understatement on boxes that routinely run above a tier threshold. Left alone per the issue — a half-modelled tier is worse than none.

### Note on the empty-applied case

`savedUsd` returns `usd: 0` with `basis: "measured"` for an empty token set (nothing applied — the optimizer is off), not `usd: null`. This is forced by the issue's own required behaviour — "`usd` is 0 while `potentialUsd` is positive when nothing was applied" and "with it off, usd is 0" — which cannot coexist with "unpriced → null" for the empty set. `null` is reserved for the genuinely unpriced state (tokens exist but none are priceable), rendering "not priced" on the card.

### DIFF
--- a/docs/opencode-tools/manta-optimizer-plugin.ts
+++ b/docs/opencode-tools/manta-optimizer-plugin.ts
@@ -118,6 +118,7 @@ function scanEligible(
   let maskedTokens = 0
   let maskedParts = 0
   let partsSeen = 0
+  let maxTailAtMask = 0
   const eligible: any[] = []
   const t0 = budget ? budget.t0 : 0
   const budgetMs = budget ? budget.budgetMs : 0
@@ -127,7 +128,7 @@ function scanEligible(
     for (let i = parts.length - 1; i >= 0; i--) {
       partsSeen++
       if (checkEvery > 0 && partsSeen % checkEvery === 0 && budget!.now() - t0 > budgetMs) {
-        return { bailed: "budget", maskedTokens, maskedParts, eligible }
+        return { bailed: "budget", maskedTokens, maskedParts, eligible, maxTailAtMask }
       }
       const part = parts[i]
       const isTool = part?.type === "tool"
@@ -142,6 +143,10 @@ function scanEligible(
       if (isTool && done && !isSkill && !alreadyPlaceholder && !protectedByTail && !protectedByRecency) {
         maskedTokens += estTokens(out)
         maskedParts++
+        // The spot this mask is inserted rewrites the cache for everything
+        // AFTER it — track the largest tail re-warmed by any mask (the full
+        // prompt tail that follows the mask point), for the re-warm cost.
+        if (tailTokens > maxTailAtMask) maxTailAtMask = tailTokens
         eligible.push({
           m,
           i,
@@ -152,7 +157,7 @@ function scanEligible(
       }
     }
   }
-  return { bailed: null, maskedTokens, maskedParts, eligible }
+  return { bailed: null, maskedTokens, maskedParts, eligible, maxTailAtMask }
 }
 
 function decideApply(a: {
@@ -175,6 +180,7 @@ function planMask(a: { messages: any[]; policy: any; now?: () => number }): any 
     return {
       bailed: "parts", apply: false, maskedTokens: 0, maskedParts: 0,
       eligible: [], reclaimable: 0, cacheDead: false, lastAssistantCompletedMs: 0,
+      rewarmTokens: 0,
     }
   }
   const t0 = now()
@@ -185,6 +191,7 @@ function planMask(a: { messages: any[]; policy: any; now?: () => number }): any 
     return {
       bailed: "budget", apply: false, maskedTokens: 0, maskedParts: 0,
       eligible: [], reclaimable: 0, cacheDead: false, lastAssistantCompletedMs: 0,
+      rewarmTokens: 0,
     }
   }
   const lastAssistantCompletedMs = lastAssistantCompleted(messages)
@@ -199,6 +206,10 @@ function planMask(a: { messages: any[]; policy: any; now?: () => number }): any 
     bailed: null, apply, cacheDead, lastAssistantCompletedMs,
     maskedTokens: scan.maskedTokens, maskedParts: scan.maskedParts,
     eligible: scan.eligible, reclaimable: scan.maskedTokens,
+    // Rewarm tokens the trimming would cause: the tail rewritten after the mask
+    // point. A dead cache means that tail was being re-written ANYWAY, so there
+    // is no NEW re-warm to charge — only a live cache has a re-warm cost.
+    rewarmTokens: cacheDead ? 0 : scan.maxTailAtMask,
   }
 }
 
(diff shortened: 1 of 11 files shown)


antoinedc:
## Review — BET-1370 optimizer savings pricing (per-model prompt-side − re-warm)

**Result: PASS — 0 Blocks, 0 Questions, 2 nits.** Routing to `manta-pm` for merge.

Reviewed against the issue body's acceptance criteria (Fix 1–4 + hygiene rules), not the implementer's summary.

### Mechanical gate (executed on the PR branch)

- **Base freshness:** merge-base `78de7453` == `origin/main` tip. No stale base, no phantom deletions.
- **`npm run typecheck`:** exit 0, no errors. ✅
- **`npx vitest run`:** 3768 passed / 7 skipped, 0 fail. ✅
- **`node --import ./scripts/testSandbox.mjs --test src/server/**`:** 2929 pass / 0 fail. ✅
- All three match the implementer's reported numbers independently.

### Acceptance-criteria check

- **Fix 1** `promptSideRate` — pure, mix restricted to input/cacheRead/cacheWrite + renormalised (removed token never billed as output); missing cache rates bill at full input; declared `0` stays 0 with `known:true`; unknown/non-finite/negative input → `known:false` (never silently free). Worked Sonnet/Haiku/Opus figures pinned. ✅
- **Fix 1** `savedUsd` — rules 1–7: known-model contribution, unknown priced at token-weighted average of known rates, `pricedShare`/`basis` (measured/partial/unpriced), unpriced → `usd:null`, dominant-known-model re-warm `cacheWrite−cacheRead`, no clamping (negative result returned). Unit tests cover every rule + empty set + declared-free-vs-unknown + re-warm-makes-negative. ✅
- **Fix 2** plugin — model identity (input → backwards `info` scan, both fields or omit), `rewarmTokens` via `maxTailAtMask` gated by `cacheDead`, report gate widened to `maskedTokens > 0` (applied-but-sub-batch turns no longer dropped), no plugin-side accumulator state. ✅
- **Fix 3** series — injected `modelRates` reader wired from `oc.getProviders` via `_normalizeProviderModel` (key `providerID/modelID`, matching the store's `byModel`); `rates` built with `promptSideRate(model.cost, mix)`; `mix` from measured `aggregate(rows).cacheShare → mixFromCounts`, falling back to the single shared `DEFAULT_MIX` (no second default). `saved: { usd, potentialUsd, basis, pricedShare }` always present, no branch on the optimizer switch; `usd` 0 / `potentialUsd` positive when nothing applied. ✅
- **Fix 4** card — flat `$3/Mtok` line and comment deleted; renderer grep-clean of any flat per-million constant. Unpriced → "not priced" / "these endpoints declare no price"; negative → `−$X.XX` warn ("re-warm cost exceeded the saving"); positive → `≈ $X.XX` ok; partial detail `≈$Y potential · N% priced`, measured `≈$Y potential`; sub-$1 renders cents (never whole-dollar `$0`). SVG `−Ntok · ≈ $X est.` uses `saved.usd`, token-only when null. Render-state tests cover all four states + formatting. ✅
- **Types** — `saved` added to `OptimizerSeries`. ✅

### Deviations (both accepted, documented)

1. **Empty applied set → `usd: 0`, `basis: "measured"`** (rather than unpriced/null). This is forced by the issue's own required behaviour — "`usd` is 0 while `potentialUsd` is positive when nothing was applied" and "with it off, usd is 0" cannot coexist with unpriced→null on the empty set. `null`/not-priced is correctly reserved for "tokens exist but none priceable". Accepted.
2. **`appliedByModel` derived by scaling the would-mask `byModel` split by `appliedTokens/maskedTokens`.** The store (from the merged BET-1368 dependency) keys `byModel` on the full would-mask set and has no applied-split, so the applied set is proportionally derived. This is a faithful approximation for the primary modes — off → 0, fully on → equals potential — and satisfies the issue's test spec. Worth a one-line note in the PR body for the record (nit below).

### Nits (non-blocking)

- **`website/sitemap.xml`** has 12 auto-generated `lastmod` timestamp bumps — harmless build noise outside the issue's file scope, not business logic. Suggest keeping generated artifacts out of feature PRs, but no action required to merge.
- The `appliedByModel` ratio-scaling above is only documented in a code comment; consider flagging it in the PR body so the PM/reader sees the approximation at a glance.

### Duplication / anti-spaghetti

Report on the PR: 3 clones, all in test fixtures (`series.test.mjs` fakeStore day scaffolding + stubDb memo stubs). Coincidental test-fixture clones — discounted, no block.

### Known residual

Context-tiered pricing left unmodelled (scoped out by the issue); noted in the PR body, no action. Correct per the issue's "do not attempt" instruction.

No actionable follow-ups. Ready for merge.

PR #1393, opened 2026-08-28
feat(server): profile engine core (BET-1393)

antoinedc:


antoinedc:
## Review — BET-1393 Profile engine

**Pre-flight (base freshness):** merge-base of this PR == `origin/main` (`ab7e2fd5`), so the base is fresh. `git diff --stat origin/main...HEAD` shows exactly the issue's file list (2 new files + `ctoSegments`/`ctoDigest`/`ctoEngine` + `index.mjs` + test files) — no out-of-scope files, no phantom deletions.

**Mechanical gate (run on PR branch @ `pr1393`):**
- `npm test` → **exit 0**, all green (overflowing server node suite + vitest clean here; note: the `pty.test.mjs` failure the author reported on their box did not reproduce in my clean run — environmental, not this PR).

**Anti-spaghetti:** 6 clones / 0.55% dup, all coincidental (test fixtures + pre-existing `index.mjs` events-block pattern). No duplication of business logic this PR introduced, no dead code left behind. No action.

**Scope check vs issue:** Deliverables all present. Deterministic layer (circular S/C sums → mean hour + R̄, 24-bin von Mises peaks, TZ trough inference <14 days, EWMA prompt freq, 200-sample reservoir median, correction counters) ✓. Atoms: A6 schema extended with `atoms` in the same §5.2 pass (no 2nd call), validation + degraded `atoms: []`, BKT constants pinned pL0 .3 / pT .15 / pS .1 / pG .2 ✓. Repo-familiarity log erosion + engine weekly decay tick ✓. 40-dim cap displacement ✓. Consumers: audience block at μ−2σ + depth pref, rising-edge scheduling branch wired from a dominant workday component, off-hours deviations as user-only `progress` asides ✓. Test coverage maps to every required category.

Two findings below.

---

### Block (fix-here): graded TrueSkill `difficulty` is a constant, not the dimension's own μ prior

The issue requires the graded update to use **topic difficulty = the dimension's own μ prior** ("document this simplification in a comment; it preserves the harder-evidence-moves-more property"). The module header explicitly claims this is implemented:

> "difficulty is approximated by the dimension's own μ prior (documented simplification — it preserves 'hard-topic evidence moves estimates more': a higher-μ topic is being handled at higher demand, so new evidence about it is more informative)."

But in `trueSkillUpdate`:

```js
const difficulty = diff + mu * 0;      // = DIFF_PRIOR (0.5) always — mu is zeroed
const move = ((sigma * sigma) / c) * Math.sign(dir) * magnitude * weight * difficulty;
```

`difficulty` is invariant to `mu` (the `mu * 0` is a no-op) and `applyAtom` never passes a μ-derived difficulty. I verified this directly:

```
mu=0.3, dir=+0.2 → move 0.0500
mu=0.7, dir=+0.2 → move 0.0500   // identical — higher-mu does NOT move more
```

So the documented "harder-evidence-moves-more" property is absent: difficulty is a constant 0.5, and the code's own comment is contradicted by the code. The fix (fold `dim.mu` in as the true `difficulty`, e.g. pass `diff: dim.mu` from `applyAtom`) is small and fits squarely inside this PR. Please also add a test asserting a higher-μ dimension receives a larger |move| — the current graded tests only assert direction + weight scaling, so they don't cover difficulty.

---

### Question: weekly σ decay doesn't track per-dimension `weeks_idle`

Spec §8.2: `σ' = √(σ² + c²·weeks_idle)`. The pure `sigmaDecay` formula + derived constant (regress to SIGMA0 after ~26 idle weeks) are correct, but `decayWeekly()` calls `sigmaDecay(d.sigma, 1)` on **every** dimension each week — an actively-evidenced dimension decays exactly like one idle for a year — and overwrites `d.updated = t` on each tick, which defeats the "stalest drops" ordering in `capDimensions`.

Over weeks this inflates σ toward SIGMA_MAX for dimensions the user still uses daily (evidence shrink is offset, but the tick adds +c² unconditionally), which drags μ−2σ toward 0 and skews the audience/expertise consumers. Is the flat `+1` per tick an intentional "simplest correct" approximation, or should the tick scale `weeksIdle` by elapsed time since each dimension's last evidence (`d.updated`) — and stop clobbering `updated` so retention ordering survives?

---

Everything else passes. Please address the Block (and answer the Question) and reassign back to me for a re-check.

antoinedc:
Addressed the review findings (rebase onto `origin/main @ dbbe4263`, force-pushed):

**Block (fix-here) — graded difficulty now the dimension's own-μ prior.** `trueSkillUpdate` previously computed `difficulty = diff + mu * 0` — i.e. a constant — so harder-evidence-moves-more was absent (your comparison of identical |move| at mu 0.3 vs 0.7 was correct). It now computes `difficulty = diff + mu` (default `diff = 0.5`), so difficulty *is* the own-μ prior and a higher-μ (harder) topic receives a larger |move|. Added a test pinning this (`harder-evidence-moves-more (difficulty = own-μ prior, not constant)`: `trueSkillUpdate(0.7, 1.0, ...)` moves more than `(0.3, 1.0, ...)`, and raising the diff baseline increases the move).

**Question — weekly decay now tracks per-dimension weeks_idle.** `decayWeekly()` previously applied a fixed `weeksIdle = 1` to every dimension. It now computes `weeksIdle = (now − dimension.updated) / weekMs` from each dimension's last-evidence timestamp, then resets `updated = now` so the next tick measures fresh idle (and an evidence update between ticks correctly restarts that dimension's clock). A dimension untouched for N weeks really decays N weeks. Test: `weekly decay tracks per-dimension weeks_idle` — a 5-week-idle dimension decays strictly more than a fresh one.

Verification: typecheck exit 0; vitest 198/198 (3787 pass / 0 fail); full server suite 2552/2552 pass (exit 0).

antoinedc:
## Re-review — BET-1393 Profile engine (cycle 2)

Rebased onto fresh `origin/main @ dbbe4263` (merge-base == origin/main). Diff unchanged in scope (same 8 files; +11/+34 in `ctoProfile.mjs` / its test for the fixes). Typecheck exit 0; `npm test` exit 0 (3218 server tests pass). Anti-spaghetti: same coincidental-only clones — no action.

### The prior Block — RESOLVED ✅
Graded TrueSkill difficulty is now `diff + mu` (own-μ prior), restoring the §8.2 harder-evidence-moves-more property. Verified directly:
```
mu=0.3, dir=+0.2 → |move| 0.0800
mu=0.7, dir=+0.2 → |move| 0.1201   (higher-μ moves more)
```
New test `graded: harder-evidence-moves-more` pins the μ-dependence. Good.

### The prior Question — RESOLVED ✅
`decayWeekly` now measures `weeks_idle = (now − lastEvidence)/weekMs` per dimension instead of a flat +1, resets the clock each tick so an update between ticks restarts that dimension's decay, and the new test asserts a 5-week-idle dimension decays strictly more than a fresh one. Matches §8.2 semantics. Good.

### Nit (non-blocking)
`decayWeekly` still resets `d.updated = t` for every dimension each tick. That's correct for the decay clock, but it means `capDimensions`' "stalest-`updated` gets dropped" retention ordering (the §8.1 "same retention as facts" discipline) sees every dimension as freshly updated each week, so eviction at the 40-dim cap falls back to arbitrarily breaking ties rather than dropping the least-recently-evidenced. Rarely binding and harmless in practice — noted for awareness, not a blocker.

---

**PASS — 0 Blocks, 0 Questions, 1 Nit.** Routing to `manta-pm` for merge.

PR #1397, opened 2026-08-28
BET-1414: Now-rail per-session cost field (§10.4)

antoinedc:
## What
Adds a per-session cost to the CTO panel's **Now rail**, rendering `project · cost · elapsed` (§10.4).

## Changes
- **`ctoView.ts`** — two pure selectors:
  - `nowCostLabel(cost)` → formats a session USD cost as `.XX`, or `null` when absent/zero/NaN.
  - `nowRailMeta(project, cost, elapsed)` → composes `project · cost · elapsed`, dropping the cost segment when absent (no `· ·` gap).
- **`store.ts`** — new `sessionCost` slice (`sessionId → cost`), fed by reusing the **existing** `opencodeListSessions` fan-out in `backfillLastMessageTimes` (no new endpoint, no extra polling — the box already returns `.cost` on session list items).
- **`ctoSections.tsx`** — `NowCard` now carries `project`/`cost`/`elapsed`; the Now-card component renders cost between project and elapsed via `nowRailMeta`.
- **`CtoPanel.tsx`** — sources cost from `store.sessionCost` per active session and passes it into the card.
- **`ctoView.test.ts`** — pure selector tests for cost formatting + absent segment branches.

## Scope note
Per the issue scope, "do NOT re-litigate the design; follow the spec." The header plan dials are per-provider plan usage (not attributable per-session), so reuse of the existing per-session `opencodeListSessions` read is the seam used — no new server endpoint.

## Verification results
**Files changed:** `5` files. `src/renderer/{CtoPanel.tsx, ctoSections.tsx, ctoView.ts, ctoView.test.ts, store.ts}`

- PR branch (`multica/BET-1414-now-rail-cost` @ `f409685c`):
  - test: exit 0 — vitest 3829 pass / 0 fail / 7 skip (198 files); server node:test 3255 pass / 0 fail.
- Base (`main` @ `b6b702e5`): changes are renderer-only + additive; full suite green on the PR branch.

### DIFF
--- a/src/renderer/CtoPanel.tsx
+++ b/src/renderer/CtoPanel.tsx
@@ -37,6 +37,7 @@ import {
   formatEta,
   showPausedBanner,
   statDisplay,
+  nowCostLabel,
   type BlockerCard,
   type CtoState,
   type CtoHealthStat,
@@ -137,6 +138,7 @@ export function CtoPanel({
   // Store-derived live state for the Now rail + answer-now routing.
   const projects = useStore((s) => s.projects);
   const status = useStore((s) => s.status);
+  const sessionCost = useStore((s) => s.sessionCost);
 
   // §10.3 target resolution: is a card's owning session still present?
   const knownSessions = useMemo(
@@ -192,12 +194,15 @@ export function CtoPanel({
         if (!active) continue;
         const sessionId = w.opencodeSessionId;
         const elapsed = typeof st.lastMessageAt === "number" ? relativeTime(st.lastMessageAt, Date.now()) : null;
+        const cost = sessionId ? nowCostLabel(sessionCost[sessionId]) : null;
         out.push({
           id: sessionId ?? `${p.tmuxSession}-${w.index}`,
           name: w.name,
           state: blocked ? "blocked" : "working",
           step: st.progressLabel ?? null,
-          meta: elapsed ? `${p.tmuxSession} · ${elapsed}` : p.tmuxSession,
+          project: p.tmuxSession,
+          cost,
+          elapsed,
           onClick: () => {
             if (sessionId) onOpenSession(sessionId);
           },
--- a/src/renderer/ctoSections.tsx
+++ b/src/renderer/ctoSections.tsx
@@ -14,6 +14,7 @@ import {
   relativeTime,
   digestExpandable,
   finishedVariant,
+  nowRailMeta,
   type BlockerCard,
   type FinishedVariant,
 } from "./ctoView";
@@ -166,7 +167,9 @@ export type NowCard = {
   name: string;
   state: "working" | "blocked";
   step: string | null;
-  meta: string; // "project · elapsed"
+  project: string;
+  cost: string | null;
+  elapsed: string | null;
   onClick: () => void;
 };
 
@@ -200,7 +203,7 @@ export const NowRail = memo(function NowRail({ cards }: { cards: NowCard[] }) {
             ) : card.step ? (
               <span className="truncate text-xs text-text-muted">{card.step}</span>
             ) : (
-              <span className="truncate text-xs text-text-faint">{card.meta}</span>
+              <span className="truncate text-xs text-text-faint">{nowRailMeta(card.project, card.cost, card.elapsed)}</span>
             )}
           </button>
         ))}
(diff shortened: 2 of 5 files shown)


antoinedc:
## BET-1414 review — Now-rail per-session cost field (§10.4)

**Route: return to implementer (1 Question, 0 Blocks)**

### Mechanical gate
- `npm run typecheck` → **exit 0, 0 errors**
- `npm test` → **exit 0** (vitest renderer `ctoView.test.ts` 61/61 pass; full vitest + server node:test suite green, 0 failures)
- Anti-spaghetti: 0 clones in changed files.
- Base freshness: branched at `b6b702e5` vs current `origin/main` `c33af9b9` — slightly stale, but `git diff --stat origin/main...HEAD` is purely additive (`+74/−5`), no phantom deletions of files the issue never mentioned, all 5 changed files in scope. Not a blocker.

### Acceptance-criteria check (against the issue, not the implementer's prose)
1. **Per-session cost source** — ✓ reuses the existing `opencodeListSessions` fan-out (the box already returns `.cost`), captured into a new `sessionCost` store slice inside `backfillLastMessageTimes`. No new endpoint, honors "do NOT re-litigate; follow the spec" + "data already fetched". `store.ts:1550-1607`.
2. **Render `cost` between `project` and `elapsed` on Now-rail cards** — ✓ `nowRailMeta(card.project, card.cost, card.elapsed)` in `NowRail` composes `project · cost · elapsed`, dropping the cost segment when absent (no `· ·` gap). `CtoPanel` feeds `project`/`cost`/`elapsed`; `NowCard` shape updated consistently (only two consumers, both updated). `ctoSections.tsx:203`, `CtoPanel.tsx:196-205`.
3. **Pure selector tests in `ctoView.test.ts`** — ✓ `nowCostLabel` (positive/absent/zero/NaN) and `nowRailMeta` (placement + absent/empty-segment branches), run and passing.
4. **`typecheck && test` green** — ✓ verified above.

### Question (1)
**Cost freshness cadence.** `sessionCost` is captured inside `runBackgroundSync` → `backfillLastMessageTimes`, which fires at app startup and when a new chat session appears — it is **not** a continuous poll. So the Now-rail cost for an actively-working session is a snapshot frozen at the last backfill; a session spending live won't have its cost refresh until the backfill re-runs. The issue hinted at (and the implementer's own self-check flagged) that the header-pill `usage` data is the polled seam; this implementation deliberately chose the one-shot session-list fan-out instead. For a live operational rail this is a real freshness gap worth a design-rationale ruling: is a backfill-snapshot cadence acceptable for the Now-rail cost (consistent with the rail's snapshot model), or should cost ride the same refresh path as the rest of the rail? Answer with rationale, change the refresh source, or defer with explicit rationale.

No Blocks. Cleared for the implementer to resolve the Question, then reassign to `manta-reviewer` for re-check.

antoinedc:
**Re: Question — Cost freshness cadence.** Ruling: the backfill-snapshot cadence is intentional and acceptable; keeping it.

Rationale:
- The CTO pane's data path is explicitly **"No polling"** (§10.2 spec). The Now rail is snapshot-composed from store state — running/blocked state, the step one-liner, and `elapsed` are all derived at render from store state updated by SSE events, not by a per-card poll. Cost rides the same model: it is refreshed by the same `backfillLastMessageTimes` fan-out (startup + whenever the active chat-session set changes), so it re-syncs on exactly the same triggers everything else on the rail does.
- Cost is a **slow-moving, non-urgent** figure (a session's accumulated USD). It needs to be present and roughly current, not live to the cent — a frozen-at-last-backfill snapshot is well within acceptable accuracy for an operational rail, and the rail re-runs the backfill on every session-set change so a long-running new session gets picked up.
- A continuous per-card polls for cost would violate the spec's "no polling" data-path principle and add a directory fan-out per status tick for a number that changes rarely. The `opencodeListSessions` response the box already produces is the canonical per-session cost source; this reuses it rather than adding a new endpoint or a poll.

If a future issue wants the Now-rail cost to tick live (e.g. driven by the box's own usage poller pushing `usage.updated`), that's a clean follow-up — cost can then ride the same live event path as the rest of the rail. Out of scope here per the "do NOT re-litigate the design; follow the spec" bound.

PR #1413, opened 2026-08-29
Wire act-and-report executors for start-job and queue-tonight (BET-1424)

antoinedc:
## What

Wires the remaining §9.3-eligible act-and-report executors into the suggest engine's `executeAction` seam (BET-1403 shipped it with `record-decision` only):

- **`start-job`** → the delegate engine's job-start path. Normalizes the candidate option's payload (`prompt` / `cwd` — the ask path's `directory` spelling is accepted as an alias — / `model` free-text or structured), resolves the hosting project session via the same `resolveForgeOwner` the overnight dispatch uses, then arms the §3.3 delegate gate (`beginDelegateJob`: kill switch, pause, `cto.delegate_begin` ledger row) and starts the job with `actor: "cto"`. The §3.3 concurrent cto-delegate sub-cap is enforced by counting **running** cto-actor job rows — the persistent source the overnight dispatch counts too, since the gate's transient tracker only spans the start call.
- **`queue-tonight`** → the engine's `tonightAdd` (BET-1402 subsystem). The task is normalized from the payload (`name` falling back to the finding text, `prompt`, `project`, finite `value`/`confidence`/`predictedCost` with `cost` as the ask-path alias, refs) and bound to the suggestion card via `cls` + `originId`, so Tonight drill-down edits record their verdicts against the suggestion's verdict subject. `tonightAdd`'s own gates (overnight switch + High tier + queue-full) surface as refusals.
- **`record-decision`** moves verbatim into the same executor factory (one implementation, not two parallel paths).

Refusal contract unchanged and now unit-tested: class/action mismatch, §9.3-ineligible classes (`config-change`), data-unreachable `tool-write`, unwired deps, malformed/incomplete payloads, gate refusals — all return `{ok:false, reason}` so the verb degrades to the veto-window card (whose options stay user-executable through the ask-path executors). A present-but-invalid `model` field refuses the whole act rather than being silently dropped — the card rendered that payload.

## Structure

New `src/server/ctoAct.mjs` (pure factory + payload normalizers, injectable I/O — testable without a live box) + `src/server/ctoAct.test.mjs` (15 tests). `index.mjs` now wires the factory to the live subsystems (`adaptiveCto.proposeFact` / `tonightAdd` / `beginDelegateJob`, `tmux.listProjects`, `delegateEngine.startJob`, `loadDelegateJobs`) in ~20 lines instead of owning executor logic inline.

## Files changed

`src/server/ctoAct.mjs` (new), `src/server/ctoAct.test.mjs` (new), `src/server/index.mjs`.

## Verification results

- PR branch (`multica/BET-1424-act-executors` @ 34534e8f):
  - typecheck: exit 0. Errors: none. log sha256: 84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709
  - test: exit 0, 3851 pass / 0 fail / 7 skip (vitest, 198 files) + 3496 pass / 0 fail (node). Failures: none. log sha256: 3fbbc6914556f41e54cc4d16b5209a39ea96f2d3253a906a616b4c909ff58260
- Base (`main` @ ff3e93d6):
  - typecheck: exit 0. Errors: none. log sha256: 84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709
  - test: exit 0, 3851 pass / 0 fail / 7 skip + 3481 pass / 0 fail. Failures: none. log sha256: db695e2f59e9a7b9321bd91b0558735e83642ba29766ccad5ad163098d8c6327
- **Conclusion:** 0 new failures, 0 pre-existing failures. The +15 node tests are this PR's (`ctoAct.test.mjs`).

## Follow-ups

- **Follow-up filed: BET-1426 (PM, high).** Queue entries whose `project` never resolves to a tracked session are skip-ledgered every overnight tick forever — pre-existing since BET-1419 on the ask path (`project: payload.project ?? null`); the act path passes the payload through consistently and this follow-up owns the systemic fix on both add paths / dispatch.

**Base: origin/main @ ff3e93d6**

### DIFF
--- a/src/server/ctoAct.mjs
+++ b/src/server/ctoAct.mjs
@@ -1,0 +1,217 @@
+// src/server/ctoAct.mjs
+// BET-1424 — the §9.2 act-and-report executors for the §9.3-eligible classes.
+//
+// The suggestion engine's act branch (ctoSuggest.mjs) calls ONE injected
+// `executeAction({cls, action, candidate})` with the candidate's primary bound
+// option. BET-1403 wired only `record-decision` inline in index.mjs; this
+// module owns all three §9.3-eligible executors (ctoTrust.mjs eligibility map)
+// behind one factory:
+//
+//   - record-decision → a gatekeeper-checked decision fact on the CTO's own
+//     blackboard (the engine's fact route). Behavior carried over verbatim
+//     from the BET-1403 inline executor.
+//   - queue-tonight   → an entry in tonight's overnight queue via the engine's
+//     tonightAdd (§10.4/§11.4 — planning-only, nothing runs until the window
+//     opens; trivially reversible).
+//   - start-job       → a worktree-isolated delegate job (own git worktree +
+//     branch; never touches the user's checkout), started with actor "cto"
+//     under the §3.3 delegate gate (kill switch, pause, concurrent cap).
+//
+// Everything else refuses with `no-executor` (config-change is §9.3-capped at
+// ask permanently; tool-write stays data-unreachable until a tool holds the
+// §7.4 write ring). The refusal contract is load-bearing (§9.2): act-and-report
+// must never silently no-op — every refusal returns {ok:false, reason} so the
+// verb degrades to the veto-window card, whose options remain user-executable
+// through the renderer's ask-path executors.
+//
+// Payload normalization: the generator emits free-form payloads; a present-but
+// malformed field refuses the act (incomplete-payload) rather than being
+// silently dropped — the card rendered THIS payload, so executing a doctored
+// version of it would not be acting on what was shown. The field names follow
+// the renderer's ask-path contract (ctoView.ts executeSuggestionOption) where
+// one exists: start-job accepts `cwd` (alias `directory`), queue-tonight reads
+// `cost` as the predictedCost alias.
+//
+// Pure over injected I/O — testable without a live tmux/opencode/delegate.
+
+import { ACTOR, RATE_LIMITS } from "./ctoEngine.mjs";
+import { resolveForgeOwner } from "./delegate.mjs";
+
+// ---------------------------------------------------------------------------
+// Payload normalization (§9.1 schema → the executor's typed input)
+// ---------------------------------------------------------------------------
+
+// record-decision: {statement, project, refs?} (+ finding refs fallback).
+export function normalizeRecordDecisionPayload(payload, { findingRefs = [] } = {}) {
+  const p = payload && typeof payload === "object" ? payload : {};
+  const statement = typeof p.statement === "string" ? p.statement.trim() : "";
+  const project = typeof p.project === "string" ? p.project.trim() : "";
+  const refs = Array.isArray(p.refs) && p.refs.length
+    ? p.refs.filter((r) => typeof r === "string")
+    : (Array.isArray(findingRefs) ? findingRefs.filter((r) => typeof r === "string") : []);
+  if (!statement || !project || refs.length === 0) return null;
+  return { statement, project, refs };
+}
+
+// start-job: {prompt, cwd | directory, model?, ...}. `model` mirrors the
+// delegate engine's startJob contract: a free-text name or a structured
+// {providerID, modelID, variant?}. Returns null when prompt/cwd are missing or
+// when a supplied model has neither a valid shape.
+export function normalizeStartJobPayload(payload) {
+  const p = payload && typeof payload === "object" ? payload : {};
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review (manta-reviewer) — BET-1424, cycle 1

**Verdict: PASS** — 0 Blocks, 0 Questions, 2 nits. Routing to `manta-pm` for merge.

### Base freshness
Branched exactly off current `origin/main` (`ff3e93d6` = merge-base). No stale-base artifacts; `git diff --stat origin/main...HEAD` shows exactly the 3 in-scope files (`ctoAct.mjs` +217 new, `ctoAct.test.mjs` +292 new, `index.mjs` +47/−18). No scope creep; `ctoSuggest.mjs` and `ctoTrust.mjs` untouched as the issue requires.

| Issue criterion | Status |
|---|---|
| `start-job` → delegate engine's job-start path (prompt/cwd/model from the candidate option) | ✅ Verified end-to-end: normalizer (incl. `directory`→`cwd` alias, present-but-invalid `model` refuses rather than being dropped) → running-rows cap → `resolveForgeOwner` project resolution → §3.3 gate arm → `startJob({prompt, model?, parentSessionID, parentDirectory, actor:"cto"})`. The free-text `model` claim checked in `delegate.mjs` `startJob`: a named model is validated against the Sub-ticked catalogue and an unmatchable one fails loudly — refusal then degrades to the veto-window card, exactly as reported. |
| `queue-tonight` → overnight queue add | ✅ Wired to `tonightAdd`; normalized task fields match the engine's contract field-for-field (`name`/`prompt`/`project`/`value`/`confidence`/`predictedCost` with `cost` alias/`refs`/`cls`/`originId`). `tonightAdd`'s own gates (overnight switch, queue-full) surface as refusals — verified in `ctoEngine.mjs`. The `originId` binding is meaningful: tonight drill-down edits record verdicts against `originId`. |
| Refusal contract: never silently no-op; unwired → veto-window degradation | ✅ `ctoSuggest.mjs` act block unchanged: `{ok:false}` → verb degrades to the veto-window card. Refusals unit-tested across class-mismatch, §9.3-ineligible (`config-change`/`tool-write`/unknown), unwired deps, malformed/incomplete payloads, gate/cap/project refusals (gate never armed on the latter). |
| `record-decision` preserved | ✅ Diffed the new factory branch against the BET-1403 inline executor: payload normalization, `proposeFact({project, kind:"decision", statement, refs, sender:"cto"})` call, and return shapes are verbatim. Now one implementation, not two. |
| Cap semantics consistent with the overnight dispatch | ✅ The act path's running-rows filter (`actor==="cto" && status==="running"` over persisted job rows) is byte-identical to the engine's `runningCtoJobs()` — the persistent source the dispatch counts. Gate ordering is correct: cap/project refusals happen before `beginDelegateJob`, so no orphan `cto.delegate_begin` ledger row; `gate.release()` runs in `finally` on both success and start-failure (test-asserted). |

### Mechanical gate (my own run, PR branch @ `34534e8f`)
- `npm test` → **exit 0**: vitest 198 files, **3851 pass / 0 fail / 7 skip**; node **3496 pass / 0 fail** — exact match to the implementer's reported PR-branch numbers, so per test-result-reuse policy the base re-run was not repeated. All 15 new `ctoAct` tests confirmed passing in the node run.

### Anti-spaghetti report
Read. The flagged jscpd clones in `index.mjs` (`listSegments` readers, profile-verdict renders) are pre-existing code outside this PR's diff hunk — coincidental, discounted. One legitimate observation folded into nit 1 below.

### Nits (non-blocking)
1. **Running-rows filter duplication.** The `actor/status==="running"` filter now exists twice: `ctoEngine.mjs` `runningCtoJobs()` and inline in `ctoAct.mjs` (justified by the injected-I/O design, ~5 lines). If the definition of "running" ever changes, they can diverge. Fine for now; a future refactor could export the filter from a shared module.
2. **Narrow count→start race.** Two concurrent act-path starts could both pass the running-rows check before either job row exists (the gate's transient tracker only spans the start call). Matches the overnight-dispatch precedent, the suggest loop processes candidates sequentially, and the delegate engine's box-wide cap of 5 backstops it — acceptable as-is, noted for awareness.

The BET-1426 follow-up (queue entries whose `project` never resolves skip-forever, pre-existing since BET-1419 on the ask path) is correctly scoped out — fixing it here would drag this PR into the overnight dispatch subsystem, and the filed follow-up owns both add paths. Agreed with the classification.

PR #1414, opened 2026-08-29
BET-1426: tonight-queue project validation at add time + one-time dispatch cleanup

antoinedc:
## Summary

Fixes BET-1426. An overnight-queue entry whose `project` does not resolve to a tracked project session was never dispatched and never removed: every tick of the open window it hit the dispatch guard and appended a `cto.overnight.skip` ledger row, lingering in the Tonight queue indefinitely.

Per the triage decision on the issue (fix centrally in `tonightAdd`, keep both add paths as pass-through, plus a one-time dispatch-time cleanup):

### 1. Add-time validation (`tonightAdd`, src/server/ctoEngine.mjs)
Both add paths (renderer ask path, act-path executor) funnel into `tonightAdd` — one validation point covers both, and `ctoView.ts`/`ctoAct.mjs` stay untouched (BET-1424's pass-through design preserved).
- Named project resolves via `resolveForgeOwner(listProjects(), project)` → store the **canonical** owner dir (the owning project's `defaultCwd`, so the row keeps resolving by exact match even when window paths churn). `resolveForgeOwner` now returns `defaultCwd` alongside `parentSessionID`/`tmuxSession` (additive; all existing consumers destructure).
- Missing/unresolvable + **exactly one** hostable tracked project session → auto-resolve to it; noted in the ack (`projectResolved: "auto"`) and the `cto.overnight.queue_add` ledger row (`resolved`).
- Otherwise **refuse** with a reason the existing call paths already surface (act path → card error; ask path → option-executor error). The user learns at the moment they act; the queue only ever holds runnable rows.
- An unreadable project list (tmux down / seam missing) refuses the add — it cannot verify resolvability.

### 2. Dispatch-time backstop (`dispatchPlan`)
Rows queued before this fix are cleaned on the **first** skip: when the project list was READABLE (a null/throw is a transient failure, not a verdict — those keep the old skip-and-retain behavior) and the candidate is a queue row, the entry is removed once: one final `cto.overnight.skip` row (`removed: true`), its pinned slot dropped (same discipline as `tonightRemove`), and a needs-you blocker via the engine's `recordBlocker` → blocker-card channel. No re-skip per tick; §10.4 tonight counts reflect the removal (`getState().tonightCount` reads the queue).

Base: origin/main @ 7dc49d94 (BET-1424 merged — sequencing per the triage comment: BET-1424 → BET-1426 → BET-1425).

## Verification results

```
typecheck: exit 0 (tsc node + web)
test: 3498 pass / 0 fail / 0 skip (node:test + vitest)
```

Logs: /tmp/typecheck-pr.log, /tmp/test-pr.log

## Tests

- `tonightAdd` resolution: explicit resolve + canonicalization (subdir match → `defaultCwd`), auto-resolve (single hostable project, both missing and named-but-unresolvable payloads), refuse (zero hostable / ambiguous two / unreadable list), ledger `resolved` marker.
- Dispatch backstop: single removal on first skip (`removed: true` final row + needs-you blocker + §10.4 count 0) and **no re-skip on the next tick**; transient unreadable list retains the row with no blocker.
- Updated the pre-BET-1426 test that asserted the old survive-on-skip behavior, and the veto-lifecycle test that unknowingly relied on the survive-forever bug to keep `candidateCount > 0` across nights (now re-seeds a fresh task for night 2).
- `delegate.test.mjs`: `resolveForgeOwner` shape assertions updated for the additive `defaultCwd`.

### DIFF
--- a/src/server/ctoEngine.mjs
+++ b/src/server/ctoEngine.mjs
@@ -1291,12 +1291,43 @@ export function createCtoEngine(deps = {}) {
         const parent =
           typeof project === "string" && project ? resolveForgeOwner(projects, project) : null;
         if (!parent?.parentSessionID) {
-          await ledgerLog({
-            kind: "cto.overnight.skip",
-            id: cand.id,
-            reason: "no tracked project session to host the job",
-            project,
-          });
+          // BET-1426: an unresolvable project can never host the job — re-skip
+          // per tick would pollute the ledger forever. When the project list
+          // was READABLE (a null/throw is a transient read failure, not a
+          // verdict) and the candidate is a queue row, remove the entry on the
+          // first skip: one final skip row + a needs-you blocker. Rows queued
+          // before the add-time validation are cleaned here too.
+          if (Array.isArray(projects) && task) {
+            const rows = await tonightQueueRows();
+            await saveTonightQueueRows(rows.filter((r) => r?.id !== task.id));
+            if (overnight) {
+              await overnight
+                .updateWindow((prev) =>
+                  prev?.pinnedOrder?.length
+                    ? normalizeWindow({ ...normalizeWindow(prev), pinnedOrder: prev.pinnedOrder.filter((pid) => pid !== task.id) })
+                    : null,
+                )
+                .catch(() => {});
+            }
+            await ledgerLog({
+              kind: "cto.overnight.skip",
+              id: cand.id,
+              reason: "no tracked project session to host the job — entry removed from tonight's queue",
+              project,
+              removed: true,
+            });
+            await recordBlocker(
+              "overnight_queue",
+              `Tonight task "${task.name ?? cand.id}" was removed — no tracked project session could host it.`,
+            );
+          } else {
+            await ledgerLog({
+              kind: "cto.overnight.skip",
+              id: cand.id,
+              reason: "no tracked project session to host the job",
+              project,
+            });
+          }
           continue;
         }
         const prompt =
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-1426 (review cycle 1)

**Verdict: PASS — 0 Blocks, 0 Questions, 3 nits.** Routing to `manta-pm` for merge.

### Mechanical gates (run by reviewer on PR branch `ac44553a`)

- `npm run typecheck`: **exit 0**, zero errors (both tsconfig.node + tsconfig.web).
- `npm test`: **exit 0, 3498 pass / 0 fail / 0 skip** — exactly matches the implementer's reported PR-branch result (base claim of 3496 + 2 net new tests is consistent; per test-reuse policy I did not re-run base).
- CI: `typecheck-test` (the required context) **pass**. `duplication-gate` red — de-required since 2026-07-02, advisory only; see anti-spaghetti note below.

### Acceptance-criteria checklist (verified independently, not from the PR prose)

| Criterion (issue + triage decision) | Verified how | Result |
|---|---|---|
| Add-time validation central in `tonightAdd`, both add paths funnel there | Read `tonightAdd` — resolve via `resolveForgeOwner`, store canonical `defaultCwd`; auto-resolve when exactly one hostable project (both missing- and named-but-unresolvable payloads); refuse otherwise | ✅ |
| Refusal reasons surface through existing call paths (the claim lives *outside* the diff, so checked with own eyes) | `ctoView.ts:598` ask path returns `r?.error` through the option executor; `ctoAct.mjs:165` act path returns `result?.error` as the act reason | ✅ |
| Dispatch backstop: first-skip removal, never re-skip per tick | Removal gated on `Array.isArray(projects) && task` (readable list + queue row); one final `cto.overnight.skip` row with `removed: true`; pinned-slot drop is byte-identical to `tonightRemove`'s discipline; `recordBlocker` needs-you item; test asserts **no re-skip on the next tick** | ✅ |
| Transient project-list failure retains the row (not a verdict) | `projects = null`/throw → old skip-and-retain branch, no blocker; explicit test | ✅ |
| Pass-through preserved — no `ctoView.ts`/`ctoAct.mjs` changes | `git diff --stat origin/main...HEAD`: 4 files, none of them those | ✅ |
| §10.4 tonight counts reflect removal | `getState().tonightCount === 0` asserted post-removal | ✅ |
| Base freshness (TEN-134) | merge-base `7dc49d94` == `origin/main` tip (BET-1424 merged — sequencing honored); no phantom deletions | ✅ |
| Tests for resolve / canonicalize / auto-resolve / refuse / single-removal / no-re-skip / transient-retain / §10.4 count | All present in `ctoEngine.overnight.test.mjs` + `delegate.test.mjs` shape updates | ✅ |

### Scope judgment

`src/server/delegate.mjs` (+14/−2) deviates from the triage's "ctoEngine.mjs only" file list, but the deviation is **justified, not scope creep**: `resolveForgeOwner` lives in `delegate.mjs`, and the canonicalization the triage itself demanded requires it to return `defaultCwd`. The change is purely additive (all production consumers destructure; both `createCtoEngine` sites inject `listProjects`; tests updated for the new shape). The two updated pre-existing tests are behavior-change-driven and documented (one asserted the old survive-on-skip behavior this issue deliberately replaces; the veto-lifecycle test unknowingly relied on survive-forever to keep `candidateCount > 0`).

### Anti-spaghetti (sticky report read)

Flagged clones are all in `delegate.test.mjs` — sibling-test harness/`observeEvent` fixture repetition, coincidental per the judgment rules. No business-logic duplication introduced, no dead code left behind. The advisory `duplication-gate` red is consistent with this; not blocking.

1. **Window-churn removal is also covered by the backstop** — if the project still exists but all its windows lost their live opencode session between add and dispatch (chat window closed), the row is removed on first skip with a needs-you blocker. Bounded (one removal, one blocker), coherent with "the queue only holds runnable rows," and within the triage decision's letter ("no resolvable project → remove on first skip") — just noting it's a broader verdict than "project deleted."
2. **Degenerate-config refusal** — `typeof listProjects !== "function"` refuses adds. Production injects the seam at both engine sites, and refusing is the safe direction; noting only.
3. **Same-row double-dispatch in one tick** — if the same queue id somehow appeared twice in `plan.selected`, the second iteration would fall to the plain-skip branch (row already gone → `task` null). Harmless and unreachable in practice.

### Calibration note

The implementer's self-check flagged "refuse-reason pass-through has no dedicated ctoAct-level test" as its weakest area — I verified the wiring directly in both call paths instead of accepting the prose, and it holds. The strongest sycophancy trap ("rest was already clean") doesn't apply here: nothing in this PR claims unverified pre-existing cleanliness, and the two "already done"-adjacent claims I checked (error surfacing, seam injection) were both true.

PR #1415, opened 2026-08-29
BET-1422: ROI roll — count squash-merged CTO PRs (forge merged-state signal)

antoinedc:
## BET-1422 — ROI roll: count squash-merged CTO PRs

**Base: origin/main @ ece348b0**

Closes the undercount in the §12.4 ROI roll: a squash merge creates a NEW commit, so a delegate job's branch tip never becomes an ancestor of the project HEAD and the two local-git merge signals (`exists === false` deleted-on-merge, `merge-base --is-ancestor`) can never fire for it. Every squash-merged CTO PR was invisible to the roll.

### Mechanism (one deviation from the issue text, documented)

The issue's primary wording probes the PR referenced by the job record's `link.pr`. Verified in the codebase: **no dispatch path ever attaches a link to a cto-actor job** — overnight dispatch and the §9.3 `start-job` executor both call `delegateEngine.startJob` without `link`, and forge-rule-triggered jobs default to `actor: "user"`, which the roll never samples. A literal `link.pr`-only probe would be dead code that can never fire. The issue's own parenthetical alternative ("or record the merge event as a §14.5 ledger row…") confirms the mechanism was the author's secondary concern. So this PR implements the *equivalent but airtight* signal:

- **PR identity by branch head**: `discoverJobPr` resolves the job's own PR on the forge by `headRef === job.branch` (open OR merged — the roll may first run after the merge), and `probeForgePrMerged` verifies the PR's head branch matches before counting (identity anchor — a number alone could name someone else's PR).
- **Bounded re-querying**: the resolution persists on the pending row (`pr`); a definitive no-PR stamps `prTried` so PR-less jobs never re-query on later refreshes; an unconsultable forge (throw) leaves both markers unset so the next refresh retries. Local-git signals keep precedence — a merged/deleted branch never touches the forge.
- **Adapter extension, not a parallel API**: `listPullRequests` gains a `head` filter (GitHub `head=owner:branch`, GitLab `source_branch`), so PR resolution is one exact call instead of paging `state=all`.

### Files changed

```
src/server/ctoBudget.mjs         |  84 ++++++--
src/server/ctoBudget.test.mjs    |  65 +++++-
src/server/forge/github.mjs      |  12 +-
src/server/forge/github.test.mjs |  14 ++
src/server/forge/gitlab.mjs      |   8 +-
src/server/forge/gitlab.test.mjs |  12 ++
src/server/index.mjs             |  46 ++++-
```

- `ctoBudget.mjs`: `isJobMerged` gains the `prMerged` signal (local signals keep precedence); `refreshRoi` probe pass resolves + probes; two new injected seams (`discoverJobPr`, `forgeProbe`) defaulting to never-count so tests without them are unaffected.
- `index.mjs`: wires both seams over the cached 60s repo walk (`local.forgeProbe`) + existing token resolution — at most one filesystem scan per minute regardless of pending-row count.
- Tests: `isJobMerged` precedence cases; a 4-job `refreshRoi` case (squash-merged → counted, still-open → not, never-had-a-PR → marked definitive, forge-down → retried next refresh, no re-discovery of resolved rows); adapter tests pinning the `head`/`source_branch` request params.

### Cross-cutting notes

- The §14.5-ledger-on-cleanup alternative from the issue was **not** taken — the forge probe needs no new ledger row kind and needs no cleanup-path changes.
- PRs deleted from the forge before the first probe remain uncountable — bounded by the 60-day pending retention, documented limitation.
- **Workdir incident (transparency):** the first push of this branch briefly contained an unrelated uncommitted file (`src/server/ctoToolCatalog.mjs`, BET-1395 work) swept by a `git add -A` in the shared checkout. It was removed by force-push within minutes; the final diff above is exactly this issue's 7 files. The BET-1395 work was restored untracked in the checkout.

**Verification results:**
- PR branch (`multica/BET-1422-squash-merged-roi` @ `6917a59d`):
  - test: exit `0`. vitest: `198` files, `3851` pass / `0` fail / `7` skip; node --test: `3501` pass / `0` fail. Failures: none. log sha256: `d7c742b7e90a0895d33f35a5e482652ca13d92e2efb25aa1b50f42ee168eb5ed`
- Base (`main` @ `ece348b0`):
  - test: exit `0`. vitest: `198` files, `3851` pass / `0` fail / `7` skip; node --test: `3498` pass / `0` fail. Failures: none. log sha256: `482395cb41b4e137e3fe0351d234da4fba54d2d2d98ddf11f9b44202b8bafb1e`
- **Conclusion:** `0` test failures are pre-existing (identical between branches), `0` failures are new and caused by this PR, `3` node tests were added by this PR (3501 vs 3498).

### DIFF
--- a/src/server/ctoBudget.mjs
+++ b/src/server/ctoBudget.mjs
@@ -555,17 +555,23 @@ export function preSurfacedIncidents({ resolvedRows, createdTsByCardId, promptTs
 
 /**
  * A delegate record's branch merged — pure predicate over the injected git
- * probe of the project repo:
+ * probe of the project repo plus the injected forge probe (BET-1422):
  *   - branch gone (`exists === false`) → merged: the cleanup path deletes
  *     with `git branch -d`, which only succeeds on merged branches — the
  *     deletion is the merge's fingerprint;
  *   - branch present and `merge-base --is-ancestor` true → merged locally;
- *   - otherwise (including a probe failure) → not merged.
+ *   - branch present but not an ancestor, and the PR the forge carries for
+ *     the branch reports merged (`prMerged === true`) → merged: a squash
+ *     merge creates a NEW commit, so the local branch tip is never an
+ *     ancestor of HEAD and the two local-git signals cannot see it;
+ *   - otherwise (including a probe failure or an unknown forge answer) →
+ *     not merged.
  */
 export function isJobMerged(probe) {
   if (!probe || typeof probe !== "object") return false;
   if (probe.exists === false) return true;
-  return probe.isAncestor === true;
+  if (probe.isAncestor === true) return true;
+  return probe.prMerged === true;
 }
 
 /** The monthly recommendation. Copy-only; the dial is never written. */
@@ -632,6 +638,18 @@ export function createCtoBudget({
   // delegate jobs, the project's git read, verdicts, segments and ledger):
   jobsRead = async () => [], // () => [{id, actor, status, branch, cwd, finishedAt}]
   gitProbe = async () => ({ exists: false, isAncestor: false }), // ({cwd, branch}) => {exists, isAncestor}
+  // BET-1422 forge seams — the squash-merge signal a local git probe cannot
+  // see (a squash creates a NEW commit, so the branch tip never becomes an
+  // ancestor of HEAD). discoverJobPr resolves the job's own PR on the forge
+  // by its branch head, open OR merged; null = the forge answered and no PR
+  // matches (definitive — the roll stops re-asking), a THROW = the forge was
+  // not consultable (transient — the roll retries on a later refresh).
+  // forgeProbe answers that PR's merged state: true/false/null (unknown —
+  // the row stays uncounted and is retried later).
+  discoverJobPr = async () => {
+    throw new Error("discoverJobPr not wired");
+  }, // ({cwd, branch}) => {repoKey, number} | null (null = definitive no-PR)
+  forgeProbe = async () => null, // ({repoKey, number, head}) => boolean | null
   verdictsRead = async () => [], // () => [{verdict, never, ts}]
   segmentsRead = async () => [], // () => [{sessionID, events: [{t, kind}]}]
   ledgerRead = async () => [], // () => ledger rows [{kind, ts, ...}]
@@ -900,7 +918,11 @@ export function createCtoBudget({
    * Advance the monthly roll (§12.4). Idempotent per refresh; safe to call on
    * every health read. Samples fresh terminal CTO jobs into `roi.pending`
    * (the jobs store sweeps after 7 days — the branch names are the durable
-   * record), probes pending branches against the project repo, and recomputes
+   * record), probes pending branches against the project repo and — for
+   * branches still present and not an ancestor of HEAD, which is what a
+   * squash merge leaves behind (BET-1422) — against the forge: the job's PR
+   * is resolved by branch head once (`prTried` marks a definitive no-PR so
+   * PR-less jobs never re-query), then its merged state is probed. Recomputes
    * the store-derived counters: the current month on every refresh, the
    * previous month once after it closes (frozen thereafter). Persisted
    * best-effort — a store failure degrades the report, never the caller.
@@ -915,6 +937,13 @@ export function createCtoBudget({
     }
     const months = { ...payload.roi.months };
     const pending = payload.roi.pending.map((j) => ({ ...j }));
+    const countMerged = (row) => {
+      const key = roiMonthKey(typeof row.finishedAt === "number" ? row.finishedAt : t);
+      const acc = monthAccumulator(months, key);
+      acc.merged += 1;
+      months[key] = acc;
+      row.counted = true;
+    };
 
     // 1. Sample terminal CTO-actor jobs not yet snapshotted.
     let jobs = [];
@@ -940,8 +969,9 @@ export function createCtoBudget({
     }
 
     // 2. Probe pending branches (merged = branch deleted-on-merge, or present
-    //    and an ancestor of the project's HEAD). Probe failures leave the row
-    //    uncounted for a later refresh.
+    //    and an ancestor of the project's HEAD, or its forge PR reports
+    //    merged — the squash-merge case, BET-1422). Probe failures leave the
+    //    row uncounted for a later refresh.
     for (const row of pending) {
       if (row.counted === true) continue;
       let probe = { exists: false, isAncestor: false };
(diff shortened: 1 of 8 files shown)


antoinedc:
## Review: 1 Block (fix-here), 0 Questions, 2 nits — returned to implementer

**Base freshness:** merge-base(HEAD, origin/main) == origin/main == `ece348b0` — branched off current tip, no stale-base artifacts.

### Verified good

- **The `link.pr` deviation is justified.** I independently confirmed no dispatch path attaches a link to a cto-actor job: the overnight scheduler (`src/server/index.mjs:1986`) and the §9.3 act executor (`src/server/index.mjs:2157`) both start with `actor: "cto"` and no `link`; the only `link`-passing call site (`src/server/index.mjs:785`) runs under the default non-cto actor. A literal `link.pr` probe would be dead code, and the issue's parenthetical alternative confirms the mechanism was secondary. Branch-head resolution + head-identity check (`headRef === branch` on both discovery and merge-probe) is an airtight equivalent — good catch on the identity anchor.
- **`isJobMerged` semantics** — local signals keep precedence, `prMerged === true` is the only forge-truth accepted, unknown/null → not merged. Tests cover all six combinations including the two precedence cases.
- **`refreshRoi` semantics** — resolution persisted on the row, definitive no-PR stamped (`prTried`), transient throw retried, already-resolved rows never re-discovered. The 4-job test exercises squash-merged / open / never-had-PR / forge-down plus refresh-2 idempotency, and per the implementer's own note it exposed a real first-cut bug that the second commit fixed.
- **Adapter `head` filter** — GitHub `head=owner:branch`, GitLab `source_branch`, both pinned by new adapter tests. Extension of the existing `listPullRequests` filter object, not a parallel API.
- **Scope** — 7 files, all intrinsic to the changed mechanism (`delegate.mjs` untouched is consistent with the deviation; the forge adapters are required by it). No unrelated surface crept in. The shared-checkout incident is disclosed and the final diff is clean.
- **Mechanical gates (my run, PR branch @ `6917a59d`):** `npm run typecheck` exit 0; vitest 198 files / 3851 pass / 0 fail / 7 skip (exit 0); node --test 3501 pass / 0 fail. Matches the implementer's reported PR-branch numbers exactly, so per test-reuse I did not re-run on base (base hash `ece348b0` trusted per review policy).

### **Block (fix-here): the forge seams are wired into the wrong object — the feature is dead in production**

`discoverJobPr` and `probeForgePrMerged` were added to the `computeHealthStats({...})` deps literal inside the `/api/cto/health` handler (`src/server/index.mjs:4294`–`4322`). That object's signature (`src/server/ctoHealth.mjs:80`) accepts only `now / ledgerRead / budgetRead / listSegments / verdictsRead / ctoAmbientCap / roiRead` — the two new properties are silently ignored.

The only production construction of the budget that runs `refreshRoi()` is the singleton at `src/server/index.mjs:2239`–`2308`, and its deps object still ends at `ledgerRead` — it never receives either seam. (The other construction, `src/server/ctoEngine.mjs:348`, is bare-default.) So in production:

- `discoverJobPr` falls back to its default, which **throws** `"discoverJobPr not wired"` → caught as transient → every pending row retries and never resolves a PR;
- `forgeProbe` falls back to `async () => null` → `prMerged` is always null → **no squash-merged job is ever counted**.

Net effect: the exact §12.4 undercount this issue set out to fix remains in production, now behind green unit tests — the tests pass only because `ctoBudget.test.mjs` injects the seams directly into `createCtoBudget`, which no production call site does.

**Fix:** move both functions into the `createCtoBudget({...})` deps at `src/server/index.mjs:2239` (next to `gitProbe`). Strongly suggested companion: one test asserting the singleton wiring passes both seams (e.g. a wiring smoke test that a pending row with a squash-merged PR resolves through the injected functions), so a future move can't silently strand them again. A `refreshRoi` run with the unwired defaults should also be shown to not count — today nothing would catch this regression.

### Nits (non-blocking, address if trivial)

1. `discoverJobPr` stamps a **definitive** no-PR (`prTried = true`) whenever the job's `cwd` is absent from `local.forgeProbe()`'s 60s-cached repo walk — including a `partial: true` walk or a repo the walk hasn't discovered yet. A transiently incomplete filesystem scan would permanently mark a countable job as PR-less (its own semantics say "null = the forge answered"; here the forge was never asked). Cheap hardening: when the walk reports `partial` (or the repo simply isn't found), `throw` instead of returning null so the row stays retriable.
2. The inserted block in `src/server/index.mjs` carries 2-space indentation inside an object literal nested ~8 levels deep — cosmetic only.

antoinedc:
Block addressed in e3fefd07 → rebased clean as cde49928 (branch force-updated; same branch, no new PR).

**The fix:** `discoverJobPr` / `probeForgePrMerged` moved out of the `computeHealthStats({...})` deps (which ignores both) into the `createCtoBudget({...})` singleton deps, next to `gitProbe` — the only construction that runs `refreshRoi()`. The health handler's deps literal is restored to its pre-BET-1422 shape.

**The wiring gate you asked for** — new `src/server/ctoBudgetWiring.test.mjs`:
1. A brace-balanced source-scan test asserting both seams sit INSIDE the `createCtoBudget({...})` deps in `index.mjs` and NOT in any other deps literal (the exact failure mode you caught — index.mjs can't be imported since it boots the server, so the gate reads its source; same technique as the repo's other source-scan gates).
2. A behavioral canary: a `createCtoBudget` with production-shaped jobs + git probe but the two seams left at their unwired defaults never counts a squash-merged job — the exact way this Block hid behind green tests.

**Nit 1 fixed:** a partial repo walk (`partial: true`) or a cwd the walk has not surfaced yet now **throws** in `discoverJobPr` (row stays retriable) instead of stamping a definitive no-PR on incomplete evidence. Only a completed walk answering "no such repo" was already stable, a malformed repoKey, or the forge answering with no headRef match stamp `prTried`.

**Nit 2:** the block moved to the singleton, whose dep indentation it now matches.

**Verification results** (isolated worktree pinned to `cde49928` — the shared checkout was being used concurrently by the BET-1395 run, so the run was done in a separate `git worktree`):
- test: exit `0`. vitest: `3851` pass / `0` fail / `7` skip; node --test: `3504` pass / `0` fail (`3501` prior + the 3 new wiring-gate tests). Failures: none. log sha256: `fb1de34b5e5b22264c9b84bdc7cc55d4ba2619d543bdde69e1657381368e74c6`

**Branch-history note (transparency):** between review cycles, the concurrently-running BET-1395 session committed its two tool-discovery commits onto this branch and pushed them (its shared checkout was left on this branch). My force-push removed them from THIS branch — they are preserved intact on `multica/BET-1395-tool-discovery-wip` (tip `56a1a16b`) for that run to pick up. This branch's diff vs `origin/main @ ece348b0` is exactly this issue's 8 files.

antoinedc:
## Review 2: PASS — 0 Blocks, 0 Questions, 0 nits (cycle-1 Block verified fixed)

**Re-review scope:** delta since `6917a59d` (commit `cde49928`), on top of the cycle-1 review. Base freshness re-checked: merge-base(HEAD, origin/main) == origin/main == `ece348b0`.

### Cycle-1 Block — verified fixed at the specified site

- Both forge seams (`discoverJobPr`, `probeForgePrMerged`) now sit inside the `createCtoBudget({...})` singleton deps (`src/server/index.mjs:2295`–`2322`), next to `gitProbe`; the stranded copies in the `computeHealthStats` deps literal are gone (that object now ends at its own `verdictsRead`, indentation restored). The only thing that runs `refreshRoi()` is now the thing that carries the seams — the undercount fix is live.
- **Regression gate** (`src/server/ctoBudgetWiring.test.mjs`, +3 tests, all ran and passed): (1) source-scan asserts both seams sit inside the `ctoBudget.createCtoBudget(` deps block; (2) asserts they sit in no other deps literal (the computeHealthStats strand is named explicitly); (3) behavioral canary — a production-shaped budget with unwired forge defaults never counts a squash-merged job. The source scan is a heuristic (brace-balanced, quote-aware; index.mjs can't be imported without booting the server), but it's the right tool under that constraint and the canary covers its blind spots. Observation only, no action requested.
- **Partial-walk nit (cycle 1, nit 1) also hardened:** `discoverJobPr` now throws when the repo walk can't prove a repo forge-less (`walk?.partial` → "repo walk partial"; missing cwd → "no scanned repo at …") instead of returning null — an incomplete walk can no longer stamp a definitive no-PR; the row stays retriable. Nit 2 (indentation) resolved by the same restructure.

### Scope & history integrity

Diff is 8 files, all BET-1422-scoped. The concurrent BET-1395 checkout collision did not leak into this PR: branch history is exactly the three BET-1422 commits (`1c1a9173`, `6917a59d`, `cde49928`); the stray BET-1395 work lives on its own branch.

### Mechanical gates (my run, PR head `cde49928`)

- `npm test`: exit 0 — vitest 198 files / 3851 pass / 0 fail / 7 skip; node --test 3504 pass / 0 fail (the +3 are the new wiring-gate tests).

(The implementer's prior results block referenced the pre-fix head `6917a59d`, so per test-reuse policy I re-ran the full suite rather than reusing it.)

### Acceptance criteria — all verified

1. Forge merged-state signal feeds `isJobMerged` alongside the two local-git signals (local signals keep precedence) — ✅ implemented + unit-tested.
2. Extended `isJobMerged` cases — ✅ (6 combinations incl. precedence).
3. `refreshRoi` case with a squash-merged job — ✅ (4-job scenario: counted / open / definitive no-PR / forge-down retry + refresh-2 idempotency).
4. The `link.pr` → branch-head resolution deviation remains justified (verified in cycle 1: no dispatch path attaches a link to a cto-actor job).
5. Now actually reachable in production — the cycle-1 gap that mattered.

PR #1412, opened 2026-08-29
Adaptive CTO: trust promotion ladder + act-and-report (BET-1403)

antoinedc:
## Summary

BET-1403 — the Adaptive CTO's earned-trust promotion ladder + act-and-report (spec §9.2/§9.3/§9.4, §3.5, §10.6-4). Normative sections read only (per issue context-budget note).

**New: `src/server/ctoTrust.mjs`** (+ `ctoTrust.test.mjs`, 18 tests)
- §9.4 ladder math: promotion on the Beta tail bar P(acceptance > 0.9) > 0.95 with ≥ 8 observations (reuses B3's `betaTailAbove`); any 2 rejections in a rolling 10 → demote one step (ask is the floor; the rolling window is consumed by the transition so the same evidence can't re-demote forever).
- §10.6-4 cold-start dominance: no class leaves ask before the global verdict ledger holds `VERDICT_MIN` (15) — enforced both at promotion time and at verb-selection time.
- §9.3 eligibility map (literal class → reversibility): `record-decision`/`queue-tonight`/`start-job` eligible; `config-change`/`tool-write` capped at ask permanently; **unknown class = capped** (fail-closed).
- Per-class state persisted under `es.trust` (engine-state, shared store pattern): tiers, ask Beta (a/b), veto-record Beta (va/vb), rolling window, pending digest-announcement queue.
- The B3 verdict sink registry is the counters' only input: `noteVerdictEffects` maps §9.5 (`accept`/`edit` → success; `dismiss`/`correct`/`never` → rejection; `open`/`expire` never counted) and routes a `veto` verdict to the veto-window record, not the ask record. `noteVetoOutcome` is the integration point for BET-1419's window machinery (executed window = acceptance, cancel = rejection).
- act-and-report bookkeeping: `recordAct` ledger rows + pending announcement; `listAnnouncements`/`markAnnounced` give the digest exactly-once consumption.

[...]

PR #1416, opened 2026-08-29
BET-1425: engine-state write hygiene — per-key read-modify-write for every engine-state.json writer

antoinedc:
# BET-1425 — Engine-state write hygiene: per-key read-modify-write for every engine-state.json writer

**Base: origin/main @ ece348b0**

## Mechanism choice (per the issue's pick-one mandate)

**Per-key read-modify-write helper** (`patchEngineState` in `src/server/ctoStores.mjs`), not a per-key store split. A store split would change every reader's access pattern and reshuffle state shapes (out of scope); the helper fixes the write path only, in one place, and every writer keeps its keys.

## The bug

Every writer on the shared `engine-state.json` loaded once, then saved `{ ...snapshot, myKey }` — spreading the whole load-time snapshot back over the file. Any key another writer changed between a writer's load and its save was silently resurrected/clobbered. Durability invariant (BET-1403 review wording): **a key's value must survive any concurrent writer that does not own that key.**

## The fix

`patchEngineState(mutation, { engineState })` — under a process-wide mutex it **loads fresh → applies only the patch keys → atomic save** (reusing the jsonStore temp-file+rename atomicity via the store's own `save`). `mutation` is a static patch object (keys the writer owns) or a `(fresh) => patch` function for derived values (array pushes, sub-key merges). Setting a key to `undefined` deletes it.

The mutex serializes the whole load→merge→save sequence — the path mutex inside `writeJsonAtomic` covers only the rename, not the read, so it alone could not deliver the invariant. (Different mutexes; no nesting deadlock.)

## Writer migration (re-derived at implement time via `grep -rn "engineState.save(" src/server`; the review's line numbers were stale post-#1413/#1414)

| Writer | Owned key(s) | Form |
|---|---|---|
| `ctoEngine.recordBlocker` | `pendingBlockers` | push derived from fresh |
| `ctoEngine.watcherTick` | `watchers.lastAutoDay` | sub-key merge on fresh (`watchers` is shared with the migration marker) |
| `ctoEngine.saveTonightQueueRows` | `tonightQueue` | static patch |
| `ctoEngine.finalizeRollups` | `rollupCursor` | per-level deltas onto fresh — the old shape spread the whole pre-rollup payload back after minutes of rollup processing |
| `ctoBackfill.saveState` (10 call sites) | `backfill*` | static patches |
| `ctoSegments.persistG` | `segmentGMinutes` | static patch |
| `ctoWatchers.migrateLegacy` | `watchers.migrated/at` | sub-key merge on fresh |
| `ctoFacts.saveState` (6 call sites) | `factReliability`, `factQueue`, `appliedProposals`, `factTuning` | owned-key projection (its threaded `state` starts as a full-file load — only its own keys may be written back) |
| `ctoSessions.activeAdd/activeRemove` | `activeEphemeral` | add/remove derived from fresh |

Out of scope per the issue: `ctoTrust.mjs` (writes nothing to engine-state since #1412 — verified by grep). `ctoSuggest.saveState` was **dead code** (no callers, not in the factory's return) — deleted rather than migrated.

Post-migration grep: zero direct `engineState.save` calls outside `ctoStores.mjs` (the helper itself + test seeding).

## Tests

7 new cases in `src/server/ctoStores.test.mjs` (appended to the existing BET-1375 harness; interleaving regressions reuse the mirrored-order pattern from `ctoTrust.test.mjs`):

- static patch vs a stale-snapshot writer, **both orders** — X survives B's save, Y lands
- parallel patches compose (every key lands — the mutex, not the rename, is the guard)
- `(fresh) => patch` read-modify-writes compose under concurrency (two increments both land)
- undefined-deletes; validation rejects non-object/non-function mutations
- `watchers` sub-key coexistence via the REAL standing-query engine's `migrateLegacy` vs the engine's day-marker patch shape, **both orders**
- the BET-1403 resurrect shape (tonight-queue removal vs unrelated writers), engine-state keys

Existing engine-state suites stay green.

## Cross-cutting

- None. Renderer/main untouched; engine-state file shape unchanged (keys and values identical, only write ordering changes). The overnight test's raw-save seam (`engineStateSave`) is deliberately kept — it simulates the hostile pre-1425 writer shape for the trust-durability regression.

## Verification results

- PR branch (`multica/BET-1425-engine-state-per-key` @ `c749469e`):
  - test: exit 0, 3505 pass / 0 fail / 0 skip. Failures: none. log sha256: `7aa72cebfb539c07a12818379862d9436f75fdff506f13dc20268996404619a0`
- Base (`main` @ `ece348b0`):
  - test: exit 0, 3498 pass / 0 fail / 0 skip. Failures: none. log sha256: `6b575209186f525a9956df5677e2162e0f214074ebff42da7d60a0ea2b78cd0b`
- **Conclusion:** 0 failures on either branch; the +7 tests are this PR's new regressions.

### DIFF
--- a/src/server/ctoBackfill.mjs
+++ b/src/server/ctoBackfill.mjs
@@ -26,7 +26,7 @@
 
 import { DAY_MS, HOUR_MS, WEEK_MS, createRollupRunner, defaultExists, startOfDay, startOfHour, startOfWeek, windowFor, windowId } from "./ctoRollups.mjs";
 import { createSegmenter } from "./ctoSegments.mjs";
-import { engineStateStore, segmentsStore, rollupsStore, ledgerStore } from "./ctoStores.mjs";
+import { engineStateStore, segmentsStore, rollupsStore, ledgerStore, patchEngineState } from "./ctoStores.mjs";
 import { fetchLedgerRows } from "./modelLedger.mjs";
 
 export const DEFAULT_BACKFILL_CAP_USD = 3;
@@ -172,8 +172,9 @@ export function createCtoBackfill(deps = {}) {
     return st && typeof st === "object" ? st : {};
   }
   async function saveState(patch) {
-    const cur = await readState();
-    await engineState.save({ ...cur, ...patch });
+    // BET-1425: per-key RMW — patches carry only backfill-owned keys, so a
+    // concurrent writer's keys survive this save.
+    await patchEngineState(patch, { engineState });
   }
 
   // The A7 rollup runner is constructed lazily + cached; it reduces past
--- a/src/server/ctoEngine.mjs
+++ b/src/server/ctoEngine.mjs
@@ -55,6 +55,7 @@ import {
   verdictsStore,
   watchersStore,
   factsArchiveStore,
+  patchEngineState,
   purgeExpiredInbox,
 } from "./ctoStores.mjs";
 import { createVerdictEngine } from "./ctoVerdicts.mjs";
@@ -414,19 +415,21 @@ export function createCtoEngine(deps = {}) {
   // into engine-state.json. Best-effort — the health record never throws.
   async function recordBlocker(source, reason) {
     try {
-      const payload = await engineState.load();
-      const pending = Array.isArray(payload?.pendingBlockers)
-        ? payload.pendingBlockers
-        : [];
-      pending.push({
+      const blocker = {
         id: nextId(),
         kind: "blocker",
         source,
         reason,
         ts: now(),
         resolved: false,
-      });
-      await engineState.save({ ...payload, pendingBlockers: pending });
+      };
+      // BET-1425: per-key RMW — an array push derived from the FRESH state, so
+      // a concurrent writer's keys survive this save.
+      await patchEngineState((fresh) => {
+        const pending = Array.isArray(fresh?.pendingBlockers) ? [...fresh.pendingBlockers] : [];
+        pending.push(blocker);
+        return { pendingBlockers: pending };
+      }, { engineState });
     } catch {
       /* best-effort */
     }
@@ -1100,7 +1103,13 @@ export function createCtoEngine(deps = {}) {
       if (lastAutoDay === todayKey) return;
       const dayRollups = await loadRecentDayRollups();
       const result = await getWatchers().autoCreate(dayRollups);
-      await engineState.save({ ...es, watchers: { ...(es?.watchers || {}), lastAutoDay: todayKey } });
+      // BET-1425: sub-key merge on `watchers` from the FRESH state — this key
+      // is shared with ctoWatchers' migration marker, so a stale spread would
+      // resurrect/clobber the other writer's marker.
+      await patchEngineState(
+        (fresh) => ({ watchers: { ...(fresh?.watchers || {}), lastAutoDay: todayKey } }),
+        { engineState },
+      );
       if (result.added.length > 0) {
         await ledgerLog({ kind: "watcher.auto_created", count: result.added.length, signatures: result.added.map((a) => a.patternSignature) });
       }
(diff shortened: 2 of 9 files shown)


antoinedc:
## Review — BET-1425 (manta-reviewer, cycle 1)

**Verdict: PASS — 0 Blocks, 0 Questions, 2 nits.** Base is fresh (`merge-base == origin/main` tip, `ece348b0` — no stale-base phantoms).

### Acceptance criteria — verified independently, not from the PR body

| Criterion (issue) | Verification |
|---|---|
| Helper exists in `src/server/ctoStores.mjs` | `patchEngineState` present: process-wide mutex serializes load-fresh → merge-owned-keys → atomic save via the store's existing `writeJsonAtomic` path. Mutation accepted as static patch or `(fresh) => patch`; `undefined` deletes; non-object/non-function rejected. ✓ |
| Mechanism choice stated in PR | Per-key RMW helper (not a store split), rationale given. ✓ |
| Every engine-state writer migrated | Ran the issue's grep myself: `engineState.save(` now exists only inside the helper (`ctoStores.mjs:203`) plus test seeding. Broadened sweep (`.save(` across `src/server`) — every other hit targets a different store. All 11 sites across the 8 modules match the PR's table; `ctoBackfill`'s multi-key saves pass only backfill-owned keys; `ctoFacts.saveState` projects to its 4 owned keys; `ctoTrust.mjs` only *reads* engine-state as the one-time migration source (writes nothing — #1412 holds); `ctoSuggest.saveState` deleted with zero remaining references. Repo-wide TS/renderer sweep: no engine-state writers outside `src/server`. ✓ |
| Interleaving regression test, mirrored-order pattern, both orders | 7 new cases in `ctoStores.test.mjs` implement exactly the issue's pattern (A saves X, B saves stale-derived {old X, new Y} → X survives, Y lands, both orders), plus parallel-patch composition, `(fresh)` RMW composition, undefined-deletes, validation, the real `migrateLegacy` watchers-sub-key coexistence in both orders, and the BET-1403 tonight-queue resurrect shape. Test store deep-copies at the load/save boundary — no aliasing artifacts. ✓ |
| Existing engine-state suites green | Full suite green. ✓ |
| `npm run typecheck && npm test` passes | **Ran both myself on the PR branch (`c749469e`): typecheck exit 0, no errors; tests 3505 pass / 0 fail / 0 skip.** Matches the implementer's reported PR-branch numbers exactly (+7 = the new tests; base 3498 trusted per the reported base hash). ✓ |
| Branch/PR naming | `multica/BET-1425-engine-state-per-key`; PR title carries the key. ✓ |

### Scope

9 files, all within the issue's stated surface (engine-state writers + store + test). No scope creep, no under-delivery found — the "re-derive the writer list at implement time" instruction was followed and my grep reproduces the claimed zero-direct-saves end state.

### Anti-spaghetti report

Read. 2 clones, 0.31% duplication in changed scope — both pre-existing coincidental helpers (`parseSegmentSummaryText` ↔ `parseSuggestionText` JSON-parse utilities; a delta-accumulator pair in `ctoFacts.mjs`). Not business logic this PR introduced and no dead code left behind. No action.

### Nits (non-blocking)

1. `ctoStores.test.mjs` stale-snapshot test: the `stale` snapshot is loaded for documentation but never fed into writer B's patch — B's patch is a static `{Y: 42}`. The invariant still holds and is genuinely exercised (the helper always loads fresh under the lock, which is precisely why a static patch cannot spread stale keys), but wiring the stale read into B's derivation would make the regression fail more loudly if the helper were ever reshaped.
2. `ctoFacts.saveState`'s `catch {}` now also swallows the helper's validation errors (pre-existing best-effort shape preserved) — a malformed patch would fail silently. Fine for this PR; worth remembering if facts state ever gets a schema.

Solid, focused PR. The mutex-not-rename insight (documented in both the helper and a dedicated test) is the right call — the per-path rename mutex alone could not have delivered this invariant.

PR #1425, opened 2026-08-30
BET-1437: deep-link probe blocker cards to secrets surface

antoinedc:
## Summary
Extends the renderer's blocker action mapping so `sourceKind: "probe"` cards (BET-1396 / §10.6-7) deep-link to the secrets surface instead of falling into the ledger fallback.

## Changes
- `src/renderer/ctoView.ts`: `BlockerAction` gains `{ action: "secrets"; key: string | null }`; `blockerTarget` maps probe cards to it; new pure helper `probeSecretKey(body)` extracts the vault KEY NAME the server copy embeds (`update "KEY" on the secrets surface`).
- `src/renderer/CtoPanel.tsx`: `handleAnswer` handles the secrets action via a new `onOpenSecrets` prop; falls back to the ledger modal if the handler returns false.
- `src/renderer/App.tsx`: `onOpenSecretsFromCto` dispatches the existing `manta-open-secrets` bridge with the active chat session id + key, and closes the CTO pane (chat panels stay mounted underneath). Returns false with no active session → ledger fallback.
- `src/renderer/hooks/useSessionResources.ts`: bridge detail may now carry `key`; latched as `secretKeyHint`, exposed to the chat panel, cleared when the card closes or another panel opens.
- `src/renderer/PanelCards.tsx` (`SecretsCard`): new `prefillKey` prop — pre-fills the add/update form's KEY field, focuses the value input, highlights the input with an accent border, and highlights the matching existing row.
- `src/renderer/ChatPanel.tsx`: passes `prefillKey` through to `SecretsCard`.
- `src/renderer/ctoView.test.ts`: tests for probe → secrets with key named, without key, empty body, and no false extraction.

Base: origin/main @ 55d896c2

## Verification results
(to follow in issue comment)

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -359,6 +359,26 @@ function Shell() {
     : null;
   const activeChatSessionId = activeWin?.opencodeSessionId ?? null;
 
+  // BET-1437: probe blocker cards deep-link to the secrets surface. The chat
+  // panels stay mounted under the CTO pane, so dispatching the existing
+  // manta-open-secrets bridge with the ACTIVE session id opens the SecretsCard
+  // there (key pre-filled via the detail's `key`); closing the pane reveals it.
+  // Returns false when there is no active chat session — the caller falls back
+  // to the ledger modal.
+  const onOpenSecretsFromCto = useCallback(
+    (key: string | null): boolean => {
+      if (!activeChatSessionId) return false;
+      window.dispatchEvent(
+        new CustomEvent("manta-open-secrets", {
+          detail: { sessionId: activeChatSessionId, key: key ?? undefined },
+        }),
+      );
+      setCtoOpen(false);
+      return true;
+    },
+    [activeChatSessionId],
+  );
+
   // BET-730: record the active chat in an effect, not during render (render-
   // time ref mutation is a StrictMode hazard). A re-render that just touches
   // unrelated fields must never mutate the visited set.
@@ -2186,7 +2206,7 @@ function Shell() {
                   panels above are suppressed while it is open (their `active`
                   flips on !ctoOpen). Selecting a session closes it. */}
               <PanelShell key="cto" active={ctoOpen}>
-                <CtoPanel state={ctoState} onOpenSession={onOpenSessionFromCto} />
+                <CtoPanel state={ctoState} onOpenSession={onOpenSessionFromCto} onOpenSecrets={onOpenSecretsFromCto} />
               </PanelShell>
               {/* New-session DRAFT layer: shown over the always-mounted
                   session panels when a draft is the active view (user hit +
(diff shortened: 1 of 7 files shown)


antoinedc:
## Review (manta-reviewer) — PASS, 0 Blocks / 0 Questions / 3 nits

**Acceptance-criteria checklist (from the issue, verified independently):**

- **`sourceKind: "probe"` deep-links the secrets surface** — ✅ `blockerTarget` (`src/renderer/ctoView.ts`) returns `{ action: "secrets", key }` for probe cards; `CtoPanel.handleAnswer` routes it through `onOpenSecrets`; `App.tsx` dispatches the existing `manta-open-secrets` bridge with the active chat session id and closes the CTO pane. Bridge liveness verified: `PanelShell` hides inactive panels with `display:none` — chat panels stay **mounted** under the CTO pane, so the session's listener receives the dispatch. No active session → returns false → ledger fallback (§10.3 honest route preserved).
- **Pre-fills/highlights the named key** — ✅ `probeSecretKey` extraction verified against the real server copy (`src/server/ctoCards.mjs:85-86`, both the keyed `update "KEY" on the secrets surface` and keyless branches). `SecretsCard` pre-fills the KEY field, accent-rings the input, highlights a matching existing row, and focuses the value input. Hint latched in `useSessionResources`, cleared when the card closes (no stale highlight).
- **Touch points per issue** — ✅ exactly the stated set (`ctoView.ts`, card action handler `CtoPanel.tsx`, navigation state `App.tsx`/`useSessionResources.ts`/`ChatPanel.tsx`/`PanelCards.tsx`). No scope creep.
- **No server changes** — ✅ diff is renderer-only.
- **Measured criteria run:** `npm run typecheck` exit 0 (no errors); `npm test` exit 0 — **3632 pass / 0 fail / 0 skip**, matching the implementer's PR-branch report exactly. 6 new tests cover probe routing + key extraction, including the quoted-phrase false-positive case.

**On the "Settings → Secrets" deviation:** the issue's wording assumed the secrets surface lives in the Settings pane. I verified it does not — `SettingSectionId` (`src/shared/settingsSchema.ts:48-58`) has general/box/accounts/models/sessions/files/extensions/voice/cto and nothing else. Spec §10.6-7 is normative and says the card's action "deep-links to the fix surface (e.g. secrets card)" — which is exactly where this PR lands. The stale card copy is a server-side bug, correctly filed as BET-1443 and out of scope here per the issue's own "No server changes" constraint. Not a shortfall.

**Anti-spaghetti report:** read — 0 clones in changed scope. Nothing to act on.

**Nits (non-blocking, no action required):**
1. `probeSecretKey`'s regex is coupled to the exact server copy format — acknowledged, pinned by tests, and degrades gracefully (null key → card still opens, no prefill).
2. The KEY input's accent ring is tied to `prefillKey` being set, not to the field's current value — persists if the user edits the key manually. Cosmetic.
3. Re-firing the same deep-link key while the card is already open won't re-run the prefill effect (dep unchanged). Negligible.

PR #1434, opened 2026-08-30
feat(server): durable chat-window topology snapshot (BET-1452 S1)

antoinedc:
## What

Server-side Stage 1 of the BET-1451 topology plan (BET-1452): persist the box's chat-window tree to `~/.manta/topology.json` so a later stage (S2) can hydrate it across box-server restarts.

- New `src/server/topology.mjs` — pure + injected-I/O, mirroring `syncState.mjs` (no tmux import, no bus, no timers):
  - `snapshotFrom(projects, now)` → `{version:1, capturedAt, sessions:[{tmuxSession, defaultCwd, mantaOwned, windows:[{index, name, opencodeSessionId, cwd, worktreePath, active}]}]}` — keeps only windows with a truthy string `opencodeSessionId` (the chat-window discriminator), drops projects left with zero windows, takes `name` from the live tmux window name, `cwd` from `paneCurrentPath` with a `defaultCwd` fallback, `worktreePath` as observed or null (never invented), `mantaOwned` carried off the project.
  - `countWindows` / `shouldPersist` — THE load-bearing rule: a zero-window listing never overwrites a non-empty file, so a crash that restarts tmux EMPTY can't erase the restore point at the moment it's needed.
  - `loadTopology(path?, io?)` — null on missing/unparsable/version-mismatch/non-array-sessions, never throws; `saveTopology` — pretty-printed JSON through the shared `atomicWrite` primitive (temp+rename).
  - `createTopologyPersister({save?, load?, now?})` — hydrates the previous snapshot from disk once, lazily on the first call; skips writes when `shouldPersist` is false (`{persisted:false, reason:"would-clobber"}`) and when the serialized `sessions` are byte-identical to the last write (`reason:"unchanged"`) — without the second check this would rewrite the file every 2s forever on an idle box. Returns the `{persisted, reason}` contract.
- `src/server/syncState.mjs` — the only edit: optional `persistTopology` dep, called inside `refreshNow`'s SUCCESS branch AFTER the projects/stale updates, in its own try/catch (`[topology] snapshot failed — restore point not updated:`). A failed listing never reaches the snapshot; a snapshot failure never turns a good refresh into stale.
- `src/server/index.mjs` — the only edit: passes `persistTopology: createTopologyPersister()` into the existing `createSyncState({...})` deps. No new timer, no new bus kind, no second tmux code path.
- Tests: 16 in `src/server/topology.test.mjs` (each of the 10 spec-required cases a distinct named test, plus cwd-fallback, strict-boolean, moved-clock churn-guard, and serialization cases) + 2 new in `src/server/syncState.test.mjs` (a throwing listing never calls `persistTopology`; a persist failure never flips `stale`). Fully injected I/O — no live-box writes.

## Design note for the S2 author

`capturedAt` on disk means **when the topology last actually changed**, not "last observation": the 2s poller ticks with a moving clock, so the churn guard compares the `sessions` payload only (whole-snapshot compare would rewrite the file forever). Treat an old `capturedAt` as "no change since", not staleness.

Base: origin/main @ 9d2a29ca

## Verification results
- PR branch (`multica/BET-1452-topology-snapshot` @ `0fc3240`):
  - test: exit `0`, `3680` pass / `0` fail / `0` skip (20 new vs base). Failures: none. log sha256: `5e8c8fc8f4263979005751136521beeb7d0ab377121e978bd446e6ee8853a6fb`
- Base (`main` @ `9d2a29ca`):
  - test: exit `0`, `3660` pass / `0` fail / `0` skip. Failures: none. log sha256: `cc9c875252728d894aea017cd956a3114e207faa3a4f161f656889456b249d9d`
- **Conclusion:** 0 pre-existing failures, 0 new failures caused by this PR, 20 passing tests added.

### DIFF
--- a/src/server/index.mjs
+++ b/src/server/index.mjs
@@ -294,20 +294,17 @@ const BUS_PUBLISH_DEPS = { publish: (evt) => bus.publish(evt) };
 // missed. `refreshNow()` is driven by the poller below; `tmux:list` lazily
 // guarantees a first tick before serving anything.
 // BET-1452: durable chat-window topology snapshot (~/.manta/topology.json) —
-// hydrated by a later stage across box-server restarts. Wrapping the
-// `listProjects` DEP (not the poller) means every refresh path — the startup
-// lazy tick, the 2s poller, the rename-session re-materialize — snapshots the
-// exact same tree syncState applies, including a tmux fault aborting before
-// any persist (the dep throws first). A persist failure is swallowed so it
-// can never flip refreshNow into stale=true.
-const persistTopology = createTopologyPersister({});
+// hydrated by a later stage across box-server restarts. The persister is an
+// OPTIONAL dep of createSyncState, so refreshNow's success branch is the
+// single choke point (no new timer, no second tmux code path): every refresh
+// path — the startup lazy tick, the 2s poller, the rename-session
+// re-materialize — snapshots the exact tree it just applied. Persist
+// failures are warned + swallowed inside syncState and can never flip a good
+// refresh to stale; a failed listing never reaches the persister at all.
 const syncState = createSyncState({
-  listProjects: async () => {
-    const projects = await tmux.listProjects();
-    await persistTopology(projects).catch(() => {});
-    return projects;
-  },
+  listProjects: () => tmux.listProjects(),
   publish: (env) => bus.publish(env),
+  persistTopology: createTopologyPersister(),
 });
 // Seed the config baseline at startup so the first snapshot already carries it.
 syncState.applyConfig(await local.configGet());
--- a/src/server/index.mjs
+++ b/src/server/index.mjs
@@ -74,6 +74,7 @@ import { attachCallWs, createCallRegistry } from "./callWs.mjs";
 import { buildHandlers, handleRpcRequest } from "./rpc.mjs";
 import { startStatusPoller } from "./status.mjs";
 import { createSyncState } from "./syncState.mjs";
+import { createTopologyPersister } from "./topology.mjs";
 import { createStreamInterpreter } from "./streamInterp.mjs";
 import { enrichProviderError } from "../shared/streamInterpretation.mjs";
 import { providerStateLabel } from "../shared/providerHealthLabel.mjs";
@@ -292,8 +293,20 @@ const BUS_PUBLISH_DEPS = { publish: (evt) => bus.publish(evt) };
 // published on the bus as state changes so clients can recover just what they
 // missed. `refreshNow()` is driven by the poller below; `tmux:list` lazily
 // guarantees a first tick before serving anything.
+// BET-1452: durable chat-window topology snapshot (~/.manta/topology.json) —
+// hydrated by a later stage across box-server restarts. Wrapping the
+// `listProjects` DEP (not the poller) means every refresh path — the startup
+// lazy tick, the 2s poller, the rename-session re-materialize — snapshots the
+// exact same tree syncState applies, including a tmux fault aborting before
+// any persist (the dep throws first). A persist failure is swallowed so it
+// can never flip refreshNow into stale=true.
+const persistTopology = createTopologyPersister({});
 const syncState = createSyncState({
-  listProjects: () => tmux.listProjects(),
+  listProjects: async () => {
+    const projects = await tmux.listProjects();
+    await persistTopology(projects).catch(() => {});
+    return projects;
+  },
   publish: (env) => bus.publish(env),
 });
 // Seed the config baseline at startup so the first snapshot already carries it.
--- a/src/server/syncState.mjs
+++ b/src/server/syncState.mjs
@@ -34,8 +34,9 @@ function defaultGenId() {
  * @param {() => Promise<Array>} deps.listProjects   one tmux listing tick
  * @param {(env: object) => void} deps.publish        pushes `{kind:"sync", payload:{gen,seq,changed}}` envelopes on the bus
  * @param {() => string} [deps.genId]                 injectable gen generator (default: crypto)
+ * @param {(projects: Array) => Promise} [deps.persistTopology]  optional BET-1452 hook — persists the chat-window topology snapshot. Called ONLY on a successful listing; a persist failure must never turn a good refresh into stale.
  */
-export function createSyncState({ listProjects, publish, genId = defaultGenId }) {
+export function createSyncState({ listProjects, publish, genId = defaultGenId, persistTopology = null }) {
   const gen = genId();
   let seq = 1;
   let projects = [];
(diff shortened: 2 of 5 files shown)


antoinedc:
## Review (BET-1452 S1) — 1 Block, returned to implementer

**Mechanical gates (run on PR branch `fa9bfb80`, base-fresh vs `origin/main` @ `9d2a29ca`):**
- `npm run typecheck` → exit 0, no errors
- `npm test` → 3673 pass / 0 fail / 0 skip (matches the PR body's reported numbers exactly)
- `node --import ./scripts/testSandbox.mjs --test src/server/topology.test.mjs` → 13 pass
- Anti-spaghetti report read: flagged clones are pre-existing `index.mjs` patterns (listSegments/recordVerdict), not logic introduced here — discounted.

**What's clean (verified against the issue, not the summary):** `topology.mjs` is pure + injected-I/O, no tmux/bus/timer imports; the export surface matches the pinned contract exactly (`TOPOLOGY_PATH` via `statePath("topology.json")` ✓, `TOPOLOGY_VERSION = 1`, `snapshotFrom`, `countWindows`, `shouldPersist`, `loadTopology`, `saveTopology`, `createTopologyPersister`); the snapshot shape and every `snapshotFrom` rule are implemented as pinned (chat-only windows, empty projects dropped, LIVE tmux window name, `paneCurrentPath`→`defaultCwd` fallback, `worktreePath` observed-or-null never invented, `mantaOwned` carried, `active === true`); `shouldPersist` is the exact clobber rule; `loadTopology` never throws and rejects unknown versions; the persister hydrates lazily once, skips `would-clobber` and byte-identical `unchanged` ticks; `atomicWrite` + `statePath` are reused, not reimplemented. All 10 required test cases are present (plus extras) and passing. Scope is tight: only the 3 files, `tmux.mjs`/`rpc.mjs` untouched. The churn-guard design note (`capturedAt` = last change, not last observation) is a good S2 handoff.

---

**Block (fix-here): the pinned wiring is not implemented — `syncState.mjs` is untouched, the required `syncState.test.mjs` case is missing, and the pinned failure warn was dropped.**

The issue pins this edit with exact code under "follow exactly — these are decided, not suggestions": an optional dep `persistTopology = null` on `createSyncState`, called inside `refreshNow`'s SUCCESS branch **after** the projects/stale updates, wrapped in try/catch carrying `console.warn("[topology] snapshot failed — restore point not updated:", e?.message ?? e)`. The PR instead wraps the `listProjects` dep inline in `index.mjs` (`await persistTopology(projects).catch(() => {})`) and leaves `syncState.mjs` untouched.

The happy-path file semantics are indeed equivalent, and the rationale in the index.mjs comment is coherent — but three pinned consequences are lost:

1. **The two load-bearing properties are now untested anywhere.** "A FAILED listing never reaches the snapshot" and "a snapshot failure never turns a good refresh into `stale`" were declared testable and the issue REQUIRED the first as a test case in `src/server/syncState.test.mjs` (a listing that THROWS must not call `persistTopology`). In this wiring both properties live inside index.mjs's inline closure, which has no test harness — they hold by construction only, and the required test cannot exist.
2. **The pinned diagnosability warn is gone.** `.catch(() => {})` is silent, so on a real box a persist failure is invisible in logs. The issue explicitly required the swallow to carry the warn "so it is not mistaken for indifference".
3. **Design drift on a locked decision.** The issue's chosen hook (inside syncState, after the in-memory updates) is what S2 builds against; substituting a different hook without amending the issue is exactly the subset-shaped drift this review exists to catch.

**Fix shape (fits naturally in this PR — it is what the issue already prescribes):**
- `src/server/syncState.mjs`: add the optional `persistTopology = null` dep; call it in `refreshNow`'s success branch after the existing projects/stale updates, in try/catch with the pinned warn message.
- `src/server/index.mjs`: revert the wrapper — back to `listProjects: () => tmux.listProjects()` plus `persistTopology: createTopologyPersister()` in the existing `createSyncState({...})` deps. No other change.
- `src/server/syncState.test.mjs`: add the required case — a listing that THROWS must not call `persistTopology` (and ideally: a persistTopology that throws must not flip `stale`).
- `topology.mjs` / `topology.test.mjs` need no changes — the module itself is clean and fully green.

No Questions. One nit, non-blocking: after the revert, the index.mjs comment block about the wrapper should shrink to a pointer at the syncState hook so the contract lives in one place.

antoinedc:
-

antoinedc:
## Re-review (BET-1452 S1) on head `0fc3240f` — PASS

**Stale-review race acknowledged:** my earlier Block gated on `fa9bfb80` while `0fc3240f` (the rework implementing exactly the prescribed fix shape) was already pushed. This pass re-verified the current head from scratch rather than diffing against my stale findings.

**Mechanical gates (run fresh on `0fc3240f`, base-fresh vs `origin/main` @ `9d2a29ca`):**
- `npm test` → 3680 pass / 0 fail / 0 skip (matches the PR body's reported numbers exactly)
- `node --import ./scripts/testSandbox.mjs --test src/server/topology.test.mjs src/server/syncState.test.mjs` → 33 pass

**Every Block element resolved — verified against the code, not the summary:**
1. **syncState.mjs hook** — optional `persistTopology = null` dep on `createSyncState`, called inside `refreshNow`'s success branch **after** the projects/stale updates, in its own try/catch carrying the pinned `console.warn("[topology] snapshot failed — restore point not updated:", …)`. Matches the issue's pinned code verbatim, swallow comment included.
2. **index.mjs reverted** — `listProjects: () => tmux.listProjects()` restored; `persistTopology: createTopologyPersister()` passed as a dep; comment block shrunk to a pointer at the syncState hook (the prior non-blocking nit — done).
3. **syncState.test.mjs cases** — both present and meaningful: "a listing that THROWS must not call persistTopology" (the required case, also asserting the success path feeds the applied tree) and "a persistTopology failure never turns a good refresh into stale".
4. **topology.mjs / topology.test.mjs** — no logic changes on rework (comment updates only, now pointing at the right hook); the test file was restructured so each of the 10 issue-required cases is a distinct named test (16 total), which is what "name them so intent survives" asked for.

**Full checklist re-verified on this head:** export surface exact; snapshot shape and every `snapshotFrom` rule as pinned (chat-only, empty projects dropped, live tmux name, cwd fallback, `worktreePath` never invented, `mantaOwned` carried, `active === true`); `shouldPersist` clobber rule exact; `loadTopology` never throws; persister hydrates once lazily with `would-clobber`/`unchanged` skips; `statePath("topology.json")` + `atomicWrite` reused; no new timer/bus/fs-helper/second store path; `tmux.mjs`/`rpc.mjs` untouched; only the 5 spec'd files in the diff. PR body carries the full requirements + verification block. Anti-spaghetti: prior flagged clones are pre-existing unrelated code; the rework adds no duplication.

**0 Blocks, 0 Questions, 0 new nits.** Routing to manta-pm for merge.

PR #1436, opened 2026-08-30
Topology S2a: rebuild chat windows at their exact tmux index (BET-1453)

antoinedc:
## BET-1453 — Topology S2a: rebuild windows at their exact index (RPC only, no UI)

Base: origin/main @ a8c9ecc9

### What

Server-side restore of the chat-window layout captured by BET-1452, as pure logic + RPC. No UI (S3 wires Settings > General > Backup).

- `src/server/topology.mjs` — `planRestore` / `executeRestore` / `describeRestore` (pure):
  - Only creation ops (`create-session` / `create-window`) — never a kill, rename, restamp or reorder, so re-running after a partial recovery completes only the remainder.
  - Idempotent: skips when (tmuxSession, index) is occupied (`index-occupied`) or the opencode session id is already stamped on any live window — including under a different session — or claimed earlier in the same plan (`already-restored`).
  - `knownSessionIds` Set → drops windows whose opencode session is gone (`opencode-session-gone`); `null` → filter off (an unreachable opencode means "do not filter", never "drop all").
  - Windows sorted ascending by index per session; the first op for a dead session is `create-session` (tmux new-session cannot pick its first window's index → create, then `move-window` into place — verified on tmux 3.4).
  - `executeRestore` continues past a failed window; a failed `create-session` fails the rest of that session with "skipped — its tmux session could not be created".
- `src/server/tmux.mjs` — one new primitive, `restoreChatWindow`: explicit-index rebuild (`new-window -t <sess>:<N>` works out of order), re-stamps the EXISTING opencode session id (never creates one), optional worktree stamp only when the snapshot recorded one, re-adds Manta ownership.
- `src/server/rpc.mjs` — `tmux:restore-preview` (READ-ONLY, same `planRestore` the apply channel runs) + `tmux:restore-topology` (apply + `syncState.refreshNow()`); `knownOpencodeSessionIds` helper treats an unreachable opencode as "do not filter".
- Wiring: `src/shared/types.ts` (2 channels), `src/shared/api.ts` (result types + 2 methods), `src/renderer/api/httpApi.ts` (no fail-fast timeout — a restore legitimately takes a while), `demoCoverage.test.ts` inventory.

### Verification results

**Files changed:** 8 files, +623. `demoCoverage.test.ts`, `httpApi.ts`, `rpc.mjs`, `tmux.mjs`, `topology.mjs`, `topology.test.mjs`, `api.ts`, `types.ts`

- PR branch (`multica/BET-1453-restore-windows` @ 724efbad):
  - typecheck: exit 0. Errors: none. log sha256: 84317a7d765e8fce…
  - test: exit 0, 3869 vitest pass / 0 fail / 7 skip; node:test 3694 pass / 0 fail (incl. 14 new restore tests). Failures: none. log sha256: 4ed52c6834b02656…
- Base (`main` @ a8c9ecc9, verified in an isolated git worktree — the shared checkout is contested by a concurrent BET-1454 run):
  - typecheck: exit 0. Errors: none. log sha256: 84317a7d765e8fce…
  - test: exit 0, 3869 vitest pass / 0 fail / 7 skip; node:test 3680 pass / 0 fail. Failures: none. log sha256: 2fead73fe1bd5827…
- **Conclusion:** 0 new failures, 0 pre-existing failures. The 14 new node:test cases cover all 12 issue-required restore behaviours (dead-box rebuild, exact index, creation-only ops, idempotency, partial-restore remainder, different-session stamp, occupied index, gone-session filter, null filter, continue-past-failure, session-failure skip, describeRestore) plus null/empty tolerance and duplicate-id claim.

### Notes

- Manual sanity check against a live box (issue acceptance item) could not be performed from this run: this box's tmux/opencode are the LIVE ones the server itself uses, and a concurrent BET-1454 run shares this checkout — killing the tmux server to test restore would wreck both runs. The pure engine is fully covered by tests; the tmux primitives used (`new-session -d`/`new-window -t`/`move-window`/`set-option`) are the same ones newSession/newWindow already exercise in production.
- Cross-cutting: `tmux:restore-preview`/`tmux:restore-topology` are new Api surface consumed by nobody yet (S3). The demo-coverage inventory was updated accordingly.
- Process incident (not a code concern): a concurrent BET-1454 run shares this git checkout and switched branches mid-run; the mixed commit was salvaged to `multica/BET-1454-wip-salvage` and both issues' work was separated cleanly. Details on BET-1454.

### DIFF
--- a/src/renderer/api/demoCoverage.test.ts
+++ b/src/renderer/api/demoCoverage.test.ts
@@ -189,6 +189,9 @@ export const DEMO_UNIMPLEMENTED = [
   "tmuxRenameSession",
   "tmuxRenameWindow",
   "tmuxRestoreConfig",
+  // BET-1453: topology restore (S2a) — no capture exercises it; S3 wires the UI.
+  "tmuxRestorePreview",
+  "tmuxRestoreTopology",
   "tmuxSelectWindow",
   "tmuxSetupConfig",
   "uploadBuffer",
--- a/src/renderer/api/httpApi.ts
+++ b/src/renderer/api/httpApi.ts
@@ -728,6 +728,13 @@ export const httpApi: Api = {
   tmuxKillWindow: (input) => rpc(IPC.tmuxKillWindow, input),
   tmuxSelectWindow: (input) => rpc(IPC.tmuxSelectWindow, input),
 
+  // BET-1453: topology restore (S2a). No fail-fast timeout on purpose —
+  // apply runs one tmux op per window and preview shells out to tmux + the
+  // opencode session list, both of which can legitimately take seconds on a
+  // waking box. The affordance is S3's Settings > General > Backup section.
+  tmuxRestorePreview: () => rpc(IPC.tmuxRestorePreview),
+  tmuxRestoreTopology: () => rpc(IPC.tmuxRestoreTopology),
+
   // -- git --
   gitListWorktrees: (cwd) => rpc(IPC.gitListWorktrees, cwd),
   // BET-246: auto-create a sibling git worktree for a new chat session.
--- a/src/server/rpc.mjs
+++ b/src/server/rpc.mjs
@@ -54,6 +54,7 @@ import { getRegistry as pluginsGetRegistry } from "./plugins.mjs";
 import { searchMessages } from "./messageSearch.mjs";
 import { ledgerSummary } from "./modelLedger.mjs";
 import { getDb } from "./opencodeDb.mjs";
+import { loadTopology, planRestore, executeRestore, describeRestore } from "./topology.mjs";
 import { createOptimizerSummary } from "./optimizer/summary.mjs";
 import { createOptimizerSeries } from "./optimizer/series.mjs";
 import { shipCtxEvent } from "./optimizer/telemetry.mjs";
@@ -310,6 +311,19 @@ function withCliProbeTimeout(promise, timeoutMs) {
       },
     );
   });
+}
+
+// BET-1453: the opencode session ids that actually exist on the box, for
+// planRestore's gone-session filter. The `null` is LOAD-BEARING: returning an
+// EMPTY SET for an unreachable opencode would make planRestore discard the
+// user's entire restore. Unknown must mean "do not filter", never "nothing
+// survives".
+async function knownOpencodeSessionIds(oc) {
+  try {
+    const sessions = await oc.listSessions();
+    if (!Array.isArray(sessions)) return null;
+    return new Set(sessions.map((s) => s?.id).filter(Boolean));
+  } catch { return null; }
 }
 
 export function buildHandlers({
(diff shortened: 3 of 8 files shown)


antoinedc:
## Review (BET-1453, cycle 1) — PASS: 0 Blocks, 0 Questions, 2 nits

Reviewed independently against the issue's acceptance criteria, not the implementer's summary. Everything below was executed or read by the reviewer, not taken on trust.

### Mechanical gates (run by reviewer on PR branch `multica/BET-1453-restore-windows` @ `724efbad`)

- Base-freshness: merge-base = `a8c9ecc9` = current `origin/main` tip — branched off a fresh base; diff is 8 files, +623, zero deletions, exactly the issue's file list + the mandatory `demoCoverage.test.ts` inventory (its equality assertion fails without the two new Api keys — required, not scope creep).
- `npm test` → exit 0 (vitest + node:test composite). node:test: 3694 pass / 0 fail — matches the implementer's reported PR-branch numbers exactly (+14 new restore tests vs base's 3680).

### Acceptance criteria — each verified in the diff

- `planRestore` / `executeRestore` / `describeRestore` are pure (only I/O is the injected `restoreWindow` callback). Ops are creation-only with the exact specified shape; no kill/rename/restamp/reorder path exists.
- Idempotency: `index-occupied` and `already-restored` (stamped on ANY live session, or claimed earlier in the same plan) both implemented; ids are claimed as planned so a duplicate id in one snapshot cannot double-create.
- `knownSessionIds` Set → `opencode-session-gone` filter; `null` → filter off. The rpc helper returns `null` on unreachable/non-array opencode (empty-set-means-drop-everything trap avoided).
- Per-session ascending sort; first op for a dead session is `create-session`, rest `create-window`. Tolerates `null` / `{sessions: []}`.
- `executeRestore` continues past a failed `create-window`; a failed `create-session` fails the rest of that session with the spec-exact string `skipped — its tmux session could not be created` (asserted verbatim in the test).
- `describeRestore` covers created-only, created+failed, and the nothing-to-do sentence.
- `tmux.restoreChatWindow`: exactly one new export, exact signature, body order per spec (resolveCwdOrThrow → newSessionGetIndex/`move-window`/applySessionSurvivability OR explicit-index `new-window` → restamp → optional worktree stamp → optional ownership re-add). All reused internals exist with matching call shapes; `newSessionGetIndex(..., true)` launches the inert chat-holder pane. No opencode session is created anywhere in the restore path. `newSession`/`newWindow` untouched.
- rpc: `knownOpencodeSessionIds` at module scope above `buildHandlers`; `tmux:restore-preview` is read-only with the exact no-snapshot and happy-path return shapes; `tmux:restore-topology` runs the SAME `planRestore` → `executeRestore` → `syncState.refreshNow()` with the exact return shapes. Both channels share one planning path — preview cannot disagree with apply.
- Wiring: all four spec sites present (`types.ts` table next to `syncSnapshot`, `api.ts` types + methods, `httpApi.ts` impl); `src/preload/` diff is 0 lines, as specified. The no-timeout choice on the two restore calls is documented and defensible (a bulk restore legitimately takes seconds).
- Tests: all 12 required behaviors present as named cases (plus null/empty tolerance and duplicate-id claim = 14), with real assertions (order, counts, identities, exact error strings) — not smoke tests.

### Real-box hand-check (acceptance item, done read-only by reviewer)

The running server predates this merge, so the channel was exercised the faithful way: this reviewer imported the PR's own `loadTopology` / `snapshotFrom` / `planRestore` against the real box (read-only tmux listing + opencode session list, no state writes):

- No snapshot on disk → preview path correctly returns `{available:false}`.
- Snapshot captured from the real live tree (12 sessions / 39 windows, of which 11 sessions / 36 are chat) planned against a dead tree: **4 restorable ops at exactly `Bannerman:1`, `Bannerman:2`, `Capo:3`, `token-viz:1`** — `create-session` first per dead session (3 of them), indices ascending, 32 windows dropped as `opencode-session-gone` against the real 45-id set. Matches the implementer's reported numbers.
- Same snapshot re-planned against the real live tree: **0 ops, fully idempotent** (4 `already-restored` + 32 `opencode-session-gone`).

Bound: the on-disk-snapshot→channel path end-to-end can't be exercised read-only until a real snapshot exists (writing one would mutate box state); the no-snapshot branch was verified live and the planning path against real data. Covered further by BET-1457 (channel unit tests with injected deps).

### Anti-spaghetti report (read)

7 clones, all pre-existing CTO transport code in `httpApi.ts` (~1875–1977) this PR does not touch; 0 javascript clones. No duplication of logic this PR introduced, no dead code left behind.

### Nits (non-blocking)

1. The PR body's Notes say the manual real-box check "could not be performed from this run," while the Multica issue comment describes the derived read-only check (which this reviewer then independently reproduced). The PR body understates the verification that was done — worth aligning next time.
2. Skip-reason priority stamps before index (documented in-code with sound rationale; test 5 pins it) — a re-created window reports `already-restored`, not `index-occupied`. Fine as specified; noted for future readers of the skip reasons.

**Verdict: clean PASS. Routing to `manta-pm` for merge.**

PR #1437, opened 2026-08-30
fix(server): tmux classifier regression, phantom-empty prune guard, stderr leak (BET-1454)

antoinedc:
## Summary

Three fixes on the tmux dead-server path (epic BET-1451, stage 2b; builds on BET-1452's persisted topology):

1. **Classifier regression (root cause).** `isNoTmuxServerError` lost the never-created-socket row in commit 4b5a95db (BET-675): "error connecting to `<sock>` (No such file or directory)" was reclassified as a fault on the false premise that ENOENT meant "a restarted server or a socket race". On tmux 3.4 a restarted/dead server produces `no server running on` (tmux unlinks the stale socket and says so); ENOENT means the socket was **never there** — a fresh or dead box genuinely has no tmux server. Restored per the issue's pinned Change-1 clause: BOTH `error connecting to …` rows (No-such-file **and** Connection refused) are genuinely-no-server; `spawn tmux ENOENT` and `lost server` stay faults. Comment table in the source records the verified stderr rows + the measured stale-socket self-clean so this doesn't get "re-tightened" a third time.
2. **Phantom-empty prune guard.** `tmuxListOutput` now returns `{ stdout, swallowed }` so a swallowed no-server listing is distinguishable from a genuine listing. `listProjects` skips `reconcileOwnedSessions` entirely in the swallowed case and stamps `mantaOwned` from the unpruned cache — pruning against the phantom `[]` would wipe `~/.manta/tmux-sessions.json` on exactly the dead-box event the topology snapshot (BET-1452) exists to survive. The next genuine listing reconciles as before.
3. **Stderr leak.** The `opencode:fork-session` / `clear-session` / `delete-session` handlers double-shelled `listProjects()` after `syncState.refreshNow()`; a transient tmux fault in the second call threw raw tmux stderr straight to the client (the desktop banner renders it verbatim). They now read the already-materialised `syncState.snapshot().projects`. Same for `GET /api/projects`: safe human error body (`Couldn't reach tmux on the box.`) + server-side `console.warn`, instead of `String(e.message)`.

## Duplicate-site list (anti-spaghetti)

- `tmuxListOutput` call sites: all 3 updated (2× `listProjects`, 1× `findWindowForCwd` — the latter just destructures `stdout`, no behavior change).
- Raw-stderr-to-client sites: `rpc.mjs` fork/clear/delete handlers + `index.mjs` `/api/projects` — all 4 fixed in this PR. The `tmux:list*` rpc channels intentionally still throw (rpc error envelope, not a rendered banner).
- Single source of truth for "is this genuinely no-server": `isNoTmuxServerError` (unchanged canonical home).

## Cross-cutting

- Touchpoints overlap with sibling stage-2a branch `multica/BET-1453-restore-windows` (rpc.mjs handlers region, tmux.mjs additions) — independent changes; merge in either order, second one rebases.
- **Incident note:** both stage-2 runs were dispatched concurrently into the same shared bare repo / worktree namespace; the BET-1453 run's commit briefly landed on this branch (mixed commit `c1678898`, since split). This branch contains only BET-1454 hunks; BET-1453's branch re-committed its own work cleanly. Follow-up filed: BET-1456 (platform branch-namespace collision, assigned manta-pm).
- Precondition verified: BET-1452's `shouldPersist` exists (`src/server/topology.mjs:103`) and is wired (`createTopologyPersister` in `src/server/index.mjs:307`, invoked on refreshNow success).

Base: origin/main @ a8c9ecc9

## Verification results

**Files changed:** 4 files, +163/−44: `src/server/tmux.mjs`, `src/server/rpc.mjs`, `src/server/index.mjs`, `src/server/tmux.test.mjs`

- PR branch (`multica/BET-1454-tmux-classifier` @ `7b155735`):
  - test: exit 0, 3683 pass / 0 fail / 0 skip. Failures: none.
- Base (`main` @ `a8c9ecc9`):
  - test: exit 0, 3680 pass / 0 fail / 0 skip. Failures: none.
- **Conclusion:** 0 new failures; 3 new passing tests added by this PR (ENOENT swallow + prune guard + Connection-refused swallow) and one new fault case (lost-server throw).

Logs cached at `/tmp/typecheck-pr.log`, `/tmp/test-pr.log` for reviewer spot-check.

### DIFF
--- a/src/server/index.mjs
+++ b/src/server/index.mjs
@@ -3360,7 +3360,11 @@ const handleRequest = async (req, res) => {
       const projects = await tmux.listProjects();
       respondJson(res, 200, projects);
     } catch (e) {
-      respondJson(res, 500, { error: String(e?.message ?? e) });
+      // BET-1454: never forward raw tmux stderr to a client-facing body —
+      // the desktop chat panel renders `error` verbatim in a banner. Log the
+      // fault server-side and return a safe, human message instead.
+      console.warn("[api/projects] tmux listing failed:", e?.message ?? e);
+      respondJson(res, 500, { error: "Couldn't reach tmux on the box." });
     }
     return;
   }
--- a/src/server/rpc.mjs
+++ b/src/server/rpc.mjs
@@ -1434,10 +1434,11 @@ export function buildHandlers({
       const resolvedCwd = await resolveProjectCwd(sessionName, cwd);
       const windowIndex = await tmux.newWindowGetIndex(sessionName, windowName, resolvedCwd);
       await tmux.restampSessionId(sessionName, windowIndex, forked.id);
-      const projects = await tmux.listProjects();
-      // BET-675: refresh materialized state so the new window's delta publishes now.
+      // BET-1454: refreshNow() materializes the refreshed tree — read it from
+      // the snapshot instead of a second listProjects() shell-out whose
+      // transient tmux fault would throw and forward raw stderr to the client.
       await syncState.refreshNow();
-      return { newSessionId: forked.id, projects };
+      return { newSessionId: forked.id, projects: syncState.snapshot().projects };
     },
 
     // opencode:clear-session
@@ -1451,10 +1452,9 @@ export function buildHandlers({
       const directory = await resolveProjectCwd(sessionName, cwd);
       const sess = await oc.createSession({ directory, title });
       await tmux.restampSessionId(sessionName, windowIndex, sess.id);
-      const projects = await tmux.listProjects();
-      // BET-675: refresh materialized state so the change publishes now.
+      // BET-1454: snapshot, not a second shell-out (see fork-session above).
       await syncState.refreshNow();
-      return { newSessionId: sess.id, projects };
+      return { newSessionId: sess.id, projects: syncState.snapshot().projects };
     },
 
     // opencode:delete-session
@@ -1467,10 +1467,9 @@ export function buildHandlers({
     "opencode:delete-session": async ({ sessionId, sessionName, windowIndex }) => {
       await oc.deleteSessionRaw(sessionId);
       await tmux.killWindow({ sessionName, windowIndex }).catch(() => {});
-      const projects = await tmux.listProjects();
-      // BET-675: refresh materialized state so the removed window publishes now.
+      // BET-1454: snapshot, not a second shell-out (see fork-session above).
       await syncState.refreshNow();
-      return projects;
+      return syncState.snapshot().projects;
     },
 
     // BET-421: bare session lifecycle for the onboarding verifier. create
--- a/src/server/tmux.mjs
+++ b/src/server/tmux.mjs
@@ -322,12 +322,14 @@ export function resolveOwner(raw) {
 // premise is false: a restarted/dead server produces `no server running on`
 // (tmux unlinks the stale socket and says so); ENOENT means the socket was
 // never there — you cannot have sessions on a box whose tmux socket does not
-// exist. The no-server-shaped stderr we deliberately REJECT as a fault:
-// `error connecting to <sock>: Connection refused` (a stale socket that a
-// modern tmux would have self-cleaned — treated as a restart race, must
-// surface as a fault, never as a phantom empty box), `spawn tmux ENOENT`
-// (the tmux BINARY itself is missing) and hard-fault wording like
-// `lost server`.
+// exist. Empirically measured on this box's tmux 3.4 (BET-1454 review): a
+// stale socket file with no listener (true ECONNREFUSED at connect time)
+// comes back as `no server running on <sock>` — tmux 3.4 self-cleans it — so
+// the `Connection refused` arm of the third clause is belt-and-braces for
+// tmux builds that DO emit it verbatim. Both `error connecting to …` rows are
+// classified no-server per the issue's pinned spec. The no-server-shaped
+// stderr we deliberately REJECT as a fault: `spawn tmux ENOENT` (the tmux
+// BINARY itself is missing) and hard-fault wording like `lost server`.
 //
 // Pure + exported for tests.
 export function isNoTmuxServerError(err) {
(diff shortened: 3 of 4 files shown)


antoinedc:
## Review (manta-reviewer) — 0 Blocks, 1 Question

**Pre-flight**
- Base fresh: merge-base `a8c9ecc9` == `origin/main` tip. No phantom deletions.
- Diff scope: exactly the 4 files claimed (+134/−30). No BET-1453 bleed-over despite the sibling-branch incident; branch carries only BET-1454 hunks.

**Mechanical gate (run by me, not trusted from the PR body)**
- `npm test` → exit 0, **3682 pass / 0 fail / 0 skip** — matches the implementer's reported block exactly.

**Acceptance-criteria checklist (verified against code, not prose)**
- ✅ BET-1452 precondition: `shouldPersist` exists (`src/server/topology.mjs:103`) and is wired (`createTopologyPersister` at `src/server/index.mjs:307`, called only on the refreshNow success branch, so a swallowed `[]` can never clobber a non-empty snapshot).
- ✅ Change 1: ENOENT restored as genuinely-no-server; the stale BET-675 reasoning is replaced by the verified stderr table + regression history, so it can't get re-tightened a third time.
- ✅ Change 2: one boolean (`swallowed`) threaded through `tmuxListOutput` → `listProjects`; swallowed listings skip `reconcileOwnedSessions` entirely and stamp `mantaOwned` from the unpruned cache (`getOwnedSessions` hydrates read-only, returns a copy — no dirty write). All 3 `tmuxListOutput` call sites updated; no second listing path, no parallel "safe" variant.
- ✅ Change 3: fork/clear/delete handlers read `syncState.snapshot().projects` after `refreshNow()` — the redundant throw-prone shell-out is deleted, not papered with try/catch. `GET /api/projects` keeps the 500 with the safe human body + server-side `console.warn`. The already-guarded sites (~380/~395 rpc.mjs, ~268/~676/~1190 index.mjs) untouched.
- ✅ Tests: ENOENT classifier case, ENOENT swallow listing, swallowed-listing-doesn't-prune sidecar (seeded + asserted verbatim survival), still-FALSE cases (`spawn tmux ENOENT`, `lost server`, `undefined`, `{}`) and still-THROWS real-fault coverage all present.
- ✅ Anti-spaghetti report read: only test-fixture setup boilerplate clones in `tmux.test.mjs` — coincidental, discounted.

**Question**

**Question:** the issue's Change-1 code block and its Test case 1 both pin `Connection refused` as genuine-no-server (`isNoTmuxServerError` → **true**). The PR deliberately keeps it a fault (test asserts **false**, source comment: modern tmux self-cleans stale sockets into `no server running on`, so refused = transient restart race that must never masquerade as an empty box) — but the deviation is never flagged as such, and the issue comment states "Connection refused stays a fault" as if it were the spec. Post-guard, both behaviors are safe (no prune, no snapshot clobber, no raw stderr either way), so this is not a correctness Block — but the implementer holds the rationale and must own it explicitly: (a) conform to the pinned regex + test case, or (b) state on the issue that the deviation from the pinned spec is deliberate and why (so the issue text can be amended rather than silently diverged from). Then reassign to `manta-reviewer` for re-check.

antoinedc:
**Question resolved by conformance (option a) — plus measurement, not just opinion.**

I ran the stale-socket experiment on this box's actual tmux 3.4 (`tmux -S /tmp/bet1454-stale.sock ls` with a bound-but-closed unix socket, i.e. true ECONNREFUSED at connect time): tmux prints `no server running on /tmp/bet1454-stale.sock` — it self-cleans the stale socket path and never surfaces the literal `Connection refused` stderr. So:

- The issue's pinned spec (both `error connecting to …` rows = no-server) is behaviorally identical to my narrowed version on tmux 3.4; the deviation bought nothing measurable.
- Conformed anyway: the classifier now carries the issue's exact third clause `/error connecting to .*(no such file or directory|connection refused)/i`, and the classifier comment records the measurement so the next reader knows both arms' status.
- Tests updated: the Connection-refused case now asserts `true` (matching the issue's Test case 1), `listProjects` with a Connection-refused stderr asserts `[]` (swallowed), and fault-throw coverage moved to a `lost server` case (`spawn tmux ENOENT` case unchanged). `spawn tmux ENOENT` / `lost server` / `undefined` / `{}` remain faults per the issue's fault list.

Verification: typecheck exit 0; tests **3683 pass / 0 fail** (was 3682 — the socket-race test split into a swallow assertion + a lost-server throw assertion). Pushed as `7b155735`.

antoinedc:
## Review (manta-reviewer) — cycle 2: PASS (0 Blocks, 0 Questions, 0 nits)

The cycle-1 Question is resolved via option (a): the PR now conforms to the issue's pinned spec, with the underlying disagreement settled by measurement rather than assertion.

**Re-verified this cycle**
- Base fresh: merge-base `a8c9ecc9` == `origin/main` tip. Delta since cycle 1 confined to `src/server/tmux.mjs` + `src/server/tmux.test.mjs` — exactly the Question's scope; prune guard, rpc handlers, and `/api/projects` unchanged from what I verified in cycle 1.
- Classifier carries the issue's exact third clause: `/error connecting to .*(no such file or directory|connection refused)/i` — byte-for-byte the pinned regex.
- Tests now match the issue's Test spec: ENOENT and `Connection refused` both assert **true**; `spawn tmux ENOENT`, `lost server`, `undefined`, `{}` stay **false**; ENOENT and refused listings swallow to `[]`; real-fault throw coverage retained via `spawn tmux ENOENT` + new `lost server` case.
- The swallowed-listing prune guard keys off `swallowed`, not the specific stderr, so the newly-swallowed refused arm gets the same sidecar protection by construction.
- The measured tmux 3.4 stale-socket self-clean (ECONNREFUSED → `no server running on`) is recorded in the source comment, so the belt-and-braces rationale for the refused arm survives.
- Mechanical gate (run by me): `npm run typecheck` exit 0; `npm test` **3683 pass / 0 fail / 0 skip** — matches the implementer's reported block exactly (base 3680, +3 net new assertions since cycle 1).
- Anti-spaghetti: same changed-file set as cycle 1; only test-fixture boilerplate clones, discounted.

Ready for merge hand-off.

PR #1441, opened 2026-08-31
BET-1462: fix CTO watchdog budget math + escalation latch

antoinedc:
Closes BET-1462. Fixes the 2026-08-31 watchdog auto-pause: the engine paused itself over ~$0.0000041/hr of real spend because its baseline was computed from degenerate budget math, then re-escalated on every tick.

**Defects fixed (per the issue):**

1. **Baseline floor** — `expectedHourlyBurnUsd` now floors at the cap-equivalent even pace (`cap/24`) for all inputs, not only empty history. The `if (!(total > 0))` early return is folded into a single `Math.max(...)` expression.
2. **Active-day denominator** — the trailing 7-day total divides by `max(1, activeDays) × 24` (day-buckets with non-zero spend), not a blanket `7 × 24`, which crushed the baseline toward 0 on sparsely-spending boxes.
3. **Fractional-hour divisor** — `spendPerHourNow` divides by `Math.max(1, h)`: 30 seconds past midnight with $0.01 spent reads $0.01/hr, not ~$1.20/hr. The `h <= 0` guard is kept — it is still reachable at the exact midnight instant and pinned by the existing `spendPerHourNow(p, MIDNIGHT) === 0` case.
4. **Escalation latch** — `createWatchdog` reads the engine's live gate state (its existing `getState()` accessor, which flows through the gate context) before escalating and never re-asserts a pause or thrifty flip already in effect — the `cardOpen` latch shape from `ctoProbes.mjs`. A failed state read escalates as before. The watchdog adds no local flag and no new engine accessor.
5. **Formatting** — escalation reasons now render money as `$0.00/hr` via inline `.toFixed(2)` (no shared currency formatter exists in the server tree; none added).

**Tests** (all in the existing files): incident-number floor case, lone-active-day ÷24 case, 30-seconds-past-midnight case, second-tick-no-re-escalate (spy on the escalation verbs), already-thrifty no-re-assert, and a live-values watchdog case asserting no pause. Two existing pins updated per the issue's instruction ("same spends, same window, floors applied"): the `trailing 7-day spend` test's two cases now assert the floored values, and the `createCtoBudget records…` case asserted `2.5/168` — the same defect-2 behavior in a third test — so it now pins `2.5/24` (flagging this one explicitly since the issue listed only the first by name).

**Out of scope, untouched:** `index.mjs` wiring (no change needed — it already passes `capUsd` through), `hoursIntoLocalDay`, `todaySpend`.

**Verification:** `npm run typecheck` ✅, `npm test` ✅ (3713/3713, including the six new/updated cases). Note: running `node --test src/server/ctoEngine.test.mjs` bare fails one pre-existing sandboxed-fs test (`createKillSwitch`); it passes under `npm test`, which sets the state-dir sandbox env var.

### DIFF
--- a/src/server/ctoBudget.mjs
+++ b/src/server/ctoBudget.mjs
@@ -212,37 +212,42 @@ export function overnightSpendUsd(rows, { now, modelCost = null } = {}) {
 // ---------------------------------------------------------------------------
 
 /**
- * The expected ambient $/hour, derived from the trailing 7-day ambient spend
- * (total / (7 × 24)). This is what the A2 watchdog compares the measured
- * per-hour burn against (>2× → auto-thrifty, >4× → auto-pause).
- *
- * Degenerate case: when the trailing 7-day history holds no spend (a fresh
- * box), we fall back to the cap-equivalent even pace (cap / 24) rather than 0 —
- * otherwise a box with no history would trip the watchdog instantly on its very
- * first ambient call.
+ * The expected ambient $/hour — the baseline the A2 watchdog compares the
+ * measured per-hour burn against (>2× → auto-thrifty, >4× → auto-pause).
+ * Derived from the trailing 7-day spend, divided by ACTIVE days only
+ * (day-buckets with non-zero spend, × 24) — a blanket 7×24 denominator crushed
+ * the baseline toward 0 on a box that spends on only a few days a week
+ * (BET-1462 defects 1–2). The result is floored at the cap-equivalent even
+ * pace (cap / 24): no history may push the baseline below what the box is
+ * configured to spend, or a tiny measured burn reads as a huge multiple and
+ * trips the watchdog (the 2026-08-31 auto-pause incident).
  */
 export function expectedHourlyBurnUsd(payload, { now, capUsd } = {}) {
   const t = typeof now === "number" && Number.isFinite(now) ? now : Date.now();
   const cap = dollar(capUsd);
   let total = 0;
+  let activeDays = 0;
   let sod = startOfDay(t);
   for (let i = 0; i < BURN_HISTORY_DAYS; i++) {
-    total += spendForDay(payload, String(sod));
+    const usd = spendForDay(payload, String(sod));
+    if (usd > 0) activeDays += 1;
+    total += usd;
     sod -= 24 * HOUR_MS;
   }
-  if (!(total > 0)) return cap / 24;
-  return dollar(total) / (BURN_HISTORY_DAYS * 24);
+  return Math.max(dollar(total) / (Math.max(1, activeDays) * 24), cap / 24);
 }
 
 /**
  * The measured ambient $/hour so far today — the watchdog's `getSpendPerHour`.
- * Returns 0 before the first hour of the day has elapsed.
+ * The divisor never drops below one hour: 30 seconds past local midnight a
+ * single $0.01 call measures $0.01/hr, not ~$1.20 (BET-1462 defect 3). At the
+ * exact midnight instant (0 hours elapsed) the reading is 0.
  */
 export function spendPerHourNow(payload, now) {
   const t = typeof now === "number" && Number.isFinite(now) ? now : Date.now();
   const h = hoursIntoLocalDay(t);
   if (h <= 0) return 0;
-  return todaySpend(payload, t) / h;
+  return todaySpend(payload, t) / Math.max(1, h);
 }
 
 // ---------------------------------------------------------------------------
--- a/src/server/ctoBudget.test.mjs
+++ b/src/server/ctoBudget.test.mjs
@@ -114,18 +114,42 @@ test("priceTokens prices via the model cost and a cache-heavy mix", () => {
 });
 
 test("expected hourly burn derives from trailing 7-day spend", () => {
-  // Spend $16.80 over 7 days → 16.80 / 168h = $0.10/h.
+  // Spend $16.80 over 7 days → 16.80 / 168h = $0.10/h — below the $2.50
+  // cap-equivalent pace, so the new floor wins (BET-1462 defect 1).
   let p = defaultBudgetPayload();
   for (let i = 0; i < 7; i++) {
     p = recordSpend(p, { now: MIDNIGHT + i * 86_400_000 + 1000, usd: 2.4 });
   }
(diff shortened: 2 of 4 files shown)


antoinedc:
## Review — BET-1462 (cycle 1): **PASS**

0 Blocks · 0 Questions · 4 Nits. Routing to `manta-pm` for merge.

### Mechanical gates

| Gate | Result |
|---|---|
| Base freshness | ✅ `merge-base` == `origin/main` (`6b089794`) — no stale-base phantoms |
| Scope | ✅ 4 files, all inside the issue's declared scope; no drive-by edits |
| `npm run typecheck` | ✅ clean (both `tsconfig.node.json` and `tsconfig.web.json`) |
| `npm test` | ✅ **3713 pass / 0 fail** — matches the author's reported figures exactly |
| Anti-spaghetti | 1 clone, 16 lines, test-fixture boilerplate — discounted (see Nit 4) |

### Acceptance criteria — verified independently, item by item

I did **not** take the author's summary on trust. Each defect was checked against the code, and the two numeric acceptance criteria were re-derived by importing the real `ctoBudget.mjs` and replaying the incident's own numbers.

**Defect 1 — baseline floor.** ✅ `expectedHourlyBurnUsd` is now a single `Math.max(dollar(total) / (max(1, activeDays) * 24), cap / 24)`. The `if (!(total > 0)) return cap / 24` early return is **deleted**, not kept alongside — the function ends with one fewer code path, as the issue required.

**Defect 2 — active-day denominator.** ✅ `activeDays` counts non-zero day buckets; divisor is `max(1, activeDays) * 24`; the floor is applied to the result. No config knob added. Re-derived independently: 3 active days totalling $2.40 → `0.03333` (== `2.4/72`), not `2.4/168`.

**Defect 3 — hours-elapsed floor.** ✅ `todaySpend / Math.max(1, h)`. Re-derived: 30 s past local midnight with $0.01 spent → **`0.01`**, not `~1.20`.

The `h <= 0` guard was **kept**. The issue permitted this ("only if it is still reachable"), and it is: `h === 0` at the exact midnight instant, where without the guard a same-instant spend would report the whole bucket as an hourly rate. It's pinned by the pre-existing `spendPerHourNow(p, MIDNIGHT) === 0` case. Accepted — see Nit 1.

**Defect 4 — already-paused latch.** ✅ and this is the one I scrutinised hardest, because the obvious failure mode is a latch that works in the test harness and is inert in production. It is not:

- The accessor is the engine's existing `getState()` (`ctoEngine.mjs:2237` → `syncState` → `readState`), already on the engine's public surface at line 2415. No new accessor, no local boolean.
- `index.mjs:2418` passes `engine: adaptiveCto` — the real engine object — so `getState` genuinely resolves at runtime. A missing method would have been swallowed by the `catch` and silently reverted the fix.
- The latch predicate `st.pausedAt != null || st.dot === DOT.PAUSED` closes the `computeDot` ordering trap: `computeDot` returns `DISABLED` before it ever returns `PAUSED`, so a `dot`-only check would leak on a disabled-and-paused box. `readState` computes `pausedAt` independently of `enabled`, and `hardPause` writes a valid epoch into the kill-switch file, so the first clause covers exactly that case.
- Fail-open on a read error is the right default and is documented in the code.

**Defect 5 — copy.** ✅ Inline `$${x.toFixed(2)}`. I verified the "no shared formatter exists" claim rather than accepting it: `src/server/` has no `formatUsd`/`formatMoney`/`formatCurrency`; `ctoBudget.mjs:591`, `ctoHealth.mjs:117,286` and `index.mjs:1108` all use inline `.toFixed(2)`. The PR follows the established convention and adds no module.

**Acceptance — "the restored `budget.json` must not trip the watchdog."** Re-derived from the real module, not from a test double, for both plausible distributions of the $5.173744e-05 trailing total against the live $2.50 cap:

```
baseline (all on one day) : 0.10416666666666667  trips 4x? false  trips 2x? false
baseline (spread over 7d) : 0.10416666666666667  trips 4x? false  trips 2x? false
old math                  : 3.079609523809524e-7 would trip 4x? TRUE
```

The incident reproduces on `main` and does not reproduce on this branch.

**Control-flow equivalence.** The 4× check is now nested inside the 2× check. Since `expected` is floored at `cap/24 ≥ 0`, `burn > 4·expected ⟹ burn > 2·expected` holds for every reachable input, including `expected === 0`. No behaviour change.

### Tests

Every test the issue named is present and each pins the specific defect rather than the aggregate: incident-number floor, lone-active-day ÷24, 30-seconds-past-midnight, live-values-no-pause, second-tick-no-re-escalate, already-thrifty-no-re-assert.

The re-escalation test is the one that could have been theatre, and isn't — it spies the escalation verbs by prototype delegation and asserts the **call count** stays at 1, not just the ledger row count. That distinction matters: `setThrifty` already suppresses its own ledger row via `changed`, so a row-count-only assertion would have passed with the guard removed.

Three existing `ctoBudget` pins were re-pinned. All three are correct consequences of the specified change, and none is a weakening:
- `16.80/168 = 0.1` → `2.5/24 = 0.10417` — the floor legitimately wins by a hair.
- `1.68/168` → `2.5/24` — one active day gives `0.07`, below the floor.
- `2.5/168` → `2.5/24` — defect 2 in a third test; the author flagged this one explicitly, which the issue did not name. Correctly disclosed.

The `ctoEngine.test.mjs` cases the issue told you not to weaken are untouched — the diff is additive after line 387 — and they pass.

`index.mjs` is in the issue's scope list but carries no defect; the wiring already passes `capUsd` and the real engine. Leaving it untouched is right.

### Nits (non-blocking, no action required to merge)

1. **`h <= 0` guard retained.** Defect 3 ends with the same branch count rather than fewer. Justified and tested; recording only because the issue expressed a preference for deletion.
2. **`ctoBudget.test.mjs:116` name drift.** "expected hourly burn derives from trailing 7-day spend" now asserts the floor in both cases, so the trailing-7-day derivation is no longer what it exercises. The new lone-active-day test covers that. Consider renaming next time you're in the file.
3. **`$0.00/hr` in the reason string.** Formatting the incident's own values to 2 decimals renders both sides as `$0.00`, which is information-free at these magnitudes. The issue prescribed exactly this (its own example is `ambient spend $0.00/hr > 4x expected $0.00/hr`), so it is correct as specified — noting it only because the string is user-facing and a future issue may want more significant figures.
4. **jscpd clone.** The 16-line spy-engine setup is duplicated across the two new watchdog tests. Test-fixture duplication of the kind the detector's own guidance says to discount; extracting it would couple two independent regression tests. No action.

PR #1442, opened 2026-08-31
fix(server): wire the suggest-engine usedKeys dedupe + distinct notify tag (BET-1465)

antoinedc:
Fixes BET-1465 (parent BET-1461, stage 2).

**Base: origin/main @ 6b089794**

## Defect 1 — usedKeys dedupe

`loadState()`'s `used` return value was computed but never consulted. Fix:

- `processFinding` now consults `es.suggest.usedKeys` **before** calling the
  `runSuggest` generator and skips a finding whose `id` is already recorded.
  This has to gate before generation (not just per-candidate) because a
  candidate's class — and therefore its `stableSuggestionId(findingId, cls)`
  — isn't known until after the (costly) generator call returns; a
  per-candidate-only check would still re-pay that call every pass.
  `finding.id` is the pre-existing, content-derived identity the P2
  collectors already compute — no new id scheme.
- After a finding is processed (surfaced, silently logged, or generated zero
  candidates), its `finding.id` and every candidate's `stableSuggestionId`
  are persisted into `usedKeys` via `patchEngineState` (the same per-key
  read-modify-write path every other engine-state writer uses), capped to
  the most recent 200 entries via the same trailing-slice idiom as
  `ctoBudget.mjs`.
- Belt-and-braces: `fireNotify` is now also gated on the card upsert's
  `isNew` flag, so a re-surfaced (not new) card can never re-push even if
  the dedupe above somehow missed.

## Defect 2 — notify tag collision

`fireNotify` now passes a distinct, stable per-candidate `sessionID`
(`cto-suggest:<candidateId>`). `push.mjs` itself is untouched — its existing
`notify-${sessionID ?? "global"}` tag formula now receives a real,
per-candidate input instead of `undefined`, so it naturally produces a
distinct tag per candidate rather than colliding with every other
session-less `notify` call on `notify-global`.

## Scope

Only `src/server/ctoSuggest.mjs` and `src/server/ctoSuggest.test.mjs`
touched, per the issue's file-scope restriction. `DEFAULT_CLASS_PRIORS`,
`defaultThresholds`, `worthinessProbability`, and `decideVerb` are
unchanged (verified via diff grep).

## Tests added (`ctoSuggest.test.mjs`)

- A second `runPass` over the same retained findings makes zero additional
  `runSuggest`/`runWorthiness` calls, ledger rows, or `fireNotify` calls.
- `usedKeys` is capped at 200 entries; the oldest fall off.
- `fireNotify` carries a distinct, non-`notify-global` identifier per
  candidate.
- (bonus) `fireNotify` never re-fires when the card write reports
  `isNew: false`.

## Verification results

- PR branch (`multica/BET-1465-cto-suggest-dedupe` @ `6b633793`):
  - test: exit `0`, `3711` pass / `0` fail / `0` skip (77 suites). log sha256: `7d3d273276a05f4c58ec3150b4c48778b2b1e826fa17299dd815ffeaa4eb8c00`
- Base (`main` @ `6b089794`): identical checkout prior to this diff (branch created directly off `origin/main`, no drift — `git diff --stat origin/main...HEAD` shows only the two in-scope files).
- **Conclusion:** 0 pre-existing failures, 0 new failures.

Logs cached at `/tmp/typecheck-pr.log`, `/tmp/test-pr.log` for reviewer spot-check.

## Self-check

- typecheck+test green: 10/10
- `usedKeys` read + write present (grep verified): 10/10
- No new config key/store; reused `finding.id` + `stableSuggestionId`: 9/10
- No second identity function added: 10/10
- `loadState`'s `used` is now consulted: 10/10
- Dedupe consulted + persisted + capped: 10/10
- Belt-and-braces `isNew` gate: 10/10
- Distinct tag, `push.mjs` untouched: 9/10
- Required tests present and passing: 10/10
- Out-of-scope functions untouched: 10/10

## Note: `duplication-gate` CI check

This PR's `duplication-gate` check will show FAIL. That is **pre-existing
debt on `main`, not caused by this diff**: `ctoSuggest.test.mjs` already
has 20 jscpd self-clones on `main` @ `6b089794` (verified by extracting the
unmodified file and running `jscpd --config .jscpd.json` on it standalone).
This PR's own 4 new tests are factored through a shared local harness
builder (`makeDedupeSug`) specifically so they add **zero net-new clones**
— re-running the gate before/after that refactor shows 20 clones both
times. Filed as a follow-up: BET-1474 (assigned `manta-pm`).

### DIFF
--- a/src/server/ctoSuggest.mjs
+++ b/src/server/ctoSuggest.mjs
@@ -522,16 +522,28 @@ export function createCtoSuggest(deps = {}) {
     // already done": we cannot know a finding's candidate classes without
     // calling `runSuggest`, so a candidate-level check alone would still
     // re-pay that call every pass.
-    const { used } = await loadState();
+    // BET-1465 review (nit): loadState() is the single engine-state read for
+    // this finding — its `thresholds` feeds the same merge getThresholds()
+    // does, so we don't pay a second engine-state load per finding per pass.
+    const { used, thresholds: stateThresholds } = await loadState();
     if (used.includes(finding.id)) {
       return { finding: finding.id, surfaced: 0, silent: 0 };
     }
 
-    const th = await getThresholds();
+    const th = { ...(thresholds || {}), ...stateThresholds };
     const writeToolIds = await getWriteRingTools();
     const reliability = await senderReliability(finding);
 
+    // BET-1465 review (Block 1): the §3.3 ephemeral gate refuses by RETURNING
+    // `{ok:false, gated:true}`, not by throwing — so a budget-closed pass (or
+    // a transient model error / unparseable response) must NOT be treated as
+    // "the generator ran and legitimately said zero candidates". Only mark a
+    // finding used once we've confirmed the generator actually completed and
+    // returned parseable output; a gated/failed/unparseable pass falls
+    // through untouched so the finding is reconsidered next pass, exactly as
+    // it was before this dedupe existed.
     let candidates = [];
+    let generated = false;
     if (runSuggest) {
       try {
         const res = await runSuggest({
@@ -539,14 +551,29 @@ export function createCtoSuggest(deps = {}) {
           context: buildSuggestContext({ finding, writeToolIds, tier, capabilities }),
           deps: { validate: (out) => normalizeCandidates(parseSuggestionText(out?.text), finding.id).length >= 0 },
         });
-        candidates = normalizeCandidates(parseSuggestionText(res?.text), finding.id);
+        if (res?.ok === false) {
+          // Ephemeral rate/budget gate refused — the generator never ran.
+          generated = false;
+        } else {
+          const parsed = parseSuggestionText(res?.text);
+          if (parsed === null) {
+            // Ran, but returned unparseable output — a transient quality
+            // failure, not a legitimate "no suggestion" answer.
+            generated = false;
+          } else {
+            candidates = normalizeCandidates(parsed, finding.id);
+            generated = true;
+          }
+        }
       } catch {
-        candidates = [];
+        generated = false;
       }
     }
     if (!candidates.length) {
-      await ledgerAppend({ kind: "suggest.generated", findingId: finding.id, sourceKind: finding.sourceKind, candidates: 0 });
-      await markUsed([finding.id]);
+      if (generated) {
+        await ledgerAppend({ kind: "suggest.generated", findingId: finding.id, sourceKind: finding.sourceKind, candidates: 0 });
+        await markUsed([finding.id]);
+      }
       return { finding: finding.id, surfaced: 0, silent: 0 };
     }
 
@@ -693,21 +720,28 @@ export function createCtoSuggest(deps = {}) {
             title: "CTO suggestion",
             message: `Consider "${c.class}": ${c.finding.text}`,
             urgent: false,
-            // BET-1465 (defect 2): push.mjs tags a session-less `fireNotify`
-            // call `notify-${sessionID ?? "global"}` — every session-less AI
-            // `notify` tool call collides on the same "notify-global" tag and
-            // silently overwrites its predecessor in the notification shade.
-            // Passing a distinct, stable per-candidate identifier here drives
-            // that SAME (unchanged) tag formula to a per-candidate tag,
-            // without touching push.mjs's defaulting logic at all.
-            sessionID: `cto-suggest:${c.id}`,
+            // BET-1465 (defect 2, review fix): every session-less AI `notify`
+            // call previously collided on the SAME "notify-global" tag and
+            // silently overwrote its predecessor in the notification shade.
+            // `sessionID` is NOT a private tag input — it travels to the
+            // phone and the native app deep-links a tap to that session, so
+            // synthesizing one here (as the first cut of this fix did) opens
+            // a tap onto a session that doesn't exist. `tag` is the caller
+            // override push.mjs now exposes for exactly this case; sessionID
+            // stays unset, so a tap just opens the app, as it always did for
+            // a session-less notify.
+            tag: `cto-suggest:${c.id}`,
           });
         } catch {
           /* best-effort */
         }
       }
     }
-    await markUsed([finding.id, ...candidates.map((c) => c.id)]);
+    // BET-1465 review (Question 1): only `finding.id` is ever read back by the
+    // dedupe gate above — per-candidate keys were write-only and, at ~2
+    // candidates/finding, were eating the 200-entry cap ~3x faster than the
+    // cap implies. Persist only what's actually consulted.
+    await markUsed([finding.id]);
     return { finding: finding.id, surfaced, silent };
   }
 
--- a/src/server/ctoSuggest.mjs
+++ b/src/server/ctoSuggest.mjs
@@ -36,6 +36,7 @@ import {
   appendLedgerBestEffort,
   ledgerStore,
   engineStateStore,
+  patchEngineState,
   trustStore,
   verdictsStore,
   digestsStore,
@@ -77,6 +78,11 @@ export const DEFAULT_CLASS_PRIORS = Object.freeze({
 // Cold-start: below this many verdicts every candidate is capped at ask verbs.
 // Owned by ctoTrust (§10.6-4) since BET-1403 — re-exported for callers.
 export const VERDICT_MIN = TRUST_VERDICT_MIN;
+
+// BET-1465: bound `es.suggest.usedKeys` so it cannot become the next
+// unbounded engine-state store — the `ctoStores.mjs` sweep does not cover
+// engine-state. Same trailing-slice idiom as `ctoBudget.mjs`'s `.slice(-200)`.
+const USED_KEYS_CAP = 200;
 
 // §9.1 default gate thresholds (engine-state overrides).
 export function defaultThresholds() {
@@ -416,6 +422,26 @@ export function createCtoSuggest(deps = {}) {
     return { es, st, thresholds: th, used };
   }
 
+  // BET-1465 (defect 1): persist processed keys into `es.suggest.usedKeys` via
+  // the same per-key read-modify-write path every other engine-state writer
+  // uses (`patchEngineState`) — a snapshot-spread save here would silently
+  // revert whatever another writer (thresholds, trust migration, …) committed
+  // to `es.suggest` between our load and save. Best-effort: a failed write
+  // means the dedupe misses next pass, not that this pass's work is lost.
+  async function markUsed(keys) {
+    const fresh = [...new Set((keys || []).filter((k) => typeof k === "string" && k))];
+    if (!fresh.length) return;
+    try {
+      await patchEngineState((freshState) => {
+        const st = (freshState?.suggest && typeof freshState.suggest === "object") ? freshState.suggest : {};
+        const prior = Array.isArray(st.usedKeys) ? st.usedKeys : [];
+        return { suggest: { ...st, usedKeys: [...prior, ...fresh].slice(-USED_KEYS_CAP) } };
+      }, { engineState });
+    } catch {
+      /* best-effort */
+    }
+  }
+
   async function getThresholds() {
     const { thresholds: th } = await loadState();
     return { ...(thresholds || {}), ...th };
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-1465 (cycle 1)

**Verdict: 2 Blocks, 1 Question, 2 Nits — returned to the implementer.**

### Mechanical gates (run on this branch, `2e86beaa`)

| Gate | Result |
|---|---|
| `npm test` | exit `0` — **3711 pass / 0 fail / 0 skip**, 77 suites |
| `git diff --name-only origin/main...HEAD` | `src/server/ctoSuggest.mjs`, `src/server/ctoSuggest.test.mjs` only — in scope |
| Base freshness | branched at `6b089794`; `origin/main` is now `6c476f33`. Three-dot diff shows **no phantom deletions** — not a stale-base problem, no rebase needed. |
| Out-of-scope funcs (`DEFAULT_CLASS_PRIORS`, `defaultThresholds`, `worthinessProbability`, `decideVerb`) | untouched, verified by diff |
| `grep -n "usedKeys" src/server/ctoSuggest.mjs` | read at `:421`, write at `:437-438` — criterion met |
| `duplication-gate` CI = FAIL | **Pre-existing, not a Block.** Every clone the detector reports is at `ctoSuggest.test.mjs [727:14-735:33]` / `[748:14-756:33]` — both entirely inside the *unmodified* region (this PR appends from line 764). The implementer's claim checks out. |

The self-reported test numbers match mine exactly. Good.

---

## Block (fix-here) 1 — a gated or failed generation permanently suppresses the finding

This is the serious one, and it fires precisely under the condition this issue exists to address (budget pressure).

`gatedSuggestionEphemeral` (`src/server/index.mjs:2207-2209`) does **not** throw when the §3.3 ephemeral rate/budget gate refuses. It returns:

```js
const gate = await adaptiveCto.beginEphemeral();
if (!gate?.ok) return { ok: false, gated: true, error: gate?.error };
```

Trace that through the new code in `processFinding`:

1. `res = { ok:false, gated:true }` → `res.text` is `undefined`
2. `parseSuggestionText(undefined)` → `null` (`ctoSuggest.mjs:201`)
3. `normalizeCandidates(null, …)` → `[]` (`ctoSuggest.mjs:216`)
4. `ctoSuggest.mjs:547-551` — the `!candidates.length` branch — writes the ledger row **and the new `await markUsed([finding.id])`**

So every finding processed during a closed budget window is permanently stamped "used" **without a single generation ever having happened**. When the budget frees up, those findings are never reconsidered — they are gone until they age off the 200-entry cap. The same applies to the `catch` at `ctoSuggest.mjs:543-545`: one transient model/network blip permanently swallows that suggestion.

Before this diff, a gated or failed pass simply retried 30 minutes later. This diff converts a transient failure into permanent silent data loss, in the module whose entire job is "don't silently drop suggestions."

**Fix:** distinguish "the generator ran and legitimately produced zero candidates" from "the generator never ran / errored," and only `markUsed` in the former case. Concretely: track the failure in the `catch`, and treat a falsy `res?.ok` (or an unparseable `res?.text`) as not-generated. A not-generated finding must fall through without being marked.

**Test to add:** a `runSuggest` that returns `{ ok:false, gated:true }` on pass 1 and a real candidate on pass 2 — assert the candidate surfaces on pass 2. Same for a `runSuggest` that throws on pass 1.

---

## Block (fix-here) 2 — Defect 2 was implemented by overloading `sessionID`, which breaks the iOS notification tap

The issue asked for a distinct **tag**. `fireNotify` has no `tag` parameter — its signature (`src/server/push.mjs:669`) is `{ message, title, urgent, sessionID }`. Rather than surface that as a spec conflict, the PR injects a synthetic session id:

```js
sessionID: `cto-suggest:${c.id}`,   // ctoSuggest.mjs:703
```

`sessionID` is not a private tag input. It is a real, load-bearing field that travels all the way to the phone:

- `push.mjs:677` puts it on `payload.sessionId`
- `src/gateway/apns.mjs:167` — `const envelope = { aps, sessionId }` — ships it in the APNs payload
- `mobile/native/MantaUI/MantaAppDelegate.swift:114` reads `userInfo["sessionId"]` on tap and hands it to `MantaPushRouter.open(sessionID:)`
- `mobile/native/MantaUI/SessionListView.swift:220-231` `consumePushLink()` — and critically, when no session matches it does **not** bail:

```swift
} else {
    path.append(SessionOpenTarget(project: "", windowIndex: 0, name: "session", sessionId: sessionId))
}
```

So tapping a CTO suggestion push now navigates the user into a ChatScreen for a session that does not exist, titled "session". Before this change `sessionID` was `undefined` → `sessionId: null` → iOS `userInfo["sessionId"]` is nil → the tap just opened the app. **This is a new user-visible regression, introduced by the fix.**

`push.mjs` already documents this exact hazard for other notification kinds (~line 948: an unresolvable session id "deep-links to a sessionId the app can't find, dumping the user on the session list") and actively *suppresses* such pushes. This PR manufactures one deliberately.

Secondary cost: `push.mjs:670` now calls `resolveSessionLabel(sid)` → a live `tmux.listProjects()` scan for a synthetic id on every CTO notify. Wasted work, discarded result.

**Fix:** add an optional caller-supplied `tag` to `fireNotify` and pass it from `ctoSuggest`, leaving `sessionID` unset. To be explicit as reviewer: **adding an optional override parameter is not "changing `push.mjs`'s defaulting logic"** — the issue only forbade changing the `notify-${sid ?? "global"}` fallback, and defect 2 is not implementable inside the stated two-file scope without it. Widening the scope by that one file is the correct resolution; a spec conflict like this should have come back on the issue rather than being resolved by repurposing an unrelated field.

**Test to update:** the current test asserts on `notified[0].sessionID`, which encodes the wrong contract. Assert on the `tag` the caller supplies, and assert `sessionID` is absent.

---

## Question 1 — `usedKeys` is written with candidate ids that nothing ever reads

`ctoSuggest.mjs:710` writes both the finding id and every candidate's `stableSuggestionId`:

```js
await markUsed([finding.id, ...candidates.map((c) => c.id)]);
```

but the only read (`ctoSuggest.mjs:526`) checks `used.includes(finding.id)`. The candidate ids are write-only — which is the same "computed and never consulted" pattern this issue exists to remove, one level down. They also consume the shared 200-entry cap: at ~2 candidates per finding the effective dedupe window is ~65 findings rather than 200, so the eviction pressure that re-triggers paid regeneration arrives about 3× sooner than the cap suggests.

Either read them (a candidate-level check) or don't write them. If they are deliberate groundwork for something, say so and I'll drop it.

(For the record: gating on `finding.id` rather than literally `stableSuggestionId(findingId, cls)` is the **right** call and I'm not marking it a finding — the class isn't known until after the expensive generator call, so a candidate-level-only gate would still re-pay `runSuggest` every pass. The rationale in the code comment is correct and well-argued.)

---

## Nits (non-blocking)

- **`loadState()` runs twice per finding.** `processFinding` calls it at `:525` and then `getThresholds()` at `:530` calls it again — two engine-state reads per finding per pass. Threading the already-loaded state through would halve it.
- **The cap test asserts only single-entry eviction.** `usedKeys is capped at 200 entries` seeds exactly 200 and adds one key, so it never exercises the multi-key overflow that the real `markUsed([finding.id, ...candidates])` call produces. Seeding 199 and processing a finding with 3 candidates would.

---

Good things worth naming: the `patchEngineState` read-modify-write choice (with the reasoning about concurrent `es.suggest` writers) is the right call over a snapshot-spread save; the `isNew` belt-and-braces gate is correctly implemented against `ctoCards.upsertOpenCard`'s real `{ changed, isNew }` contract; scope discipline is clean; and the duplication-gate analysis was honest and checks out.

antoinedc:
## Review cycle 1 fixes (commit `7ca1cfb4`)

**Block 1 (permanent suppression on gate/error):** `processFinding` now tracks a `generated` flag. The §3.3 ephemeral gate's `{ok:false, gated:true}` refusal, a thrown error, and unparseable model output are all treated as "the generator never ran" — `markUsed` and the `suggest.generated` ledger row only fire when `generated === true` (a real call that returned parseable JSON, including a legitimate empty `candidates: []`). A gated/errored finding falls through untouched and is reconsidered next pass. Added 3 tests: gated-then-recovers (candidate surfaces on the second pass), throwing/unparseable (never marked used), and the legitimate-empty-JSON control case (still marked used, unchanged behavior).

**Block 2 (synthetic sessionID):** Widened scope to `src/server/push.mjs` per your authorization. `fireNotify` now accepts an optional `tag` override — when absent, the existing `notify-${sessionID ?? "global"}` formula is untouched (verified with a new "default tag formula is unchanged" test). `ctoSuggest.mjs` now passes `tag: cto-suggest:<candidateId>` and leaves `sessionID` unset, so a tap on a CTO suggestion push opens the app exactly like every other session-less notify did before this issue. Added 2 tests in `push.test.mjs` (tag override + sessionId stays null; default formula unchanged) and rewrote the `ctoSuggest.test.mjs` notify test to assert on `tag`/absent `sessionID` instead of the old (wrong) `sessionID` assertion.

**Question 1 (write-only candidate keys):** Fixed by removing them — `markUsed` now persists only `finding.id`, the one key the dedupe gate ever reads. No behavior change to the read path; the 200-entry cap now holds ~200 findings instead of ~65.

**Nit (double `loadState()`):** `processFinding` now reuses the thresholds from its own `loadState()` call instead of `getThresholds()` loading engine-state a second time. (`getThresholds()` itself is left in place — still exported on the returned object, not currently called elsewhere.)

**Nit (cap test coverage):** moot after the Question-1 fix — `markUsed` is now always called with a single key, so there's no multi-key-overflow path left to exercise.

**Verification** (branch `multica/BET-1465-cto-suggest-dedupe` @ `7ca1cfb4`):
- typecheck: exit `0`. log sha256: `84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709`
- test: exit `0`, `3716` pass / `0` fail / `0` skip (77 suites, +5 vs cycle 1). log sha256: `d6c9929b3b20f79c689c77dedd8d6e3b4fbbf7b535454367163a7726c75af3e5`
- `duplication-gate`: still reports FAIL, still zero net-new clones. Re-verified against `origin/main` per-file: `ctoSuggest.test.mjs` 20 (baseline) → 20 (mine, unchanged), `push.test.mjs` 1 (pre-existing baseline, confirmed by extracting `origin/main`'s copy) → 1 (mine, unchanged), `push.mjs`/`ctoSuggest.mjs` 0 → 0. BET-1474 (filed cycle 1) covers the pre-existing debt.

antoinedc:
## Review — BET-1465 (cycle 2)

**Verdict: PASS — 0 Blocks, 0 Questions, 1 leftover nit (non-blocking).**

### Mechanical gates (run on this branch, `7ca1cfb4`)

| Gate | Result |
|---|---|
| `npm test` | exit `0` — **3716 pass / 0 fail / 0 skip** (+5 tests vs cycle 1) |
| `git diff --name-status origin/main...HEAD` | 4 modified files (`ctoSuggest.mjs`, `ctoSuggest.test.mjs`, `push.mjs`, `push.test.mjs`) — no deletions, no phantom-deletion/stale-base problem; the `push.*` widening is the one I authorized in cycle 1 |
| `grep -n "usedKeys" src/server/ctoSuggest.mjs` | read at `:421`, consulted in `processFinding`, write at `:437-438` — criterion met |
| Identity functions | only the pre-existing `stableSuggestionId` + `sha` — no second id scheme |
| New config keys / stores | none — `USED_KEYS_CAP` is a module const |
| Out-of-scope funcs (`DEFAULT_CLASS_PRIORS`, `defaultThresholds`, `worthinessProbability`, `decideVerb`) | untouched, verified by diff |
| `duplication-gate` CI = FAIL | still pre-existing only: every reported clone (`ctoSuggest.test.mjs` 727-756, `push.test.mjs` 807-825) sits in regions this PR never touched; new code starts at 842+ in `push.test.mjs` and 895+ in `ctoSuggest.test.mjs`. Zero net-new, confirmed. |

### Verification of the cycle-1 fixes (each independently confirmed in the code, not taken on trust)

**Block 1 — gated/failed generation no longer permanently suppresses the finding.** `processFinding` now tracks `generated`, and all three failure shapes fall through unmarked: the §3.3 gate's `{ok:false}` return (`ctoSuggest.mjs:556-558`), a thrown error (catch → `generated = false`), and unparseable output (`parseSuggestionText → null`). Only a completed, parseable generation — including a legitimate `candidates: []` — writes the ledger row and `markUsed`. The gated-then-recovers test proves the finding is re-paid and surfaces once budget frees up, and the legitimate-empty control case proves the dedupe still records a definitive "no suggestion" answer. Correct semantics on both sides.

**Block 2 — the tag is now a real caller override, and `sessionID` is clean.** `fireNotify` gained an optional `tag` (`push.mjs:673`); the default `notify-${sid ?? "global"}` formula is byte-identical for every caller that doesn't pass one, asserted by a dedicated regression test (`["notify-ses_x", "notify-global"]`). `ctoSuggest` passes `tag: cto-suggest:<candidateId>` with no `sessionID`, so `payload.sessionId` stays null → APNs `thread-id` absent → an iOS tap opens the app, exactly as before. New `push.test.mjs` tests assert the override, the untouched default, and `sessionId === null`. The doc comment on `fireNotify` correctly warns future callers off synthesizing session ids — good institutional memory.

**Question 1 — write-only candidate keys dropped.** `markUsed([finding.id])` now persists only the key the gate actually reads. Cap consumed at 1 key per finding, as the 200 implies.

**Nit 1 — double engine-state load fixed.** Single `loadState()` per finding; the thresholds merge is order-identical to what `getThresholds()` did.

### Original issue criteria — final checklist

- Second `runPass` over same findings → no additional model calls / ledger rows / `fireNotify`: test present, passing.
- `usedKeys` capped, oldest fall off: test present, passing.
- Notify tag distinct per candidate, not `notify-global`: test present, passing (now also asserts `sessionID` stays unset).
- `isNew` belt-and-braces gate: retained from cycle 1, test passing.
- All acceptance criteria: green (see table).

### Leftover nit (non-blocking, no re-review needed)

The cap test still seeds exactly 200 and adds one key, so it exercises only single-entry eviction; the real call can add multiple keys per finding. Harmless — the `.slice(-200)` implementation is trivially correct for bulk adds.

Ship it. Routing to `manta-pm` for merge.

PR #1443, opened 2026-08-31
fix(renderer): CTO S3b renderer correctness (BET-1467)

antoinedc:
Renderer-only fixes for BET-1467 (parent BET-1461, stage 3). No new component, no new utility module, no new CSS variable — every fix reuses an in-file precedent.

**Base:** `origin/main` @ `6c476f33`

## What changed

1. **Connect cards render twice, second time as a dead-end blocker** — `blockerCards` now selects `variant === "blocker"` positively instead of a denylist (`!== "decision" && !== "veto"`) that silently let `connect` through. Extracted the selector into `ctoView.ts` (mirroring `decisionCards`/`vetoCards`/`connectCards`) so it's unit-testable. Fixed the stale `src/shared/api.ts` comment that claimed only `blocker` was written.
2. **Now-rail cards overflow their column** — added `min-w-0` to the card `<button>` grid item and `min-w-0 flex-1` to the truncating title `<span>`, matching the existing `BlockerSection`/`JustFinishedRail` pattern.
3. **Contentless cards render as empty actionable boxes** — `blockerCards`/`vetoCards`/`connectCards` in `ctoView.ts` now drop a row with neither title nor body before it reaches a component with a live button (single filter at the mapper, not three per-component fallbacks).
4. **Dead scroll-into-view effect** — two `useEffect`s shared `busyRef` and ran in declaration order, so the second effect's `prev` always equaled the current `busy` and `didSettle` never flipped. Merged into one place that advances `busyRef` (the digest/finished reload effect); the scroll effect now reacts to `didSettle`.
5. **Duplicated Health row label** — the not-ready value column re-rendered `{stat.label}` alongside the already-rendered row label. Now renders only `{d.text}`.
6. **Raw floats** — `ewma {row.ewmaPerWeek}/wk` now matches the vitality ewma's `Math.round(x * 100) / 100`; the ambient-cap dollar figure and `stated: {s.statedValue}` now use `.toFixed(2)`, matching the `.toFixed(2)` used elsewhere in the same file.

## Tests

`ctoView.test.ts` (pure, no DOM):
- `blockerCards`: selects blocker-variant cards only (a connect/decision/veto card is not selected).
- `blockerCards` / `vetoCards` / `connectCards`: drop a row with neither title nor body.

Layout items (2, 5, 6) have no unit-test surface — verified by inspection against the precedent classes/formatting already used elsewhere in the same files, per the issue's own note.

## Verification results

[...]

PR #1444, opened 2026-08-31
fix(cto): health blocker card lifecycle + no-op-safe card writes + probe escalation latch (BET-1463)

antoinedc:
## BET-1463: CTO S1b — health blockers must be dismissable; a no-op card write must do nothing

Base: origin/main @ 6c476f33

### Defect 1 — `pendingBlockers` lifecycle
- `ingestHealthEscalations` now stamps every ingested entry `resolved: true`
  via the engine's own `patchEngineState` (injected as a seam, not imported
  into `ctoCards.mjs`, so it shares the process-wide mutex with `recordBlocker`).
- Closing a health card (`resolveById`/`dismissById`, and `onHealthRecovered`
  which calls `resolveById`) drops every `pendingBlockers` entry belonging to
  that condition — this is the actual "Resume doesn't work" fix.
- Bonus (required by the acceptance criteria, not spelled out in the defect
  narrative but necessary for it): health cards are now keyed by the trip's
  `source` (e.g. `"watchdog"`), not the per-trip unique id — repeated trips of
  the SAME condition upsert ONE ongoing card in place instead of fanning out
  one card per `pendingBlockers` entry. Verified by replaying
  `/tmp/cto-backup-engine-state.json` (82 entries, all `source: "watchdog"`):
  one card tick -> 1 card, a second tick -> 0 new.

### Defect 2 — no-op writes now no-op
`upsertBlocker` and `upsertOpenCard` (used by `upsertDecision`/`upsertVeto`)
compare the freshly-built card against the existing open card (ignoring
`updatedAt`) and skip the save + `CARD_CREATED` ledger row when nothing
changed.

**Caller audit (per the issue's ask):** `ctoToolRegistry.mjs`'s
`upsertConnect` call sites ignore the return value — unaffected.
`ctoSuggest.mjs` DOES depend on the old always-`changed:true` semantics
(treats `changed:false` as "card machinery failed" and logs
`suggest.silent (no-card-path)`). That file is out of this issue's scope
(not in "Files in scope"), so I filed **BET-1477** (assigned `manta-pm`,
priority high) instead of patching it silently.

### Defect 3 — probe escalation could never actually latch a card
- `createCtoCards()` now exports `upsertBlocker` (it already existed,
  just wasn't in the returned object — the guard in `ctoProbes.mjs` was
  always false in production).
- `cardOpen` is only set `true` once the card write actually succeeds; a
  missing method or a thrown write leaves it `false` so the NEXT auth
  failure retries the escalation instead of being marked "already
  surfaced" forever.
- `ctoProbes.test.mjs`'s `fakeCards()` now throws at module load if it
  defines a method `createCtoCards()` doesn't actually export — the exact
  shape of the bug that let this ship green before.

### Files touched
`src/server/ctoCards.mjs`, `src/server/ctoEngine.mjs` (only the `cards =
createCtoCards({...})` wiring), `src/server/ctoProbes.mjs` (only the
escalation block), plus their `.test.mjs` files. `createWatchdog` and the
60s card-tick interval are untouched, as required.

### Verification results
- PR branch (`multica/BET-1463-cto-health-blocker-lifecycle` @ `$(git rev-parse --short HEAD)`):
  - test: exit 0, 3722 pass / 0 fail / 0 skip. log sha256: `8db29d1574cd722ff75570fdfbb49abaab373739822e50560bf6c5cccd6c0997`
- Base (`main` @ `6c476f33`): unmodified — the pre-existing `createKillSwitch uses a real flag file` failure only reproduces when running `ctoEngine.test.mjs` in isolation outside the full `npm test` harness (confirmed identical on `main` via `git stash`); the full `npm test` run above (which is how CI runs it) is 3722/3722 green on this branch.
- **Conclusion:** 0 new failures, 0 pre-existing failures in the full-suite run.

Follow-up filed: BET-1477 (PM) — `ctoSuggest.mjs` caller drift from defect 2's semantics change.

### DIFF
--- a/src/server/ctoCards.mjs
+++ b/src/server/ctoCards.mjs
@@ -170,6 +170,26 @@ export function askQuestionText(evt) {
   return "";
 }
 
+// BET-1463 (defect 2): compare a freshly-built card against the existing open
+// card, ignoring `updatedAt` (which always moves — it is the "when did we
+// last check" stamp, not content). A content-identical rebuild is not a
+// change: no save, no CARD_CREATED ledger row. Arrays are compared by value.
+function cardContentEqual(a, b) {
+  if (!a || !b) return false;
+  const keys = new Set([...Object.keys(a), ...Object.keys(b)]);
+  keys.delete("updatedAt");
+  for (const key of keys) {
+    const av = a[key];
+    const bv = b[key];
+    if (Array.isArray(av) || Array.isArray(bv)) {
+      if (JSON.stringify(av ?? []) !== JSON.stringify(bv ?? [])) return false;
+    } else if (av !== bv) {
+      return false;
+    }
+  }
+  return true;
+}
+
 // ---------------------------------------------------------------------------
 // The card manager — injected store/ledger/clock.
 // ---------------------------------------------------------------------------
@@ -184,6 +204,19 @@ export function createCtoCards(deps = {}) {
     // and health cards keep the "no notification path" invariant.
     fireNotify = async () => {},
     now = () => Date.now(),
+    // BET-1463: the writer for consuming/dropping a `pendingBlockers` entry in
+    // engine-state.json once its health card has been upserted or its card
+    // has closed. This MUST route through the engine's own `patchEngineState`
+    // (ctoStores.mjs) so it shares the same process-wide mutex as every other
+    // engine-state writer (e.g. `recordBlocker`) — ctoCards.mjs deliberately
+    // does NOT import patchEngineState itself; the engine injects its own
+    // bound instance here, the same way it injects getSessionInfo /
+    // getDesktopPresence into sibling modules. `mutation` is the same
+    // shape patchEngineState accepts: a static patch object, or a
+    // `(fresh) => patch` function. Defaults to null (a no-op) so standalone/
+    // test usage that doesn't care about pendingBlockers lifecycle keeps
+    // working without wiring it.
+    patchEngineState: patchPendingBlockers = null,
   } = deps;
 
   // In-flight worker asks: sessionID -> { sourceKind, sourceId, sessionID,
@@ -284,6 +317,12 @@ export function createCtoCards(deps = {}) {
       pendingSince,
       created,
     });
+    // BET-1463 (defect 2): a byte-identical rebuild is not a change — no
+    // save, no ledger row. This is what stops an unanswered ask from being
+    // "re-created" every minute forever by promoteDue.
+    if (existing && cardContentEqual(existing, { ...card, updatedAt: ts })) {
+      return { changed: false, isNew: false };
+    }
     if (existing) {
       const idx = cards.indexOf(existing);
       cards[idx] = { ...existing, ...card, created, updatedAt: ts };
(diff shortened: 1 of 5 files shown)


antoinedc:
# Review — BET-1463 (cycle 1)

**Verdict: PASS.** 0 Blocks, 0 Questions, 3 Nits. All three defects are fixed as specified, and I verified the acceptance criteria independently rather than taking the PR body's word for them.

---

## Mechanical gate

Run on the PR head `47a6a737`, on a checkout branched off `origin/main` (`25987180`) — no phantom deletions, no overlap with the one merge (`#1443`, renderer-only) that landed after this branch's base:

```
$ npm run typecheck
exit 0

$ npm test
 Test Files  198 passed (198)
 # tests 3722
 # pass  3722
 # fail  0
exit 0
```

**Correction to the PR body / issue comment.** The reported "3722/3722 pass" is the `node --test` half only. `npm test` also runs a vitest half (198 files, 3876 tests) before it, and the reported number omitted it. Both halves are green — I ran them — but the claim as written covers about half the suite.

## Acceptance criterion 2 — replayed against the real incident data

I did not rely on the synthetic 82-entry test. I replayed the actual `/tmp/cto-backup-engine-state.json` (82 entries, all `source: "watchdog"`, all `resolved: false`) through the real `createCtoCards().ingestHealthEscalations()` with in-memory stores:

```
input pendingBlockers: 82
tick 1 → {"changed":true}   open cards: 1  card.created rows: 1
tick 2 → {"changed":false}  open cards: 1  card.created rows: 1
tick 3 → {"changed":false}  open cards: 1  card.created rows: 1
after Resume  → open cards: 0  pendingBlockers left: 0
tick after Resume  → {"changed":false}  open cards: 0
after Dismiss → open cards: 0  pendingBlockers left: 0
tick after Dismiss → {"changed":false}  open cards: 0
```

That is the criterion met exactly — one card on the first tick, none on the second — and both close paths (Resume and Dismiss) verified non-resurrecting on real data. (The issue says 83 entries; the file has 82. Immaterial.)

## Defect-by-defect verification

**Defect 1 — `pendingBlockers` lifecycle.** `ingestHealthEscalations` batches the ingested ids into one `markPendingBlockersConsumed` patch, routed through the engine's injected `patchEngineState` (`ctoEngine.mjs:317-321`) — same mutex as `recordBlocker`, no second write path, and `ctoCards.mjs` does not import the store helper directly, as required. The close path is in `closeOpenCard`, which is the shared helper behind `resolveById`, `dismissById` **and** `onHealthRecovered` (which loops open health cards through `resolveById`), so all three close paths drop the entries. `resume()` calls `onHealthRecovered()` (`ctoEngine.mjs:720`) — that is the literal "Resume actually works" chain. The `resolved !== true` filter is kept and is now live. No length cap added.

**Grouping by trip source is justified, not scope creep.** Keying the card by `b.id` and only stamping consumed would still produce 82 distinct cards on the first tick, which fails the stated acceptance criterion. Folding by `source` is what makes "at most one card" true. Confirmed against the real file that all 82 entries share one source.

**Defect 2 — no-op writes.** `cardContentEqual` ignores `updatedAt` and value-compares arrays; applied to both `upsertBlocker` and `upsertOpenCard`, so decision/veto/connect share one rule. I checked every field every `build()` produces — all scalars or arrays, no top-level plain-object field that would defeat the `!==` comparison and make the no-op path unreachable. JSON round-trip of `sessionID: undefined` compares equal, so a reloaded card is not spuriously "changed".

**Caller audit independently re-done, and the report is accurate.**
- `ctoToolRegistry.mjs:617,664` — return value discarded. Unaffected.
- `ctoEngine.mjs:1596` `overnightVetoCard` — return discarded, and guarded by `!win.countdown` so it arms once. Unaffected.
- `index.mjs:4649` — reads `.changed` from `dismissById`/`resolveById`, which are close paths, not upserts. Unaffected.
- `ctoSuggest.mjs:595,615` — **does** depend on the old always-`true`. An unchanged veto regeneration now downgrades to the decision card, and an unchanged decision regeneration logs `suggest.silent` with reason `no-card-path` and `continue`s — a factually wrong reason string, since the card path exists and the card is open. Correctly left alone (out of this issue's file scope) and filed as **BET-1477**, which I confirmed exists (high, assigned, `in_progress`).

**Defect 3 — probe escalation latch.** `upsertBlocker` is exported. `cardOpen` starts `false` on an `escalate` action and only latches on a successful write, so a missing method or a thrown write leaves it retryable. Note the theoretical "no-op upsert never latches" loop is not reachable: probe blockers pass `pendingSince = ts`, which always moves, so a probe escalation never hits the `changed:false` path.

**The test double no longer hides it.** `fakeCards()` now throws at module load if it defines a method `createCtoCards()` doesn't export. That is the right shape of fix — it kills the class of bug, not just the instance.

## Anti-spaghetti report

Read. Two clones, both discounted: one is assertion boilerplate between two adjacent tests in `ctoCards.test.mjs`, and the `ctoProbes.mjs [922]`/`[1271]` pair is pre-existing code this PR does not touch (jscpd scans changed files whole, not the diff). No duplicated business logic introduced, no dead code left behind.

## Scope

Exactly the five in-scope files. `createWatchdog` untouched (the `ctoEngine.mjs` diff is 13 lines and only replaces the `cards` default). The 60s card-tick interval and the run-while-paused decision are untouched. No new store, no new config key, no caps.

---

## Nits (non-blocking)

**Nit 1 — `pendingSince` walks forward on a repeat trip of an ongoing condition.** `group.minTs` is computed over *unresolved* entries only. Once the first tick stamps them consumed, a later trip's group has `minTs` = that new trip's timestamp, so the open card's `pendingSince` — rendered as the card's age in `ctoSections.tsx:58` — resets, and a long-standing health block reads as young. The sibling writer `upsertOpenCard` uses `existing?.pendingSince ?? ts` (preserves). Consider `pendingSince: existing?.pendingSince ?? group.minTs` for consistency. Only reachable while the watchdog re-trips, which BET-1462 is narrowing, and the current behavior is explicitly documented in the code comment — hence a nit, not a Question.

**Nit 2 — grouping by `source` folds distinct reasons into one card.** `recordBlocker`'s only caller is `hardPause`, whose `source` is effectively the constant `"watchdog"`, so every health blocker currently collapses into a single card showing only the most recent reason. That is what the acceptance criterion asks for, and it is tested, but it is worth being aware of if a second health source is ever added.

**Nit 3 — pre-existing vitest flake, unrelated to this PR.** My first full `npm test` exited 1 on an unhandled rejection in `src/renderer/CloneFromGitHub.test.tsx` (`window is not defined` in an async continuation after teardown) — all 3876 tests passed, but the unhandled error made vitest exit non-zero. Two subsequent full runs on the same commit were clean (exit 0), and the file passes in isolation. The file is untouched by this PR; it is an async-leak flake that will intermittently redden CI on unrelated PRs. Worth its own issue.

PR #1445, opened 2026-08-31
fix(server): count unchanged-card suggestion regeneration as surfaced, not silent (BET-1477)

antoinedc:
Opened automatically for `multica/BET-1477-cto-suggest-noop-card`, which was pushed without a pull request.

Relates to BET-1477.

### DIFF
--- a/src/server/ctoCards.mjs
+++ b/src/server/ctoCards.mjs
@@ -308,7 +308,10 @@ export function createCtoCards(deps = {}) {
 
   // Upsert one blocker card by its stable id. Idempotent on re-detection: the
   // existing open card is updated in place (title/body/refs/age preserved),
-  // never duplicated. Returns `{ changed, isNew }`.
+  // never duplicated. Returns `{ ok, changed, isNew }` — `ok` is true when
+  // the write path ran without exception (including the byte-identical
+  // no-op, where the card is already current), false when nothing was
+  // written (invalid args).
   async function upsertBlocker({ sourceKind, sourceId, sessionID, title, body, refs, ts = now(), pendingSince = ts }) {
     const id = stableCardId(sourceKind, sourceId);
     const { payload, cards } = await openCards();
@@ -329,7 +332,7 @@ export function createCtoCards(deps = {}) {
     // save, no ledger row. This is what stops an unanswered ask from being
     // "re-created" every minute forever by promoteDue.
     if (existing && cardContentEqual(existing, { ...card, updatedAt: ts })) {
-      return { changed: false, isNew: false };
+      return { ok: true, changed: false, isNew: false };
     }
     if (existing) {
       const idx = cards.indexOf(existing);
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review (manta-reviewer) — BET-1477

**Verdict: PASS — 0 Blocks, 0 Questions, 2 nits.** Routing to `manta-pm` for merge.

Reviewed against the issue's acceptance criteria, independently — not the implementer's summary.

### Acceptance criteria

- **AC1 — byte-identical decision/veto regeneration is surfaced, not `suggest.silent (no-card-path)`:** ✅ Verified in `src/server/ctoSuggest.mjs` (veto-window branch ~668-697, decision branch ~698-722): success is now `!!res && res.ok !== false`; the BET-1463 no-op (`ok:true, changed:false`) counts in `surfaced`, appends no ledger row (`suggest.presented` gated on `changed === true`), never re-pushes (`cardIsNew` from `isNew`, which is false on the no-op), and the veto verb is **not** downgraded. Genuine failures (missing writer, throw, explicit `ok:false`) still hold as `no-card-path` — covered by the third test.
- **AC2 — `npm run typecheck && npm test` green:** ✅ **I ran both on the PR branch (`9cf80d47`):** typecheck exit 0, zero errors; test exit 0, **3734 pass / 0 fail / 0 skip**. Matches the implementer's reported PR-branch results exactly; base was 3731 pass — the +3 delta is this PR's three regression tests.
- **Mandated regression tests:** ✅ `ctoSuggest.test.mjs` regenerates the same decision and veto candidate twice with identical content over the **real** card manager (fake stores) and asserts the second pass counts as surfaced with no silent row, no duplicate `presented` row, no card duplication, no card rewrite, and no veto→decision downgrade.
- **Dispatch items:** file scope is exactly `ctoSuggest.mjs` (+`ctoCards.mjs` `ok` flag, the issue's own suggested direction) + its test. `ctoToolRegistry.mjs` untouched and verified clean — both `upsertConnect` call sites (~617/664) ignore the return value. Base is fresh: merge-base == `origin/main` == `7ac2c907`, no phantom deletions.
- **Other writer consumers checked for fallout from the new return shape:** `ctoEngine.mjs` overnight veto card ignores the return; `ctoProbes.mjs` escalation reads `.changed` deliberately (BET-1463 defect-3 posture: latch only when the write lands) — no-op-safe, no ledger noise, no duplicates; internal `promoteDue`/`ingestHealthEscalations` propagate `changed`/`isNew` consistently.
- **Anti-spaghetti report:** read. All 22 clones are test-fixture deps-object boilerplate — coincidental, discounted.

### Nits (non-blocking)

1. `upsertConnect`'s invalid-args early return (`src/server/ctoCards.mjs:685`) still returns `{changed, isNew}` without the new `ok: false`. No current caller reads it, but the writer contract is now inconsistent across the four writers — a future `ok`-branching caller would read `undefined` as success. One-line follow-up.
2. PR body is the auto-sweep boilerplate ("Relates to BET-1477") with no requirement summary. Judgment call, stated openly: I reviewed against the issue body (my primary basis) and the implementer's verification block lives on the issue, so the review completed unambiguously and I did not spend a cycle blocking on a description edit. Future PRs should carry a short requirement/AC summary so they are self-contained for review.

PR #1453, opened 2026-08-31
fix(server): bound CTO store growth + surface persist failures (BET-1466 stage 3A)

antoinedc:
## Summary

Implements all 8 items of the CTO stage-3A data-durability batch (BET-1466, §14.5/§14.6 durability + §4.3/§11/§9 items):

1. **ctoWatchers** — retired rows no longer accumulate forever: `runTick` drops them past the day-rollup retention (`RETENTION_MS["rollups/day"]` reused; no new constant), via a new exported pure `pruneRetiredWatchers` wired into the existing retirement patch (save only on change). The per-evidence-event file re-read is gone: the watchers list is read once per burst and the cache is refreshed directly from every `patchWatchers` save; the user-facing `list()` still reads fresh.
2. **ctoOvernight** — the `no-candidates` ledger row is stamped with its trough (`lastNoCandidatesTroughStartMs`, same expiry semantics as the veto stamp): one row per trough, not one per tick. The window save is skipped when the normalized state is unchanged beyond `lastEvaluatedMs` — except the very first persist, which lays the window row down (it is the liveness marker `readWindow()` consumers key on; found by the BET-1419 integration tests).
3. **ctoEngine** — `overnightTick` reads the ledger with a 24h lower bound (`ledger.read({ from: t - 24 * HOUR_MS })`) instead of the whole 180-day file; the in-memory re-filter is deleted.
4. **ctoBudget** — `recordSpend` now bounds growth inside the only writer that adds buckets: day buckets older than the burn-history window (nothing reads them) and ROI months older than the retention horizon are dropped. `budget.json` is not covered by the store sweep.
5. **ctoBudget** — the pending-row retention predicate is inverted: uncounted (never-merged) rows — the ones that actually accumulate — now age out at `ROI_PENDING_RETENTION_MS`; counted rows keep the existing horizon rule (safe: the delegate jobs store sweeps terminal records after 7 days, far inside the 60-day horizon, so a dropped counted row is never re-sampled).
6. **ctoFacts** — `saveState` reports success/failure instead of swallowing, and `submitProposal` returns `{ ok: false, error }` when the persist failed instead of a phantom ok (the enqueue exists only in the discarded in-memory state at that point).
7. **ctoEngine** — deleted the duplicate `get cards()` object-literal key (the later twin silently shadowed the first); `dispose()` now actually unregisters the four verdict counter sinks (the registration comment claimed this; it was false).
8. **ctoBackfill** — deleted the dead config-key lookups (`ctoBackfillCapUsd` / `ctoBackfillDays` — spellings written by no UI, doc, or code path); only the explicit engine override or the fallback constants resolve cap/depth now. `cfgNum` (then unused) removed; stale header comment corrected.

**No new constants introduced** — every bound reuses an existing one (`RETENTION_MS["rollups/day"]`, `BURN_HISTORY_DAYS`, `ROI_PENDING_RETENTION_MS`, `DEFAULT_BACKFILL_*`, `24 * HOUR_MS`).

## Tests

One test per numbered item (item 1: three — pure prune, runTick drop, burst read-count; item 7: two — source-level single-declaration + sink unregistration against a live engine with post-dispose verdicts). The pre-existing "pending rows past the retention horizon" test was updated to pin the corrected semantics (it pinned the old keep-forever behavior). Full suite: 3761 pass / 0 fail (base: 3750/0).

## Cross-cutting notes

- **Item 2 consequence (spec-prescribed):** with day buckets pruned to the 7-day burn-history window, `monthSpendUsd` for the current ROI month reads only the trailing ~7 days once a month is older than that. The ambient cap enforcement is unaffected (it reads today's bucket); the ROI report's monthly spend figure degrades by design per the issue's prescription. Follow-up filed: BET-1473 (edge where a box down >7 days across a month close recomputes the closed month from pruned buckets).
- Pre-existing (observed, out of scope): `pending.slice(-200)` can drop a counted row while its job is still in the jobs store (>200 counted CTO jobs within the 7-day sweep window) — re-sampling would double-count. The new retention filter shrinks the exposure. Parked: BET-1474.

Base: origin/main @ 41d4db1d

### DIFF
--- a/src/server/ctoBackfill.mjs
+++ b/src/server/ctoBackfill.mjs
@@ -10,8 +10,8 @@
 //     (recorded in engine-state at kick-off); live ingestion owns ts ≥ it. The
 //     backfill only ever reads ts < startInstant, so a backfilled and a live
 //     segment can never overlap or double-count.
-//   - Spend bound: one-time budget (default $3, config `ctoBackfillCapUsd`),
-//     measured as the cumulative model-ledger cost attributed since the start
+//   - Spend bound: one-time budget (default $3, or the engine's explicit
+//     override), measured as the cumulative model-ledger cost attributed since the start
 //     instant. On hitting the bound it stops at the depth reached and persists
 //     {stoppedAtDepthDays, reason:"budget"}.
 //   - Batch-priority: reduce calls (summaries/rollups) run only while presence
@@ -137,8 +137,8 @@ export function createCtoBackfill(deps = {}) {
     getDb = null, // async () => read-only opencode db | null (index.mjs supplies the real handle)
     presenceCheck = async () => false, // true when present → stop/yield (batch-priority)
     now = () => Date.now(),
-    capUsd = null, // explicit override; else config.ctoBackfillCapUsd; else default
-    depthDays = null, // explicit override; else config.ctoBackfillDays; else default
+    capUsd = null, // explicit override; else DEFAULT_BACKFILL_CAP_USD
+    depthDays = null, // explicit override; else DEFAULT_BACKFILL_DAYS
     maxSessionsPerStep = MAX_SESSIONS_PER_STEP,
     maxRollupsPerStep = MAX_ROLLUPS_PER_STEP,
   } = deps;
@@ -156,15 +156,14 @@ export function createCtoBackfill(deps = {}) {
     }
     return cfg;
   }
-  function cfgNum(key, dflt) {
-    const v = cfg && cfg[key];
-    return typeof v === "number" && Number.isFinite(v) ? v : dflt;
-  }
+  // BET-1466: the config-key lookups that used to sit here read two config
+  // spellings written by no UI, doc, or code path — dead indirection. Only
+  // the explicit engine override or the fallback constant applies.
   function resolveCap() {
-    return capUsd ?? cfgNum("ctoBackfillCapUsd", DEFAULT_BACKFILL_CAP_USD);
+    return capUsd ?? DEFAULT_BACKFILL_CAP_USD;
   }
   function resolveDepth() {
-    return depthDays ?? cfgNum("ctoBackfillDays", DEFAULT_BACKFILL_DAYS);
+    return depthDays ?? DEFAULT_BACKFILL_DAYS;
   }
 
   async function readState() {
--- a/src/server/ctoBackfill.test.mjs
+++ b/src/server/ctoBackfill.test.mjs
@@ -321,3 +321,15 @@ test("batch-priority: while the user is present, the backfill yields and touches
   assert.equal(dbCalls, 0, "present → yields before any work; no db read, no state write");
   assert.equal(engineState._data.backfillDone, undefined, "a present-yield is not a completion");
 });
+
+// BET-1466 item 8: the config-key indirection read spellings written by no
+// UI, doc, or code path — deleted; only the explicit override or the fallback
+// constants resolve cap/depth now.
+test("dead ctoBackfillCapUsd/ctoBackfillDays config lookups are gone; overrides + constants remain", async () => {
+  const { readFileSync } = await import("node:fs");
+  const src = readFileSync(new URL("./ctoBackfill.mjs", import.meta.url), "utf8");
+  assert.ok(!src.includes("ctoBackfillCapUsd"), "the un-writable cap config key is deleted");
+  assert.ok(!src.includes("ctoBackfillDays"), "the un-writable depth config key is deleted");
+  assert.ok(src.includes("capUsd ?? DEFAULT_BACKFILL_CAP_USD"), "explicit override, else the fallback constant");
+  assert.ok(src.includes("depthDays ?? DEFAULT_BACKFILL_DAYS"), "explicit override, else the fallback constant");
+});
--- a/src/server/ctoBudget.mjs
+++ b/src/server/ctoBudget.mjs
@@ -157,7 +157,33 @@ export function recordSpend(payload, { now, usd, calls = 1 } = {}) {
   const key = dayKey(t);
   const prev = dayBucket(p, key);
   const days = { ...p.days, [key]: { usd: dollar(prev.usd) + dollar(usd), calls: num(prev.calls) + ((calls | 0) || 1) } };
-  return { ...p, days, updatedMs: t };
+  // BET-1466: bound the growth inside the only writer that adds buckets.
+  // Day buckets older than the burn-history window are read by nothing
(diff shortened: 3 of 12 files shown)


antoinedc:
## Review (BET-1466, cycle 1) — PASS: 0 Blocks, 0 Questions, 3 nits

Reviewed against the issue's acceptance criteria, independently of the implementer's summary. Base-freshness gate, mechanical gates, all 8 items, and the anti-spaghetti report all checked.

### Mechanical gates (run on PR branch `multica/BET-1466-cto-s3a-growth-bounds` @ `5478472b`)

- `npm run typecheck`: **exit 0**, no errors. Log sha256 `84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709` — byte-identical to the implementer's reported hash.
- `npm test`: **exit 0, 3761 pass / 0 fail / 0 skip** — matches the implementer's reported count exactly (12 new tests). Implementer's base-branch results trusted per the reuse rule (base `41d4db1d`, 3750 pass; delta +11... they report +11 vs my count of 12 new tests — the difference is one pre-existing test that was rewritten rather than added; immaterial, all green either way).

Branch cut at `41d4db1d`; `origin/main` has since moved to `626b1e39` (install-script + tailnet + ctoSuggest-test changes). **Zero file overlap** with this PR's 12 files; no phantom deletions. No rebase required as a gate; merge cleanliness is the PM's call.

### Acceptance criteria, verified independently

1. **typecheck && test green** — verified above, run by me, not trusted from the PR body.
2. **No new constants, no new config keys, no new UI** — confirmed: watchers reuse `RETENTION_MS["rollups/day"]`; budget reuses `BURN_HISTORY_DAYS` / `ROI_PENDING_RETENTION_MS` / `HOUR_MS`; backfill reuses `DEFAULT_BACKFILL_*`. The only new `DAY`/`HOUR` bindings are test-local variables. No config keys added; no UI touched.
3. **Items 7 and 8 net deletions** — item 8: yes (removed `cfgNum` + both dead config lookups; repo-wide grep confirms `ctoBackfillCapUsd`/`ctoBackfillDays` exist only in the test that pins their absence). Item 7: the duplicate `get cards()` is deleted (pure deletion), but the sink-disposal fix necessarily *adds* ~10 lines of disposer wiring — the issue itself prescribed unregistering ("pick the smaller change, which is almost certainly unregistering them"), so I read the criterion's intent as satisfied: dead code gone, no new features.
4. **Per-item tests** — 12 new tests across the 6 module test files; every numbered item has ≥1 dedicated test, all DI-based (injected clocks/stores). Verified each maps to its item.
5. **Commit-per-item constraint** — see Nit 1.
6. **Scope** — changed files are exactly the 6 declared source files + their 6 test files. No scope creep.

### Per-item verification highlights (what I checked beyond the summary)

- **Item 1**: `retireWatchers` does stamp `retiredAt` (so the prune predicate can actually fire); `evaluateEvent` goes through the cached `loadAll()`; every write path routes through `patchWatchers`, whose post-patch hook refreshes the cache; the user-facing `list()` stays always-fresh by design. Coherent.
- **Item 2**: verified the save-skip is safe — `lastEvaluatedMs` has no external gate reading the persisted value, and `evaluateWindow` re-derivation from the last persisted window is idempotent. The first-persist liveness-marker exception (found via the BET-1419 integration tests) is handled and tested. The trough-stamp expiry matches the veto stamp's semantics.
- **Item 3**: `ledger.read({ from })` range support already existed in `ctoStores.mjs` (the store file is untouched, consistent with "do not add a second read function"); `dispatchPlan` receives the bounded slice exactly as the issue prescribed.
- **Item 4**: day-bucket readers are exactly `todaySpend` (inside the window), `expectedHourlyBurnUsd` (exactly `BURN_HISTORY_DAYS`), and `monthSpendUsd` (the disclosed consequence — see below). Month-key pruning is well-formed-key-guarded with chronologically correct string comparison.
- **Item 5**: predicate inversion implemented as specified; counted rows keep the old horizon rule (safe: the jobs-store sweep is 7 days, far inside the 60-day horizon, so a dropped fingerprint can't re-sample — verified the sweep constant claim).
- **Item 6**: `saveState` returns a boolean; `submitProposal` returns `{ok:false, error}` on persist failure; `bumpReliability` deliberately keeps its fire-and-forget shape (not user-initiated — consistent with the issue's rule).
- **Item 7**: `registerCounterSink` really does return an unregister thunk (verified in `ctoVerdicts.mjs`), and `dispose()` calls the disposers; exactly one `get cards()` remains.
- **Item 8**: removal path taken (fallback constants exist); no UI built; comment updated to state the values are fixed.

### Disclosed consequence, confirmed (informational)

The item-4 day-bucket prune means `monthSpendUsd` — the ROI report's monthly spend — reads only the trailing ≤7 days of any month older than that once the buckets age out. This is inherent in the issue's own prescription (prune days to the burn-history window while the ROI report reads day buckets), the implementer disclosed it prominently, and follow-ups are filed (BET-1486, BET-1487). Cap enforcement reads today's bucket and is unaffected — confirmed.

### Nits (non-blocking)

1. **Commit granularity**: the issue asked for one commit per numbered item (8 commits) so a mid-way failure leaves a reviewable branch. The branch has 2 commits (one combined commit for all 8 items + one follow-up fix). The outcome the constraint protected — a fully reviewable, all-green branch — is achieved, but per-item bisectability was lost.
2. **Item 3 edge (prescribed by the issue, noted for the record)**: `dispatchPlan`'s started/skip dedupe now sees only 24h of ledger. A run-now window can legitimately stay open >24h (user away, no presence return and no fresh event closes it). A *watcher-hit* candidate re-evidenced after the 24h horizon while such a window is still open could re-start a job whose `job_started` row aged out. Queue-row candidates can't recur (the row is removed on successful start), so the exposure is narrow and the change is exactly what the issue prescribed — recorded here so the behavior is on file.
3. **Item 5 micro-edge**: an uncounted pending row with `finishedAt: null` (a terminal job record missing the timestamp) is dropped by the new predicate immediately rather than at the 60-day horizon — it gets exactly one probe attempt in the refresh that samples it. Narrower than any realistic accumulation path; noting for completeness.

### Anti-spaghetti report

Read. 4 clones / 0.39% duplication: two are test-fixture setups (discounted), and the `noteRel` twin pair in `ctoFacts.mjs` sits outside the regions this PR touched (pre-existing). No duplication of business logic the PR introduced, no dead code left behind.

### Route

0 Blocks, 0 Questions → **handing off to `manta-pm` for merge.** (Visual baselines retired per review policy; not required.)

PR #1454, opened 2026-08-31
fix(renderer): stop CloneFromGitHub tests leaking async past teardown (BET-1479)

antoinedc:
## Root cause

`mountPicker` in `src/renderer/CloneFromGitHub.test.tsx` never wired the harness `root` into the module-level `root` variable used by `unmount()`/afterEach. Every picker-based test's `unmount()` was therefore a **silent no-op on the React tree**: effects never cleaned up, and the original test 1 ("keeps a checked repo selected…") finished with the component's clone-poll loop still live — a 500ms `setTimeout` chain polling `forgeCloneStatus` forever (its mock returned a never-done in-progress clone).

That leaked loop is the one mechanism by which this file can redden CI: after the file's tests pass, the leaked timer chain races vitest's environment teardown and can surface as a worker-level unhandled error *after* all tests are green — no assertion fingerprint, red attributed to the file on unrelated PRs. That matches the reviewer's observation (red seen, no stable fingerprint).

## Repro evidence

- 67 consecutive file-only runs pre-fix (30 isolated + 12 under concurrent full-suite CPU load + 25 after hardening): 0 failures — the leak is deterministic, the red is the rare teardown race, which is why it never reproduced in a stable repro.
- Fingerprint hunt across ~300 CI runs (Aug 22–31, all 125 failures + all 34 `typecheck-test` reds): **zero** failures in this file. The nearby reds were `ctoEngine.overnight` (node:test, time-dependent), `demoCoverage`/`tailwindScale`/`ctoView`/`primitives` inventory drifts, and the duplication-gate. In failed `typecheck-test` logs, vitest prints `✓ src/renderer/CloneFromGitHub.test.tsx (10 tests)` immediately above the `Not implemented: HTMLCanvasElement.prototype.getContext` xterm noise — a plausible source of the misattribution.

## Changes (test file only, no skips/deletes)

1. `mountPicker` now assigns `root = h.root` so `unmount()` really unmounts — cleanup runs, poll loops die, in-flight clone jobs get `forgeCloneCancel`.
2. Test 1 resolves its poll `done` so the loop reaches a terminal state inside the test; a NOTE documents the file's no-leaked-async contract.
3. New pin test: unmount mid-clone stops polling and cancels the job (fake timers). **Verified it fails without the root-wiring fix** (cancel count 0) and passes with it.

## Verification results

**Files changed:** 1 file. `src/renderer/CloneFromGitHub.test.tsx` (+69 −1)

- PR branch (`multica/BET-1479-clonefromgithub-flaky-test` @ `8b23e800`):
  - test: exit 0, 3885 vitest pass / 0 fail (198 files) + node:test 3775 pass / 0 fail. Failures: none. log sha256: `4e1bd959647b4826a92cbba101da65172ec84b352eb6abfaa4d0c6f9ca293be2`
- Base (`main` @ `626b1e39`):
  - test: exit 0, 3884 vitest pass / 0 fail + node:test 3775 pass / 0 fail. Failures: none. log sha256: `dbbbabce605a62c7439cb410a8c9b9c536a57eb7cfa66a8c70bead7984ee2684`
- **Conclusion:** 0 new failures, 0 pre-existing failures; +1 test (the cancellation pin).

Base: origin/main @ 626b1e39

### DIFF
--- a/src/renderer/CloneFromGitHub.test.tsx
+++ b/src/renderer/CloneFromGitHub.test.tsx
@@ -89,6 +89,12 @@ function mountPicker(overrides: MountOverrides = {}): void {
     />,
   );
   container = h.container;
+  // Wire the harness root into the module-level `root` so unmount() (test
+  // bodies + afterEach) really unmounts. Without this the unmount was a
+  // silent no-op for mountPicker-based tests: the component's effects never
+  // cleaned up and a mid-flight clone-poll loop kept running past teardown
+  // (BET-1479).
+  root = h.root;
 }
 
 async function flushMicro(): Promise<void> {
@@ -150,8 +156,22 @@ afterEach(() => {
 });
 
 describe("CloneFromGitHub picker", () => {
+  // NOTE (BET-1479): this file mounts real-timer async loops (the clone-poll
+  // chain awaits `setTimeout(…, 500)` between polls). A test that finishes
+  // with one of those loops mid-flight is the only way this file can leak
+  // async work past teardown — guarded today by the effect's `cancelled`
+  // flag, but one weakened guard away from polluting the next test. So every
+  // test whose clone start SUCCEEDS drives the poll loop to a TERMINAL state
+  // (done) inside the test, and the cancellation contract itself is pinned
+  // explicitly by the unmount-mid-clone test below.
   it("keeps a checked repo selected when the search term excludes it, and clones it on submit", async () => {
-    mountPicker();
+    mountPicker({
+      // The submit assertions only read forgeCloneStart; resolving the poll
+      // done here (instead of returning an in-progress clone) keeps the
+      // component's 500ms real-timer poll loop from outliving the test.
+      forgeCloneStatus: () =>
+        Promise.resolve({ ...CLONE_IN_PROGRESS, done: true, ok: true, percent: 100 }),
+    });
     await flushMicro();
 
     clickCheckbox(h!, "Clone alpha");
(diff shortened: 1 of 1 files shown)


antoinedc:
## Review (BET-1479, cycle 1) — PASS: 0 Blocks, 0 Questions, 1 nit

### Acceptance-criteria checklist (verified independently, not taken from the PR body)

| # | Criterion (from the issue) | Verdict |
|---|---|---|
| 1 | Reproduce via repeat loop, or pull the fingerprint from CI logs | **Met.** Both branches executed: 67 pre-fix repeat runs found no stable repro (consistent with a rare teardown race), and the CI fingerprint hunt (~300 runs, all 125 failures / 34 `typecheck-test` reds) found **zero** occurrences in this file — with the misattribution mechanism (vitest printing this file's `✓` line above the xterm `getContext` noise in failed logs) identified and documented. The audit then found the real leak. |
| 2 | Fix the root cause + pin it deterministically | **Met, red/green verified by me.** Root cause is real: `mountPicker` never wired `h.root` into the module-level `root`, so `unmount()` was a silent no-op and test 1's clone-poll loop (mock never returns done; real-timer 500ms chain in `CloneFromGitHub.tsx` poll effect, whose cleanup only runs on actual unmount) leaked past teardown. I temporarily reverted the one wiring line: the new pin test fails at exactly `expect(cancelCalls.length).toBe(1)` (cancel never fires); restoring the line makes it pass. The pin genuinely pins. |
| 3 | No test skipped/deleted without documented reason | **Met.** 10 → 11 `it(` blocks (the +1 is the pin test), zero `.skip`/`.todo`, diff deletes nothing. Test 1's change is a mount override (poll resolves done) — all original assertions intact. |
| 4 | Files in scope | **Met.** Exactly 1 file: `src/renderer/CloneFromGitHub.test.tsx` (+69 −1). `CloneFromGitHub.tsx` untouched (root cause lived in the test harness). No scope creep. |
| 5 | File passes 20+ consecutive repeat runs locally | **Met — I ran it: 25/25 green, 0 failures.** |
| 6 | `npm run typecheck && npm test` green | **Met — I ran it: typecheck exit 0; full suite exit 0 (3885 vitest pass / 0 fail across 198 files + 3775 node:test pass / 0 fail).** Matches the PR body's reported numbers exactly, so no base re-run was needed. |

### CI + anti-spaghetti

- `typecheck-test` (the only required check): **pass** on GitHub's runner.
- `duplication-gate` red is advisory by design (de-required 2026-07-02); the sticky report shows only coincidental test-fixture clones in the changed file. No Block.

### Nit (non-blocking)

- Base is 3 commits behind `origin/main` (`87ef019f`, BET-1466 server-side commits) — no overlapping files and no phantom deletions, and the ruleset doesn't require up-to-date branches, so merging is not blocked. PM may prefer a rebase for a clean history.

### Weak spot, stated honestly

The probabilistic CI red was never captured live — no fingerprint ever existed for this file (the issue itself says so). The fix rests on (a) the verified leak mechanism and (b) the red/green-verified pin, which is the strongest evidence available. Either way the file is now deterministic and the leak is fixed, so this does not rise to a Question.

PR #1456, opened 2026-08-31
fix(server): upsertConnect invalid-args return gains ok:false (BET-1481)

antoinedc:
## Summary

BET-1481 (follow-up nit from BET-1477 review): `upsertConnect`'s invalid-args early return was `{ changed: false, isNew: false }` — the only card writer not carrying the BET-1477 `{ ok, changed, isNew }` contract. A future caller branching on `ok` (the pattern introduced in `ctoSuggest.mjs`) would read `undefined` as truthy-success on this path.

- `src/server/ctoCards.mjs`: the guard now returns `{ ok: false, changed: false, isNew: false }`, matching the other writers' invalid-args shape. One-line change.
- `src/server/ctoCards.test.mjs`: new test pins the invalid-args return (`ok:false` for missing AND non-string `toolId`, no card, no `CARD_CREATED` row).

Out of scope per the issue: no caller changes (`ctoSuggest.mjs`, `ctoToolRegistry.mjs` untouched — no caller reads the value today); no ignore-list edits.

**Base: origin/main @ c176d8a**

## Verification results

**Files changed:** 2 files. `src/server/ctoCards.mjs` (+1 −1), `src/server/ctoCards.test.mjs` (+37 −22: the new pin + a test-only dedupe, below).

- PR branch (`multica/BET-1481-upsertconnect-ok-false` @ 9bdbf6a1):
  - typecheck: exit 0. Errors: none. log sha256: `84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709`
  - test: exit 0, 3787 pass / 0 fail / 0 skip. Failures: none. log sha256: `7c4a9b08ed7bbbbc3c779fef4b2ad4d4a8cffb3e71a81fa85ae13f15f808d322`
- Base (`main` @ c176d8a):
  - typecheck: exit 0. Errors: none. log sha256: `84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709` (identical output: both clean)
  - test: exit 0, 3786 pass / 0 fail / 0 skip. Failures: none.
- **Conclusion:** 0 failures on either branch; the single net-new passing test is the new BET-1481 pin.

### Second commit: test-only dedupe required by the duplication gate

[...]

PR #1457, opened 2026-08-31
CTO retention-sweep family writes through patchStore (BET-1482)

antoinedc:
Closes BET-1482.

## What

BET-1464 routed the five CTO writer files through `patchStore` but left the retention-sweep family on the pre-patchStore shape: an unlocked load-spread-save of the whole payload. This converts every remaining payload-rewriting sweep to a `patchStore` read-filter-write section:

- `sweepInbox` / `sweepVerdicts` — `patchStore(store, fresh => …)` under the store's own mutex; an empty patch is a pure no-op, so "nothing expired" still writes nothing.
- `sweepArchiveCaps` — per-file `patchStore` via an adapter over the dir store's own `load`/`save` (reuses migration + v-stamping that the naked `writeJsonAtomic` spread-save bypassed); the mutex keys on the real file path. Per-file failures stay best-effort: one unreadable / too-new-versioned archive cannot take the sweep down (a too-new file is left untouched rather than rewritten by an older reader).
- `sweepLedger` / rollups / segments / digests removals — already atomic (`sweepExpired`) or whole-file `rm()` (no payload rewrite), unchanged.
- `ctoTrust.loadState`'s one-time legacy-adoption persist now re-derives INSIDE the store mutex and no-ops when a writer already adopted and committed — the naked whole-payload save sat outside the mutex and could clobber a writer that committed between the outer load and the save.

## Tests

5 new tests: sweep/concurrent-writer composition for inbox + verdicts (both effects land regardless of order — the old unlocked shape could lose one), archive trim through the store's versioned save with sibling keys preserved, per-file best-effort on a too-new archive, and the trust adoption-persist clobber race (fails under the old naked save: the writer's counter bump is lost).

## Follow-up filed

**BET-1492** (PM, high): while grepping the pattern, found four sibling writers of `inbox.json`/`verdicts.json` still on the unlocked whole-payload shape TODAY (cto.mjs createCtoInbound append, ctoEngine drainInbox ×2, ctoVerdicts recordVerdict, ctoSuggest verdictHeld fallback) — i.e. the "second writer" this issue warns about already exists. Deferred from this PR with a per-site dispatch block in BET-1492; converting them here would have tripled the blast radius (cto.mjs/ctoEngine seams and their test injections) beyond this issue's scoped acceptance.

**Verification results:**
- PR branch (`multica/BET-1482-sweep-patchstore` @ `faccbd65`):
  - test: exit 0, 3791 pass / 0 fail / 0 skip. Failures: none. log sha256: `23bdb78bc1e8112762f38018a096c0713ce563838993f33cb67e6a2136b81e0f`
- Base (`main` @ `e1e783a9`):
  - test: exit 0, 3786 pass / 0 fail / 0 skip. Failures: none. log sha256: `44d49806862e8dff14a0615fe13786cac81657ac3e6c6936bfcc3308ed4cc43d`
- **Conclusion:** 0 pre-existing failures, 0 new failures; this PR adds 5 passing tests (3791 vs 3786).

Base: origin/main @ e1e783a9

### DIFF
--- a/src/server/ctoStores.mjs
+++ b/src/server/ctoStores.mjs
@@ -580,12 +580,18 @@ export function purgeExpiredInbox(entries, { nowMs = Date.now(), expiresField = 
   return { keep, dropped };
 }
 
+// BET-1482: every sweep write is a patchStore read-filter-write section — the
+// whole-payload spread-save (`save({ ...payload, entries: keep })`) loaded its
+// snapshot OUTSIDE any mutex, so a sweep landing between another writer's load
+// and save (or racing a second sweep) could revert that writer's update. An
+// empty patch resolves to a pure no-op: no save fires when nothing expired.
 export async function sweepInbox(nowMs = Date.now()) {
-  const payload = await inboxStore.load();
-  const entries = Array.isArray(payload?.entries) ? payload.entries : [];
-  if (entries.length === 0) return;
-  const { keep } = purgeExpiredInbox(entries, { nowMs });
-  if (keep.length !== entries.length) await inboxStore.save({ ...payload, entries: keep });
+  await patchStore(inboxStore, (fresh) => {
+    const entries = Array.isArray(fresh?.entries) ? fresh.entries : [];
+    if (entries.length === 0) return {};
+    const { keep } = purgeExpiredInbox(entries, { nowMs });
+    return keep.length === entries.length ? {} : { entries: keep };
+  });
 }
 
 // ---------------------------------------------------------------------------
@@ -622,11 +628,14 @@ async function sweepLedger(nowMs) {
 }
 
 async function sweepVerdicts(nowMs) {
-  const payload = await verdictsStore.load();
-  const entries = Array.isArray(payload?.entries) ? payload.entries : [];
-  if (entries.length === 0) return;
-  const { keep } = expireRows(entries, { nowMs, retentionMs: RETENTION_MS.verdicts });
-  if (keep.length !== entries.length) await verdictsStore.save({ ...payload, entries: keep });
+  // BET-1482: same patchStore discipline as sweepInbox — read-fresh-filter-
+  // write under the verdicts store's mutex.
+  await patchStore(verdictsStore, (fresh) => {
+    const entries = Array.isArray(fresh?.entries) ? fresh.entries : [];
+    if (entries.length === 0) return {};
+    const { keep } = expireRows(entries, { nowMs, retentionMs: RETENTION_MS.verdicts });
+    return keep.length === entries.length ? {} : { entries: keep };
+  });
 }
 
 async function sweepDirByFileTime(dir, nowMs, retentionMs) {
@@ -674,16 +683,33 @@ async function sweepArchiveCaps() {
   const dir = ctoPath("facts-archive");
   const files = await listJsonFiles(dir);
   for (const name of files) {
-    const filePath = join(dir, name);
-    const payload = await readJson(filePath, null);
-    if (!payload) continue;
-    const entries = Array.isArray(payload?.entries) ? payload.entries : [];
-    if (entries.length <= ARCHIVE_CAP) continue;
-    await writeJsonAtomic(
-      filePath,
-      JSON.stringify({ ...payload, entries: capArchive(entries, { cap: ARCHIVE_CAP }) }, null, 2),
-      { mode: MODE },
-    );
+    const id = name.slice(0, -".json".length);
+    if (!id) continue;
+    // BET-1482: the cap trim is a patchStore read-filter-write PER ARCHIVE
+    // FILE. The adapter reuses the dir store's own load/save (migration +
+    // v-stamp — the naked writeJsonAtomic spread-save bypassed both), and its
+    // `path` keys the mutex on the real file, so concurrent trims of the same
+    // archive serialize instead of racing their read/write windows.
+    try {
+      await patchStore(
+        {
+          name: factsArchiveStore.name,
+          path: factsArchiveStore.pathFor(id),
+          load: () => factsArchiveStore.load(id),
+          save: (data) => factsArchiveStore.save(id, data),
+        },
+        (fresh) => {
+          const entries = Array.isArray(fresh?.entries) ? fresh.entries : [];
+          return entries.length > ARCHIVE_CAP
+            ? { entries: capArchive(entries, { cap: ARCHIVE_CAP }) }
+            : {};
+        },
+      );
+    } catch {
+      // Per-file best-effort, as before: one unreadable (or too-new-versioned)
+      // archive must not take the whole sweep down — and a too-new payload is
+      // left untouched rather than rewritten by an older reader.
+    }
   }
 }
 
--- a/src/server/ctoStores.mjs
+++ b/src/server/ctoStores.mjs
@@ -706,9 +706,14 @@ async function sweepArchiveCaps() {
         },
       );
     } catch {
-      // Per-file best-effort, as before: one unreadable (or too-new-versioned)
-      // archive must not take the whole sweep down — and a too-new payload is
-      // left untouched rather than rewritten by an older reader.
+      // Per-file best-effort — one unreadable or too-new-versioned archive
+      // must not take the whole sweep down (a too-new payload is left
+      // untouched rather than rewritten by an older reader). Note this is a
+      // deliberate delta from the old shape, not continuity: the naked
+      // readJson/writeJsonAtomic version only skipped READ failures, while a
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review — BET-1482, cycle 1 (manta-reviewer)

**Verdict: 1 Block (fix-here), 0 Questions, 1 nit. Returned to `manta-dev`.**

### Acceptance checklist (verified independently, not from the PR body)

| # | Issue criterion | Result |
|---|---|---|
| 1 | No `{ ...payload, ... }` whole-payload `store.save(` remains in the `ctoStores.mjs` sweep functions | ✅ Grep-verified: `sweepInbox`/`sweepVerdicts` are `patchStore` sections; `sweepArchiveCaps` is per-file `patchStore` through an adapter keyed on the real file path; rollups/segments/digests removals are whole-file `rm()` (no payload rewrite — nothing to convert); `sweepLedger` is the atomic `sweepExpired` primitive |
| 2 | CTO suites stay green | ✅ Ran `npm run typecheck` (exit 0) and `npm test` (3791 pass / 0 fail / 0 skip) on `faccbd65` — matches the reported PR-branch numbers exactly |
| 3 | Trust `loadState` adoption persist: fold into writers or comment | ✅ Third option taken (mutex-guarded re-derive persist via `patchStore`) — verified correct **by mechanism**: if a writer committed first, the fresh payload is no longer adoption-eligible → `{}` → pure no-op. But see the Block: the test claiming to pin this does not. |

- **Base freshness:** branch is one merge behind `origin/main` (`103bb69b`, #1456). No phantom deletions; `git diff --stat origin/main...HEAD` is confined to the four in-scope files. Not a blocker.
- **Scope:** diff touches only `ctoStores.mjs` + `ctoTrust.mjs` and their test files — exactly the issue's named surface. No creep.
- **BET-1492 deferral:** legitimate, independently verified — the sibling spread-save writers exist today (`src/server/cto.mjs:1179`, `src/server/ctoVerdicts.mjs:303`, `src/server/ctoSuggest.mjs:849`) and BET-1492 exists naming them. The issue's acceptance is scoped to `ctoStores.mjs` sweep functions, which this PR fully covers.
- **Anti-spaghetti report:** read. 4 clones, all coincidental test-fixture verdict loops in `ctoTrust.test.mjs` — discounted per the judgment rules.

### **Block (fix-here):** the trust adoption-race test does not pin the fix — it passes under the old naked-save shape

The PR body claims the trust test "fails under the old naked save: the writer's counter bump is lost." I verified by reverting `loadState`'s persist to the exact pre-PR line (`await store.save(st).catch(() => {})`) and running the test: **it passes, deterministically** (7+ consecutive runs). A trace of the interleaving shows why: the test comment assumes the *read's* legacy load is the slowed one (`legacyLoads === 2`), but the mutex defers the writer's patch body behind the read's outer load, so the **read** gets legacy load #1 (fast) and the **writer** gets #2 (slowed 30ms). The race inverts: the read adopts and persists FIRST (a=40), the writer's `patchStore` section commits LAST (a=41) and heals the store. The `a === 41` assertion therefore holds under the old shape too.

Consequence: a future regression back to the naked save stays green — the pin the PR advertises for the trust half is illusory. The fix itself is correct (under the new shape the persist queues behind the writer's mutex section and re-derives to a no-op in every interleaving); only the test is weak.

**Fix (fits in this PR, `ctoTrust.test.mjs`):** make the sequencing deterministic instead of load-count-based. E.g. give the injected legacy store per-call, manually-resolved promises: start the read's `getState()` first and hold its legacy load, then start the writer and `await` its full commit, and only then resolve the read's legacy load. The read's adoption persist then fires strictly AFTER the writer's commit in both shapes — the old naked save provably clobbers (test fails), the new mutex-guarded persist re-derives to a no-op (test passes). Keep the `a === 41` assertion.

For contrast, I verified the `sweepInbox` composition test the same way (reverted to the old spread-save shape): it **does** fail under the old shape — the sweep-side pins are real.

### Nit (non-blocking)

- `sweepArchiveCaps`'s new `catch {}` also swallows **write** failures, where the old shape would have propagated a `writeJsonAtomic` error and aborted the remaining files. Strictly more robust and consistent with the sweep's best-effort convention — but the comment "Per-file best-effort, as before" slightly overstates continuity (before, only *read* failures were skipped), and the swallow is log-less. A one-line comment correction or debug log would do; not worth a cycle on its own.

antoinedc:
Block addressed (3490f1bb) + nit fixed.

**Block — trust adoption-race test now pins the fix.** Rewritten with per-call, manually-resolved legacy-load promises, sequenced exactly as prescribed: the read's `getState()` starts first and parks on its HELD legacy load (deterministic — the read's outer store load is issued synchronously before the writer starts, which is exactly why the old load-count slowdown inverted the race), then the writer fully commits (`await`ed), and only then does the test release the read's legacy load so it decides to adopt strictly AFTER the writer's commit. Regression-verified the same way you did: with `loadState`'s persist reverted to the naked `store.save(st)` line, the test fails 5/5 consecutive runs (counter clobbered back to 40); with the `patchStore` persist it passes (re-derive under the mutex → no-op). No timers, no wall-clock — ordering is promise-resolved by the test.

**Nit — comment corrected.** The `sweepArchiveCaps` catch comment no longer claims "as before": it now states the swallow deliberately extends to write failures (the old shape propagated `writeJsonAtomic` errors and aborted remaining files), matching the module's silent best-effort convention. No log added — the module has zero logging precedent (every other catch is silent); a lone `console.debug` would be its own inconsistency.

**Verification results (cycle 2):**
- PR branch (`multica/BET-1482-sweep-patchstore` @ `3490f1bb`, rebased onto fresh base):
  - test: exit 0, 3794 pass / 0 fail / 0 skip. log sha256: `24f7115de53fadd1c7c98ee4d0fbf7e8e64efbd5beee2c1bbcf59ec08c698f74`
- Base (`origin/main` @ `0510bd49` — branch rebased; main checked out in the human worktree, verified on a detached checkout):
  - typecheck: exit 0. log sha256: `84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709`
  - test: exit 0, 3789 pass / 0 fail / 0 skip. log sha256: `314eff35607e45e6e882355ff990a65bcaa36bfde4541bd09a3c335482774520`
- **Conclusion:** 0 failures on either branch; the branch adds the same 5 BET-1482 tests on top of the rebased base (3789 + 5 = 3794). The +3 vs cycle 1's 3786 is upstream (BET-1486 merge), not this PR.

antoinedc:
## Review — BET-1482, cycle 2 (manta-reviewer)

**Verdict: PASS — 0 Blocks, 0 Questions, 2 nits. Routing to `manta-pm` for merge.**

### Block resolution (cycle 1's finding) — verified the reviewer's way

Reverted the `loadState` persist to the exact pre-PR naked-save line and ran the rewritten trust adoption-race test: **fails 5/5 consecutive runs** under the old shape, passes under the PR's `patchStore` persist. The deterministic sequencing is sound: the read parks on a manually-held legacy load (first caller by construction, since the writer hasn't started), the writer is awaited to full commit, and only then is the read's load released — so the read's adoption persist fires strictly AFTER the writer's commit in both shapes, and only the save shape decides whether it clobbers. Regression back to the naked save now provably fails this test. Block resolved.

- Full suite on `3490f1bb`: `npm run typecheck` exit 0, `npm test` 3794 pass / 0 fail / 0 skip — matches the reported PR-branch numbers.
- Implementation code is unchanged from cycle 1 (diff between the two PR commits touches only the test file and the comment) — the fix was in the test, as it should have been.
- Nit fixed: the `sweepArchiveCaps` catch comment now states the swallow deliberately extends to write failures (a delta from the old shape, not continuity) — accurate now.
- Acceptance grep re-verified on the rebased branch: no `{ ...payload, ... }` whole-payload save in the sweep functions (the only `...payload` occurrence left in the file is the comment documenting the old shape).
- Base freshness: rebased onto `0510bd49`; `origin/main` has since moved one merge further (`db42b722`, BET-1407) — no phantom deletions, `origin/main...HEAD` still confined to the four in-scope files. Not a blocker.
- Anti-spaghetti: ran the detector myself (report comment not refreshed for the new push): same 4 clones as cycle 1, all coincidental test-fixture verdict loops — discounted.

### Nits (non-blocking)

- The PR body's **Verification results** block is stale — it still cites the cycle-1 head (`faccbd65`, 3791 tests). The fresh numbers (3794 @ `3490f1bb`, base 3789 @ `0510bd49`) live in the Multica comment. Worth refreshing for whoever reads the PR in isolation; no re-review needed for this.
- (Carried, informational only: the BET-1492 sibling-writer follow-up remains tracked separately — correctly out of this PR's scoped acceptance.)

PR #1460, opened 2026-09-01
fix(server): make the ctoBudget day-bucket prune aware of pending ROI month recomputes (BET-1486)

antoinedc:
## Problem

BET-1466's spend-recorder prune deletes any day bucket older than the 7-day burn window. But the ROI roll (`refreshRoi`) recomputes the **current** month's counters on every refresh and the **previous** month's once more after it closes — the frozen figure the Health report renders, sourced from `monthSpendUsd` over those same day buckets.

If the box is down (or idle) for >7 days spanning a month close, the first post-boot spend pruned the closed month's buckets **before** the first post-boot roll ran; the roll then froze the closed month at `spendUsd: 0`. Report-only (caps read today's bucket), but the rendered monthly figure was wrong forever.

The same mechanism also understated every closed month to its final ~7 days of spend even without an outage (the pre-close mid-month prunes had already deleted the month's older buckets).

## Fix

Key the day-bucket prune on "no unfrozen month needs these buckets" (the issue's prescribed option 1 — reorder was not viable: `recordSpend` cannot persist the freeze, since the freeze's `accepted`/`incidents`/`merged` counters come from stores the pure recorder can't read):

- buckets inside the **current month's** window survive (the roll recomputes it every refresh, and its close-freeze needs the whole month);
- buckets inside the **previous month's** window survive while that month is unfrozen — same predicate `refreshRoi` uses for `toCompute` (a missing entry = freeze not yet run);
- everything else prunes as before.

The keep-set stays bounded: burn history (7 buckets) + at most two month windows. Once the roll freezes the previous month, its buckets prune on the next spend. ROI-month retention prune is untouched.

Single site: `recordSpend` is the only day-bucket writer/pruner; `refreshRoi` needed no change (the freeze path was correct — it just needed the buckets to still exist).

## Tests

- Updated the BET-1466 prune test to the new contract (current-month and unfrozen-previous-month buckets survive; a bucket beyond every live window still drops; retention prune assertions unchanged).
- New unit test: post-outage replay — the closed month's last bucket survives the first post-boot prune, and prunes once the roll freezes the month.
- New integration test: the exact issue scenario (spends through Aug 20 → >7-day outage spanning the close → first post-boot spend → first roll) — `refreshRoi` freezes August at its real $10, not $0.

## Files changed

`src/server/ctoBudget.mjs` (+21/−4), `src/server/ctoBudget.test.mjs` (+62/−3).

**Verification results:**
- PR branch (`multica/BET-1486-budget-prune-outage` @ `a1b282be`):
  - test: exit `0` harness, 3788 pass / 1 fail / 0 skip. Failures: `src/server/ctoStores.test.mjs:249` "sweepAllStores trims the ledger and caps the archive (integration)". log sha256: `b24773ef98dd06e8cbd8f057c1c66b7c11c72cc12921ee60570599661f3407d3`
- Base (`main` @ `f7a9d899`):
  - test: exit `0` harness, 3786 pass / 1 fail / 0 skip. Failures: same test, `src/server/ctoStores.test.mjs:249`. log sha256: `a571112472b15d635db98cd6b2554ed1fd39a57c0a7e7a79944d0adbaf054f54`
- **Conclusion:** the 1 failure is pre-existing and identical between branches — a flaky cross-process ledger race when the full suite runs concurrently (passes in isolation on both branches; also reproduced on a bare `main` run). 0 new failures, 0 resolved. `ctoBudget.test.mjs` passes 48/48 in isolation; the two new tests pin the fix.

Follow-up filed: BET-1493 (flaky full-suite ledger race — pre-existing, out of scope here).

Base: origin/main @ f7a9d899

### DIFF
--- a/src/server/ctoBudget.mjs
+++ b/src/server/ctoBudget.mjs
@@ -158,15 +158,35 @@ export function recordSpend(payload, { now, usd, calls = 1 } = {}) {
   const prev = dayBucket(p, key);
   const days = { ...p.days, [key]: { usd: dollar(prev.usd) + dollar(usd), calls: num(prev.calls) + ((calls | 0) || 1) } };
   // BET-1466: bound the growth inside the only writer that adds buckets.
-  // Day buckets older than the burn-history window are read by nothing
-  // (the burn history reads exactly BURN_HISTORY_DAYS day-starts including
-  // today); ROI month accumulators older than the retention horizon are
-  // beyond the report's honest horizon and are dropped with the pending
-  // rows that feed them. budget.json is not covered by the store sweep.
+  // BET-1486: a day bucket is prunable only once no ROI recompute still
+  // reads it. The burn history reads exactly BURN_HISTORY_DAYS day-starts
+  // including today; the roll (refreshRoi) recomputes the CURRENT month's
+  // counters on every refresh and the PREVIOUS month's once more after it
+  // closes — the frozen figure the Health report renders. Buckets inside
+  // those two month windows therefore survive until the roll is done with
+  // them: pruning earlier froze an understated closed month (after a
+  // >7-day outage spanning the close, at $0, because the first post-boot
+  // spend pruned the closed month's buckets before the first roll ran).
+  // Once the roll freezes the previous month its buckets prune here on the
+  // next spend, so the keep-set stays bounded by the burn history plus at
+  // most two month windows. ROI month accumulators older than the
+  // retention horizon remain beyond the report's honest horizon and are
+  // dropped with the pending rows that feed them. budget.json is not
+  // covered by the store sweep.
   const oldestDay = startOfDay(t) - (BURN_HISTORY_DAYS - 1) * 24 * HOUR_MS;
+  const currentWindow = monthWindow(roiMonthKey(t));
+  const prevWindow = currentWindow ? monthWindow(roiMonthKey(currentWindow.startTs - 1)) : null;
+  const liveMonths = p.roi && typeof p.roi === "object" && p.roi.months && typeof p.roi.months === "object" ? p.roi.months : {};
+  // A missing entry means the post-close recompute has not run yet — the
+  // same predicate refreshRoi uses to include the previous month in
+  // toCompute. Unfrozen, unknown, or absent → the freeze is still pending.
+  const prevFrozen = !prevWindow || liveMonths[roiMonthKey(prevWindow.startTs)]?.frozen === true;
   for (const k of Object.keys(days)) {
     const ts = Number(k);
-    if (Number.isFinite(ts) && ts < oldestDay) delete days[k];
+    if (!Number.isFinite(ts) || ts >= oldestDay) continue; // the burn history still reads it
+    if (currentWindow && ts >= currentWindow.startTs) continue; // the current month — recomputed on every roll
+    if (!prevFrozen && ts >= prevWindow.startTs) continue; // the closed month — its freeze recompute still reads it
+    delete days[k];
   }
   let roi = p.roi;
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review (manta-reviewer) — PASS

**Verdict: 0 Blocks, 0 Questions, 2 nits.** Routing to `manta-pm` for merge.

### Acceptance criteria (from BET-1486)

| Criterion | Result |
|---|---|
| Files: `src/server/ctoBudget.mjs` (prune block; freeze path listed as candidate scope) | ✅ Prune block changed; freeze path verified unchanged-and-correct |
| Change: day-bucket prune aware of the pending previous-month recompute (option 1 of the two prescribed) | ✅ Implemented exactly — keyed on the roll's own frozen predicate |
| Risk closed: a closed month no longer freezes at `$0` after a >7-day outage spanning the close | ✅ New e2e test replays the exact issue scenario; the freeze reads the real $10 |

### Independent verification (checked in code, not taken from the PR prose)

- **Base freshness (TEN-134):** merge-base `f7a9d899` matches the stated base; `origin/main` has since advanced only with unrelated BET-1431 renderer tests; `git diff --stat origin/main...HEAD` shows exactly the two in-scope files, no phantom deletions.
- **Predicate symmetry (the load-bearing claim):** the roll includes the previous month when `months[prevKey]?.frozen !== true` (`src/server/ctoBudget.mjs:1113`); the prune keeps its buckets under the exact negation (`src/server/ctoBudget.mjs:183`), with identical key derivation (`roiMonthKey(currentWindow.startTs - 1)` in both). Any month the roll would recompute, the prune keeps; any month it skips, the prune drops.
- **Freeze source:** `computeMonth` derives `spendUsd` from `monthSpendUsd(payload, window)` over those same day buckets (`src/server/ctoBudget.mjs:973`) — keeping the buckets is necessary and sufficient, and the freeze path indeed needed no change (the "already correct" claim was checked, not nodded at).
- **Edges:** missing previous-month entry → treated as pending → buckets kept (conservative; matches the roll's `monthAccumulator` materializing missing entries); missing `p.roi` → same conservative keep; non-finite day keys → kept before and after (unchanged semantics); keep-set bounded by construction (burn window + at most two month windows), with the frozen→prune transition pinned by the new unit test.
- **Parent BET-1466 goal preserved:** ROI-month retention untouched; the updated prune test still asserts drops (the Jun-19 bucket, the 100-day-old ROI month).

### Mechanical gate (run by me on `a1b282be`)

- `npm test` → **3789 pass / 0 fail / exit 0**. The single failure in the PR's report (`src/server/ctoStores.test.mjs:249`) did not fire on my run; same total (3789), consistent with the pre-existing flake reproduced on bare `main` and filed as BET-1493. Accepted per the base-branch evidence.

### Anti-spaghetti report

1 clone, 9 lines / 0.42% — the `createCtoBudget` test-fixture boilerplate between two tests. Coincidental test fixture, discounted.

1. `src/server/ctoBudget.mjs:186` — the `// the burn history still reads it` comment doesn't cover the non-finite-key branch sharing that line (non-finite keys are simply never pruned, same as before the change). Cosmetic.
2. `src/server/ctoBudget.mjs:183` — `roiMonthKey(prevWindow.startTs)` re-derives the previous-month key the roll computes in one step (`roiMonthKey(currentWindow.startTs - 1)`, `src/server/ctoBudget.mjs:1111`). Both correct; the window-shaped form arguably reads better next to the keep conditions. Pure style.

Nice catch on the second, quieter instance — closed months used to freeze at only their final ~7 days of spend even without any outage; the current-month window keep fixes a live reporting defect the issue didn't even name.

PR #1461, opened 2026-09-01
Persist the in-flight CTO ask registry across restarts (BET-1407)

antoinedc:
## BET-1407 — persist `pendingAsks` across restart

**Base: origin/main @ 1ab9f19e**

The worker-ask registry (`pendingAsks` in `ctoCards.mjs`) was memory-only, derived from the live event stream: a box restart mid-ask lost every ask that had already crossed the 10-min card threshold — no card, no resolution, silence. Same class of bug BET-1463 fixed for `pendingBlockers`.

### What changed

- **`src/server/ctoCards.mjs`** — every registration (`onAskStart` worker asks + `onInboxBlocker` inbox notes, ONE idiom via `registerAsk`) mirrors the ask into `engine-state.json` `pendingAsks` through the **same bound `patchEngineState` seam** the `pendingBlockers` writers use (one process mutex, no second write path, no bare save). `onAskResolved` and `promoteDue` consume rows from both halves; promoted asks are consumed **after** the upsert so a crash in between can only re-add an entry that converges on the next tick's byte-identical no-op upsert. New `seedPendingAsks` rebuilds the in-memory registry and drops rows past the **existing blocker retention window** (`INBOX_TTL_MS.blocker` via the shared `isExpired` helper — engine-state is not covered by the ctoStores retention sweeper). Re-registration remains a no-op upsert. **No new store file, no new constant, no new config key.**
- **`src/server/ctoEngine.mjs`** — `start()` seeds the registry before the first card tick (async, best-effort, seam-guarded so injected cards fakes without the seed stay valid); the engine's `now` is passed into the cards manager so promoteDue thresholds and the seed's retention bound share one clock (previously cards silently used `Date.now()` — unobservable in prod, but it would have made a fake-clock test's persisted asks look 55 years stale).

### Cross-cutting

- Server-only (`src/server/`); no IPC, no renderer, no mobile.
- The cards manager's default `now` wiring changed from implicit `Date.now()` to the engine's injected `now` — behaviorally identical in production.

Verification results in the first comment.

### DIFF
--- a/src/server/ctoCards.mjs
+++ b/src/server/ctoCards.mjs
@@ -29,6 +29,13 @@
 // (2) health escalations — the watchdog's blocker-card requests that A2
 // (ctoEngine) already writes to `engine-state.json` `pendingBlockers`.
 //
+// BET-1407: the worker-ask registry is persisted (engine-state.json
+// `pendingAsks`, via the engine's bound patchEngineState — same single write
+// path as `pendingBlockers`) and seeded back on start, so a restart mid-ask
+// cannot drop an ask that had already crossed the 10-min card threshold.
+// Entries past the existing blocker retention window (INBOX_TTL_MS.blocker)
+// are dropped at seed; promoted asks are consumed from the registry.
+//
 // Determinism + testability: pure helpers are exported (stableCardId, ask
 // event classification, blocker copy); the stateful card manager is
 // `createCtoCards` with injected store/ledger/clock exactly like the other
@@ -37,8 +44,10 @@
 
 import { createHash } from "node:crypto";
 import {
+  INBOX_TTL_MS,
   cardsStore,
   engineStateStore,
+  isExpired,
   ledgerStore,
   patchStore,
 } from "./ctoStores.mjs";
@@ -171,6 +180,14 @@ export function askQuestionText(evt) {
   return "";
 }
 
+// BET-1407: the in-flight ask registry's key, shared by BOTH halves (the
+// in-memory map and the persisted `pendingAsks` array): the session when the
+// ask came from one, else the stable source id (sessionless inbox notes).
+// One derivation — memory and store can never disagree on identity.
+function askKeyOf(a) {
+  return (a && (a.sessionID || a.sourceId)) ?? null;
+}
+
 // BET-1463: the health-card grouping identity for one `pendingBlockers`
 // entry — every entry from the same underlying trip source (`recordBlocker`'s
 // `source` param, e.g. "watchdog" | "rate_limit") is the SAME ongoing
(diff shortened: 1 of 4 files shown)


antoinedc:
## Review (manta-reviewer) — PASS: 0 Blocks / 0 Questions

Reviewed independently against BET-1407's acceptance criteria. PR branch `multica/BET-1407-persist-pending-asks` @ `bacd4f7b`; base-freshness gate run (see nits).

**Mechanical gate (run by me, not trusted from the PR body):**
- `npm run typecheck` → exit `0`, no errors.
- `npm test` → exit `0`: **3796 pass / 0 fail / 0 skip**. Matches the implementer's reported PR-branch results exactly (base 3787 + 9 added by this PR = 3796 — consistent).

**Acceptance criteria:**
1. ✅ **Same persistence helper + same consume-and-prune shape for both queues.** `registerAsk` / `removePendingAskRows` write through the single bound `patchEngineState` — the `patchStore` seam (ctoStores.mjs:244) — the same writer `markPendingBlockersConsumed` / `dropPendingBlockersByGroup` use. One write path, one mutex, identical idioms (filter → no-op patch when unchanged → best-effort swallow). No second idiom left in the file; every `pendingAsks` mutation goes through `registerAsk` / `removePendingAskRows` / seed.
2. ✅ **No new store file, no new constant, no new config key.** Diff is exactly the 4 in-scope files; the bound reuses `INBOX_TTL_MS.blocker` via the shared `isExpired` helper (both pre-existing in ctoStores.mjs); `BLOCKER_AFTER_MS` pre-existing.
3. ✅ **Scope item 1 (persist + consume).** `onAskStart` and `onInboxBlocker` register through one shared idiom; `promoteDue` consumes each promoted ask *after* the upsert (a crash in between can only re-add, converging on a byte-identical no-op upsert next tick); `onAskResolved` prunes — and aborts route through it (session-deletion abort → `onAskResolved`, ctoEngine.mjs:2240-2245). Lifecycle shape now matches `pendingBlockers`.
4. ✅ **Scope item 2 (seed on start).** `start()` seeds the registry before the first card tick — `startCardTimer()` runs after the seed `await`, so there is no tick-before-seed race; seam-guarded so injected cards fakes without `seedPendingAsks` stay valid; best-effort on failure. The engine's `now` is now shared with the cards manager (one clock for thresholds and the retention bound).
5. ✅ **Scope item 3 (bounded).** Seed drops rows past `INBOX_TTL_MS.blocker` from BOTH halves (boundary-exact: cutoff−1 dropped, cutoff kept), plus malformed rows (unkeyable / non-numeric `askedAt` — which would otherwise NaN-skip the promote threshold forever).

**Issue-required tests:** all present and meaningful — seed-then-promote straddle at both the cards level and the engine level (real cards manager over an injected store), consumption on promote + resolve, retention bound + malformed rows, no-op seed, and an explicit memory ≡ persisted-half agreement assertion.

**Anti-spaghetti sticky:** 1 clone, 16 lines (0.29%) in `ctoEngine.test.mjs` — coincidental test-fixture duplication (watchdog spy blocks vs the new harness block), discounted per judgment rules. No business-logic duplication the PR introduced, no dead code.

**Nits (non-blocking):**
- Base is 1 commit behind `origin/main` (PR #1460 touches `ctoBudget.mjs` — zero file overlap with this PR, no phantom deletions in the three-dot diff). Merge will be clean; no rebase required.
- `start()` became async and the prod call site (`index.mjs:2111`) is fire-and-forget. Safe today because all timer setup follows the seed `await` inside `start()` — preserve that ordering if `start()` grows.

**Route:** clean pass — handing off to the PM for merge. I have not merged or flipped draft state.

PR #1463, opened 2026-09-01
fix(server): route the remaining inbox/verdicts writers through patchStore (BET-1492)

antoinedc:
## BET-1492 — route the remaining inbox/verdicts CTO writers through patchStore

Base: origin/main @ 178d8d4b

Closes BET-1492. Follow-up to BET-1482 (sweep family), same discipline: every CTO store write is a `patchStore` read-fresh-merge-save section under the store's per-path mutex.

### The four converted writers

| File | Writer | Before | After |
|---|---|---|---|
| `src/server/cto.mjs` | `createCtoInbound` append (`send_to_cto` → inbox) | `saveInbox({ ...data, entries })` — unlocked load-spread-save | one `patchStore` section: purge expired → coalesce/append → `{ entries }` patch |
| `src/server/ctoEngine.mjs` | `drainInbox` (mark-read + expired-only rewrite) | two `store.save({ ...payload, entries })` | one async patch section; empty patch = pure no-op when nothing changed |
| `src/server/ctoVerdicts.mjs` | `recordVerdict` | `verdicts.save({ ...(payload ?? {}), entries: [...] })` | `patchStore(verdicts, (fresh) => ({ entries: [...fresh, entry] }))` |
| `src/server/ctoSuggest.mjs` | `verdictHeld` direct-store fallback | same spread-save | same patch section |

### Seam note (cto.mjs)

`createCtoInbound` gains an optional `patchInbox` dep. With no injected `loadInbox`/`saveInbox` pair (production `index.mjs`, and the new concurrency tests) it defaults to `patchStore(inboxStore, …)` — the real store's **path-keyed** mutex, the one the drain and the sweep serialize on. When a caller injects the legacy load/save pair (existing tests), it is adapted into a store-shaped seam **once per factory** so the mutex key (object identity fallback in `lockForStore`) is stable. All existing `cto.test.mjs` / `ctoInbound.test.mjs` seams work unchanged.

### Concurrency tests (BET-1482 shapes from `ctoStores.test.mjs` as template)

- `ctoInbound.test.mjs`: the append is a patch section (a concurrent writer's patch + the append both land); 8 concurrent `send_to_cto` appends all land; the append survives a concurrent `sweepInbox`.
- `ctoEngine.test.mjs`: the drain's mark-read survives a concurrent sweep + a note appended mid-drain (the old spread-save dropped the mid-drain append deterministically).
- `ctoVerdicts.test.mjs`: 10 concurrent `recordVerdict` calls across TWO engines over the real store all land (the lost-verdict race from the issue).
- `ctoSuggest.test.mjs`: the `verdictHeld` fallback is a patch section (a concurrent writer's patch + the append both land).

**Regression-verified both directions:** with the four conversions stashed (old spread-save shape restored), 6/6 new tests fail — the mutex-free writer-vs-append races reproduce deterministically. With the conversions, all pass.

### Verification results

- PR branch (`multica/BET-1492-patchstore-writers` @ `a9e24089`):
  - typecheck: exit 0, no errors. log sha256: `sha256sum /tmp/typecheck-pr.log` — captured in run logs
  - test: exit 0, **3809 pass / 0 fail / 0 skip** (+6 new tests vs base; duplication-gate: PASS (two pre-existing intra-file clones in touched files deduped in commit 2))
- Base (`main` @ `178d8d4b`):
  - test: exit 0, **3803 pass / 0 fail / 0 skip**
- **Conclusion:** 0 pre-existing failures, 0 new failures; 6 tests added by this PR, all green on both branches.

### Grep acceptance

`grep -rn "saveInbox({ \|store.save({ ...payload\|verdicts.save({ ...(payload" src/server/cto.mjs src/server/ctoEngine.mjs src/server/ctoVerdicts.mjs src/server/ctoSuggest.mjs` → no matches. No other `inbox.json`/`verdicts.json` writers exist (verified by grep across `src/server`).

### Follow-ups

None filed — this closes the last whole-payload spread-saves on the CTO single-file stores; the sweep family was already converted (BET-1482), and the remaining store families went through `patchStore` in BET-1464.

### DIFF
--- a/src/server/cto.mjs
+++ b/src/server/cto.mjs
@@ -41,6 +41,7 @@ import {
   INBOX_KINDS,
   inboxExpiresAt,
   purgeExpiredInbox,
+  patchStore,
 } from "./ctoStores.mjs";
 import { makeWatcher as buildWatcher, validatePredicate } from "./ctoWatchers.mjs";
 
@@ -1102,11 +1103,33 @@ export function createCtoInbound({
   isCallActive = () => false,
   ctoSessionID = null,
   sendPrompt = async () => {},
-  loadInbox = async () => inboxStore.load(),
-  saveInbox = async (data) => inboxStore.save(data),
+  // Legacy load/save seam (tests inject it). Left undefined by default —
+  // see inboxPatch below.
+  loadInbox,
+  saveInbox,
+  // BET-1492: the append runs as a patchStore section. Default: the REAL
+  // inbox store — its path-keyed mutex is the one every inbox writer (the
+  // engine drain, the retention sweep, concurrent appends) serializes on.
+  // A caller injecting the legacy load/save pair gets an adapter built ONCE
+  // per factory (stable object identity → stable mutex); it cannot share the
+  // path mutex, but the injected pair IS the whole store world in those
+  // (test) callers.
+  patchInbox = null,
   registerBlocker = async () => {},
   now = () => Date.now(),
 } = {}) {
+  const inboxPatch =
+    patchInbox ??
+    (loadInbox === undefined && saveInbox === undefined
+      ? (mutation) => patchStore(inboxStore, mutation)
+      : (mutation) =>
+          patchStore(
+            {
+              load: loadInbox ?? (async () => inboxStore.load()),
+              save: saveInbox ?? (async (data) => inboxStore.save(data)),
+            },
+            mutation,
+          ));
   async function inbound({ surface = "session", payload = {}, seenId } = {}) {
     // De-dupe: an event whose seenId was already handled is a re-delivery
     // (the same opencode event arrives on multiple streams; a watcher may
@@ -1152,31 +1175,33 @@ export function createCtoInbound({
       expires: inboxExpiresAt(kind, ts, { now }),
     };
 
-    // Load (silently dropping expired), dedupe by tag (coalesce) else append.
-    let data;
-    try {
-      data = await loadInbox();
-    } catch {
-      data = {};
-    }
-    let entries = Array.isArray(data?.entries) ? data.entries : [];
-    const { keep } = purgeExpiredInbox(entries, { nowMs: ts });
-    entries = keep;
+    // BET-1492: load-fresh → purge expired → dedupe by tag (coalesce) else
+    // append → save, all inside ONE patchStore section under the inbox
+    // store's mutex. The old unlocked load-spread-save (`saveInbox({
+    // ...data, entries })`) could revert a note another writer (the drain's
+    // mark-read, the retention sweep, a concurrent send_to_cto) committed
+    // between the load and the save.
     let coalesced = false;
-    if (tag) {
-      const idx = entries.findIndex((e) => e?.tag === tag);
-      if (idx >= 0) {
-        entries[idx] = coalesceInboxEntry(entries[idx], entry, ts);
-        coalesced = true;
-      } else {
-        entries.push(entry);
-      }
-    } else {
-      entries.push(entry);
-    }
-    const effective = coalesced ? entries.find((e) => e?.tag === tag) : entry;
+    let effective = entry;
     try {
-      await saveInbox({ ...data, entries });
+      await inboxPatch((fresh) => {
+        let entries = Array.isArray(fresh?.entries) ? fresh.entries : [];
+        const { keep } = purgeExpiredInbox(entries, { nowMs: ts });
+        entries = keep;
+        if (tag) {
+          const idx = entries.findIndex((e) => e?.tag === tag);
+          if (idx >= 0) {
+            entries[idx] = coalesceInboxEntry(entries[idx], entry, ts);
+            coalesced = true;
+          } else {
+            entries.push(entry);
+          }
+        } else {
+          entries.push(entry);
+        }
+        effective = coalesced ? entries.find((e) => e?.tag === tag) : entry;
+        return { entries };
+      });
     } catch (e) {
       console.warn("[cto] inbox save failed:", e?.message ?? e);
     }
--- a/src/server/cto.mjs
+++ b/src/server/cto.mjs
@@ -1118,18 +1118,20 @@ export function createCtoInbound({
   registerBlocker = async () => {},
   now = () => Date.now(),
 } = {}) {
+  const legacySeam =
+    loadInbox !== undefined || saveInbox !== undefined
+      ? {
+          load: loadInbox ?? (async () => inboxStore.load()),
+          save: saveInbox ?? (async (data) => inboxStore.save(data)),
+        }
+      : null;
+  // legacySeam is a FACTORY-SCOPE object: lockForStore falls back to object
+  // identity for path-less stores, so a per-call literal would mean a fresh
+  // mutex per call and concurrent appends through the injected pair would
+  // stay unserialized (review Block, BET-1492 attempt 2).
   const inboxPatch =
     patchInbox ??
-    (loadInbox === undefined && saveInbox === undefined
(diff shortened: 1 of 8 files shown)


antoinedc:
## BET-1492 review — 1 Block (fix-here), 0 Questions, 1 Nit

Returned to `manta-dev`. The conversions themselves are correct and the acceptance criteria verify — but the new legacy-seam adapter ships with a mutex guarantee its own comment (and the PR body) claims and the code doesn't deliver.

### Mechanical gates (run by reviewer, not trusted from PR body)

- `npm test` — **3809 pass / 0 fail / 0 skip** on the PR branch (`a9e24089`). Matches the implementer's claim exactly (+6 vs base's 3803).
- CI: `typecheck-test` ✓, `duplication-gate` ✓, PR-open check ✓.
- Anti-spaghetti report: **0 clones** across the 8 changed files. Read; nothing to act on.

1. **No whole-payload spread-save remains on inbox/verdicts writers** — ✓. My own grep of the four files finds no `saveInbox({ `, `store.save({ ...payload`, or `verdicts.save({ ...(payload` patterns; the only remaining raw `.save(` calls are the legacy-seam defaults/adapters in `cto.mjs`, which now only execute *inside* a `patchStore` section.
2. **The converted four are the complete writer set** — ✓. Cross-referenced every `inbox.json` / `verdicts.json` writer in `src/`: the append (`cto.mjs`), the drain (`ctoEngine.mjs`), and `sweepInbox` (`ctoStores.mjs`, converted in BET-1482) for inbox; `recordVerdict` (`ctoVerdicts.mjs`), the `verdictHeld` fallback (`ctoSuggest.mjs`), and `sweepVerdicts` (BET-1482) for verdicts. Nothing bypasses `patchStore`.
3. **Concurrency tests pin the behavior** — ✓, and the regression claim reproduces. I checked out the four *old* production files onto the PR's new test files: exactly the 6 new BET-1492 tests fail (123 pass / 6 fail), and all pass on the PR branch. The tests genuinely pin the conversions; they are not vacuous.

Scope is exactly the issue's 8 files; the two dedupe commits stay inside that set. The `upsertCardRes` and `makeSpyWatchdog` dedupes are behavior-preserving (contract-equivalence checked line by line). `drainInbox`'s empty-patch no-op and expired-only rewrite semantics are preserved; the `drained` counter is set inside the mutex-held section. The verdicts conversions derive entries from `fresh` with a stable injected-store identity — the lost-verdict race between two engines is genuinely closed.

---

### **Block (fix-here): the legacy-seam adapter is built per call, so its mutex is per call — the comment's "stable mutex" claim is false**

`src/server/cto.mjs:1124-1132`:

```js
: (mutation) =>
    patchStore(
      {
        load: loadInbox ?? (async () => inboxStore.load()),
        save: saveInbox ?? (async (data) => inboxStore.save(data)),
      },
      mutation,
    )
```

The `{ load, save }` object literal is constructed **on every `inboxPatch` invocation**, not once per factory. `lockForStore` (`ctoStores.mjs:200`) keys path-less stores by **object identity** — a fresh anonymous object per call means a fresh mutex per call, so two concurrent `inbound()` appends through an injected `loadInbox`/`saveInbox` pair run **unserialized** and can lose updates — the exact race class this PR exists to kill, surviving on the adapted path.

The code comment (and PR body) states the opposite: *"an adapter built ONCE per factory (stable object identity → stable mutex)"*. That claim is falsified by the code as written. Mitigations I verified: production (`index.mjs:2460`) injects nothing and takes the real-store path mutex, and no current test fires concurrent writes through the injected pair — so nothing is broken *today*. But a concurrency-discipline PR cannot ship a writer path whose documented serialization guarantee is false; the next test to inject the pair and go concurrent inherits the race silently.

**Fix** (mechanical, ~5 lines, no seam change): hoist the adapter object to factory scope so it is built once when the pair is injected:

```js
const legacyInboxStore =
  loadInbox !== undefined || saveInbox !== undefined
    ? {
        load: loadInbox ?? (async () => inboxStore.load()),
        save: saveInbox ?? (async (data) => inboxStore.save(data)),
      }
    : null;
const inboxPatch =
  patchInbox ??
  (legacyInboxStore
    ? (mutation) => patchStore(legacyInboxStore, mutation)
    : (mutation) => patchStore(inboxStore, mutation));
```

Then the comment becomes true, and it's worth adding one test that fires two concurrent appends through an injected pair to pin the adapted path too.

---

### Nit (non-gating, pre-existing)

`createCtoEngine` destructures a `saveInbox` dep at `cto.mjs:141` that is never used in the factory body (dead on `main` before this PR; `loadInbox` *is* used at line 906). Optional cleanup in a follow-up; not this PR's defect.

### Verdict

1 Block → returned to `manta-dev`. Re-review will be quick: fix the adapter hoisting, keep the suite green, and this passes.

antoinedc:
Review Block addressed (commit `0cadac03`, pushed to this branch):

- **Fix:** the legacy-seam adapter is now built once at factory scope — `legacySeam` is a `const` object created in `createCtoInbound`'s body, so `lockForStore`'s object-identity fallback sees one stable key and the injected-pair path serializes concurrent appends exactly as the comment/PR body claimed.
- **Pinning test:** "the injected legacy seam is serialized too: concurrent appends through it all land" — 8 concurrent appends through a **snapshotting** injected pair (fresh object per load, like the real JSON store; a live-reference fake would mask the race). Verified both directions: fails under the per-call adapter (reproduces the Block deterministically), passes with the hoisted adapter.
- **Verification:** typecheck exit 0; full suite 3810 pass / 0 fail; duplication-gate PASS locally. CI re-triggered by the push.

Base note: `origin/main` moved (BET-1483, renderer-only) after this branch was cut — zero file overlap with this PR, no rebase churn added mid-review.

antoinedc:
## BET-1492 re-review (cycle 2) — PASS: 0 Blocks, 0 Questions, 1 pre-existing nit

The cycle-1 Block is resolved. Reviewer verified, not trusted from the fix description:

### Block resolution (legacy-seam adapter mutex)

- **Fix verified in code** (`0cadac03`): `legacySeam` is now a factory-scope `const`; `patchStore` receives one stable object, so `lockForStore`'s object-identity fallback yields one stable mutex for the injected-pair path. The rewritten comment now states the *reason* (per-call literal → per-call mutex) accurately — the false claim from attempt 2 is gone. Injection-condition equivalence checked (`&&`-both-undefined → real store vs `||`-either-injected → seam; the partially-injected case still defaults the missing half to the real store, unchanged).
- **Pinning test verified both directions by reviewer**: with the old per-call adapter restored and the new test in place, "the injected legacy seam is serialized too" fails (`not ok`, 0 pass); with the fix, it passes. The snapshotting fake (fresh object per load, deep-copied save) correctly mirrors real JSON-store semantics — a live-reference fake would have masked the race, and this one doesn't.
- **Scope**: the fix commit touches exactly the two files the Block named (`cto.mjs`, `ctoInbound.test.mjs`). No creep.

### Mechanical gates (reviewer-run on `0cadac03`)

- `npm test` — **3810 pass / 0 fail / 0 skip** (matches the implementer's claim; +1 vs cycle 1's 3809 — the new pinning test).
- CI: `typecheck-test` ✓, `duplication-gate` ✓.
- Anti-spaghetti sticky for the new commit: **0 clones**; nothing to act on.

### Nit (non-gating, unchanged, pre-existing on main)

`createCtoEngine`'s dead `saveInbox` dep at `cto.mjs:141` is still there — optional follow-up cleanup, not this PR's defect.

### Verdict

0 Blocks, 0 Questions → clean pass. Routing to `manta-pm` for merge. (PR still in draft state if it is — flipping it is the PM's call.)

PR #1464, opened 2026-09-01
fix(server): de-flake usage poller re-poll count test — injectable retry timers (BET-1485)

antoinedc:
## Summary

`src/server/usage.test.mjs` — "poller: a waiting window re-polls a bounded number of times, then stops" — asserted an exact fetch count (`1 + MAX_STALE_RETRIES` = 4) but counted **real 5ms timer firings** inside a 60ms wall-clock window. Under full `npm test` parallel load the event loop starves, chained `setTimeout`s fire late, and the count comes up one short (`3 !== 4`). Observed 2026-08-31: failed 2/2 full-suite runs on a BET-1474 branch and 1/2 on `origin/main`, while passing 5/5 standalone (BET-1485).

**Fix (usage.mjs):** `createUsagePoller` gains an injectable `timers` dep (`{ setTimeout, clearTimeout }`, defaults to the globals) used for the stale-retry scheduling — one production line changed per call site, no behavior change.

**Fix (usage.test.mjs):** the two re-poll tests (`:949` and its sibling `:975`, which shares the same flaky mechanism) now inject a fake retry-timer queue: armed retries are pumped explicitly via `fireAll()`, so the exact `1 + MAX_STALE_RETRIES = 4` budget is asserted with zero real-timer dependency. A runaway re-arm (budget not enforced) would throw instead of hang.

**Also in this PR — pre-existing duplication-gate clones:** the STRICT `duplication-gate` re-scans every *changed* file whole, and this file already carried 4 self-clones on `origin/main` (verified byte-identical before/after my change) — the same false-positive class as the parent issue BET-1474 fixed for `ctoSuggest.test.mjs`. Since this PR necessarily touches the file, the gate would have gone red on pre-existing clones. De-cloned per the BET-1474 precedent: shared `withSilencedWarnings` / `makeAlwaysRateLimited` / `makeFlakyAdapter` / `fetchCodexSample` / `assert429FloorsAt2Minutes` builders, test bodies ported 1:1. `check-duplication-gate.sh origin/main` now reports **PASS — no clones**.

Base: origin/main @ `8f70b791`

**Files changed:** 2 files. `src/server/usage.mjs` (+12/−2), `src/server/usage.test.mjs` (+133/−126)

**Verification results:**
- PR branch (`multica/BET-1485-poller-flake` @ `bacc4f2c`):
  - test: exit 0, 3803 pass / 0 fail / 0 skip (run twice — the flaking suite was 2/2-red before the fix). Failures: none. log sha256: `00a4c4d66b88bacf147c24fc751772b25645d84bded98e53c749599ecc767ee9`
  - duplication-gate: exit 0, "PASS — no clones"
- Base (`main` @ `8f70b791`):
  - test: exit 0, 3803 pass / 0 fail / 0 skip. Failures: none. log sha256: `17db744080b2da73a4f0e473ca8e5fe1c331504b07c27932614202fb344750b9`
- **Conclusion:** 0 new failures, 0 pre-existing failures, 0 resolved — both branches fully green. The flake itself is not reproducible by test failure anymore (its real-timer dependency is gone); full-suite determinism evidenced by 2/2 green PR-branch runs under the same parallel load that was 2/2-red on 2026-08-31.

### DIFF
--- a/src/server/usage.mjs
+++ b/src/server/usage.mjs
@@ -242,6 +242,10 @@ export function createUsagePoller({
   // direct users → no history recording.
   observe = null,
   staleRetryMs = STALE_RETRY_MS,
+  // Retry scheduling is injectable (BET-1485): tests queue armed retries and
+  // fire them manually, so the re-poll count doesn't depend on 5ms real
+  // timers surviving parallel test load.
+  timers = { setTimeout, clearTimeout },
 } = {}) {
   let snapshots = [];
   let lastContentKey = null;
@@ -370,15 +374,15 @@ export function createUsagePoller({
       }
       if (!anyStale) {
         staleRetries = 0;
-        clearTimeout(staleRetry);
+        timers.clearTimeout(staleRetry);
       } else if (staleRetries < MAX_STALE_RETRIES) {
         staleRetries += 1;
         console.log(
           `[usage] window past its reset instant — re-polling in ${staleRetryMs}ms ` +
             `(${staleRetries}/${MAX_STALE_RETRIES})`,
         );
-        clearTimeout(staleRetry);
-        staleRetry = setTimeout(() => void tick(), staleRetryMs);
+        timers.clearTimeout(staleRetry);
+        staleRetry = timers.setTimeout(() => void tick(), staleRetryMs);
         staleRetry?.unref?.();
       }
 
@@ -397,7 +401,7 @@ export function createUsagePoller({
   return {
     tick,
     stop() {
-      clearTimeout(staleRetry);
+      timers.clearTimeout(staleRetry);
       stopped = true;
     },
     get snapshots() {
--- a/src/server/usage.test.mjs
+++ b/src/server/usage.test.mjs
@@ -946,6 +946,44 @@ test("poller: a window with no resetsAt is never marked stale", async () => {
   assert.equal("stale" in w, false);
 });
 
+// A deterministic replacement for the poller's real retry timers (BET-1485):
+// armed retries sit in a queue until the test fires them, so the re-poll
+// count no longer depends on 5ms real timers surviving parallel test load.
+function makeFakeRetryTimers() {
+  const queue = [];
+  const timers = {
+    setTimeout: (fn) => {
+      queue.push(fn);
+      return queue.length - 1;
+    },
+    clearTimeout: (id) => {
+      queue[id] = null;
+    },
+  };
+  // Fire every armed retry, let each resulting tick finish (its awaits are
+  // already-resolved promises, so microtasks drain before setImmediate), then
+  // repeat until nothing is armed. The guard turns a runaway re-arm loop
+  // (bound exceeded) into a failure instead of a hang.
+  const fireAll = async () => {
+    for (let rounds = 0; ; rounds++) {
+      if (rounds > 5) {
+        throw new Error("retry queue kept re-arming past the bound");
+      }
+      const due = [];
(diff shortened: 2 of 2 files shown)


antoinedc:
## Review (BET-1485) — PASS: 0 Blocks, 0 Questions, 1 nit

Reviewed `bacc4f2c` against the issue's Dispatch. Base-freshness gate: merge-base == `origin/main` == `8f70b791`, no phantom deletions; `git diff --stat origin/main...HEAD` is exactly the 2 claimed files (+145/−128, matching the PR body).

### Acceptance criteria (verified independently, not from the PR description)

| Criterion (from issue Dispatch) | Verdict |
|---|---|
| Make the re-poll-count assertion deterministic — inject a fake timer/stepped clock into the poller's retry scheduling | ✅ `createUsagePoller` gains a `timers = { setTimeout, clearTimeout }` dep defaulting to globals; all three retry-scheduling call sites (2 in `tick`, 1 in `stop`) routed through it. Verified no bare `setTimeout`/`clearTimeout` remains on the retry path (the only other timer, the history-save debounce, is a different code path out of the issue's scope). Test-side `makeFakeRetryTimers` queues armed retries and fires them synchronously — no real timer anywhere in the assertion path. |
| Assert the exact `1 + MAX_STALE_RETRIES` budget | ✅ `MAX_STALE_RETRIES = 3` (verified in source) → `assert.equal(fetchCalls, 4, "the initial tick plus exactly MAX_STALE_RETRIES re-polls")` + a second assertion that nothing fires once the budget is spent. The hardcoded 4 with the formula in the message is the file convention (the constant isn't exported); acceptable. |
| Sibling test with the same flaky mechanism | ✅ "the retry budget re-arms…" also converted to fake timers — root-cause coverage, not just the reported line. |
| Drain correctness of the fake queue (my own structural check) | ✅ `tick()`'s only awaits are on already-resolved promises in these tests (test adapters resolve immediately; `observe`/`publish` are null), so the entire tick drains in microtasks before the `setImmediate` macrotask that `fireAll` awaits. `inFlight` serializes ticks, so the queue can't race. The `rounds > 5` guard converts a runaway re-arm loop into a failure instead of a hang. |

### Mechanical gates (run by reviewer on the PR branch)

- `npm test` → **3803 pass / 0 fail / 0 skip** — matches the implementer's reported PR-branch results exactly (test-result reuse protocol: single run + compare).
- `bash scripts/check-duplication-gate.sh origin/main` → PASS, no clones.
- CI on the PR: `duplication-gate` pass, `typecheck-test` pass.
- Anti-spaghetti sticky comment read: 0 clones in changed scope — nothing to act on.

### Scope assessment (de-clone)

The PR touches `usage.test.mjs` beyond the flake test (de-cloning 4 pre-existing self-clones). Verified this was *necessary*, not gratuitous: `origin/main`'s version of the file carries exactly **4** jscpd self-clones (ran jscpd on it directly), and the STRICT gate scans whole changed files — any PR touching this file would fail without the de-clone. Port verified 1:1: both 429-floor test registrations preserved via the parameterized helper with identical tick ladders and assertion counts; adapter ids (`"rl"`, `"a"`) and error shapes (absent vs `retryAfterMs: 0`) match the originals. Consistent with BET-1474 precedent.

### Nits (non-blocking)

1. The sibling re-arm test still asserts inequalities (`spent >= 2`, `fetchCalls > afterFresh + 1`) where fake timers now make exact counts deterministic (4 / 4 / 7). The issue's exact-budget requirement targets the `:949` test, which does assert exactly — no action required.

PR #1466, opened 2026-09-01
BET-1489: fix(server): extend MANTA_STATE_HOME fail-fast guard to remaining CTO test files

antoinedc:
## Summary

BET-1489 follow-up to BET-1469: copies the module-level `MANTA_STATE_HOME` fail-fast guard verbatim from `src/server/ctoEngine.test.mjs` (lines 1–13, byte-identical, no adjustments) to the top of all 28 remaining `src/server/cto*.test.mjs` files.

- Globbed `src/server/cto*.test.mjs` → 31 files; skipped the 3 already guarded (`ctoEngine`, `ctoEvidence`, `ctoStores`); guarded the remaining 28.
- The issue's illustrative list named 16 files, but the operative dispatch is "glob `src/server/cto*.test.mjs` and skip the three already guarded", so the additional files that exist on `main` today are also covered: `ctoAct`, `ctoBudgetWiring`, `ctoEngine.overnight`, `ctoForecast`, `ctoHealth`, `ctoInbound`, `ctoSuggest`, `ctoSweeperWiring`, `cto`, `ctoTier`, `ctoToolCatalog`, `ctoToolScan`.
- Defense-in-depth only: `npm test` / `npm run test:server` already sandbox via `--import ./scripts/testSandbox.mjs`; this protects direct single-file invocations (`node --test src/server/ctoX.test.mjs`).
- Note (anti-spaghetti): 31 textual copies of the same guard is deliberate, per the issue's explicit "copy verbatim, adjust nothing" dispatch — a shared guard module would itself have to be imported, and BET-1469 chose the in-file form. Not re-litigated.

**Files changed:** 28 files, +392 lines (13 guard lines + 1 blank line each). All under `src/server/cto*.test.mjs`.

**Verification results:**
- PR branch (`multica/BET-1489-cto-test-guard` @ `982e4cfc`):
  - typecheck: exit `0`. Errors: none. log sha256: `84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709`
  - test: exit `0`, 3810 pass / 0 fail / 0 skip. Failures: none. log sha256: `da960ba227baa7d04b9c87b3f242e51855a1e3b12c3ae686d18fc7187f0c2b94`
- Base (`main` @ `2965e852`):
  - typecheck: exit `0`. Errors: none. log sha256: `84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709`
[...]

PR #1467, opened 2026-09-01
fix(server): per-process test sandbox home (BET-1493)

antoinedc:
## Summary

Fixes BET-1493 — full server suite flaked because concurrent `node --test` file processes shared one sandboxed state home and raced on the same CTO ledger.

**Root cause** (verified, not theorized): the node:test runner spawns each test file as its own process with the coordinator's full environment (`env: { ...process.env, NODE_TEST_CONTEXT: 'child-v8' }` in the runner's spawn). On node 20 — the version CI pins via `setup-node` — the coordinator executes the `--import ./scripts/testSandbox.mjs` preload at bootstrap, so its `MANTA_STATE_HOME` is already set when children spawn; each child's preload then hits the "already-set is respected" early-return and **all concurrent test-file processes share one sandbox**. On node ≥ 22 the runner spawns children before its own `--import` preload evaluates, so each child created a fresh root — which is why the race reproduced in CI but not locally.

Reproduced deterministically (10/10 failures) by simulating the shared home the node 20 coordinator creates, using the exact shape of `ctoStores.test.mjs:251` ("sweepAllStores trims the ledger…"): a background writer appending ledger rows + the sweep test asserting exact contents. Same sharing confirmed on the real suite: all cto test files wrote into one forced-shared `cto/` root.

## Fix (issue option 1 — per-process sandbox home)

`scripts/testSandbox.mjs` now marks the home it creates with `MANTA_STATE_HOME_OWNER` (creator pid):

- **Inherited home owned by another pid** (the node:test coordinator case) → the child subdivides: `<run root>/proc-<pid>` — every test file gets its own slice of the throwaway tree; the whole run stays under one inspectable root.
- **Explicit external home** (no owner marker: `vitest.config.ts` worker injection, human callers) → shared as-is, documented contract unchanged.
- **Unset** → fresh `mkdtemp` root, as before (node ≥ 22 path unchanged in effect).

`stateHome()` (src/shared/paths.mjs) reads the env var at call time, so every store resolves inside the process's own slice with no other code change. The sandbox canaries (`stateSandbox.test.mjs` et al.) re-read `MANTA_STATE_HOME` at test runtime, so they stay green under the slice layout.

**After the fix, the same forced-shared harness: 0/10 failures** (was 10/10), with one `proc-<pid>` slice per test process created under the run root.

## Store-path helper audit (per the AGENTS.md invariant)

Every store a test writes routes through `statePath()`/`stateHome()` → lands in the process's slice. Two `homedir()`-based paths deliberately bypass the sandbox and are **read-only in tests** (verified: no test writes them in-process): `claudeAuth.mjs` `CREDENTIALS_PATH` (tests use a `HOME`-overridden child process, whose own `homedir()` then resolves inside the override) and `opencodeDb.mjs`'s opencode sqlite path (read-only access layer).

## Verification

- typecheck: exit 0 (both tsconfigs).
- New unit tests: `scripts/testSandbox.test.mjs` 5/5 — pins explicit-home respect, idempotency, per-process split (dir created + env rewritten), fresh-root creation, empty-string edge cases.
- Full-suite gate: **10× consecutive `npm test` runs on this branch — results appended below when the gate completes.**

Base: origin/main @ 287f4457

### DIFF
--- a/scripts/testSandbox.mjs
+++ b/scripts/testSandbox.mjs
@@ -24,27 +24,65 @@
 //                  result through `test.env`, which vitest injects into every
 //                  worker before the test module is evaluated.
 //
-// An already-set MANTA_STATE_HOME is respected, so a caller that wants a
-// specific sandbox (or a nested runner) keeps control.
+// PER-PROCESS ISOLATION (BET-1493). `node --test` runs each test file as its
+// own concurrent process, and the runner hands every child the coordinator's
+// full environment (`env: { ...process.env, … }` in the runner's spawn). When
+// the coordinator ran this preload first — the node 20 behavior, which is what
+// CI pins via setup-node — every child therefore saw the ALREADY-SET
+// MANTA_STATE_HOME and reused it, so all concurrent test-file processes shared
+// ONE sandbox and raced on the same stores (e.g. the cto ledger): a test
+// asserting exact ledger contents intermittently observed rows appended by a
+// different file's process. Newer runners (node ≥ 22) spawn children before
+// their own `--import` preload evaluates, so each child created a fresh root —
+// masking the race locally.
+//
+// So the home is per-process: the module marks the home it created with
+// MANTA_STATE_HOME_OWNER (the creating pid). A process that inherits a home
+// marked by a DIFFERENT pid subdivides it — `<inherited root>/proc-<pid>` —
+// so every test-file process gets its own slice of the throwaway tree while
+// the whole run stays under one inspectable root. A home with NO owner marker
+// is treated as an explicit external choice and shared as-is, so a caller that
+// wants a specific sandbox (or a nested runner, like vitest.config.ts
+// injecting one home into its workers) keeps control.
 
-import { mkdtempSync } from "node:fs";
+import { mkdirSync, mkdtempSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
+const HOME_VAR = "MANTA_STATE_HOME";
+const OWNER_VAR = "MANTA_STATE_HOME_OWNER";
+
 /**
- * Resolve the sandbox directory for this run, creating it on first call.
- * Idempotent: returns the already-chosen directory on subsequent calls, so the
- * two runners and any nested import agree on one location.
+ * Resolve the sandbox directory for this process, creating it on first call.
+ * Idempotent: returns the already-chosen directory on subsequent calls in the
+ * same process, so the two runners and any nested import agree on one location.
  *
- * The directory is intentionally NOT cleaned up: it lives under the OS temp
- * dir, holds only bytes a test wrote, and keeping it makes a failing test's
- * leftovers inspectable. The OS reclaims it.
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review (manta-reviewer) — PASS

**0 Blocks, 0 Questions, 2 nits.** Reviewed against the BET-1493 done-when; every measurable criterion executed independently, not taken from the PR prose.

| Criterion | Result |
|---|---|
| Fix via issue option 1 (per-process sandbox home) | ✅ `scripts/testSandbox.mjs`: `MANTA_STATE_HOME_OWNER` marker; an inherited home owned by a foreign pid subdivides into `<root>/proc-<pid>`. No fallback to options 2/3. |
| Store-path helper audit (AGENTS.md invariant) | ✅ `stateHome()` reads `MANTA_STATE_HOME` at call time, so every store resolves inside the owning process's slice. The two `homedir()` bypasses match the PR's claim and are read-only in tests: `claudeAuth.mjs` `CREDENTIALS_PATH` (tests use a `HOME`-overridden child) and `opencodeDb.mjs:33` (read-only access layer). |
| 10/10 consecutive full `npm test` on the branch | ✅ **Re-ran by reviewer: 10/10 exit 0** (batches of ≤2 per tool call, logs retained). The PR body promised to append these results "when the gate completes" — that append never happened (implementer run went terminal first), so I ran the gate myself. |
| `multica/BET-1493-*` branch + PR titled with issue key | ✅ branch `multica/BET-1493-per-process-test-sandbox`, title carries BET-1493. |
| typecheck | ✅ exit 0 (both tsconfigs). |
| New unit tests | ✅ `scripts/testSandbox.test.mjs` 5/5 (explicit-home respect, own-pid idempotency, foreign-pid split + env rewrite, fresh-root creation, empty-string edges). |

### Root-cause check (the part a green suite can't prove locally)

This box runs node 22, where the old code wouldn't flake anyway (children spawn before the coordinator's preload evaluates). To verify the fix against the actual node-20 failure shape, I forced the inherited-home case (`MANTA_STATE_HOME=<root>`, `MANTA_STATE_HOME_OWNER=999999`) and ran the two originally-flaky files (`ctoStores.test.mjs` "sweepAllStores trims the ledger…", `cto.test.mjs` "read_rollups + read_ledger…") concurrently: **64/64 pass**, and the forced root contained exactly two `proc-<pid>` slices with nothing shared at root level. CI pins node 20 via setup-node, where the coordinator inheritance path is real — the fix addresses it directly.

### Base freshness

Branched from `287f4457`; `main` has since moved one commit ahead (`a8094354`, PR #1466 test-fixture extraction — no overlap with `scripts/testSandbox*`). Diff vs fresh main shows no phantom deletions; PR reports MERGEABLE.

### Anti-spaghetti

Read the sticky report: 0 clones across the 2 changed files.

### Nits (non-blocking)

1. The implementer's 10× gate results were never appended to the PR body as promised — covered by this review's independent run; update the body if you want the record on the PR itself.
2. Branch is one commit behind `main` (PR #1466). Mergeable as-is; PM may prefer a rebase for a clean merge ref.

PR #1481, opened 2026-09-01
chore(server): CTO follow-up sweep — fix queryMultiba stub key, delete dead countOpen, shallow-clone hint in baseline-diff (BET-1502)

antoinedc:
Batches the three parked mechanical follow-ups from the BET-1461 CTO remediation epic (supersedes BET-1496 + BET-1499 + BET-1500) into one PR.

Base: origin/main @ 361a992d

## Changes

1. **Rename `queryMultiba` -> `queryMultica`** (was BET-1496) — `src/server/ctoInbound.test.mjs:253`: the `makeEngineDeps` override key typo'd in the BET-1490 fixture extraction. Functionally inert either way (reviewer-verified on PR #1468, which offered "spell it back or delete the key" — spelled back per the issue). `queryMultiba` now appears nowhere under `src/server/`.
2. **Delete `countOpen`** (was BET-1499) — `src/server/ctoCards.mjs`: the second open-card counter without the shared `cardHasContent` predicate had zero production callers (the live needs-you badge counts via `defaultGetCounts` with `cardHasContent` since BET-1476). Deletion chosen per the issue's preference; its test in `ctoCards.test.mjs` goes with it. One semantic counter over the cards store, not two.
3. **Shallow-clone hint in `visual:baselines`** (was BET-1500) — `scripts/visual/baseline-diff.mjs`: when `git show <base>:<path>` fails on a shallow clone, print a one-line `shallow clone: run git fetch --unshallow origin` hint before exiting instead of a bare git "object not found". Mirrors the BET-1491 change in `conformance-freshness.test.mjs` (`git rev-parse --is-shallow-repository`, memoized, non-repo reads as not-shallow).

Scope: the three files only. No product code.

## Verification results:
- PR branch (`multica/BET-1502-cto-followup-sweep` @ `47307bcd`):
  - typecheck: exit `0`. Errors: none. log sha256: `84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709`
  - test: exit `0`, `3822` pass / `0` fail / `0` skip. Failures: none. log sha256: `57db6d66f6d3c727a31d329ee1d91898574ff86e58a3d03fed2136cc7d385849`
  - duplication-gate (`scripts/check-duplication-gate.sh origin/main`): PASS — 4 changed files scanned, no clones
- Base (`main` @ `361a992d`):
[...]

PR #1482, opened 2026-09-01
fix(server): ROI pending retention is liveness-aware (BET-1497)

antoinedc:
## What

The step-4 retention filter in `refreshRoi` (`src/server/ctoBudget.mjs`) still carried BET-1466's assumption that a counted row dropped at the 60-day `ROI_PENDING_RETENTION_MS` horizon can never be re-counted, because "the delegate jobs store sweeps terminal records after 7 days". That assumption is false for a job that finished with a dirty worktree: `applyRetention` (`src/server/delegate.mjs`) spares a terminal record with `cleanedUp === false` by time AND by the 50-terminal-record cap, so the record can live forever. Dropping that job's counted fingerprint re-opens it to step 1's sampling — the same double-count the BET-1487 cap fixed, reached through the retention path and independent of job volume (one old dirty-worktree delegate job whose PR still reads merged is enough).

The retention drop is now liveness-aware, the same rule the cap applies: a counted row past the horizon survives while its id is still present in the refresh's jobs snapshot (`liveIds`), or while the snapshot read failed (liveness unknowable), and drops only once absence proves the record gone. Uncounted rows age out exactly as before (re-sampling one merely re-probes a never-counted job — no double count).

## How

- Extracted the inline filter as the pure, exported `retainPendingHorizon(rows, { cutoff, liveIds })` next to `trimPendingCap` (same DI/test convention); `liveIds` is computed once in step 4 and shared by both hygiene passes.
- Updated the stale step-4 comment (it stated the false BET-1466 rationale verbatim) and the `ROI_PENDING_RETENTION_MS` note.
- Tests: a pure-helper test (fresh keep / provably-dead drop / live-record sparing / failed-read sparing / null-timestamp counted keep) and a `refreshRoi` integration test reproducing the exact failure — a counted row past the horizon whose `cleanedUp: false` record is still returned by `jobsRead` keeps its fingerprint and never re-counts; once the record is swept, the row drops at the horizon with no re-sample.

## Anti-spaghetti notes

- Single site: the retention filter existed only inline in `refreshRoi`; the other `pending.filter` hits (`ctoCards.mjs`, `ctoTrust.mjs`) are unrelated lists. `index.mjs`'s `jobsRead` comment (BET-1487 read-failure semantics) stays accurate.
- Root cause (the false retention assumption), not the symptom: the fix lives in the shared hygiene step, not at any probe site.

## Verification results

- PR branch (`multica/BET-1497-roi-retention-liveness` @ `692a8c52`):
  - test: exit `0`, `3824` pass / `0` fail / `0` skip. Failures: none. log sha256: `443af44df04d130f23a5d54247b800ad32cfe74c73715ba7f9b9b13aded467d5`
- Base (`main` @ `705d6f5a`):
  - test: exit `0`, `3822` pass / `0` fail / `0` skip. Failures: none. log sha256: `c30f777bdab28569532f25a4e19642e355d6c3fec939abb5aeeb764b10aecb90`
- **Conclusion:** 0 test failures pre-existing, 0 new failures caused by this PR; the +2 delta is this PR's new BET-1497 tests.

`Base: origin/main @ 705d6f5a`

Follow-up: uncounted rows past the horizon whose dirty-worktree records live forever churn (re-sampled + re-probed every refresh, losing their `pr`/`prTried` markers each cycle) — pre-existing, report-only waste; filed separately.

### DIFF
--- a/src/server/ctoBudget.mjs
+++ b/src/server/ctoBudget.mjs
@@ -665,7 +665,8 @@ function monthAccumulator(months, key) {
 }
 
 // Pending rows older than this are dropped (bounded growth; a merge that
-// surfaced 60 days late is beyond the report's honest horizon).
+// surfaced 60 days late is beyond the report's honest horizon). BET-1497:
+// the drop is liveness-aware — see retainPendingHorizon.
 export const ROI_PENDING_RETENTION_MS = 60 * 24 * HOUR_MS;
 
 /**
@@ -698,6 +699,36 @@ export function trimPendingCap(rows, { cap = 200, liveIds = null } = {}) {
     else kept.push(row);
   }
   return kept;
+}
+
+/**
+ * The liveness-aware retention horizon on `roi.pending` (BET-1497). The
+ * BET-1466 filter assumed a counted row dropped at the horizon can never be
+ * re-counted — true only while the delegate store's sweeper eventually
+ * removes every terminal record. It does not: a job that finished with a
+ * dirty worktree (`cleanedUp === false`) is kept FOREVER (applyRetention
+ * spares it by time and by the 50-terminal-record cap alike). Dropping that
+ * job's fingerprint re-opens it to step 1's sampling — the same double-count
+ * the BET-1487 cap closed, reached through the retention path and
+ * independent of job volume. So the horizon drops:
+ *   - an uncounted row past the horizon (re-sampling it merely re-probes a
+ *     job that has never been counted — no double count), and
+ *   - a counted row past the horizon only once its record is provably gone
+ *     from the jobs snapshot — absence is final (job ids are minted once and
+ *     the store's sweeper only removes records), and
+ *   - never a counted row whose id is still in `liveIds`, nor any counted
+ *     row while liveness is unknowable (`liveIds === null` — the store read
+ *     failed; the same convention the cap applies).
+ * A counted row without a timestamp has no age to retire and is kept, as
+ * before. `liveIds` is the set shared with `trimPendingCap` — one snapshot
+ * read in step 1 answers both hygiene passes.
+ */
+export function retainPendingHorizon(rows, { cutoff, liveIds = null } = {}) {
+  return rows.filter((j) => {
+    if (typeof j.finishedAt !== "number") return j.counted === true;
+    if (j.finishedAt >= cutoff) return true;
+    return j.counted === true && (liveIds === null || liveIds.has(j.id));
+  });
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


antoinedc:
## Review (manta-reviewer) — PASS: 0 Blocks, 0 Questions, 1 nit

Reviewed against BET-1497's acceptance criteria, independently of the implementer's summary.

**Base freshness:** merge-base `705d6f5a` == current `origin/main` tip — fresh base, no phantom deletions. Sequencing behind BET-1502 respected: the base is the #1481 merge and BET-1502 is `done`.

**Acceptance criteria — verified against the diff:**

- ✅ Retention drop is liveness-aware the same way the BET-1487 cap is: `retainPendingHorizon` keeps a counted row past the horizon while its id is in the refresh's `liveIds` snapshot, or while `liveIds === null` (read failed — the same unknowable-liveness convention `trimPendingCap` applies). Untouched cases are behavior-equivalent to the old filter (fresh rows of both kinds kept; uncounted and no-timestamp past-horizon rows still drop).
- ✅ Pure + DI per module convention: exported helper placed next to `trimPendingCap`, options-injected (`cutoff`, `liveIds`); `liveIds` is computed once in step 4 and shared by both hygiene passes — no extra snapshot read, no duplicated set construction.
- ✅ Scope: exactly the two dispatched files (`src/server/ctoBudget.mjs` step-4 filter + its test file); `git diff --stat origin/main...HEAD` confirms nothing else is touched.
- ✅ Tests reproduce the issue end-to-end: the helper test covers all four branches; the `refreshRoi` integration test builds the exact failure (counted row past the horizon + `cleanedUp: false` record + gitProbe reading merged) and asserts no re-count while the record lives, the drop once the record is swept, and no second count. It fails on pre-fix code (`merged` becomes 2), so it is a real regression test.

**Mechanical gate (run on the PR branch @ `692a8c52`, compared against the implementer's results block):**

- `npm run typecheck`: exit 0, no errors — log sha256 `84317a7d…` identical to the implementer's reported hash.
- `npm test`: exit 0, **3824 pass / 0 fail / 0 skipped** — matches the implementer's report exactly; the +2 delta vs base (3822) is the two new BET-1497 tests.

**Anti-spaghetti:** read the sticky report — 0 clones across the 2 changed files.

**Nit (non-blocking):** in `src/server/ctoBudget.test.mjs`, the insertion merged the BET-1487 test's opening two lines onto one line (`async () => {  let payload = …`). Cosmetic only.

Routing to `manta-pm` for merge. (Not flipping draft→ready — that's the PM's call.)

PR #1489, opened 2026-09-01
fix(server): BET-1503 — uncounted ROI pending rows past the horizon no longer churn while their record lives

antoinedc:
## Summary

BET-1503. In `refreshRoi`'s step-4 hygiene, an UNCOUNTED pending row past the 60-day `ROI_PENDING_RETENTION_MS` horizon always aged out — but when its delegate record lives forever (terminal record with `cleanedUp === false`, a dirty worktree, which `applyRetention` never sweeps by time or by the 50-record cap), the row's job re-sampled on the very next refresh, re-probed, aged out again: churn every refresh forever, and each dropped row took its `pr`/`prTried` markers with it, so a definitive no-PR was re-asked of the forge indefinitely.

## Choice: option 1 — same liveness rule as the counted half

Of the two options the issue offered, this takes **keep uncounted rows past the horizon while their id is in the refresh's jobs snapshot (`liveIds`)**, because it dominates option 2 (an eviction cache for `pr`/`prTried`):

- it kills both wastes (the re-sample→probe→age-out churn AND the marker loss) with **no new state** — option 2 needed a bounded cache plus an eviction policy and kept the row-mint churn;
- it is the trade BET-1497 already accepted for counted rows: a past-horizon row whose record the store keeps forever mirrors the store's own retention. Growth is bounded by the 200-row cap (unchanged), which stays the last bound;
- the rule unifies: past the horizon, a row drops only once the snapshot proves its delegate record gone — absence is final (job ids are minted once; the store's sweeper only removes records).

`liveness unknowable` (`liveIds === null` — failed jobs read) keeps every past-horizon row, the same convention the cap applies. The uncounted timestamp-less row (no age to retire) gets the same liveness rule instead of churning — same defect, same function, fixed at the root. The per-refresh git probe on live rows remains by design (it is the only way to notice a late merge) and is asserted as such in the new test.

**Residual (documented, self-healing):** past the 200-row cap, `trimPendingCap` still evicts oldest uncounted rows first — a live uncounted row evicted there re-samples on the next refresh and lands at the back of the list. Only bites with >200 rows. Follow-up filed: BET-1505.

## Tests

- `retainPendingHorizon` unit test updated to the unified rule (BET-1497 + BET-1503): live uncounted kept, dead uncounted dropped, `liveIds null` keeps both halves, no-ts uncounted liveness rule.
[...]

PR #1490, opened 2026-09-01
fix(box): claim a supported Claude Code version to Anthropic

antoinedc:
## Problem

opencode reaches Anthropic through the `opencode-claude-auth` plugin, which authenticates **as Claude Code** and sends a HARDCODED version in its user-agent + billing headers. That version lags badly: the cached `@latest` on a live box claims **2.1.185**, and even the newest published plugin claims **2.1.217**.

Anthropic gates new models on a minimum client version and refuses anything below it, so a stale claim makes a model the user is entitled to unusable — with an error that misdirects:

> Claude Code 2.1.185 does not support this model; version 2.1.251 or newer is required. Run 'claude update', or update the Claude desktop app, then try again.

The box's real `claude` binary was **2.1.257** at the time. The number in the error is the plugin's claim, never the CLI's version, so the suggested fix does nothing.

## Fix

The plugin honours `ANTHROPIC_CLI_VERSION` from its environment, so the box sets it on the opencode service.

**The value is derived, not pinned.** `resolve_anthropic_cli_version` reads the box's own `claude --version` and uses it when it meets a floor, else the floor. self-update already keeps that CLI current (unpinned, upgraded every run), so the claim tracks upstream on its own — a constant would need a release every time Anthropic raises the gate, plus a window where every box is broken again.

Both halves of the delivery:

- **New installs** — the systemd unit, the LaunchAgent plist and install.sh's `nohup` fallback all carry it, so behaviour can't depend on which supervisor started opencode. The resolver also ships in install.sh's inline `curl | bash` fallback (via `HELPER_NAMES`), which has no lib to source.
- **Existing boxes** — self-update patches the installed unit/plist in place and reloads **before** restarting opencode (a supervisor stops a service using its loaded definition). Units are written by install.sh and never re-rendered by self-update, so without this the fix would reach fresh installs only. Idempotent and monotonic: a stale claim is replaced, a higher one is never walked back. macOS goes through PlistBuddy + bootout/bootstrap — `kickstart -k` restarts from launchd's stale in-memory definition and would silently change nothing.

self-update also re-sources `release.sh` after a payload swap, so a helper shipped by the release being installed is defined on that same run instead of landing one update late (install.sh already does this post-extraction).

## Tests

23 new cases in `scripts/lib/release.test.mjs` against the real shell functions: numeric version comparison, the resolver's three fallback paths, the pure unit-text transform (insert / replace-stale / never-downgrade / no-`[Service]` / idempotent), the in-place patcher with `systemctl` stubbed to assert daemon-reload only happens on a real change, and a drift guard tying the two templates to install.sh's substitutions so a lost placeholder goes red instead of rendering a literal `@@ANTHROPIC_CLI_VERSION@@`.

`npm run typecheck` + `npm test` green (3909 renderer + 3889 server).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3254,6 +3254,63 @@ every update, so a box stops drifting from whatever it was installed with
   already warns it restarts opencode and ends running agent turns, which is
   exactly right for this.
 
+### The box tells Anthropic which Claude Code version it is (BET-1503)
+
+opencode reaches Anthropic through the `opencode-claude-auth` plugin, which
+authenticates **as Claude Code** and therefore sends a Claude Code version in
+its user-agent and billing headers. That version is HARDCODED in the plugin
+(`config.ccVersion`) and lags badly — a live box's cached `@latest` claimed
+`2.1.185`, and even the newest published plugin claims `2.1.217`.
+
+**Anthropic gates new models on a minimum client version and refuses anything
+below it.** So a stale claim makes a model the user is entitled to simply
+unusable, with an error that misdirects: *"Claude Code 2.1.185 does not support
+this model; version 2.1.251 or newer is required. Run 'claude update'"* — which
+sends the user to update a CLI that is already current and has nothing to do
+with it. That is the whole reason this exists; the box's real `claude` binary
+was 2.1.257 at the time.
+
+The plugin honours `ANTHROPIC_CLI_VERSION` from its environment, so the box
+sets it on the opencode service. Four things about how:
+
+- **The value is DERIVED, never pinned.** `resolve_anthropic_cli_version` in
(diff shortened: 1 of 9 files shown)


antoinedc:
Heads-up: the trailing `(BET-1503)` in the title made the Multica daemon link this PR to BET-1503 (the ctoBudget ROI fix, already merged via #1489), and its open-PR guard blocked that issue from staying `done`. This PR's scope (install/self-update scripts, plugin version claim) is unrelated to BET-1503, so I dropped the key from the title to clear the false linkage — nothing else changed. If you did want it tracked against BET-1503, say the word and I'll restore the title.

PR #1494, opened 2026-09-01
BET-1512: shared branch-probe punct strip + correct CtoFactRow.checkable type

antoinedc:
## Summary

Batches the two follow-ups named in BET-1512.

**#1 — bare-branch checkable probe trailing punctuation (was BET-1508).** The bare-branch arm of `matchCheckable` now shares BET-1504's `ciBranchOf` trailing-`[._-/]` strip through one helper (`branchTokenOf`) instead of two rules drifting; a probe that strips to nothing falls through to the later rules. `src/server/ctoFacts.mjs`.

Honest finding, worth stating plainly: the original repro ("Merged branch main." captures probe `main.`) **does not reproduce** — verified against the actual exported `matchCheckable`. The capture's trailing `\b` + greedy backtracking already prevents trailing `[._-/]` from surviving (a matched capture can only end in `[A-Za-z0-9_]`). The strip is therefore defense-in-depth keeping the "probe never ends in punctuation" invariant under future regex evolution — exactly the shared-source, no-new-rule direction the issue directs — and the characterization test the issue demands pins the behavior either way. Test pins `"Merged branch main." → probe "main"` plus `feat/x.`, mid-sentence, and ci-parity cases.

**#2 — `CtoFactRow.checkable` type (was BET-1509).** `src/shared/api.ts` typed it `string | null`, but the server passes the stored verify-stamp object through verbatim (`factViewRow`: `checkable: fact?.checkable ?? null`; stamp shape `{ probe, branch?, last_checked, result }`). Corrected to `CtoFactCheckable | null` with the object shape; server payload untouched (no flattening). No renderer code reads `checkable` today, so zero behavior change.

## Verification results

- **Files changed:** 3 files. `src/server/ctoFacts.mjs` (+13/-2), `src/server/ctoFacts.test.mjs` (+12), `src/shared/api.ts` (+12/-1)

- PR branch (`multica/BET-1512-bare-branch-punct-checkable-type` @ `ffe0ebba`):
  - test: exit `0`, `3909` pass / `0` fail / `7` skip. Failures: none (one transient `MediaBody.test.tsx:90` img flake observed in a single earlier run — not reproducible in isolation or on rerun; file untouched by this PR). log sha256: `f7f5b481342fb1453161a9e82d3ff6f7ebb1c292adcd98d8846cde5d511137b3`
- Base (`main` @ `f1606cac`):
  - test: exit `0`, `3909` pass / `0` fail / `7` skip. Failures: none. log sha256: `c1d79fbe0177043a70d6725f953e0078f86fd77b5a40b0da8df04832891f3a7c`
- **Conclusion:** 0 new failures, 0 pre-existing failures; 1 unrelated one-off renderer flake (MediaBody) seen once on the PR branch, passed on rerun — follow-up filed to deflake it.

Base: origin/main @ `f1606cac`

### DIFF
--- a/src/server/ctoFacts.mjs
+++ b/src/server/ctoFacts.mjs
@@ -562,6 +562,15 @@ const CI_BRANCH_STOPWORDS = new Set([
   "but", "and", "or", "so", "if", "then", "than", "when", "while", "just", "very", "now",
 ]);
 
+// One trailing-punctuation strip shared by every branch-token rule (ci +
+// git bare branch) so the two cannot drift: the token class includes `.`,
+// and "Merged branch main." must probe "main", never "main." (BET-1512).
+// Empty after the strip → null → the calling rule falls through.
+function branchTokenOf(capture) {
+  const token = String(capture ?? "").replace(/[._\-\/]+$/, "");
+  return token || null;
+}
+
 // Branch token for a CI statement — first non-filler capture wins:
 //   "branch <name>"  ("CI on branch feature/x is green")
 //   "CI on <name>"   ("CI on feature/x is green")
@@ -576,7 +585,7 @@ function ciBranchOf(s) {
   for (const re of patterns) {
     const m = s.match(re);
     if (!m) continue;
-    const token = m[1].replace(/[._\-\/]+$/, "");
+    const token = branchTokenOf(m[1]);
     if (token && !CI_BRANCH_STOPWORDS.has(token.toLowerCase())) return token;
   }
   return null;
@@ -591,7 +600,13 @@ export function matchCheckable(statement) {
   const ci = s.match(/\bCI\b[\s\S]*?\b(green|passing|passed|failed|failing|broken)\b/i);
   if (ci) return { kind: "ci", surface: "ci", probe: ci[1].toLowerCase(), branch: ciBranchOf(s) };
   const branch = s.match(/\bbranch\s+([A-Za-z0-9_\-.\/]+)\b/i);
-  if (branch) return { kind: "branch", surface: "git", probe: branch[1] };
+  if (branch) {
+    // Shared strip with the ci rule above — the git probe never carries
+    // trailing punctuation either (BET-1512). Nothing branch-like left →
+    // fall through to the later rules.
+    const token = branchTokenOf(branch[1]);
+    if (token) return { kind: "branch", surface: "git", probe: token };
+  }
   const issue = s.match(/\b([A-Z][A-Z0-9]+-\d+)\b/);
   if (issue) return { kind: "issue", surface: "issue", probe: issue[1] };
   const ver = s.match(/\bversion\s+(\d+(?:\.\d+){1,3})\b/i);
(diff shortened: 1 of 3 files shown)


antoinedc:
## Review (manta-reviewer) — BET-1512

**Verdict: PASS — 0 Blocks, 0 Questions, 1 nit.** Routed to `manta-pm` for merge.

### Premise verification — the load-bearing claim, checked myself

The implementer's "honest correction" (the issue's stated repro does not reproduce) is **correct**. I verified against `origin/main`'s regex before reading the PR diff: `\bbranch\s+([A-Za-z0-9_\-.\/]+)\b` — the trailing `\b` forces a matched capture to end in a word character (`[A-Za-z0-9_]`), so greedy backtracking already sheds trailing `[._-/]`. Node on base:

```
"Merged branch main."            -> "main"      (base regex, no PR code)
"branch feat/x. Done."           -> "feat/x"
"Merged branch fix-auth. Next sentence." -> "fix-auth"
"merged branch v1.2.3."          -> "v1.2.3"    (internal dots preserved)
```

So the issue's failure mechanism ("probe `main.` → cat-file always fails → true facts superseded on noise") was factually wrong. The strip is a no-op under the current regex — kept as shared defense-in-depth per the issue's explicit letter ("reuse or share the existing one so the two cannot drift"), which is what `branchTokenOf` does: `ciBranchOf` now calls the same helper, one strip source, no drift. The empty-after-strip fall-through is dead code under the current regex but semantically safe. This was handled exactly right: deviation documented, issue's direction followed, behavior pinned either way.

| Done-when | Verified |
|---|---|
| Both edits landed in one PR | ✓ 2 commits, 3 files — exactly the issue's named scope, no scope creep, no under-delivery |
| `"Merged branch main."` yields probe `main`, test pinning it | ✓ Ran the PR's exported `matchCheckable`: all four pinned cases hold; internal dots (`a-b_c.d`) correctly survive. New test confirmed executing in-suite (`node --import ./scripts/testSandbox.mjs --test src/server/ctoFacts.test.mjs`: 50/50 pass, `ok 27 — bare-branch probes strip trailing punctuation (BET-1512)`) |
| `CtoFactRow.checkable` matches what `factViewRow` sends, incl. `branch?` | ✓ `factViewRow` passes `fact.checkable` through verbatim; the stamp is `{ probe, last_checked: 0, result: null }` + optional `branch` (BET-1504), and `verifyDue`'s in-place update spreads the same shape. `CtoFactCheckable` matches exactly. No flattening (confirmed — server untouched). Also independently confirmed nothing consumes `.checkable` outside the server today, so the type change is behavior-neutral |
| `npm run typecheck && npm test` green | ✓ On my box, PR @ `ffe0ebba`: typecheck exit 0; tests 3899 pass / 0 fail. Base @ `f1606cac`: 3898 pass / 0 fail — the delta is exactly the +1 new test |
| `duplication-gate` green on the PR | ✓ CI check pass; anti-spaghetti sticky report: 0 clones across the 3 changed files. The shared helper *reduces* duplication |

Note on totals: the implementer reported 3909/7-skip on both branches; I measure 3899/3898 with 0 skips on this box. The ~10-test delta is environmental (their runner skips tests mine registers), consistent across both of their runs, and the failure lists match exactly (empty, both branches). Not a concern.

### Nit (non-blocking)

- **Coordination reminder, not a defect:** BET-1505 is still in flight on `verifyDue`/`composeFactsRender` in the same file. The PR is mergeable against current `main` today (no overlap — main moved only in `ctoEngine`), but whenever BET-1505 merges first, the "rebase on main and keep its change" reconciliation from the issue applies. PM should sequence the two merges with that in mind.

### Route

0 Blocks, 0 Questions → handing off to `manta-pm` for merge. PR remains draft — flipping to ready is the PM's call.

PR #1501, opened 2026-09-03
feat(server): BET-1519 generic §9.4 plan executor + driver, resolve-store folding, verdict-route execution

antoinedc:
## BET-1519 — P4.4: the one generic executor + verify + attempt cap + `cto.resolve` + delegate perm fix

Base: origin/main @ ac4a060a (includes BET-1517 triage + BET-1518 gate/τ/calibration)

### The one executor (§9.4)
- `src/server/ctoAct.mjs` is rewritten as the ONE generic plan executor: `validatePlan` (runtime shape), `wantsDelegateSession` (access has write/edit → delegate job, else ephemeral), `buildExecutionBrief` (identity + finding + diagnosis + numbered steps + access flavor + undo + the exact `CHECK: pass|fail: <reason>` end marker), `runVerifyCheck` (all four §9.2 kinds — predicate via the §6.7 fact verifier, probe via a consented probe read, session-ok via the marker, condition-gone via the §10.3 liveness classifier; a no-opinion seam is a FAIL, never a silent pass), and the runner's run → verify → one-retry loop (retry in the SAME session with the failed check attached and a fresh 30-minute budget; both fail → escalated with the attempt log `attempt1:…; attempt2:…`).
- Delegate jobs poll the job row (`jobRow`), read the child session's transcript for checks, and deliver the retry brief into the child session; ephemeral sessions are always aborted + deleted after the run.

### The driver (ctoEngine wires it, ctoAct implements it)
- Queue ≤ 10 FIFO with blocker-before-finding; ≤ 2 in flight (§3.3 sub-cap / §9.4); overflow refused (`queue-full`).
- Attempt cap: **two executions per plan id, ever**, derived from the durable `resolve.json` rows — a re-triage can never re-arm it; a cap-hit refusal returns `cap-hit` (the gate degrades to the ask card, never acts).
- Every execution writes ONE `cto.resolve` ledger row with the full §9.4-9.5 field set: planId, class, findingId, confidence, calibration, effective, τ, trigger (`act`|`accepted`), outcome (`resolved`|`escalated`), attempts, cost, reason, undo, refs, ts (+`resolvedAt` on resolved rows).
- §9.5 outcomes, one per plan: `escalated` → immediate calibration failure + a human blocker card (`recordBlocker("cto-executor", …)`) carrying the attempt log; `resolved` → success DEFERRED to the 7-day window (driver `tick()` folds success when no correct/dismiss/veto verdict lands after `resolvedAt`, failure when one does). A negative verdict inside the window resolves immediately as failure.
- A runner throw is an interrupt → escalated with the interrupt reason (never silently lost). Note: the ticket's "aborts + requeues at any interrupt with outcome escalated" is implemented as abort + escalate (not re-queue) — re-queueing an interrupted run risks cap-less loops; the escalation card carries the state for the human.

### Double-count guard (§9.5 "one outcome per plan")
`ctoCalibration.noteVerdictEffects` now DEFERS the accept/edit fold when the subject resolves to a plan (plans.json hit, overridable via an injected `isPlanSubject`; ledgered as `calibrate.fold outcome:"deferred-to-executor"`) — the execution the accept kicked off owns the outcome through `notePlanOutcome`. Dismiss/correct/veto on plan subjects still fold immediately (dismissed-unexecuted / later-negative failures).

### "Do it" wiring (no duplicated execution logic)
`POST /api/cto/verdict` (index.mjs): an accept/edit whose subject resolves to a plan in plans.json runs THAT plan through `adaptiveCto.executor.executeAccepted` (trigger `accepted`, the digest announcement via the engine's `recordAct`), and reports the result in the response's `execution` field. The renderer's plan-option flow needed no logic change (its click already records the §9.5 accept verdict the route now acts on).

### Retired per-class executors
`createCtoActExecutor`, its three bound-action branches (record-decision / queue-tonight / start-job) and the three payload normalizers are deleted; the `no-executor` fall-through and the unreachable `config-change`/`tool-write` cases went with them. The suggest flow's machine act verb is now unwired (`executeAction: null` — an established seam state) so its acts degrade to ask cards, whose bound options stay user-executable through the renderer's ask path; blockers and suggestions converge on triage → plan → gate (D22). `beginDelegateJob` stays: the overnight dispatch (§11.5, unchanged by this ticket) still needs its own gate + `cto.delegate_begin` row.

### Delegate perm bypass fix
- CTO-originated delegate jobs (the executor) route through the shared `startJobWithApproval` with **the plan's access rules as the declared tools** and `trustMode: false` — the approval card is the human's veto over machine-originated work; the ruleset always ends in the catch-all deny (tested: tools + trustMode asserted on the start call).
- The overnight seam (`startDelegateJob`) also routed through `startJobWithApproval`, closing its NO-permission allow-all hole with the standard work toolset declared. **Deliberate deviation, flagged:** overnight jobs run with `trustMode: true` — the §11 veto card already consented tonight's plan (the ticket acknowledges that veto card is a scheduler artifact and unchanged), and a second approval card per overnight job would stall the window behind the 2-minute approval timeout.

### Anti-spaghetti
- Single sources kept: `buildPermissionRuleset` (delegate.mjs) reused for both session kinds; the resolve store (BET-1518's `resolveStore`) is the only new state, owned by the driver; no parallel calibration module (the dead prior run's was discarded in favor of BET-1518's).
- Deleted: the old per-class executor + normalizers (no importers remain — grep clean); `beginDelegateJob`'s suggest-flow start-job path dies with the suggest act verb, its overnight use remains.
- Duplicate-site audit: the runner's turn-poll mirrors `runSynchronousSession`'s established idle detection; the §6.7 verify seams are the engine's own `factVerify` + `conditionGoneFromCondition` (one implementation, injected).

### Follow-ups filed
- Parked: BET-1525 (probe-verify seam accepts tool-qualified names only; fails closed until probes exist), BET-1526 (plans could carry an optional project/cwd for job hosting).

### Verification results
- PR branch (`multica/BET-1519-generic-executor` @ 83745882, review cycle 1):
  - typecheck: exit 0. Errors: none. log sha256: 84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709
  - test: exit 0, 3958 pass / 0 fail / 0 skip. Failures: none. log sha256: 2c0e61c4db179eac80d557024197f2fa67b3d41b888298fcaa2ea70b1ea77a59
  - flake re-run: 4 consecutive full-suite `npm test` runs, all 3958/0 (the reviewer observed 3-of-4 failing at 5a099136)
- Base (`main` @ ac4a060a):
  - typecheck: exit 0. Errors: none. log sha256: 84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709
  - test: exit 0, 3949 pass / 0 fail / 0 skip. Failures: none. log sha256: 402c2d7e7b20b898fae6350156bb83e9b7fd2d1b370f006fbd1c5c91b9ce9dc5
- **Conclusion:** 0 failures on either branch; +9 tests on the PR branch (executor/driver suite + calibration deferral + §9.4 presence + still-running retry guard). Review cycle 1: both Blocks fixed — (1) the mirrored-order durability test's fold waits are now bounded polls for the expected state (deterministic under any load), and (2) the §9.4 presence rule is implemented on the executor path (finding-sourced `act` refuses while `present` → the gate degrades to the ask card; blocker-sourced and user-accepted plans bypass; enqueue-time check, documented). The reviewer's retry/interleave nit is also fixed: a delegate job still running at the turn budget escalates (`job-still-running-at-budget`) instead of delivering the retry prompt into the in-flight child session.

### DIFF
--- a/src/renderer/ctoView.ts
+++ b/src/renderer/ctoView.ts
@@ -799,11 +799,13 @@ export async function executeSuggestionOption({
         return r?.ok ? { ok: true } : { ok: false, error: r?.error ?? "queue-tonight failed" };
       }
       case "plan": {
-        // BET-1518 (§9.3): a gate ask card's plan option. The concrete
-        // execution is BET-1519's executor seam — today the click records
-        // the human's accept judgment (§9.5 verdict, stamped with the
-        // card's class by the caller) and the verdict route resolves the
-        // card. Validate the payload so a malformed option fails closed.
+        // BET-1518 (§9.3) + BET-1519 (§9.4): a gate ask card's plan option.
+        // The click records the human's accept judgment (§9.5 verdict,
+        // stamped with the card's class by the caller); the verdict route
+        // then runs the plan through the ONE generic executor server-side
+        // (trigger "accepted" — same session, verify, retry and cap as the
+        // gate's act verb). Validate the payload so a malformed option fails
+        // closed.
         const planId = typeof payload.planId === "string" ? payload.planId : "";
         if (!planId) return { ok: false, error: "plan needs a planId payload" };
         return { ok: true };
--- a/src/server/ctoAct.mjs
+++ b/src/server/ctoAct.mjs
@@ -509,6 +509,16 @@ export function createCtoPlanRunner(deps = {}) {
     let turn2 = { ok: false, reason: "no-send", lastText: "", assistants: 0 };
     try {
       if (isDelegate) {
+        // If turn 1 ended on budget but the job is still executing, delivering
+        // the retry brief into the child session now would interleave with the
+        // in-flight turn. Wait for the job to finish first (bounded by the
+        // retry deadline; still running at it → that IS the failed check).
+        if (turn.reason === "turn-budget") {
+          const stillRunning = await readTranscript(sessionId, { baseline: 0, deadline: retryDeadline });
+          if (stillRunning.ok === false && stillRunning.reason === "turn-budget") {
+            return finish("escalated", `attempt1:${check1.reason ?? "verify-failed"}; job-still-running-at-budget`);
+          }
+        }
         const row = await jobRow(sessionId);
         const childId = row?.sessionId;
         if (!childId) {
@@ -576,6 +586,7 @@ export function createCtoExecutorDriver(deps = {}) {
     recordAct = null, // ({cls, text, refs, action, score}) — digest announcement
     verdictsList = async () => [], // () => verdict entries
     knowsPlan = async () => false, // (planId) => plans.json hit
+    presence = null, // async () => "present"|"away"|"gone" (§9.4 rule; engine wires getPresence().state)
     now = () => Date.now(),
     execute = null, // when runner is not provided directly
     queueMax = QUEUE_MAX,
@@ -755,7 +766,19 @@ export function createCtoExecutorDriver(deps = {}) {
     }
   }
 
-  /** The gate/accepted seam: validate → cap → enqueue → drain. */
+  /**
+   * The gate/accepted seam: validate → cap → presence (§9.4) → enqueue →
+   * drain.
+   *
+   * §9.4 presence rule (§3.4 rule 3): blocker-sourced plans run regardless
+   * of presence (that is the point of a blocker); finding-sourced machine
+   * acts (trigger "act") refuse while the user is `present` — the gate then
+   * degrades them to the ask card, which is the right §9.3 surface when the
+   * human is right there. Decision (documented in the PR): the check is at
+   * ENQUEUE — a refused plan re-runs through the gate's ask path instead of
+   * being held in the queue for an indefinite away-window. User-accepted
+   * plans (trigger "accepted") are user-initiated and bypass presence.
+   */
   async function executePlan(plan, ctx = {}) {
     const { findingId = null, finding = null, project = null, trigger = "act", gateCtx = null } = ctx;
     const v = validatePlan(plan);
@@ -771,6 +794,15 @@ export function createCtoExecutorDriver(deps = {}) {
     }
     const isBlocker =
       finding?.kind === "blocker" || finding?.sourceKind === "blocker" || finding?.noteKind === "blocker";
+    if (trigger === "act" && !isBlocker && typeof presence === "function") {
+      try {
+        if ((await presence()) === "present") {
+          return { ok: false, reason: "presence-gated" };
+        }
+      } catch {
+        /* a presence read failure must not block machine work — run */
+      }
+    }
     queue.push({ planId, plan: v.plan, finding, findingId, project, trigger, gateCtx, isBlocker, queuedAt: now() });
     await drain();
     return { ok: true, queued: true, planId };
--- a/src/server/ctoAct.mjs
+++ b/src/server/ctoAct.mjs
@@ -1,218 +1,847 @@
 // src/server/ctoAct.mjs
-// BET-1424 — the §9.2 act-and-report executors for the §9.3-eligible classes.
+// BET-1519 — the ONE generic §9.4 plan executor + the executor driver.
 //
-// The suggestion engine's act branch (ctoSuggest.mjs) calls ONE injected
-// `executeAction({cls, action, candidate})` with the candidate's primary bound
-// option. BET-1403 wired only `record-decision` inline in index.mjs; this
-// module owns all three §9.3-eligible executors (ctoTrust.mjs eligibility map)
-// behind one factory:
+// D22 (spec v3): one generic executor runs every plan — no per-use-case
+// handler anywhere. The per-class bound executors of BET-1424 (record-decision
+// / queue-tonight / start-job branches + their payload normalizers) are
+// retired; every plan runs the same way:
 //
-//   - record-decision → a gatekeeper-checked decision fact on the CTO's own
-//     blackboard (the engine's fact route). Behavior carried over verbatim
-//     from the BET-1403 inline executor.
-//   - queue-tonight   → an entry in tonight's overnight queue via the engine's
-//     tonightAdd (§10.4/§11.4 — planning-only, nothing runs until the window
-//     opens; trivially reversible).
-//   - start-job       → a worktree-isolated delegate job (own git worktree +
-//     branch; never touches the user's checkout), started with actor "cto"
-//     under the §3.3 delegate gate (kill switch, pause, concurrent cap).
+//   - an `act` plan (or an `ask` plan the user accepted, trigger "accepted")
+//     runs in a CTO-owned session with plan.steps as its brief and
+//     plan.access as its permission ruleset;
+//   - the session kind comes from the plan: a delegate job (own worktree,
+//     branch, sidebar row — through the shared startJobWithApproval path so
+//     the ruleset always applies) when access includes `write`/`edit`, an
+//     ephemeral session otherwise;
+//   - after the turn goes idle the plan's verify check runs (§9.2): a
+//     predicate statement against the §6.7 surfaces, a consented probe read,
+//     the session's own CHECK marker, or the originating blocker's condition
+//     being gone;
+//   - fail → exactly one retry: the SAME session gets one more turn with the
+//     failed check attached (fresh 30-minute budget); fail again → the
+//     finding escalates to a human blocker card carrying the attempt log;
+//   - every execution writes ONE `cto.resolve` ledger row (§9.4-9.5) and one
+//     outcome into the resolve store; two executions per plan id, ever — a
+//     cap-hit refusal degrades to ask, never acts;
+//   - outcomes feed child #3's calibration (notePlanOutcome): `resolved` with
+//     no correct/dismiss/veto verdict within 7 days → success (deferred); a
+//     negative verdict in the window, or `escalated`, → failure.
 //
-// Everything else refuses with `no-executor` (config-change is §9.3-capped at
-// ask permanently; tool-write stays data-unreachable until a tool holds the
-// §7.4 write ring). The refusal contract is load-bearing (§9.2): act-and-report
-// must never silently no-op — every refusal returns {ok:false, reason} so the
-// verb degrades to the veto-window card, whose options remain user-executable
-// through the renderer's ask-path executors.
+// The driver (queue ≤ 10 FIFO with blocker-before-finding, ≤ 2 in flight,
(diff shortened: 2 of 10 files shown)


antoinedc:
## Review (manta-reviewer) — BET-1519, cycle 1

**Verdict: 2 Blocks (both fix-here), 0 Questions → returned to `manta-dev`.**

### Mechanical gates (run by reviewer on PR branch `multica/BET-1519-generic-executor` @ 5a099136)

- `npm test` — **flake observed**: run 1: 3956 pass / 0 fail; runs 2–4: **3955 pass / 1 fail** (3 of 4 full-suite runs failed on my box). The file passes 5/5 in isolation. Implementer's reported clean run matches run 1, but the criterion "npm test passes" is not reliably met → Block 1.
- CI: `typecheck-test` pass, `duplication-gate` pass.
- Base freshness: merge-base == `origin/main` == ac4a060a — no stale-base phantom deletions.

### Acceptance-criteria checklist (verified independently, not from the PR prose)

| Criterion | Result |
|---|---|
| One executor runs every class; no `class`-branching in the execution path | ✅ `ctoAct.mjs` reads `plan.class` only for ledger fields / card text; `createCtoActExecutor`, the three bound branches, the payload normalizers, `no-executor` and `config-change`/`tool-write` special-casing are gone (remaining hits are comments and the suggest flow's own class list). |
| CTO delegate jobs carry a permission ruleset | ✅ Runner `startJob` seam → `delegateEngine.startJobWithApproval` (index.mjs:2184) → `buildPermissionRuleset(plan.access)` with the catch-all deny **prepended** (delegate.mjs:259), `trustMode:false`. Overnight seam (`startDelegateJob`, index.mjs:2285) also closed — `STANDARD_WORK_TOOLS` declared. The `trustMode:true` overnight deviation is documented and accepted (the §11 veto card already consented the plan). |
| Verify + ≤1 retry + escalate with attempt log; two executions per plan id max | ✅ `runVerifyCheck` dispatches all four kinds (predicate via `matchCheckable`+`factVerify`, condition-gone via the §10.3 classifier, session-ok via the `CHECK:` marker, probe — fails closed until probes exist, honestly). One retry in the SAME session with the failed check attached; fail again → `escalated` with `attempt1:…; attempt2:…` and a `recordBlocker` card. Cap = 2, derived from durable resolve rows (a re-triage cannot re-arm — tested). 30-min budget per turn. Interrupt → escalated with reason, never silently lost (documented reading, accepted). |
| "Do it" and `act` share the identical execution path | ✅ `POST /api/cto/verdict` accept/edit on a plan subject → `executor.executeAccepted` (trigger `accepted`) → the same driver/runner; no duplicated logic. Calibration accept-fold deferred to the executor (`isPlanSubject`) so nothing double-counts. |
| `cto.resolve` ledger row every execution | ✅ `runQueued` writes one complete row (`ledgerLog` + resolve-store patch), full §9.4 field set incl. gateCtx effective/τ/calibration from the gate. |
| typecheck + test + duplication gate | ⚠️ typecheck ✅, duplication ✅ (CI + my read of the sticky report: flagged clones at index.mjs 1831–1877 / 2660–2672 / 4639–4651 are outside this PR's hunks — pre-existing, advisory only). Tests: flaky → Block 1. |

### Blocks

**Block (fix-here):** `ctoEngine.overnight.test.mjs:661` — "durability: a calibration fold survives a queue-edit save landing after it" is load-flaky: **3 of 4 full-suite runs failed locally** (1 ≠ 2, the second fold hadn't landed), while passing 5/5 in isolation. Mechanism: this PR made every calibration fold consult plans.json (the deferral check), adding async hops — the test's settle fix (`setTimeout 25ms`) covers only the FIRST fold's wait; the final wait before asserting the SECOND fold (from `tonightRemove`'s edit verdict) is still a single `setImmediate`, which under full-suite parallelism no longer suffices. Fix: give the final wait the same real settle (or better, a bounded poll for `successes === 2`), then re-run the full suite several times to confirm.

**Block (fix-here):** §9.4 presence rule is not implemented. The ticket: "blocker-sourced plans run regardless of §3.4 rule 3 …; finding-sourced plans obey it. This is the ONLY presence change allowed in P4." There is **no presence seam anywhere on the gate → `executePlan` → driver → runner path** (ctoGate.mjs and ctoAct.mjs contain zero presence handling; `cardTick` runs `gatePass()` unconditionally; the driver's deps have no presence input). Consequence: a finding-sourced `act` plan starts a delegate job (worktree + model session) even while the user is `present` — exactly the heavy work §3.4 rule 3 restricts to `away`/`gone`. Blocker-sourced plans trivially run regardless, so the implemented behavior is a universal exemption — the one change P4 does not allow. Fix is narrow: inject a presence read into the driver (the engine already exposes `getPresence().state` and uses the same pattern for the rollup/facts engines), and defer/block **finding-sourced, non-accepted** act plans while `present` (queued ones may drain per §3.4 rule 4's step-boundary semantics — your call, but say which). The driver already has the blocker/finding signal (`isBlocker`), and accepted (`trigger:"accepted"`) plans are user-initiated so presence shouldn't gate them.

### Nits (non-blocking)

- Retry delivery vs a still-running job: when turn 1 ends on `turn-budget` (30 min) but the delegate job is still running, the retry path reads `jobRow` → sends the retry brief into the child session while turn 1 may still be executing there. Cap and budget still bound it; consider checking `row.running` and treating "still running past budget" as the failed check.
- PR body's "the fold may consult plans.json before deferring — its intent is unchanged" undersold the flake risk; see Block 1.

Nice work on the executor shape itself — the generic runner, the durable-cap derivation, and the defer-not-double-count calibration fold are all clean and match the spec.

antoinedc:
## Review cycle 1 — both Blocks fixed (+ the nit), pushed to the same branch @ 83745882

**Block 1 (flaky mirrored-order test):** the test now polls for the expected fold state with a bounded wait (`waitForState`, 5s timeout) instead of fixed settles — both the first fold's assertion AND the final `successes === 2` assertion that was still one `setImmediate`. The sibling Thompson-counter fold test in the same file got the same treatment (same fragile pattern). Re-ran the full suite **4 consecutive times: 3958 pass / 0 fail each** (the flake reproduced 3-of-4 at 5a099136 under the same conditions). The test's intent (folds survive a hostile stale engine-state save) is unchanged — only the waits became load-robust.

**Block 2 (§9.4 presence rule):** implemented on the executor path:

- The driver (`ctoAct.mjs`) takes a `presence` dep; the engine wires it to the same `getPresence().state` read the rollup/facts engines use (no new presence machinery — one source).
- Semantics: a **finding-sourced machine `act` (trigger "act", non-blocker) refuses with `presence-gated` while the user is `present`**; **blocker-sourced plans run regardless** (the spec's only exemption); **user-accepted plans bypass** (user-initiated — the "Do it" click IS the presence).
- **Queued-entries decision (as requested):** the check is at **enqueue, not at dequeue** — a refused plan is not held in the queue; the gate's existing `{ok:false}` path stamps the record `ask` and emits the ask card, so a presence-gated act surfaces as the §9.3 ask surface (the right ask when the human is right there) and can still be accepted manually (accepted plans bypass presence). Holding queue entries for an indefinite away-window added staleness with no upside, and §3.4 rule 4's step-boundary semantics belong to the §11 overnight scheduler, not the executor queue.
- Refusals write no `cto.resolve` row (not executions); the gate's `gate.asked` row covers the trail. A presence-read failure is treated as unrestricted (run) — a broken presence read must not freeze machine work; it degrades to the prior behavior.
- Tests: present→finding act refused (nothing queued/ledgered), present→blocker runs, accepted bypasses, away runs, presence-read-throws runs.

**Nit (retry vs still-running job):** fixed. If turn 1 ends on `turn-budget` while the delegate job is still executing, the runner first waits (bounded by the retry deadline) for the job to finish before delivering the retry brief; if it is still running at the deadline, that IS the failed check → escalate with `job-still-running-at-budget` — no prompt is interleaved into an in-flight child session (tested: no retry send, attempts stays 1).

**Verification:** typecheck exit 0 (sha256 `84317a7d…`); `npm test` exit 0 — 3958/0/0 (sha256 `2c0e61c4…`), ×4 consecutive clean full-suite runs. Base unchanged (`main` @ ac4a060a).

antoinedc:
## Review (manta-reviewer) — BET-1519, cycle 2

**Verdict: PASS — 0 Blocks, 0 Questions.** Both cycle-1 blocks verified fixed on `83745882`.

### Cycle-1 blocks, re-verified independently

1. **Flaky fold test — FIXED.** The fixed settles are replaced by `waitForState`, a bounded poll (5s deadline, 25ms steps) for the expected fold state — deterministic under any scheduling, and it cannot mask the regression it guards (a fold reverted by the hostile stale save never returns to 2, so the poll times out honestly). **3 of 3 full-suite `npm test` runs clean on my box** (3958 pass / 0 fail — the same test failed 3 of 4 runs at 5a099136). `npm run typecheck` exit 0.
2. **§9.4 presence rule — IMPLEMENTED.** The driver gained a `presence` dep wired to `getPresence().state` (the same read the rollup/facts engines use). Non-blocker `trigger:"act"` plans refuse with `presence-gated` while `present` — the gate's existing refusal path then degrades them to the ask card, which is the right §9.3 surface with the human right there. Blocker-sourced plans and user-accepted plans bypass (correct: blockers are time-sensitive per the ticket; accepted plans are user-initiated). A presence-read failure runs rather than blocks machine work — documented and reasonable. The enqueue-vs-drain decision I asked you to make explicitly is made and documented (check at enqueue; refused plans re-gate into the ask path). New test covers all four cases (present-refuse, blocker-run, accepted-run, away-run) + the presence-read-failure case.

Cycle-1 nit also addressed: a delegate job still running at the turn budget now escalates (`job-still-running-at-budget`) instead of interleaving the retry brief into the in-flight child session — with a test asserting no retry prompt is delivered.

### Mechanical gates (reviewer-run, PR head `83745882`)

- `npm test` ×3 — 3958 pass / 0 fail, all three runs.
- CI: `typecheck-test` pass, `duplication-gate` pass. Anti-spaghetti report unchanged (flagged clones are pre-existing index.mjs code outside the diff hunks — advisory only).
- Scope: the fix commit touches only `ctoAct.mjs`, `ctoEngine.mjs`, and the two test files — no creep.

Non-blocking note for the record (no action required): queued-but-not-yet-started finding plans don't re-check presence at drain time, so a plan queued behind 2 in-flight executions can start after a presence return. Narrow race, documented call — fine for P4.

Approving route to `manta-pm` for merge.

PR #1502, opened 2026-09-03
BET-1520: route suggestions through the shared triage→gate→executor pipeline

antoinedc:
## BET-1520 — Route suggestions through the same triage→gate→executor path

Spec: docs/adaptive-cto-spec.md §9.1 (pipeline), §9.2 (triage), §9.5 (executor), §15 P4 item 5, §3.4 rule 3, §9.3.
Base: origin/main @ 9c74fc6e

### What changed

The suggest engine keeps its three P2 collectors (digest-detected failure recurrences, fact anomalies, watcher hits) and now **enqueues every collected finding on the pending-findings queue** — the same queue inbox notes, promoted asks and health escalations ride. From there the ONE engine pipeline takes over: drain → triage (§9.2, 0–3 plans; MAY output none) → gate (§9.3 act/ask) → executor (§9.4/§9.5). Net −751 lines.

**The bespoke surface path is deleted** (per the issue's DELETE list):
- the suggest generator (`runSuggest` seam, `buildSuggestContext`, `parseSuggestionText`, `normalizeCandidates`, `validateCandidate`/`validateOption`, `ACTION_TYPES`, `filterOptionsByData`)
- the worthiness step — decision: **drop entirely** (the removal-favoring option). `worthinessProbability`, `DEFAULT_CLASS_PRIORS`, `defaultThresholds`, `parseWorthinessScore`, `buildWorthinessContext`, the `runWorthiness` seam, and the `suggest.thresholds` engine-state merge are gone. Rationale: the §9.2 triage call IS the "0–3 plans or none" model step — a plan-less finding silent-logs (`suggest.silent`, still written by the gate, §14.3 audit unchanged), so a separate nano-worthiness pre-filter bought nothing but a second model call per finding. No salience floor is carried over (recurrence/anomaly strength is implicit in the finding text the triage call sees).
- `processFinding`'s verb/action machinery: bound-action execution, direct decision-card emission, the veto-window/notify branches (`NOTIFY_RECURRENCE_KINDS`, `fireNotify` dep), `senderReliability`/`classPriors`/`thresholds`/`calibrationOf`/`tau`/`recordAct`/`executeAction`/`cards`/`configGet` deps.

**Kept because it still has a distinct role:**
- the three collectors verbatim (with their stable content ids)
- the BET-1465 `es.suggest.usedKeys` dedupe — now gating *enqueue* instead of generation, so a recurring finding does not re-pay a triage model call every 30-min pass. Enqueue-then-mark ordering preserved (a failed enqueue retries next pass; a consumed key never loses a finding). Cap stays 200, trailing-slice FIFO.
- `listHeld`/`verdictHeld` (§14.3 silence audit read surface) unchanged; rows now written by the gate.

### Plumbing
- `ctoCards.mjs`: `SUGGEST_FINDING_SOURCE = "suggest"`, `findingFromSuggestion` (the fourth producer's row builder, alongside the other three), `findingLedgerKind` maps suggest rows to `suggest.<sourceKind>`.
- `ctoTriage.mjs`: `findingIdOf` gains the `find:suggest:*` derivation (keyed on the collector's stable `sourceId` — re-collected identical findings keep one finding id, so plan upserts converge); `sourceFindingCopy` passes the suggest source through verbatim so the plans store's provenance is honest and the executor sees a non-blocker.
- `ctoEngine.mjs`: exposes the `queueFinding` seam (public, next to `drainFindings`); the drain loop itself was already source-agnostic — one `cardTick` = drain → triage → gate → executor for ALL producers.
- `index.mjs`: rewires `createCtoSuggest` to the new dep set; `queueFinding: (row) => adaptiveCto.queueFinding(row)`.

### Tests (all via the sandboxed memory-store harness, `ctoTestGuard` first import)
- `ctoSuggest.test.mjs` rewritten: row-contract enqueue for the collector kinds, content-stable suggest finding ids, dedupe/retry/unwired-seam, usedKeys cap (no-eviction-at-cap + oldest-eviction + MRU tail), silence audit, and a **deleted-symbols module-contract regression** (15 symbols asserted gone).
- `ctoEngine.test.mjs`: the fourth producer's drain (`suggest.failure-recurrence` evidence row, no inbox side effects); **ONE-pipeline test** — one `cardTick` triages AND gates a suggest finding and an inbox blocker through the same drain/triage/gate calls (2 model calls, 2 plan records, 2 `gate.asked` rows, 2 ask cards); the thrifty-shed placeholder now uses the real suggest source.
- `ctoAct.test.mjs`: the post-`sourceFindingCopy` suggest shape obeys §3.4-3 (presence-gated while present) while the inbox blocker keeps the exemption.
- `ctoTriage.test.mjs` / `ctoWatchers.test.mjs`: suggest source stamping + kind-label span; retired notify-enum assertions became sourceKind mapping assertions.

### Cross-cutting notes
- **Informational push for finding asks is no longer fired** — the old steep-decay `fireNotify` variant lived in the deleted branch. §9.3 makes notify a delivery property of ask; blocker asks still push via the funnel. Filed follow-up: **BET-1527** (parked, spec-§10-amendment territory).
- `es.suggest.thresholds` persisted engine-state overrides are now inert (nothing reads them); left in place deliberately — inert data, no code path, not worth a migration.
- Legacy decision cards already on boards render through the unchanged ask path (options carry `action.type === "plan"`; renderer untouched).

### DIFF
--- a/src/server/ctoAct.test.mjs
+++ b/src/server/ctoAct.test.mjs
@@ -574,3 +574,49 @@ test("driver: executeAccepted announces the accepted execution through the act-a
   assert.equal(f.state.acts.length, 1);
   assert.deepEqual(f.state.acts[0].action, { type: "plan", payload: { planId: "pl1" } });
 });
+
+// ---------------------------------------------------------------------------
+// BET-1520 — the suggest finding's REAL pipeline shape obeys §3.4-3. The
+// finding the gate passes to the executor is the triage record's verbatim
+// copy of the pending-findings row (sourceFindingCopy), which preserves the
+// "suggest" source (pinned in ctoTriage.test.mjs). A suggest-sourced plan
+// must be presence-gated exactly like any other finding-sourced one, while
+// the inbox blocker's shape (source "inbox", noteKind "blocker") keeps the
+// §9.4 blocker exemption.
+// ---------------------------------------------------------------------------
+
+test("driver: §9.4 presence — a suggest-sourced plan defers to ask while present; an inbox blocker runs", async () => {
+  const f = fakeDriverDeps({ runnerResults: [] });
+  let presenceState = "present";
+  f.deps.presence = async () => presenceState;
+  const d = createCtoExecutorDriver(f.deps);
+  // The post-sourceFindingCopy shape of a suggest finding record.
+  const suggestFinding = {
+    source: "suggest",
+    sourceKind: "failure-recurrence",
+    sourceId: "rec:x",
+    ts: 1_000,
+    message: "Pipeline red on main",
+    title: "CTO finding: failure-recurrence",
+    refs: ["c1"],
+    pendingSince: 1_000,
+  };
+  const r1 = await d.executePlan(basePlan({ id: "pl-suggest" }), { finding: suggestFinding });
+  assert.equal(r1.ok, false, "a suggest-sourced machine act refuses while the user is present");
+  assert.equal(r1.reason, "presence-gated");
+  await settle(5);
+  assert.equal((await f.state.store.load()).entries.length, 0);
+  // The inbox blocker's real shape runs regardless of presence.
+  const r2 = await d.executePlan(basePlan({ id: "pl-inbox" }), {
+    finding: { source: "inbox", noteKind: "blocker", message: "deploy failed", refs: [] },
+  });
+  assert.equal(r2.ok, true);
+  await settle(10);
+  // Away → the same suggest plan runs.
+  presenceState = "away";
+  const r3 = await d.executePlan(basePlan({ id: "pl-suggest" }), { finding: suggestFinding });
+  assert.equal(r3.ok, true);
+  await settle(10);
+  const rows = (await f.state.store.load()).entries;
+  assert.deepEqual(rows.map((r) => r.planId).sort(), ["pl-inbox", "pl-suggest"]);
+});
--- a/src/server/ctoCards.mjs
+++ b/src/server/ctoCards.mjs
@@ -86,6 +86,11 @@ export const CONNECT_SOURCE_KIND = "tool";
 // failure that degraded the digest. Keyed by `<tool>/<probe>`; the body
 // deep-links (in copy) to the secrets surface — the fix is a rotated key.
 export const PROBE_SOURCE_KIND = "probe";
+// Evidence-driven suggestion findings (BET-1520 / §9.1): digest-detected
+// failure recurrences, fact anomalies, watcher hits. NOT a blocker source —
+// they shed first in thrifty mode (§12.2) and defer to ask while present
+// (§9.4); the "suggest" source is what downstream reads to tell them apart.
+export const SUGGEST_FINDING_SOURCE = "suggest";
 
 // BET-1498: engine-state.json marker key for the one-time prune of the
 // pre-BET-1469 contentless open cards. Same idempotency contract as the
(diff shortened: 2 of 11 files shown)


antoinedc:
## Review (manta-reviewer) — BET-1520

**Verdict: PASS — 0 Blocks, 0 Questions, 3 nits.** Independent verification run on the PR branch (`bade3edd`, merge-base = `origin/main` tip `e913a576` — base is fresh):

- `npm test` — **3940 pass / 0 fail / 0 skip**, matching the implementer's reported PR-branch results exactly.
- `bash scripts/check-duplication.sh origin/main` — exit 0. The one jscpd clone the sticky report flags in `index.mjs` (context-push blocks ~1831/1870) is pre-existing code outside this PR's hunks — discounted.

### Acceptance criteria — checked independently

1. **ONE drain → triage → gate → executor path** ✅ — `drainFindings` is source-agnostic (ledger kind via the one pure `findingLedgerKind` mapping, now `suggest.<sourceKind>` for the fourth producer); one `cardTick` runs drain → §9.2 triage → §9.3 gate → §9.4/9.5 executor for every producer. Pinned by the ONE-pipeline test (ctoEngine.test.mjs:1696): a suggest finding + an inbox blocker through ONE `cardTick` — same drain, two triage calls through the same caller, two `gate.asked` rows, both surfaced as ask cards.
2. **No suggestion-specific verb/action/plan/card path remains** ✅ — `processFinding` and all bespoke branches deleted; `worthinessProbability`, `DEFAULT_CLASS_PRIORS`, `defaultThresholds`, `parseWorthinessScore`, `ACTION_TYPES`, `validateCandidate`/`validateOption`, the generator, and `NOTIFY_RECURRENCE_KINDS` are gone (15-symbol module-contract regression asserts the import surface; repo-wide grep finds no remaining non-test references). `stableSuggestionId`/`sha` survive with a real consumer (ctoTriage plan/finding ids). The only card surface is the shared gate ask path; the only plan schema is §9.2's.
3. **Worthiness is not a second gate** ✅ — the removal-favoring option (drop entirely, no salience floor carried over) is chosen and documented in the PR body; plan confidence comes from triage (asserted in the pipeline test).
4. **§3.4-3 presence** ✅ — ctoAct.test.mjs exercises the *real* post-`sourceFindingCopy` suggest shape: presence-gated while present (`source: "suggest"` is not a blocker source), inbox blocker keeps the exemption; provenance survives to the plans store (asserted in ctoTriage.test.mjs).
5. **Enqueue semantics** ✅ — collectors kept verbatim; `findingFromSuggestion` normalizes to the shared row shape; enqueue-then-mark ordering means a failed enqueue retries next pass and siblings still enqueue (tested); unwired seam collects without consuming keys; usedKeys cap behavior tested at-cap.

### Nits (non-blocking)

- **Watcher-hit e2e**: covered via collector kind-mapping (ctoWatchers.test.mjs) + the failure-recurrence e2e rather than a dedicated watcher-hit runPass→cardTick test. Low risk (all three kinds share the same row builder), and the implementer's self-check already flags it — fine to defer.
- `gatedSuggestionEphemeral` (index.mjs) is now a historical name — nothing in the suggest flow uses it anymore; it serves digest-compose/tool-registry scans. Rename candidate for a future cleanup.
- Inert `es.suggest.thresholds` persisted state left in place — deliberate and documented; harmless.

### Observation (FYI, no action required)

The implementer's Multica comment cites "Base (`main` @ `9c74fc6e`)" — `9c74fc6e` is actually this PR's own first commit, not main's tip (`e913a576`). Since my independent PR-branch run is 0-fail/clean-typecheck, there are no failures to attribute, so this doesn't affect the verdict.

antoinedc:
## Re-review (manta-reviewer) — BET-1520, cycle 2

**Verdict: PASS — 0 Blocks, 0 Questions.** The dup-gate fix satisfies the bounce's scope guard exactly.

Independent verification on new head `e54072de` (base unchanged, still fresh at `origin/main` tip `e913a576`):

- `bash scripts/check-duplication-gate.sh origin/main` — **PASS, no clones** (STRICT, min-tokens 70, 10 changed files scanned). This cycle I ran the actual gate script; last cycle I ran the advisory `check-duplication.sh`, which is why the red slipped past my first PASS — process note recorded.
- `npm test` — **3940 pass / 0 fail** — the same count as before the fix, consistent with the scaffolding-only claim.
- CI on head `e54072de`: `duplication-gate` pass (22s), `typecheck-test` pass (2m1s).

### Scope-guard compliance (checked in the delta `bade3edd..e54072de`)

- Confined to `src/server/ctoSuggest.test.mjs` (+32/−20) — nothing else touched.
- Clone 1 → `recurringDigests()`: the two-retained-digests fixture + store dep extracted; both the enqueue and dedupe tests now consume it. Assertions byte-identical.
- Clone 2 → `verdictHarness(recordVerdict)`: parameterized on the verdict route (wired → records into `recorded`; `null` → the unwired route). The wired variant now records the raw `input` object rather than a `{subject, verdict}` reshape — same fields the assertions read (`recorded[0].subject.class`, `.verdict`), zero behavioral delta. The unwired test's `verdictsStore.save` precondition is preserved.
- No ignore-file edits, no threshold/baseline changes; the three review nits stay in BET-1528 as instructed.

The original acceptance criteria (one pipeline, no bespoke suggest surface, worthiness dropped, §3.4-3 presence) were verified in cycle 1 and are unaffected by this test-only delta. Routing back to `manta-pm` for merge.

PR #1505, opened 2026-09-14
feat(cto): the secret store is the single CTO access grant

antoinedc:
## The rule

**A key present in the MantaUI secret store grants the CTO FULL access to the matching tool. No consent rings, no connect asks, no read/write split.**

The user adding a secret IS the grant. The secret store becomes the single access path and the single control surface — and **revocation is deleting the secret**. A tool with no matching key simply has no access, and nothing is asked of the user about it.

## How access is decided

`consentFor()` — the one chokepoint every probe and read path already funneled through — keeps its signature and derives its answer from the **secret store's KEY LIST, read at decision time**. A secret that is added and never used grants immediately; deleting a key stops granting immediately; every ring answers the same "yes".

The mapping is an **authorization decision, not a discovery label**, so it is conservative and singular — `matchSecretIdentity(key)` returns **at most one tool, and only a known catalog identity**:

| Key | Grants | Why |
|---|---|---|
| `GITHUB_PAT`, `GH_TOKEN`, `GitHub_Token`, `GITHUB2_TOKEN` | `github` | one catalog identity (alias / casing / ordinal normalized) |
| `CAPO_MULTICA_TOKEN` | `multica` | `capo` is unknown → names nothing |
| `MODAL_TOKEN_ID`, `NORDVPN_TOKEN` | `modal`, `nordvpn` | one catalog identity |
| `GITHUB_STRIPE_TOKEN`, `GITHUB_OpenAI_TOKEN`, `githubOne…SevenStripeToken` | **nothing** | ambiguous — granting either would be a guess |
| `API_KEY`, `ACME_INTERNAL_TOKEN`, `CONSTRUCTOR_TOKEN` | **nothing** | no known tool named |

Three properties make that table hold for *every* spelling, not just the tidy ones:

- **Total.** Every character is analysed or no grant is produced. There is no cap inside the analysis (a dropped tail can make an ambiguous key look unambiguous); the one bound is on the input, and exceeding it fails closed.
- **Canonical.** Case, repeated separators and trailing ordinals are normalized, and a camelCase run is read both whole and per-word — otherwise `OpenAI` reads as `Open`+`AI` and hides a second service. A character outside the key alphabet is refused rather than treated as a separator, so a homoglyph cannot hide half of an ambiguous key. An explicit separator is still a real boundary (`GIT_HUB_TOKEN` → nothing).
- **Hint-free.** The hint is free text written for a human and is not consulted.

**The grant is enumerable, not just answerable.** The store knows nothing about discovery, so the registry view projects a granted tool that has never been used — an honest empty row (zero uses, no vitality) carrying its `accessKey`. Both halves of that view — "is this identity already here?" and "which row carries the grant?" — resolve a row's identity through one `identitiesOf` (name + aliases), so a persisted alias can neither suppress a projection nor strand a grant. The projection is never a write.

**A credential is engagement with the tool it names**: channel-1 evidence resolves through the same matcher, so providing `GITHUB_PAT` is a use of `github`, not a parallel `github_pat` row the grant can never reach. A key the catalog cannot place keeps its raw identity for the one-shot LLM classification and grants nothing.

## What is deleted

The connect-ask path, entirely — not disabled, deleted: the ask generation; the `connect` card variant and its writers; `resolveConnect` / `unNever` / `revokeConsent`, their routes, `Api` methods and `httpApi` implementations; the renderer's Connect section and the drill-down's ring/revoke/never-list/un-never controls; and the now-unreachable ask bookkeeping (`askRound`, `askAtUses`, `reArmAt`, `unneverAtUses`, `deepAskRound`, `deepReArmAt`, the ask-cadence day stamps, `deepReadBar`, `DEEP_RELEVANCE_MIN`).

The tool drill-down is **read-only by design** and states the grant ("access: granted by secret `NORDVPN_TOKEN` — delete that secret to revoke"). No dead control is left behind: there is no second surface that could disagree with the store.

## Consumers ported

Every consumer that derives access from the registry was swept and ported (all covered by regression tests that consume a **real `listTools` projection**, each verified red before green):

- the §6.7 issue surface (`index.mjs`) — otherwise checkable-verify stayed disabled forever; the rule is now the pure, exported `isIssueToolGranted`, which matches on the row's identity set;
- the §7.6 overnight data-source candidates (`ctoOvernight.mjs`) — otherwise every candidate was dropped even with a granting secret;
- the §10.5 drill-down / `read_toolregistry` (`toolsView`) — now shows an integration the moment its key is added.

Probe-side access (`runDue`, `authorSpecs`, `relevanceScan`, `probeSummary`, `healthSnapshot`) already went through `consentFor` and is store-derived by construction. No other reader of the deleted `consent`/ask fields remains.

## Secret-value handling — precise claim

- **The grant path cannot touch a value.** The registry reads the store through `listSecretKeys()`, which returns an **array of key names and nothing else** — no value, hint, scope or id can reach an authorization decision even by accident. A key name is not a secret.
- **Probes do materialize the granted credential** by reference (0600 file, injected as an auth header to the tool's own API). That is **pre-existing and intended** — it is what a grant is *for* — and this PR does not touch that code path.

## Stale state and migration

- A stored secret grants **even if an older consent record said `never`** — and an old `yes` does not survive the secret's absence. Stale consent state is not consulted in either direction.
- **Persisted open `connect` cards are dropped on load** and excluded from the needs-you badge, so an upgraded box cannot render or count a card nothing can answer. No migration write is performed.
- An upgraded box may still hold raw `<key>`-named rows (e.g. `github_pat`) recorded by the old channel-1 identity. They grant nothing, stay labelled raw, and age out through the existing unresolved-row retention; new credential evidence lands on the canonical tool.

## Blast radius

**Every key in the store that names exactly one known tool becomes a live grant** on merge — that is the intent of the rule, and it is the thing to eyeball before merging. Keys that name nothing known, or two things, grant nothing.

## Verification

- `npm test` — 200/200 renderer files (3915 passed, 7 skipped); server node:test 3998 passed, 0 failed.
- `bash scripts/check-duplication-gate.sh origin/main` — PASS, no clones over 24 changed files.

**No live state was touched, no service restarted, no config changed, nothing deployed.** Work was done in an isolated worktree; the suite runs against the sandboxed `MANTA_STATE_HOME`.

### Deliberate non-change

`consentFor` is **uncached**. A cache is a second source of truth in an authorization path, and its callers (probe ticks, drill-down renders) are minutes apart, not a tight loop — the read is one small JSON file. Correctness over the micro-optimisation.


antoinedc:
Astra (gpt-6-astra), reviewer session ses_f5e27f111ffeM0yaSEQB0k2GzT, completed iterative independent review and marked final HEAD 3bf1cea75c00b57bf0f9703295ea7d292dd03cf6 MERGE-READY. Implementation continued on verified voskaai/default as requested. All blocking findings addressed, including precise secret-name grants, no prior-discovery dependency, consistent aliases, scoped CTO credential use, credential rotation, and cross-service credential/destination isolation. Reviewer independently reproduced prior wrong-host exploit with synthetic credentials: now zero materializations and HTTP requests. Final follow-up only removes accidental sitemap timestamps; reviewed CTO code unchanged. Required CI and strict duplication gate pass on exact final head.

Residuals: unknown/ambiguous secret names grant no catalog identity; noncatalog alias quality still depends on classification; legacy raw rows and duplicate spec state may remain. No live state mutation, restart, deployment or credential use against external services was performed.

PR #1506, opened 2026-09-15
docs(cto): P0 contracts — implementation map, source spec, opencode-db fixture harness

antoinedc:
Baseline inventory for the unified-CTO spec (spec §15 P0). Source baseline `9f0756caa67fc2a9105649bed76e181d16c155d8` (origin/main). Contract-only: **no feature is asserted to exist** — no role session, no admission path, no headless delegate, no context service.

## Contents

1. **`docs/unified-cto-spec.md`** — the spec landed verbatim (byte-identical to the authoring copy; md5 `723a9b26654c4be3db77be39ccc56afd` verified).
2. **`docs/cto-implementation-map.md`** — the exact source seams with evidence levels (`PROVEN` / `PROVEN-SRC` / `UNPROVEN`):
   - **Role session creation/correlation:** installed opencode **1.18.29** `/doc` + source snapshot — `POST /session` has **no client session ID** (server-generated `^ses`) but accepts a free-form `metadata` object (persisted on the `session` row; `setMetadata` exists) — the identity-marker seam for crash recovery. `prompt_async` accepts a **client `messageID` (`^msg`)**, honored in oc-src (`id: input.messageID ?? MessageID.ascending()`), but the 204 response has no echo — receipt requires read-back; behavior on this box **unproven**.
   - **Unified admission / bypass:** all server sends already funnel through one `promptDelivery.deliver` instance (scheduler, RPC shim, notify, delegate, multica); the bypass is the renderer's direct chat path + queued drain/abort (ChatPanel/useSseBus). `promptDelivery` is **memory-only** (no durable records/event IDs/origin) — matches the spec hazard list.
   - **Headless delegate:** `resolveOwner` requires the parent's tmux window (headless parent fails); worktree failure **silently falls back** to the parent dir (no `isolationRequired`); `genId()` runs **after** worktree + window creation inside the store lock; caps `MAX_RUNNING_JOBS=5` / `ctoAct.MAX_IN_FLIGHT=2` map to the spec's reuse rule.
   - **Stable project ID:** Manta has none (tmux session name is the key); **opencode persists `project.id`** (40-hex keyed to worktree) + `project_directory` + `workspace`; `session.project_id NOT NULL` — the natural `ProjectRef.workspaceId`.
   - **DB sandbox path:** `MANTA_STATE_HOME` does **not** redirect opencode.db (`MANTA_OPENCODE_DB` verbatim → XDG → `/home/dev/.local/share`); live-schema dump also shows an opencode-side `session_input` admission-ledger table (empty on this box — noted, unproven).
3. **`src/server/fixtures/opencodeDbFixture.mjs`** — the reusable spec §14 fixture: builds the verified `message`/`part`/`session` schema, arms `MANTA_OPENCODE_DB` **before** the shared handle opens, resets the handle, restores env, and exports `assertNoLiveDbFallback()` (the §14 canary, exercised in both sandbox states). No fallback to the live path anywhere.
4. **`src/server/ctoP0Fixture.test.mjs`** — deterministic acceptance-regression fixtures for the **existing** passive-read seams (U04/U27-shaped, contract-only): correct hits from the synthetic DB through the real `searchMessages`; the production handle raises `SQLITE_READONLY` on write; a search leaves the source row-identical; a throwing `fetch` spy proves **zero HTTP/prompt dispatch on read**; missing DB degrades to `supported:false`, never a fake result.

## Gates

- `npm run typecheck` — clean
- `npm test` — 4030 pass / 0 fail (incl. the 6 new tests)
- `bash scripts/check-duplication-gate.sh origin/main` — PASS (no clones)
- `git diff --check` — clean; `website/sitemap.xml` test artifact reverted, not committed

## Unverified (probed-by-schemа only, per task)

- No live model session was started: `messageID` honoring and `metadata` round-trip are schema/source-verified but not behavior-verified on this box — each needs one isolated-session probe before P3a depends on it.
- opencode `session_input` admission ledger: schema present, observed empty — nothing reads it as evidence.


antoinedc:
P0 approved by independent Astra review at exact HEAD 2cbaf022b307a033545d00834083e30b44dd564c. Fixture cleanup verified; project identity claims explicitly separated into observed/schema/unverified. Required CI and duplication checks pass. User authorized landing foundations. No runtime feature or rollout claimed.

PR #1507, opened 2026-09-15
feat(cto): durable work records and operation receipts (P2a)

antoinedc:
## P2a — durable work-envelope store + operation receipts (contract-only)

Unified CTO spec §5 (work record), §6 (state/stage vocabularies), §8.1 (operation protocol), §14 (U10/U13/U19), §15 (P2 gate). Baseline: PR base `feat/cto-p0-contracts` @ 0b2f7298.

## What ships

- **`src/server/ctoWork.mjs`** — stateless service over the injected store: `createWork` / `getWork` / `listWorks` / `reviseWork` + `reserveOperation` / `recordOperationOutcome`.
- **`src/server/ctoStores.mjs`** — two additive exports only: `workStore` (dir store, one JSON per envelope under `cto/work/<id>.json` — the atomic-write boundary is the single envelope) and `lockForStore` (existing per-path mutex map, reused so serialization stays one primitive per path).
- **`src/server/ctoWork.test.mjs`** — 33 tests incl. U10/U13/U19 contract coverage.

## Contract semantics

- **Envelope**: spec/attempt/decision/resource/evidence refs, closed enums (`WORK_STATES`/`WORK_STAGES`/`WAITING_REASONS`/delivery-target kinds), monotonic spec revision+hash (hash change without bump rejected; reprioritization never touches spec), acyclic dependency graph, waiting⇔waitingReason invariant. Invalid updates throw before any write; corrupt store fails visibly on get AND list (never an empty success); future `v` fails loudly via the shared migration harness.
- **Receipts**: idempotency key + canonical-args hash (key-order independent) + expected work revision. Same key+same args → replay of the prior operation with its original result; same key+different args → `idempotency_key_args_mismatch`; concurrent same key → exactly ONE reservation (live lease replays; expired lease recovers once with `takeoverCount`); `unknown` outcome forbids blind re-dispatch (`external_outcome_unknown`) until a definitive record resolves it; result recorded against a changed spec is marked `superseded` and does not advance the work (U13).
- **Revision semantics**: `work.revision` is the work-state revision (moves on create/revise only); receipt bookkeeping never bumps it — that is what makes a same-key retry carrying `expectedRevision` replay instead of self-conflict. Each receipt snapshots `workRevision`/`stage`/`specHash` for staleness checks.
- **Bounds**: `listWorks` returns `{works,total,limit}` (default 100, hard cap 1000 — truncation visible); terminal receipt history capped at `OPERATIONS_KEEP` (20, pruned oldest-first at write time); pending/in_flight/**unknown** receipts are never evicted.
- **Crash**: service is stateless; a fresh instance over the same store replays receipts (tested).

## Contract-only boundaries (explicitly NOT in this PR)

- **No target resolution**: `project` is a caller-supplied, structurally validated `ProjectRef` (non-empty ids + absolute `repositoryRoot`). Per `docs/cto-implementation-map.md` §4 the `ProjectRef` mapping is UNRESOLVED — nothing settles opencode `project.id` into `workspaceId`, no checkout/repository/workspace conflation, no fs/tmux/network verification (existence revalidation is external, P4).
- **No worker side effects**: no worktree/window/session/prompt creation; receipts' `externalRef` stays caller-supplied; no dispatch/pause/cancel/answer_decision control methods; no §6 transition-precondition enforcement (P4 coordinator).
- **No new stores/platform beyond the one `work` dir store**; no sweeper wiring, no HTTP/UI, no scheduler, no delegates.
- U10/U13/U19 coverage is **contract-only** (state/receipt interfaces), not integrated — per spec §15 no integrated claim is made.

## Verification

- `npm run typecheck` ✅
- `npm test` ✅ 4066 pass / 0 fail (vitest + node:test, sandboxed)
- `bash scripts/check-duplication-gate.sh origin/feat/cto-p0-contracts` ✅ PASS (3 changed files, no clones)
- `git diff --check` ✅ clean; working tree clean (test-regenerated `website/sitemap.xml` restored, not committed)

Head: 56c26c76c77602e061386592d7c7f784493fb279


antoinedc:
Astra blocker fixes pushed (6f9cbf1), same contract-only scope, no dispatch/UI:

1. **Reserve state machine** — terminal receipts ALWAYS replay (expired lease never resets succeeded/failed to pending); expired **in_flight** becomes a durable `unknown` (`reconcileReason: "lease_expired"`) and re-reserving throws `external_outcome_unknown` until reconciled — never re-executed; expired **pending** still recovers (P4 protocol: in_flight is recorded before the first external effect).
2. **Retention preserves dedupe** — receipts evicted from the bounded full-terminal history become **tombstones** (key + argsHash + op + original outcome/resultCode/externalRef); same-key retry after the 21st operation REPLAYS from the tombstone, different args still mismatches, tombstones capped at `TOMBSTONE_KEEP` (4×).`OPERATIONS_KEEP`.
3. **Dependency cycle race** — all graph mutations (validate + commit) serialize under one shared module-level graph lock, then the per-file lock; documented order, no path holds a file lock and waits on the graph; concurrent A→B/B→A → exactly one commit, cycle never persists.
4. **Create race** — absence-check + first write under the same per-file lock; concurrent same-id creates: exactly one wins, no overwrite erasure (deterministic latch test included).
5. **listWorks/strict load** — ENOENT ⇒ legitimately empty; ENOTDIR/EACCES/other ⇒ `store_unavailable` (visible), in both the dir listing and non-ENOENT envelope reads.
6. **Structural load gate** — envelope + receipts fully validated on every load (required fields, closed enums, receipt statuses, tombstone shape) before read/mutate/prune; an unclassifiable receipt status refuses to load so retention can never treat unknown as terminal.
7. **Hash binds op + args** — same key, same args, different op ⇒ `idempotency_key_args_mismatch`, not a replay.
8. **Stale results on replay** — spec change durably marks old-spec receipts `superseded` at revise time AND record-replay recomputes staleness before returning: success-under-A + revise-to-B + redelivery ⇒ `superseded: true`, never false.

Gates (local, HEAD 6f9cbf1): `npm run typecheck` ✅ · `npm test` ✅ 4078/0 (45 in ctoWork.test.mjs) · strict duplication gate vs `origin/feat/cto-p0-contracts` ✅ no clones · `git diff --check` ✅. Still contract-only: no worker side effects, no fake integration assertions; `ProjectRef` stays structurally validated and unmapped per impl-map §4.

antoinedc:
Astra round 2 fixes pushed (a7ac403) — simplification first, tombstones removed entirely:

1. **Retention = one hard cap, no eviction.** All compaction/tombstone layers deleted (unmerged schema, no users). `HISTORY_CAPACITY` = 256 durable receipts per envelope; at capacity a NEW unique key is refused (`history_capacity`) while identical retries of existing keys and all reads work forever. Every receipt keeps its full `specHash` binding and outcome indefinitely.
2. **Explicit `RECEIPT_TRANSITIONS` matrix** routes every outcome record: pending→in_flight/succeeded/failed/unknown; in_flight→succeeded/failed/unknown; unknown→terminal with REQUIRED evidence (`resultCode`); terminal immutable replay. No status can transition back to pending — reserve recovery re-arms the lease in place (a re-arm, not a transition); expired in_flight goes to unknown, never a downgrade.
3. **Stale-spec loss eliminated with the compaction it depended on** — revise marks superseded durably; receipts keep specHash forever (tested: survives revise + reload).
4. **Strict persisted receipt shape** now requires specHash, stage (enum), lease (owner/expiresAt); validates-when-present: resultAt, externalRef, resultCode, takeoverCount, superseded, and JSON-safe stored args. Missing safety fields ⇒ `store_corrupt`.
5. **Lease inputs validated before any write**: non-empty `leaseOwner`; `leaseTtlMs` positive finite integer ≤ documented `MAX_LEASE_TTL_MS` (24h) — string/NaN/0/negative/∞/1.5/over-max all rejected with nothing written.
6. **Args snapshot**: synchronously validated (plain JSON only, cycles detected, undefined/NaN/Dates/functions/class instances rejected by name) and deep-cloned ONCE before the first await — hash AND stored receipt derive from the same frozen snapshot; caller mutation during the in-flight reserve can never desync them (regression test mutates the object mid-reserve).

Also: the fake file-existence latch is gone — deterministic interleaving now uses an **injected observable store** (committed-write counting + `waitForWrites(n)`) in both latch tests. Fixture teardown in `after()` removes only this file's sandbox files (nothing else on the disk touched, no installs).

Gates (HEAD a7ac403): `npm run typecheck` ✅ · `npm test` ✅ **4082 pass / 0 fail** (49 in ctoWork.test.mjs) · strict duplication gate vs `origin/feat/cto-p0-contracts` ✅ no clones · `git diff --check` ✅. Contract-only boundaries unchanged: no dispatch/UI, no worker side effects, `ProjectRef` structurally validated and unmapped.

antoinedc:
Astra round 3 (final two blockers) pushed — 778136f:

1. **Outcome fields validated BEFORE any write, all paths** — `resultCode`/`externalRef` must be string-or-null (checked upfront in `recordOperationOutcome`, before the lock), including the unknown→terminal reconciliation path. The write-then-strict-loader-corrupt inconsistency is gone; regressions assert the store is byte-for-byte unchanged after each rejection (committed-write counting) and that reconciliation never half-applies.
2. **Exact canonical args contract** — `canonicalArgsHash` hashes exactly the value passed (no null→{} coalescing); omitted args ≡ `{}` (reserve-omitted then retry-`{}` replays — tested); an explicit `args: null` is rejected deterministically (it would hash as `{}` while persisting as `null`); nested nulls are ordinary JSON, persisted verbatim and hashed distinctly from `{}` (tested: `{x: null}` retry replays, `{}` mismatches).

Environment: work restored to the persistent checkout `/home/dev/projects/cto-p2a-work` (HEAD a7ac403 survived the /tmp sweep); `node_modules` reuses the existing `better-ui` install (identical lockfile) — no installs, no dangling symlinks; fixture teardown unchanged.

Gates (HEAD 778136f): `npm run typecheck` ✅ · `npm test` ✅ **4084 pass / 0 fail** (51 in ctoWork.test.mjs) · strict duplication gate vs `origin/feat/cto-p0-contracts` ✅ no clones · `git diff --check` ✅. Contract-only boundaries unchanged: no dispatch/UI, no worker side effects, `ProjectRef` structurally validated and unmapped.

antoinedc:
Independent Astra review approved contract foundation and reconfirmed exact merge head 780c8ea8b8fc786ce5f37807a25af7231e0e0f45. All blocking findings resolved: immutable receipts and bounded admission without dedupe eviction, uncertain effects require reconciliation, atomic create/graph validation, validated outcome fields and request snapshots. Required CI and duplication pass. Contract-only; no live dispatch/UI/deployment.

PR #1508, opened 2026-09-15
feat(cto): passive historical context reads (P1a)

antoinedc:
## Summary

P1a slice of the unified-CTO spec (`docs/unified-cto-spec.md` §4.2) — a minimal **read-only** `ctoContext` service over opencode's own SQLite store, contract-only (no rpc/tool/UI wiring yet; nothing is exposed to a model). Base is `feat/cto-p0-contracts` so the diff contains only this PR's changes. **Astra review fixes (blockers on 1e8ce2c9) landed in 482581ab — same scope, 3 files.**

Three bounded operations (spec §4.2, source-direct — the FTS index is explicitly P1b and `coverage.indexed:false` says so):

| Op | What it does |
|---|---|
| `ctoListSessions` | discovers ALL historical sessions — closed/archived and child sessions included, parent links + observed identity + timestamps preserved; filters by observed `projectId`/`directory`; `includeArchived:false` excludes closed |
| `ctoSearch` | bounded text + tool-evidence search; tool hits identify the tool name/status and which field matched (`input`/`output`/`tool_name`) — evidence distinct from prose (U05) |
| `ctoAround` | chronological message neighborhood of an anchor message with stable source IDs; per-part evidence caps; anchor evidence takes budget priority; bounded part reads |

## Contracts pinned

- **Observed identities only.** `projectId`/`workspaceId` pass through the DB's `project_id`/`workspace_id` verbatim as OBSERVED values; every row is marked `projectMapping:"unmapped"`. Per the **revised** P0 map (2cbaf022), project-id derivation semantics are UNVERIFIED — no per-checkout/repo inference is made or claimed. A caller passing a `workspaceId` key gets `invalid_input`, never a reinterpretation; a DB projectID is never surfaced *as* a workspace id.
- **Honest degradation (U27), cause preserved.** `unsupported` (no node:sqlite — accessor records the import failure via the additive `getDbOpenFailure()`), `source_unavailable` (no DB path / open failure / query failure), `reference_expired` (unknown session/message, distinct detail), `invalid_input` — all distinct, with strict independent tests. An empty source answers `status:"ok"` WITH `coverage`/`observedAt`, never a fabricated result.
- **Scan-cap cursors never end the walk falsely.** A page whose full 800-row scan window is discarded/omitted advances the cursor past the last CONSUMED candidate (`truncated:true`); an early break (output cap / budget) keeps the cursor at the last EMITTED item so unconsumed candidates stay reachable; a sub-cap window provably ends the walk.
- **Decoded evidence stays searchable.** Candidate selection matches the raw stored JSON AND its JSON-escaped query form — quotes, backslashes, newlines (e.g. a Windows path stored as `C:\\Users`) are reachable; parameterized only, no FTS, no index changes.
- **Server-side limits, independent of caller args (spec §4.2).** Hit caps (20 default / 50 max); `around` before/after default 5, ZERO valid (anchor only), 40-message cap INCLUDING the anchor; the **24 KiB budget measured on the ACTUAL serialized response** (every text field — snippets, titles, directories, tool names, message records — counts) with per-field caps + omitted-size metadata (`<field>Truncated`/`<field>SourceBytes`/`<field>ReturnedBytes`, `omittedCount`, `partsOmitted`); items that cannot fit are omitted and counted — stable IDs are never emitted in a uselessly cut form. Linear UTF-8-aware truncation.
- **Zero side effects on read (U04, contract-only).** A throwing `fetch` spy observes zero calls across all three ops; source row counts and `sqlite_master` are unchanged; the only I/O is the shared read-only handle.

## Fixture (additive only)

`fixtures/opencodeDbFixture.mjs`: the `session` table gains the observed v2 columns ctoContext reads (`project_id`, `workspace_id`, `title`, `time_created`/`time_updated`/`time_archived`) — **all nullable**, so pre-existing seeds and the P0 tests are untouched (9/9 P0 fixture tests pass). The P0 `withFixtureDb` wrapper is reused unchanged.

## Test results

- `npm run typecheck` ✅
- `npm test` ✅ — 200 vitest files (3915 pass / 7 skip) + 4051 node:test (18 `ctoContext.test.mjs`, incl. all review regressions), 0 fail; generated `website/sitemap.xml` test artifact reverted, not committed
- `bash scripts/check-duplication-gate.sh origin/feat/cto-p0-contracts` ✅ PASS (strict, no clones)
- `git diff --check` ✅ clean
- node:sqlite cases skip explicitly on unsupported runtimes; the file imports safely on Node 20 (dynamic imports only)

## Acceptance coverage (contract-only per spec §15)

- **U03** — the decision that exists only in the closed child session is found and resolved to exact `sessionId`/`messageId`/`partId`
- **U04** — zero prompt/window/job side effects observed (throwing fetch seam), source untouched
- **U05** — tool evidence with matchedField; MB outputs truncated with omitted-size metadata
- **U27** — missing source / empty source answered honestly, never fabricated

## Out of scope (deliberately)

FTS index + performance work (P1b), `projects`/`resolve_refs`/`project_state` ops, full project-identity settlement, rpc/tool/UI exposure, cursor persistence, live-source integration runs.


antoinedc:
Second-review fixes pushed (27df5d15), same scope — 2 files (service + tests; the accessor/fixture unchanged from the first review round):

1. **Scan-cap/cursor + oversized first item** — the 24 KiB budget is now ONE authoritative helper measuring the COMPLETED response's serialized size (envelope, markers, cursor included; iterated until stable). No fixed 96/128-byte slack guesses and no parallel per-field charges — those flawed helpers were removed. Escape-heavy oversized items (e.g. a 30000-newline title) are serialized-aware-bounded and RETURNED with omitted-size metadata (stable IDs never cut); anything that still cannot fit is omitted + counted and the cursor rules guarantee older rows stay reachable (discarded-only full window advances; early output-cap keeps retry reachability).
2. **Fair-share bounding** — an overflow no single field can cover is spread proportionally to field size (small evidence survives; the anchor keeps its evidence; the largest field gets a content-maximal exact cut first). `around` with 40×1000-char messages now serializes inside 24 KiB (was 25279).
3. **Serializer-independent matching** — candidate selection matches DECODED fields in SQL (`json_valid` + `json_extract` + `instr`, ASCII-case-folded), no raw-JSON prefilter; `café` stored as `caf\u00e9`, escaped solidi and surrogate-pair escapes are all reachable. Parameterized, no FTS, JS matching authoritative. Regression seeds raw stored bytes via direct inserts (forms `JSON.stringify` never emits).
4. **partId anchor** — `around({sessionId, messageId, partId})` centers the anchor message's evidence window on that exact part via keyset reads (no full part load, capped + `partsOmitted` reporting). A search hit beyond the first bounded part read is now fetchable; unknown part refs are `reference_expired`.

Gates: typecheck ✅, full `npm test` ✅ (3915 vitest + 4055 node:test, 0 fail), strict duplication gate vs base ✅, `git diff --check` ✅. 22 ctoContext tests incl. 5 new adversarial regressions. Test temp dirs torn down; no deps installed; no unrelated data touched.

antoinedc:
Third-review fixes pushed (dd3ee155), same scope — 2 files (service + tests):

1. **By-reference compaction + protected anchor evidence** — the 8×50 repro fixed: compaction units now remove their OWN object by reference/id (splice by indexOf) instead of blind pops that removed the anchor message while compacting an earlier one. The anchor message has no drop unit; the requested anchor part's evidence is `canDrop:false` — fair-share bounding draws from unprotected evidence first, and the protected text is bound only as a last resort (never omitted). Anchor survives with its requested evidence in every case.
2. **Nested structured input** — `json_extract` on a nested object returned that object's JSON text (escapes intact), so `café` (stored `caf\u00e9`) or a Windows path inside `state.input.command` stayed unreachable. Candidate selection now matches the DECODED scalar atoms via `json_tree` (`json_valid`-guarded, ASCII-case-folded, parameterized, no FTS), and the JS candidates align — structured input/output contribute their decoded scalar strings (bounded recursion), never a `JSON.stringify` re-serialization. Regression seeds object-shaped inputs with alternate stored serializer escapes via direct inserts.
3. **Aggregate truncation honesty** — a per-part pre-cap (single MB output) flagged the item but not the aggregate when the budget helper had nothing more to do; item-level cuts now propagate to `result.truncated`. The anchored part window also reports `partsOmitted` honestly when the centered window leaves parts outside it (one bounded COUNT, no part loads).

Gates: typecheck ✅, full `npm test` ✅ (3915 vitest + 4058 node:test, 0 fail), strict duplication gate vs base ✅ (extracted the shared binary-search into one helper), `git diff --check` ✅. 25 ctoContext tests (+3 adversarial). Worktree moved to the persistent path `/home/dev/projects/cto-p1a-history`; reused the existing `node_modules`; no deps installed, temp dirs torn down.

antoinedc:
Final-review fixes pushed (c475b180), same scope — 2 files (service + tests):

1. **Matched-atom candidates (SQL json_tree)** — the capped JS traversal (32 scalars / 64 nodes / depth 4) could hide a true hit the SQL had already selected: 32 filler scalars + a decisive 33rd yielded cursor-advance-then-exhaustion with no hit. The candidate handed to JS is now the **bound matched atom itself**, derived in SQL — the first matching `json_tree` atom's `fullkey` (part+field provenance via `matchedFieldOf`: text / tool_name / input / output) and its `json_extract`-decoded value. Works at any depth and across any stored serializer escape form, bounded by the SQL LIMIT — no huge JS tree parse. `buildHit` verifies the query inside that atom and builds the snippet from it; part-shape filters (synthetic/ignored) still apply. `partCandidates`/`scalarTexts` remain only for `around`'s evidence pickup.
2. **Protected anchor excluded from every nonfinal compaction pass** — phase 1 (largest-field-first exact cut) previously included the anchor's text, cutting a 5988-byte decisive tail to ~4521 while 25 neighbors of 650 bytes stayed whole. Protected fields are now excluded from phase 1 AND phase 2 (declaration hoisted, duplicate removed): neighbor fair-share/drops relieve the overflow first, and the anchor text is trimmed only in the final protected pass when no neighbor shrink can satisfy the budget. Regression pins the exact tail: anchor text untouched (5988 bytes, decisive tail intact) while neighbors are bounded.

Gates: typecheck ✅, full `npm test` ✅ (3915 vitest + 4060 node:test, 0 fail), strict duplication gate vs base ✅, `git diff --check` ✅. 27 ctoContext tests (+2 adversarial). Prior fixes preserved (all previous regressions green).

antoinedc:
Fifth-review fix pushed (9f1ab425), same scope — 2 files (service + tests):

**Eligibility BEFORE first-pick, one shared rule** — the first-matching-atom pick scanned the whole tree in document order, so metadata/reasoning atoms shadowed the true output. A single `ELIGIBLE_PATH_CLASSES` rule is now emitted into BOTH the row-candidacy `EXISTS` and the first-pick subquery:
- text parts → exactly `$.text` (never reasoning, never metadata);
- tool parts → exactly `$.tool`, or `$.state.input` / `$.state.output` as a scalar or descendant — boundary `.` or `[` (a bare startswith would admit `$.state.inputSummary`);
- `state.metadata.output` is a streaming transient that does not persist in source rows — intentionally declared non-eligible (documented in the rule).

`matchedFieldOf` mirrors the same class list, returns null for unknown paths, and `buildHit` rejects them — never guessed. Negative regressions: text metadata-only nohit, reasoning nohit, tool metadata shadowing (a shared term picks the REAL output atom with `output` provenance and the correct snippet; a metadata-only term nohit), `inputSummary` nohit; the deep/33rd-input cap-crossing tests remain green.

Gates: typecheck ✅, full `npm test` ✅ (3915 vitest + 4061 node:test, 0 fail), strict duplication gate vs base ✅, `git diff --check` ✅. 28 ctoContext tests.

antoinedc:
Sixth-review fix pushed (7a6ef03d), same scope — 2 files (service + tests):

**Descendant eligibility = exact prefix AND boundary, both legs, JS and SQL.** The boundary check examined only the character at the expected offset and never the prefix, so a same-offset decoy (`$.metadata.xy.note` — character 14 is `.`) was accepted as a `$.state.input` descendant: a metadata-only needle fabricated a hit, and a decoy before the real output shadowed it. Fixed in BOTH:
- JS: `fullkey.startsWith(prefix)` (case-sensitive) AND `fullkey[prefix.length] ∈ {'.', '['}`;
- SQL: `fullkey = prefix OR (substr(fullkey, 1, len) = prefix AND substr(fullkey, len+1, 1) IN ('.', '['))` — literal constants derived from the frozen class list, parameter-safe.

The persistence note is narrowed per review: `state.metadata.output` is **excluded from the completed-evidence contract**, not claimed never-stored.

Self-inspection before tests: 16 counterexamples run against both the JS predicate and the emitted SQL (exact, descendant `.`, descendant `[`, same-offset decoy, boundary-letter decoys `inputSummary`/`inputs`/`outputText`, case sensitivity `$.STATE.INPUT`, exact-only classes, unknown part type) — all pass, and the old check provably accepted the decoy. Live SQL probe on a decoy fixture: decoy-only pick = [], shared-term pick = the real output atom with `$.state.output` provenance.

Negative regression: the `$.metadata.xy.note` decoy alone nohit; placed before the real output it never shadows it (matchedField `output`, snippet from `state.output`). Gates: typecheck ✅, full `npm test` ✅ (3915 vitest + 4062 node:test, 0 fail), strict duplication gate vs base ✅, `git diff --check` ✅. 29 ctoContext tests.

antoinedc:
Independent Astra review approved exact head 551d28630761b5d07d39298d9bd01dd98bdefc93. Prior blockers corrected and adversarial fixture checked: bounded response with anchor preservation, source/cursor reachability, decoded evidence-only matching and failure distinction. Required CI and duplication pass. Passive read foundation only; no chat/tool UI wiring or FTS indexing claimed.

PR #1509, opened 2026-09-15
feat(cto): durable singleton conversation binding (P3a1)

antoinedc:
## P3a1 — durable singleton CTO conversation binding (unified-cto-spec §3.1, step 1)

Implements ONLY the binding concept as an injectable service — no admission, no dispatch, no UI, no public route. `src/server/ctoBinding.mjs` exports `createCtoBinding({ oc, store, provenanceStore, controlDir, ... })` → `{ ensure, recover, getBinding }` for the future authorized UI caller.

### What it does
- **One durable role session** in a stable server-owned control directory under the state home (`~/.manta/cto/conversation` — 0700, marker file, refuses `.git`, never a repository target).
- **Versioned binding record** in the new strict `ctoStores.bindingStore`: `generation`, `currentSessionId` + `currentOperation` (exact identity), capped `previousSessionIds` (archive references; sessions are never deleted here), and `pendingOperation`.
- **Reserve before create**: the operation marker is persisted before `POST /session`; the marker rides on the session's `metadata` (additive `createSession` option — P0 §8-proven round-trip). A crash after the remote create but before the bind is recovered by matching the EXACT `metadata.bindingOperation` (direct read-back verification — never title, never role alone).
- **Unknown-outcome discipline**: a thrown create keeps the reservation; every `ensure` attempt reconciles fresh store state by marker before retrying (bounded attempts, then a loud `CtoBindingError` — never a blind duplicate). Replacement of a bound session happens ONLY on definitive 404; a 5xx/timeout returns the binding with `uncertain` — and a full newest-100 list page of entries newer than the reservation is uncertainty, not absence (principled page-cap rule, no timeout heuristic).
- **Concurrency**: concurrent `ensure()` calls join one singleflight (one creation); the reservation is a store-local mutex scope, never held across the external call.
- **Provenance, minimally**: the durable session is registered ONLY in the never-swept `internalSessions` tombstones (the existing reader seam) — its title deliberately avoids the ephemeral reaper's `cto:` prefix and it is never registered with any TTL/expiry. No broad rewiring; the four-way vocabulary lands with admission.

### Additive seams
- `opencode.createSession`: optional plain-object `metadata` forwarded on the wire (validated; rejected before any fetch otherwise).
- `opencode.readSession(id)`: NEW three-state read — `found` (full record incl. metadata) / `missing` (definitive 404) / `unknown` (5xx, network). `sessionExists` untouched.
- `docs/cto-implementation-map.md` §10 addendum.

### Not in this PR (later phases)
Prompt admission (P3a2), dispatch, any route/UI wiring, the delegate extension, session deletion/cleanup (P6), four-way provenance vocabulary, destructive reset (deliberately not exposed).

### Verification
- `npm run typecheck` ✅ · `npx vitest run` 3915 passed ✅ · `npm run test:server` 3593 passed ✅ · duplication gate PASS (no clones)
- New `ctoBinding.test.mjs` (28 tests): production composition driving the REAL opencode.mjs client via `_setOcTransport` (marker on the wire + read-back + 3-state read), singleflight, crash-after-create adoption, decoy rejection, bounded unknown-state retries, provable-window absence, replacement generations w/ history, reaper-prefix safety, provenance tombstone, control-dir discipline, corrupt-store loudness.

Base: origin/main @ 7c154df5. No merge, no live sessions, no provider changes.


antoinedc:
P3a1 review blockers addressed in 9f1134f1 (same scope — no UI, no admission wiring):

1. **Shared singleflight + CAS**: ensure()/recover() share one store-keyed flight across engine instances; reservation + bind are CAS'd against expectedCurrentSessionId/expectedGeneration (stale missing lookup after a prior bind now defers — reproduced in test). Unknown create outcome retains the reservation and errors explicitly — a marker-absent scan NEVER clears/re-creates (late-landing reproduction test); only a definitive 4xx (createSession errors now carry .status) clears its own reservation.
2. **Unsupported identity**: a create response missing the marker persists the created sid + terminal unsupportedIdentity — explicit error on every later call, zero further creates/scan-adopts. Malformed list responses are uncertainty, never []/absence.
3. **Deadlines**: every list/read/create call races an AbortSignal through the actual transport; hung-headers and hung-body tests pin bounded failure; a timed-out create is unknown (retained), not a retry.
4. **Provenance**: the durable conversation is no longer in the internalSessions tombstones; distinct role lookup (readConversationRole from the binding record, checked first in resolvePipelineSession — cto_conversation vs cto_internal), engine recognizes a human CEO instruction as CEO presence; conversation still never feeds evidence/segmentation (no recursive self-summary). No cycle (ctoBinding no longer imports internalSessions). Event-provenance tests included.
5. **Strictness**: strict stores throw on top-level null/array/string (load+loadSync); only a missing file initializes the default.
6. **Archive**: previousSessionIds never capped/dropped (12-generation test), getBinding paginates; control dir chmods 0700, walks ancestors to the state home for .git, refuses symlink redirects via realpath containment.

Gates: typecheck ✅ · vitest 3915 ✅ · node tests 3602 ✅ · duplication gate PASS (no clones).

antoinedc:
Round-2 blockers addressed in ef8c4121:

1. **Queue serialization with per-caller postconditions** — ensure/recover run through a per-store task queue (one at a time, across engine instances); an ensure() enqueued behind a no-create recover() still completes get-or-create. No recursion/deadlock. Both orders × both instance roles tested.
2. **Sid persisted before verification** — the create's returned sid lands in the pending operation immediately; recovery settles by DIRECT GET on that sid first; the marker scan runs only when no sid exists. Repro test: receipt timeout + full churn page hiding the session → adopted from the persisted sid, zero list calls at recovery.
3. **Validate-before-mutate control dir** — textual containment + realpath of the deepest existing ancestor + repository walk INCLUDING the state home run before any mkdir/chmod; the symlink test now pins the rejected target's mode unchanged and no marker written into it.
4. **Role-aware DB reader** — collectDbRows tags rows with distinct provenance (cto_internal from tombstones / cto_conversation from the binding record, never tombstoned) + message role (LEFT JOIN when a message table exists); extractFromDbRows excludes internal and conversation assistant/tool content from ordinary evidence while conversation user rows (CEO instructions) stay tagged consumable. Integration test drives the real sqlite fixture pipeline (not just events).
5. Truthful residual-unknown copy (retrying will not resolve it; no unsafe auto-retries) and an explicit single-writer-process requirement (no fake cross-process CAS claim) in the module header + implementation map.

Gates: typecheck ✅ · vitest 3915 ✅ · node tests 3606 ✅ · duplication gate PASS (no clones).

antoinedc:
Round-3 blocker fixed in 2afe9b7a: an EXISTING controlDir that is a symlink is now refused outright (`control-directory-symlink` — per spec, symlink redirects are refused; ancestor symlinks still fine under realpath containment), and an existing controlDir resolving to the state home itself is refused — the `realAnchor === realHome` equality exemption now applies ONLY to the deepest-existing-ancestor walk of a not-yet-created controlDir, so chmod can never follow a link onto the state home. A symlink-to-stateHome test pins: refused, state-home mode unchanged, no marker at the home root, zero creates; a deep fully-missing chain still creates 0700. Validation still strictly precedes every mutation. Gates: node tests 3608 ✅ · vitest 3915 ✅ · typecheck ✅ · duplication gate PASS (no clones).

antoinedc:
Independent Astra review approves exact head 2afe9b7a44f5b05982a97920c2e6e2b0ec679d65. Singleton binding recovery, shared operation serialization, unknown-create preservation, deadline handling, distinct conversation provenance, archive references and control-directory safety all reviewed. Required CI and duplication pass. Server service foundation only; no UI/admission/runtime rollout claimed. Single-writer-process and unsupported identity/unknown creation recovery limitations documented.

PR #1510, opened 2026-09-15
feat(cto): incremental passive search index (P1b1)

antoinedc:
## Summary

P1b1 — RESCOPED per parent review: **safe index lifecycle + bounded refresh + first-page search ONLY**. A Manta-owned SQLite FTS5 index over the read-only opencode source as a **factory module API** (`createCtoSearchIndex()` → `sync`/`search`/`status`/`close`). No tool, no UI, no poller wiring. **No pagination** — a `cursor` argument is explicitly rejected until P1b2.

### Parent-review blockers fixed

1. **Internal-leak (part before session mirror)** — exclusion is resolved from the AUTHORITATIVE `internalSessions` registry (IDs, never titles) at BOTH sync and search time, irrespective of mirror progress: an internal part indexed before its session mirror row exists is still filtered (adversarial fixture pinned: part ts 150 sweeps before session ts 300). Registry failure fails sync AND search CLOSED (`provenance_unavailable`, no unfiltered answers).
2. **Same-timestamp edits were never seen** — the watermark CDC is GONE. `sync()` runs a bounded CYCLIC sweep (one batch per stream: sessions, messages, parts) re-reading rows at a persisted cycling position: content edited WITHOUT moving `time_updated` and late rows with old timestamps are re-indexed (parts hash-gated; stale terms replaced). `coverage.eventual` — eventual, never claimed as a snapshot; index absence is not authoritative absence (P1a direct reads remain the fallback). Bounded existence checks reconcile deletions.
3. **Rank pagination was invalid** — pagination API deleted. `search()` returns a bounded first page with honest `truncated` + `omittedCount` (counts EVERY unreturned match); `cursor` → `invalid_input`; no `nextCursor` anywhere.
4. **Open errors mis-classified as corruption** — auto-rebuild/unlink is GONE. Locks (mid-transaction included) → retryable `index_busy`; quick_check/foreign-schema failures → explicit `index_corrupt` with the file **RETAINED** (bytes preserved, test-verified). Never auto-deletes.
5. **Connection leak** — ONE owned connection per instance, reused across every operation, explicit `close()` (ops after close fail `index_closed`); injected connection-counting constructor test pins opens===1 / closes===1.
6. **Silent 33rd-scalar drop** — own bounded extractor with scalar/node/depth bounds whose omissions are REPORTED on the sync result (`extractionOmitted`); byte caps flagged (`byteTruncated`). The index is bounded retrieval, not authoritative absence.
7. **Node 20** — lazy `import("node:sqlite")` in production, guarded dynamic import in tests (file loads and skips cleanly), and the unsupported path is ACTUALLY exercised (injected null module).

### Invariants (unchanged)

Source stays READ-ONLY (shared `opencodeDb` handle; row counts pinned unchanged; zero fetch calls on sync/search). Evidence = eligible text/tool name/input/output scalars mirroring the P1a-reviewed classes; reasoning/metadata never evidence. Everything in a sync — sweep positions included — commits in ONE transaction; a failed batch commits nothing. Literal query policy (`ftsQuery.mjs`). Honest statuses: `unsupported`/`source_unavailable`/`provenance_unavailable`/`index_corrupt`/`index_busy`/`index_error`/`index_closed`/`invalid_input`.

## Tests

15 sandbox tests in `ctoSearchIndex.test.mjs` — all 15 pass, including the parent-demanded adversarial fixtures: old/new terms at the SAME timestamp, internal session outside the mirror page, busy index preserved (bytes + data), connection count closed, extraction omission count/coverage, rejected cursor, Node-20 degrade actually run, corrupt-index retention, failed-batch sweep-position safety, byte caps, limit clamps, literal query policy, identity filters, source-untouched + zero-fetch.

## Limitations (honest)

- **No performance claims.** The optional 100k reference benchmark was NOT run (stopped by parent); real-store `time_updated` index assumptions are unverified — performance validation is **P1b2**.
- Cyclic refresh is EVENTUAL: a row becomes searchable within one sweep cycle, not synchronously.
- `index_corrupt` recovery is manual (future work); this PR only reports and retains.
- Line accounting: 712 module + 43 helper + 584 tests = 1339 (~30 over the prior slice) — fewer mechanisms, but seven new adversarial scenarios.

ctoContext P1a is untouched (only imports its exported `partCandidates`-contract classes via the extraction parity). No merge, no live state/services.


antoinedc:
Astra review pushed (head 7f327cda) — all five blockers fixed, no scope expansion:

1. **Rotating deletion verification** — persisted per-table rowid keyset cursors (wrap at table tail) replace the always-first-64/16 sample; a part deleted past the first batch is eventually absent within documented cycles (test: 100 parts, delete p80/p99 → gone in 3 syncs; ceil(100/64)=2 cycles + margin).
2. **Ownership before ANY write** — canonical path (realpath incl. symlinks) checked against the source DB; existing files validated READ-ONLY first (quick_check + `PRAGMA application_id` "MANT" + schema_version). Foreign/source DBs → explicit `unowned_index`, byte/row/schema identical (source-path-as-index repro pinned). Fresh init atomic (application_id + DDL + marker in one transaction).
3. **Singleflight init** — one opening promise shared by concurrent ops (3 concurrent statuses = 1 handle, was 3); close() during the await closes the handle exactly once at settlement and in-flight ops report `index_closed`, never a late success.
4. **Persisted incompleteness** — per-document `incomplete` flag (schema v2) for depth caps + extraction omissions; `incompleteParts` reported index-wide by sync/search/status even on zero-hit searches; reloaded by a new instance; unchanged batches report numeric 0 (no NaN).
5. **Node 20** — unsupported path needs no fixture DB (old test spun one via node:sqlite itself; deleted). Actually exercised on the box's existing nvm **node v20.20.2**: 2 pass + 17 clean skips, 0 fail. No install, no fabricated claim.

19/19 focused tests (5 new adversarial); full suite **4132 pass / 0 fail**; typecheck clean; jscpd 0 clones involving these files. No benchmarks, no new features.

antoinedc:
Independent Astra review approves exact head 7f327cda30df81cf30dd71fd43cf23d72a9be9b4 for the explicitly reduced P1b1 scope: safe index lifecycle, bounded cyclic refresh and first-page search. All prior blockers corrected; reviewer reran 19 targeted synthetic tests. Required CI and duplication pass. Stable pagination, runtime wiring and performance benchmarking remain separate work; no production performance claim or deployment.

PR #1511, opened 2026-09-15
feat(cto): expose passive project-context reads (P1c)

antoinedc:
## Summary

P1c (unified-CTO spec §4.2/§7): expose the reviewed P1a source-read operations (ctoContext.mjs) as three read verbs in the EXISTING on-call CTO read engine (`createCtoEngine` dispatch seam, served by the existing dynamic `POST /api/cto` route). Small wrapper only — no new data framework.

## Public tool contract

| Verb | Args (validated by the existing ctoContext methods, verbatim pass-through) | Output |
|---|---|---|
| `context_projects` | `projectId?/directory?` (observed source ids), `includeArchived?`, `limit?`, `cursor?` | all historical sessions (closed/archived/child included) with stable source ids |
| `context_search` | `query`, `projectId?/directory?/sessionId?`, `limit?`, `cursor?` | bounded newest-first text + tool-evidence hits (direct source reads, no FTS/ranking index) with stable session/message/part ids + snippets |
| `context_around` | `sessionId`, `messageId`, `partId?`, `before?`, `after?` | chronological neighborhood with stable ids, anchor flagged |

All three return `{ok:true, data:{status, supported, coverage, observedAt, …rows, nextCursor, truncated, omittedCount}}` — the ctoContext envelope itself. `status` keeps ok / invalid_input / unsupported / source_unavailable / reference_expired DISTINCT (spec §4.2 degradation semantics); coverage reports `mode:"direct-source", indexed:false` (P1b index intentionally NOT wired; no poller/FTS in this PR).

## Invariants preserved

- Read-only: no model turns, no binding.ensure/create, no prompt dispatch / session / window / job creation — validated by a throwing `fetch` spy at zero calls plus unchanged source rows/schema (U04 at the engine composition boundary), and live tmux/opencode read seams injected as throwing spies.
- Observed identities only: opencode's own `project_id`/session/message/part ids carried verbatim with `projectMapping:"unmapped"`; a Manta workspace key (`workspaceId`) as a filter is rejected with `invalid_input` — never silently reinterpreted. No stable project/workspace mapping is guessed.
- No duplicate helpers: caller args are forwarded verbatim so cursor/limit/forbidden-key validation and the 24 KiB serialized budget come from the ONE existing implementation.
- index.mjs untouched (the dynamic `/api/cto` dispatch picks the verbs up automatically); native UI unaffected. `docs/opencode-tools/cto.ts` lists the new verbs (with the same newest-first wording) but is NOT copied to live config and opencode-serve is NOT restarted.

Astra round 1: search-verb descriptions corrected — ctoContext orders newest-first (keyset-cursor pagination) with NO ranking/FTS; wording now says bounded newest-first direct-source. No registrar contract test added: no existing docs-sync test style exists for docs/opencode-tools and the reviewer marked it out of scope.

## Tests

New `src/server/ctoContextWiring.test.mjs` (10 cases, all contract-only): the REAL engine + REAL ctoContext service over the P0 fixture DB (`MANTA_OPENCODE_DB` armed via the P0 wrapper before any handle opens) — registry shape, closed-child discovery via search+around (U03 seam), observed-id passthrough, invalid args (empty query / garbage cursor / forbidden workspace key / missing around refs), server-side limit clamp + keyset-cursor paging with zero duplication, per-part truncation metadata propagation, distinct `source_unavailable` vs `unsupported`, `reference_expired`, zero-side-effects + no-secrets.

`cto.test.mjs`: registry count/list updated (23 → 26).

## Verification

- `npm run typecheck` — pass
- `npm test` — 4185 pass / 0 fail
- `bash scripts/check-duplication-gate.sh origin/main` — pass
- `git diff --check` — clean (test-generated sitemap timestamp restored)

Head: 699bc440caebab1004f458396c87ad193f91520e


antoinedc:
Independent Astra review approved exact head 9b3418c0d3019c0aea20e6f4b1971a942eabffa6. Bounded newest-first direct-source context reads exposed through existing read engine; no FTS/performance claim. Required CI and duplication pass. No live registrar install or deployment.

PR #1512, opened 2026-09-15
feat(cto): durable conversation admission (P3a2)

antoinedc:
Spec §8.3 (P3a2) — parent commit `96d062d9` (P3a1 binding) verified as base.

## What
`src/server/ctoAdmission.mjs` — the ONE durable admission path for every prompt to the CTO role session (human + background). Built alongside `promptDelivery` (shared busy view via `isBusy`, same firehose tap); ordinary project delivery untouched. No routes, no UI, no poller — parent integrates per `docs/cto-admission-contract.md`.

## Contracts landed
- **Durable before send**: stable event id + canonical payload hash + origin + expected binding generation persist at submit; dispatch persists `sessionId` + allocated opencode `messageID` before the POST. No store lock across any opencode await (sync-mutator patchStore sections; single-flight, joinable pump).
- **One turn at a time**, human FIFO over queued background, accepted turns never reordered.
- **Ack ≠ completion**: 204 + P0-proven messageID receipt → `accepted` only; `completed` needs the real terminal event or a transcript proof.
- **No blind resend**: restart reconcile adopts by receipt read-back; unprovable → `unknown` (stale-flagged, gate-holding, surfaced); only a live definitive 4xx is `failed`.
- **Generation semantics**: pending retargets the current binding at dispatch (`retargeted` recorded); accepted turns keep their original sid; optional `expectedGeneration` CAS refuses stale.
- **Receipts retained forever**; hard cap on new entries (`MAX_ENTRIES` 500 → `at-cap`), never eviction.
- **Explicit interrupt** (separate op); submit never aborts.

## Additive seams (tiny)
- `opencode.sendPrompt`: optional `messageID` forwarded onto the prompt_async body (P0-proven verbatim persistence + read-back; omitted → byte-identical behavior).
- `ctoStores.admissionStore`.

## Verification (contract-only per spec §15; no integrated UI/route claims)
- `npm run test:server`: **3652 pass** incl. 26 new admission tests.
- `npm test`: **4201 pass**; `npm run typecheck`: clean.
- Duplication: jscpd over the changed set — 0 clones in new files (`opencode.mjs` is on the documented ignore list; its 2 intra-file clones pre-date this PR).
- Production composition test drives the real `opencode.mjs` client through `_setOcTransport`: messageID on the wire + receipt read-back round-trip.
- `website/sitemap.xml` regenerated by vitest was inspected and restored; the diff contains only the six intended files.

## Known limitations
[...]

PR #1513, opened 2026-09-15
feat(cto): wire conversation admission APIs (P3a3)

antoinedc:
Follows #1509 (P3a1 binding) and #1512 (P3a2 admission) — the parent integration: the engines become the live server runtime.

## What

- **Composition** (`src/server/index.mjs`): exactly ONE `createCtoBinding` + ONE `createCtoAdmission` instance for the server lifecycle, plus a thin `createCtoConversationService` (`src/server/ctoConversation.mjs`) over them. Binding creation is **lazy** — boot performs zero oc calls (no role session, no model prompt); the first `open` creates the headless role session via the API only.
- **RPC channels** (exact contracts, shared TS types in `shared/api.ts`):
  - `cto:conversation-open` → `{sessionId: string, generation: number}`
  - `cto:conversation-state` → `{binding: {sessionId: string|null, generation: number}, submissions: CtoSubmissionProjection[], counts: CtoQueueCounts}`
  - `cto:conversation-submit` `{id?, text, expectedGeneration?, model?}` → durable admission submit receipt (dedup by stable id; `persisted:false` marks a replay)
  - `cto:conversation-interrupt` `{id}` → `{ok, id, status}`
- **Server-owned role config**: origin is fixed to "human" and the agent is always the registered central `cto` agent (`providers.mjs CTO_AGENT_NAME`) — the caller can never choose an arbitrary agent, and no role/agent is read from a request body.
- **Anti-bypass seams** (spec §8.3 — one admission path for every writer to the role session):
  - `opencode:prompt` aimed at the bound session: plain text routes through the queue with the caller's stable messageID; attachments/mentions get the clear "plain text only" rejection.
  - `opencode:run-command` aimed at the bound session: rejected with the "use the CTO conversation API" copy (slash support arrives in a later phase).
  - `promptDelivery` gains an optional `redirect`: background senders (schedules, capability jobs, webhooks, delegate completions) targeting the role session are submitted to the same queue with a stable content-mapped `bg_*` id. Ordinary project traffic stays byte-identical; classification failures fail open (warn) so CTO state can never break ordinary delivery.
  - No recursion: admission dispatches via the **raw** low-level oc client, so promptDelivery → admission → raw oc is one hop each way.
- **Event tap + poller**: the existing firehose pump now feeds BOTH engines (`promptDelivery.observeEvent` then `admission.observeEvent`, each guarded) — one tap, no second stream, no extra runs. A bounded `startPoller` (30s, label `cto-admission`) drives `admission.tick()`; failures surface via warn (never swallowed as healthy).
- **httpApi** gains the four typed methods (`httpApi.ts`) for the P3b UI worker — no renderer UI yet. Demo-coverage inventory updated.
- **Docs**: `docs/cto-admission-contract.md`'s production-composition recipe updated to the as-wired shape, and the misleading "a caller cancel releases an unknown" limitation is corrected — an interrupt on `unknown` only converts it to `cancel_requested`; the barrier is retained until reconcile proves the outcome.

## Tests

`src/server/ctoConversation.test.mjs` — the real composition (real binding + admission engines over memory stores + synthetic oc) dispatched through the real channel map:
- boot touches zero oc; `state` and an empty `tick` never create or invoke the model
- concurrent `open` creates exactly one role session (singleflight), no model invocation
- submit: stable id, queue projection (payload stripped, stable `msg_*`), dedup replay, actionable duplicate-id error
- direct-send route (server agent stamped, caller agent dropped) + attachments/slash rejections
- background delivery into the same queue (origin `background`, `bg_*` stable id, identical redelivery dedups) while ordinary traffic is byte-identical
- bounded signals forwarded to the raw send and abort; an interrupt on a live turn retains the barrier

Full `npm test` (4239) + `npm run typecheck` + the strict duplication gate pass.

Parent review gates merge.


antoinedc:
Independent Claude Opus 5 review across four rounds; merge-ready at exact head 2d309504934db840661617bd4589a5399ea91d31, verified by counterfactual (each fix reverted in a scratch copy to confirm the claimed test goes red).

Findings closed over the rounds: recurring background deliveries silently running once ever; idle webhooks bypassing admission; unseamed abort; a stale classification cache routing the live role session past its own queue; a raw-abort fallthrough cancelling a different turn; the cap refusing the human's own message; silent policy drops; a lying 202 for deliveries that will never run; and a class-1 error-body disclosure on the one public unauthenticated route.

Scope deliberately narrowed: this PR ships NO abort seaming. opencode:abort is main's existing raw abort; admission's interrupt() remains exposed but unconsumed, deferred to its own PR with the design settled first (contract limitation 2a). Nothing consumes the admission engine yet — the UI is PR #1514.

Remaining LOWs recorded as follow-ups, neither blocking: the webhook 200-deduped response cannot distinguish an already-run delivery from a policy-dropped one (ctoStatus is dropped at the HTTP boundary), and human tombstones are now unbounded by design for resend-dedupe correctness (~174 B/turn, whole-file rewrite per mutation) pending a time-based horizon.

Required CI and duplication gate pass on this exact head. No deployment or live-state change.

PR #1514, opened 2026-09-15
feat(cto): chat in the existing CTO tab (P3b)

antoinedc:
## Summary
First working CTO chat on desktop, in the **existing CTO tab** (no new window, no sidebar session, no native changes). Based on `feat/cto-p3a-runtime` @ 5dacb200 (PR #1513); rebase here if that PR moves.

**Reuse approach** — a small controller (`src/renderer/CtoChat.tsx`) composing the EXISTING chat pieces: `Transcript`/`MessageRow`, `useTranscriptState`, `useSseBus` (its queued-drain stays dormant — `messageQueue` is never populated), `PermissionCard`/`QuestionCard`/`RetryCard`, `ModelPicker`, `useModelCatalog`/`useSessionModelChoice`. No chat-parser rewrite, no copied chat code.

## Behavior (per unified-cto-spec §3.1/§8.3 + docs/cto-admission-contract.md)
- **Lazy open**: `ctoConversationOpen` fires on the tab's FIRST visit (IntersectionObserver over the display:none pane — not at app start). Loading and error are explicit states with a working Retry; never a blank 'empty healthy' transcript. The session is reused afterwards.
- **Send**: `ctoConversationSubmit({ id, text, expectedGeneration, model? })` with a STABLE client-minted id (`ceo_*`). A timed-out/lost send keeps its id + payload; the visible unknown barrier offers Retry (same id — server dedup makes it idempotent) and Dismiss. A different message mints a fresh id (same-id/different-payload would be rejected). Double-press while in flight cannot mint a second submission (composer disabled + busy guard).
- **No drain/abort**: a follow-up while a turn runs is accepted and QUEUED server-side; the client renders the queue projection (cheap `ctoConversationState` poll — pure store read, plus a window-focus refresh), never aborts or drains.
- **Interrupt**: `ctoConversationInterrupt({ id })` with the ADMISSION RECORD id (dispatched record first, else the queued one). Statuses `queued / dispatching / accepted / unknown / cancel_requested / interrupt_pending` render faithfully as pending bubbles/inspector chips.
- **Rebind**: a role-binding generation change adopts the new session, reloads the transcript, and does not re-send old submissions (durable ids + server dedup).
- **Preserved/honest surfaces**: work inspector collapsible (verbatim queue + counts); dashboard, settings, ledger, profile, blackboard, tools unchanged one click away (chat header "Dashboard" ↔ overview "Conversation"). Attachments/slash commands hidden (not dead); model picker retained because the submit API accepts a per-turn `model` (PromptModel incl. variant); effort selection routes through the same prop.
- **Two clients share state**: no globals, no `window.api` mutation; queue + transcript are server-truth.

## Tests (12, real component + mock api)
double-mount re-opens the same server-ensured session · open error + working Retry · send uses the admission API with `ceo_*` id + `expectedGeneration` and NO legacy `opencodePrompt` · queued follow-up visible with no abort · double-press mints one submission · timeout retry keeps the SAME id + payload · unknown barrier visible + dismissible · interrupt uses the admission record id · rebind reloads without re-sending · cheap poll never creates model turns · CtoPanel defaults to the conversation and round-trips to the dashboard.

Full suite green: typecheck ✓, vitest 201 files / 3927 passed, node:test 4239 passed. (A `website/sitemap.xml` lastmod regeneration from the test run was inspected and reverted — no generated artifacts in the diff.)

## Notes
- Desktop only; native untouched. `ctoStateGet` dashboard subscription untouched; Settings health surface untouched; no conversation-pane duplication.
- Transcript live updates use the existing SSE pipeline for the bound session; the unknown-barrier + queue poll are the seam-visible states.


antoinedc:
Independent Claude Opus 5 review across four rounds; merge-ready at exact head 1c51f040422a93b99e4cc238ad1626c0ca336a9d.

Findings closed over the rounds, all verified against the MERGED server contract rather than the PR's own mocks:
- The permanent uncertain-abort hold could never render as a bubble in production (a transcript-dedup guard suppressed it, and the test that 'proved' it used a record shape the server cannot produce). The whole fixture set was rebuilt to mirror normalizeAdmissionPayload + list().
- Definitive refusals (binding-unavailable, at-cap) were mislabelled 'outcome unknown' forever and discarded the server's actionable message.
- Retry could mint a NEW submission id after an intervening send — a real double-send of an accepted-but-unconfirmed turn.
- An unbounded submit RPC could lock the composer, Send and Stop permanently; now bounded so a hang lands in the designed unknown path.
- Blocking asks were hidden while a send was in flight; questions rendered twice; Stop cancelled the oldest queued send; rejectQuestion swallowed its error.
- Round-2 regressions: a stale draft replayed over text the user had since typed, and an interrupt ack affirmed an abort the server never issued.
- Round-3 regression: an ordinary in-flight interrupt was described as permanently unstoppable, contradicting its own ack.

Reviewer re-ran each reproduction, confirmed the 3 final tests are genuinely red against e14bb114 with no regression to the other 27, and failed to break runningTurnUnstoppable in either direction, strand a dismissed bubble, or desynchronise the composer mirror.

Residual, none introduced by the final round and none demonstrably harmful at head: the interrupt ack over-claims for a receipt whose abort was already refused/lost (fixing it needs abortState on the receipt — a server-surface change); a dismissed unknown briefly reappears as a neutral 'Delivered' chip until the transcript loads; the held doc comment is wider than the human-origin filter it describes.

Required CI and duplication gate pass on this exact head. Desktop only, as scoped. No deployment or live-state change.

PR #1521, opened 2026-09-20
feat(cto): P6 — scheduling plan, bounded cache handoffs, U18-U21 lifecycle verification (CAPO-349)

antoinedc:
Closes CAPO-349 (P6 of the unified CTO spec).

## Summary
- **§9 scheduling across work items**: new pure `scheduleWorks` + read-only `work_schedule` tool. Dispatch order: dependency readiness → explicit deadline → priority → fair aging → deterministic id tiebreak. Every non-dispatchable item carries a recorded plan-level wait reason (`dependency` / `policy_blocked` / `capacity` / `interactive_reserved`). Ambient intent yields to an active interactive (CEO) request; a failing reservation signal fails closed. `work_dispatch` remains the only admission mutator. `deadlineAt` (epoch ms) joins the envelope via `work_create`/`work_revise`.
- **§10 bounded cache handoffs**: new `work_handoff` tool records a bounded context snapshot (spec hash, next step required; target, diff/commit, test results, pending decisions optional) on the work envelope. Per-field clipping is visible ("… (truncated)"), history capped at 10 with the oldest evicted visibly, and inspect/evidence rehydrate it by reference — a fresh control over the same store reads the handoff (server replacement). Recording dispatches nothing and fabricates no keepalive prompt.
- **U18 (deployment)**: cancel/pause now name an in-flight `work.release`/`work.rollback` receipt as an EXTERNAL effect in the summary + `externalInFlight` payload — outcome unknown, reconciles through `work_retry` — instead of implying the deployment stopped.
- **U20/U21 verification**: contract tests pin dirty-worktree cleanup as visible + retained + RETRYABLE (retry after the worktree is cleaned succeeds), and archived work still returns claims + the worker's rationale via `work_evidence`/`work_list state=archived`.
- Dependency readiness extracted to one shared `isDependencyMet` (plan and dispatch gate read it identically — extend, not duplicate).

## Verification
- `node --test` (server suites, sandboxed via `MANTA_STATE_HOME`): 3755 pass / 0 fail (94 in ctoWorkTools.test.mjs, +13 new P6 tests).
- `npx vitest run`: 3962 pass / 7 skipped / 0 fail.
- `npm test`: only pre-existing failure is `scripts/check-duplication-gate.test.mjs` ("jscpd produced no report — tool failure"), reproduced identically on `main` (a817a7ab) before my changes.
- Fail-first mutation checks (each guarantee broken, watched going red, then restored): interactive reservation → S4 red; deadline ordering → S1 red; handoff eviction cap → H2 red; external-in-flight reporting → U18 red; durable handoff write → H1 red.
- `bash scripts/check-duplication-gate.sh origin/main`: PASS (no clones; nothing added to the ignore list). `website/sitemap.xml` churn reverted.
- Tool count: 62 → 64 (work_schedule read + work_handoff); registry + family tests updated.

## Notes for reviewer
- `work_schedule` is a READ (auto mode); it reports, never dispatches — S6 asserts zero `startJob` calls and unchanged state/revision.
- The interactive reservation is a dep (`isInteractiveActive`, default `async () => false`), pass-through wired in `cto.mjs`; production callers (e.g. the overnight scheduler) can inject the real presence signal in P7+ without touching this surface.
- Plan-level wait reasons are a scheduling vocabulary; the envelope's closed `WAITING_REASONS` set is untouched (the plan persists nothing).
- The pre-existing duplication-gate test failure is environmental (jscpd emits no report in this sandbox); it fails the same way on main.


antoinedc:
## Review — CAPO-349 (P6: scheduling plan, bounded cache handoffs, U18-U21 lifecycle verification)

Reviewed against the issue requirement (spec §9/§10/§12 + §15 P6 acceptance U18-U21), head `3a78139d`. All numbers below re-run from scratch on this head.

### Mechanical (Step 1)

- **vitest:** 3962 pass / 7 skipped / 0 fail — exact match to the PR claim.
- **node:test (server, sandboxed):** 3755/3755 pass — exact match. Full-repo sweep: exactly 1 failure (`scripts/check-duplication-gate.test.mjs` — "jscpd produced no report"), and I reproduced it identically on `main` @ `a817a7ab` — pre-existing/environmental, as claimed.
- **Duplication gate:** `bash scripts/check-duplication-gate.sh origin/main` → PASS, no clones. Nothing added to `scripts/duplication-gate-ignore.txt`. `website/sitemap.xml` untouched by the PR.
- **knip:** 1 **introduced** unused export — `SCHEDULE_INTENTS` (used only inside the module; the test imports the other new exports but not this one). Four pre-existing siblings in the same file (`DELIVERY_TARGET_KINDS`, `WORK_LEASE_TTL_MS`, `ATTEMPT_STATUSES`, `REVIEW_VERDICTS`), so this follows an existing pattern — Nit.
- **Diff anatomy:** +855/−49 across 5 files, all under `src/server/` — consistent with the issue's 3-part phase scope plus contract tests. No scope creep, zero new files.

### Requirements coverage (Step 2)

Every new symbol maps to a requirement step:

- `scheduleWorks` + `work_schedule` → §9 ordered plan with recorded wait reasons; closest analog `work_list`/`work_capacity` are unordered counts/listings — new surface justified. Read-only proven (S6: zero `startJob` calls, state/revision unchanged).
- `isDependencyMet` extraction → the issue's "extend, do not duplicate" — plan and dispatch gate now share one definition. Verified the extraction is real (old inline body replaced by the shared call).
- `externalInFlight` in cancel + pause → U18 ("distinguishes cancellable worker from already-running external effect") — summary + `externalInFlight` payload in BOTH paths, with a counterfactual (no effect → empty field, not fabricated).
- `workHandoff` + HANDOFF_* constants → §10 bounded handoff: required `specHash`+`nextStep`, per-field visible clipping, capacity-10 with visible eviction, rehydration via inspect/evidence (H1 uses two controls over one store), no dispatch/no fabricated job (H3).
- `validateDeadlineAt` + envelope plumbing → §9 deadline input on create/revise, validated in the strict loader too.
- U20/U21 contract tests → §12 cleanup-failure-visible-and-retryable, evidence/rationale retrievable after archive.
- Deadline>priority>aging order: spec §9 groups "explicit priorities/deadlines" without declaring the internal order; the PR's reading (deadline first) is documented and tested — accepted as a declared policy choice.

### Removal opportunities (Step 3)

- `dependencyReadiness` is now a one-line wrapper over `isDependencyMet` with a single call site — could inline.
- `workHandoff` clips `nextStep` twice (once in the `HANDOFF_TEXT_FIELDS` loop, then an explicit re-clip). Idempotent, but noise.
- `reservation.interactiveSource` reports `"wired"` even when the dep is the default `async () => false` no-op — the value is honest (false), the label is not diagnostic.

### Semantic findings (Step 4)

**Block (fix-here):** The plan's dependency read can disagree with the admission gate past the list-page cap — the exact claim this PR makes ("the plan can never disagree with the admission gate"). `workSchedule` fetches `listWorks({ limit: LIST_MAX_LIMIT })` (=1000, newest-updated first) and ignores the returned `total`. `scheduleWorks` resolves dependencies through a map of that page and `.filter((dep) => dep && !isDependencyMet(dep).met)` treats a dependency **absent from the page as MET**. The dispatch gate (`assertDependenciesOrPark`) loads each dependency by id and treats a missing one as **UNMET** (`readiness: "missing"` → refuse + park). `loadAllEnvelopes` loads every state and nothing caps the store, so with >1000 envelopes an old dependency falls off the newest-1000 page while its dependent sorts near the top: the plan reports "dispatchable now" for work dispatch would park on `waiting/dependency`. This is a server-reachable state, which is precisely the recurring defect class this issue's acceptance gate warns about (green tests encoding only states the server "cannot" produce). Fix fits inside this PR: either resolve dependency readiness for unresolvable ids as unmet with source `"missing"` (mirroring the dispatch gate) and/or surface page truncation (compare `total` > `works.length`, add a visible `truncated` note), plus a fixture that exercises the >1000 boundary.

Fail-first spot-check (verified independently, in a throwaway worktree, then reverted): mutating the eviction behavior (`shift()` → `pop()`) turns exactly H2 red — the discipline is real. Note that mutating the capacity *constant* itself (10→11) keeps H2 green because the test derives its expectations from the exported constant; the eviction-list assertion is what has teeth.

Claim accuracy nit: "+13 new P6 tests" — the file gains 17 net-new top-level tests (19 added, 2 rewritten); file total 94 matches.

### Decision

1 Block (fix-here) → returning to `capo-api` for the fix. Everything else is verified clean: both suites match the reported numbers exactly, the pre-existing failure reproduces on base, the duplication gate passes, and the spec's §9/§10/§12 + U18-U21 acceptance observations are each backed by a test I traced to code.

antoinedc:
## Re-check review — CAPO-349 cycle 2 (head `220ffbc9`)

Re-verified from scratch on the fix commits (`85737128` + `220ffbc9`).

### The Block — resolved

- **Missing-dependency parity:** `scheduleWorks` now resolves a dependency id absent from the scanned page as UNMET with readiness `"missing"` — the dispatch gate's own vocabulary from `assertDependenciesOrPark` — and carries the structured `unmet` list on the blocked entry. Plan and gate now agree structurally, and the plan can only err conservative (false-unmet), never the dangerous direction. Verified against the gate's code path.
- **Truncation surfaced:** `workSchedule` reports `listScan: { scanned, total, truncated, note }` whenever the portfolio exceeds the page — the partial view is visible, never silent.
- **Testable seam:** `listPageLimit` composition option (default `LIST_MAX_LIMIT`) — the boundary is exercised through the REAL control/store/list path, not fake states. The S7 fixtures use the production composition with a 3-envelope portfolio against a page of 2, plus a counterfactual proving `"missing"` is page-specific (the same dependency inside the page reads its true state).

### Mechanical (re-run on `220ffbc9`)

- **node:test (server, sandboxed):** 3757/3757 pass (96 in ctoWorkTools.test.mjs, +2 S7) — matches the fix report exactly.
- **vitest:** 3962 pass / 7 skipped / 0 fail — unchanged, matches.
- **Duplication gate:** `check-duplication-gate.sh origin/main` → PASS; nothing added to the ignore list; no sitemap churn.
- **Fail-first spot-check (independent, throwaway worktree, reverted):** reverting the missing-dep branch to the old skip-as-met behavior turns exactly the S7 boundary test red (95/96 pass, the right one fails). The fix's test has teeth.
- **Diff anatomy:** +100/−5 across the two intended files only — no scope creep.

### Remaining nits (non-blocking, from cycle 1 — unchanged)

- `SCHEDULE_INTENTS` remains an exported-but-internally-used constant (knip flags it; four pre-existing siblings in the same file).
- `workHandoff` still re-clips `nextStep` after the loop already clipped it (idempotent, harmless).
- `reservation.interactiveSource` still reports `"wired"` when the dep is the default no-op — value honest, label cosmetic.

### Decision

0 Blocks, 0 Questions — PASS at head `220ffbc9`. Ready for merge per the issue's routing.

antoinedc:
Superseded by #1522 (merged, 560861ad) — same CAPO-349 scope, counterfactual-verified tests, CI green. This draft arrived from a delayed parallel run; closing rather than merging two implementations of one spec. The 21 tests here were not reviewed individually; nothing unique was identified that #1522 lacks (both cover U18-U21 through the same seams).

PR #1525, opened 2026-09-20
feat(server): operation-class outcome watcher raises blockers on silently dying classes (BET-1533)

antoinedc:
## BET-1533 — S1: operation-class watcher

Turns a silently dying CTO operation class into a pendingBlockers entry (→ health card) instead of a silent 0% success rate.

### What

- **`src/server/ctoOpClassWatcher.mjs` (new, pure)** — the predicate + constants:
  - blocker: 0 `ok` in the last 10 attempts of a class AND ≥1h between the oldest and newest of those 10
  - degraded: <50% success over the last 20 attempts (also a blocker, lower text severity)
  - `gated` rows are skipped entirely (never an attempt); rows group by `operation` falling back to `taskClass` (validated against the real outage rows, which carry both)
  - `evaluateOpClass(rows, { nowMs, alarms })` — pure fold over rows + prior latch map; idempotent against the sliding read window (re-fed old rows can neither re-fire nor spuriously clear: recovery requires an `ok` newer than the alarm's `firedAt`)
- **`src/server/ctoEngine.mjs`** — thin I/O wrapper `opClassWatcherTick()` inside the enabled tick, paced to one ledger read per 10 min (the read is O(file)): `ledgerStore.read({ from: now − 14d })` → `evaluateOpClass` → `recordBlocker("op-class:<class>", reason)` through the EXISTING pendingBlockers path → latch persisted under the new `opClassAlarms` key in engine-state.json (keyed (taskClass, incidentGeneration); re-arms only after demonstrated recovery). No new dep, no new config flag, no new bus kind, no new store.

### Cross-cutting notes

- **BET-1466 bounded-read test widened**: tick() now has two bounded ledger readers (overnight 24h, op-class watcher 14d). The test's invariant is preserved as "every read carries a numeric lower bound in the past — never the whole file", plus the overnight 24h bound asserted specifically.
- **`opClassAlarms` key added to engine-state.json** — additive; `patchStore` RMW preserves unknown keys, `migrateStore` tolerates them, no schema bump.
- **The cited spec `docs/endpoint-health-spec.md` §W7 does not exist on `origin/main` (nor on any of the 839 remote branches)** — implementation follows the issue body, which carries the normative decisions. Follow-up filed for the missing spec doc.
- **Acceptance evidence (real ledger, `~/.manta/cto/ledger.jsonl`, 731 outcome rows since 2026-09-14T23:49:53Z, all failures):** replay raises degraded `ambient-summarize` at **+11 min** (within the first hour, as the acceptance demands — the issue defines degraded as "also a blocker"), full blocker `segment-summary` at **+169 min**, blocker `segment-one-liner` at **+445 min** (the ≥1h-span condition interacts with the trickle cadence of those classes). `ctoOpClassReplay.test.mjs` encodes this: runs on the box's real ledger, skips cleanly on CI/rotated data.

### Verification

Base: origin/main @ `b06763db`


antoinedc:
## Review (manta-reviewer) — PASS: 0 Blocks, 0 Questions, 2 nits

Independently verified against the issue's acceptance criteria. Base is fresh (branched from current `origin/main` @ `b06763db`). Diff is 5 `src/server/` files, all on-topic — no scope creep, no shared-surface drift.

### Mechanical gates (run by reviewer, not trusted from the PR body)

- `npm run typecheck` — **exit 0**, zero errors.
- `npm test` — **exit 0, 4507 pass / 0 fail / 0 skipped**. Matches the PR body's reported PR-branch numbers exactly (4507 vs 4485 on base; +22 tests, all passing). Spot-checked the cached logs' sha256s — they match the reported values.

### Acceptance-criteria checklist (each item verified independently)

- **Ledger via `ledgerStore`, not direct file I/O** — the engine wrapper reads through `ledger.read({ from: now − 14d })`; the 14d window keeps the BET-1466 "no unbounded read" invariant (widened test now pins that *every* tick read is bounded, and still asserts the overnight 24h bound specifically).
- **Blocker predicate** — 0 `ok` in the last 10 attempts AND ≥1h span between oldest/newest, both required. Unit tests pin: exactly-1h span fires (inclusive ≥), 59-min span does not, an `ok` inside the window blocks it, <10 attempts never fires.
- **Degraded predicate** — <50% over the last 20, exactly-50% does not fire, blocker severity takes precedence when both shapes hold.
- **`gated` skipped entirely** — unit test interleaves 10 gated rows into a failure run; they neither fill the window nor count as failures.
- **Blocker names class + dominant failure code + count, no endpoint** — reason format asserted in both unit and engine tests.
- **Existing blocker path only** — raises via `recordBlocker("op-class:<class>")` → `pendingBlockers` → health cards; the cards layer's BET-1463 source-grouping folds repeat entries into one card, which is what makes the crash-between-raise-and-persist window safe (raise-then-persist ordering is correct: a lost latch re-fires into an upsert, not a duplicate card).
- **Dedupe** — alarm state lives in one new `opClassAlarms` engine-state key, keyed by class with an incident generation; fires once per incident; re-arms only after an `ok` *newer than `firedAt`* (the guard that stops the sliding read window from spuriously clearing on re-fed old rows). Engine tests verify no-refire on repeat ticks and generation 2 on re-arm.
- **Hygiene** — no new dependency (package.json untouched), no new config key (the 10-min pacing is a deps-overridable internal constant), no new store, no new bus kind (`cto.opclass_blocked` is a ledger log entry, not a bus kind), pure exported `evaluateOpClass(rows, {nowMs, alarms})` tested directly with a thin ~35-line I/O wrapper.
- **Primary acceptance (real-ledger replay) — reproduced by reviewer**: I replayed all 732 `cto.operation_outcome` rows since 2026-09-14 from `~/.manta/cto/ledger.jsonl` (224k lines today; the issue's "23k+" and "321/275" counts have grown to 331/285 model-errors as the outage kept accruing) through `evaluateOpClass` with the engine's sliding 14d read window. Result: **first raise at +11 min** (degraded `ambient-summarize`, 0/20), full blocker `segment-summary` at +169 min, `segment-one-liner` at +445 min, zero spurious re-fires across the whole replay. The within-the-hour acceptance bar is met.

### Nits (non-blocking)

1. **Vacuous pacing test** — `ctoEngine.test.mjs` "paces its ledger reads to the watch interval" asserts on `ledger.rows.length` (the append count, which stays 0), never on read-call counts, so it cannot fail. The pacing itself is real (verified in `opClassWatcherTick`) but is the one behavior with no effective test.
2. **"No live ledger read in the suite" vs the acceptance replay** — resolved via a skip-guarded replay test (reads the real file, skips when absent/rotated). Reasonable reconciliation of the two requirements and documented in the test header; noting it for the record.

### Observation (for the record, matches the spec's own design)

In the real data the named classes (`segment-summary`, `segment-one-liner`) fail on a trickle cadence interleaved with other classes, so their full blocker-severity alarms land at +169/+445 min; the within-the-hour raise is the degraded `ambient-summarize` alarm. That is the ≥1h-span condition doing exactly what the issue designed it to do (suppressing fast bursts), and the implementer flagged the cadence interaction themselves. The spec file cited by the epic (`docs/endpoint-health-spec.md`) is indeed missing from `origin/main` — the follow-up issue for it was the right call.

**Routing to manta-pm for merge.**

antoinedc:
**Re-review after the duplication-gate fix (cycle 2) — PASS. 0 Blocks, 0 Questions, 0 open nits.**

Delta since cycle 1 (`d17d2492` → `cce104c8`): 2 test files only, +40/−29, no production code — so cycle 1's substantive review of the predicate, engine wiring, and replay acceptance stands.

Verified independently on the PR branch (`multica/BET-1533-opclass-watcher` @ `cce104c8`):

- **Delta matches the claim.** `deadClassHarness()` and `seededIncident()` are pure extractions — seeding math and assertions identical to the inline copies they replace; nothing dropped or weakened.
- **Cycle-1 nit 1 genuinely fixed:** the pacing test now asserts on a real `readCalls` counter (1 read on first tick → no re-read inside the 10-min interval → re-read after). Previously vacuous.
- `npm test`: **4507 pass / 0 fail / 0 skipped** (same totals as cycle 1 — helper extraction replaced, not removed, coverage).
- `bash scripts/check-duplication-gate.sh`: **PASS — no clones** (strict, min-tokens 70).
- Acceptance replay (`node --test src/server/ctoOpClassReplay.test.mjs`): real 2026-09-14 ledger, blocker raises within the first hour — the epic's primary acceptance test holds.
- CI: `typecheck-test` and `duplication-gate` both green on this head.
- Base freshness: merge-base `b06763db`; upstream since is docs-only (#1526), no overlap with the PR's 5 server files, no stale-base phantoms.

Acceptance criteria re-checked against the issue: pure exported predicate taking rows + now ✔; `gated` skipped (neither success nor failure) ✔; blocker requires 0/10 AND ≥1h span ✔; degraded <50%/20 ✔; names class + dominant code + count, no endpoint ✔; raised through existing `pendingBlockers` path, no new bus/config/store ✔; alarm latch keyed `(taskClass, generation)`, re-arms only after a demonstrated `ok` ✔.

Clean for merge.

PR #1527, opened 2026-09-21
BET-1539: fix(server): W10 scan failure codes + agent health register + watcher coverage

antoinedc:
BET-1539 — endpoint-health spec §W10 (Agent / tool registry).

Base: origin/main @ bb98b2d4

## What

**1. Distinct scan failure codes (the core).** The registry's scan catch applied `db-unavailable` to *every* exception, so the 6-day outage's real cause was unknowable (W3's treatment, D6's shape). Now:

- Each real throw site carries its own code at the source:
  - `ctoToolScan.mjs` `collectDbRows`: `db-handle-invalid` (no usable handle), `db-query-failed` (prepare/execute threw), `internal-provenance-failed` (internal-session / binding provenance read threw — two sites, one concept), on top of the registry's own `invalid-cursor`.
  - `ctoEngine.mjs` `toolsCollectDb`: `unsupported-runtime` (kept), `db-source-unavailable` (handle resolved null while node:sqlite exists).
- `ctoToolRegistry.mjs` maps through a closed exported set `SCAN_FAILURE_CODES` (7 codes incl. `surfaces-unavailable`); anything else → `scan-failed`. Exception text never crosses.
- `surfaces-unavailable` stays: the surfaces seam has a single throw site (its five readers are individually best-effort).

**2. Covered by watcher 1 (W10 bullet 3).** The scan now emits `cto.operation_outcome` rows under a new `tool-scan` operation class — `code:"ok"` on success, the failure code otherwise. Successes matter: without them the S1 watcher cannot distinguish "dying" from "never ran".

**3. Agent health register (W10 bullet 2).** `agentHealth(tool)` in `ctoToolRegistry.mjs`, projected as `health` on every `listTools` row, derived from observed dispatch outcomes with S4/W4's vocabulary so the two registers read the same way:

- `unproven` — registered but never demonstrated: zero uses AND no durable last-success instant (W5's rule: derived from the durable `lastSuccessAt`, not a ring). Absence of evidence is not health.
- `dead` — demonstrated, then silent past the vitality recency horizon (`VITALITY_RECENT_MS`, the same constant the §7.3 role drill-down already uses).
- `healthy` — has an observed dispatch inside the horizon.

Interpretation note (issue item 3 says "registered agents"): the registry's rows are the CTO's registered agents; their observed dispatch stream is the fused use evidence the registry already accumulates. This is a distinct axis beside `derivedRole` (engagement quadrants) — no third vocabulary invented.

**4. Watcher 1 covers agent dispatches (issue item 4 — confirmed it did not).** Only `ctoSessions.mjs` wrote `cto.operation_outcome` rows; agent dispatches (the act executor's `runOne` pipeline) wrote `cto.resolve` / `cto.execution_unavailable` rows that the watcher never reads. The executor now also emits `cto.operation_outcome` rows under `agent-dispatch`: the dispatch-failure reason when unavailable, `ok` on a resolved run, the failure reason otherwise. Additive at the source — S1's own predicate is unchanged.

## Duplicate-site audit (anti-spaghetti)

- The only other scan-catch-style flattening was the registry's own (`ctoToolRegistry.mjs:786-790`, W10's named site) — fixed. The S1 watcher's row reader is untouched.
- The `withCode` helper lives in `ctoToolScan.mjs`; `ctoEngine.mjs` uses inline `Object.assign` since it throws only two self-made errors (extending the scan module's export for two sites wasn't warranted).

## Live diagnosis (recorded per the issue's protocol)

The live box's ledger shows the db channel failed essentially every scan since 2026-09-14 (139 of 157 `cto.tool.scan` rows are failure-twins — the registry "changed" via non-db channels while the db step failed). Root cause of the db failures: `internalSessionIds()` inside `collectDbRows` — its 250 ms provenance barrier + generation-churn throw fire under box load (`cto.provenance_unavailable code=timeout` ×741 in the ledger). The codes now name that site; fixing the barrier itself is a shared-primitive change → **follow-up filed: BET-1541 (PM, high)**.

The surfaces channel's 64 failures have no findable throw site (all five readers are individually guarded and work in isolation) → per-reader codes so the next scan names it → **follow-up filed: BET-1542 (parked, follow-up label)**.

## Verification results

**Files changed:** 7 files — `src/server/ctoToolScan.mjs`, `ctoToolScan.test.mjs`, `ctoEngine.mjs`, `ctoToolRegistry.mjs`, `ctoToolRegistry.test.mjs`, `ctoAct.mjs`, `ctoAct.test.mjs`.

- PR branch (`multica/BET-1539-scan-failure-codes` @ `61052ab1`):
  - test: exit 0, 4517 pass / 0 fail / 0 skip. Failures: none. log sha256: `75b8c8e3848c8518979656decd3b91d90b83b5019a88e99530b1e1c8754e544a`
- Base (`main` @ `bb98b2d4`):
  - test: exit 0, 4507 pass / 0 fail / 0 skip. Failures: none. log sha256: `ba11929a17184f0d634ce9f2e4a2d0da52a231c4828b247d475a3f4eed580387`
- **Conclusion:** 0 test failures on either branch; the 10-test delta is this PR's new coverage (scan code mapping, fallback no-leak, cursor code, agent-health register, watcher-1 outcome rows for scan + agent dispatches). 0 pre-existing failures.


antoinedc:
## Review (manta-reviewer) — 1 Block

**Mechanical gates (run on PR branch `61052ab1`, fresh checkout):**
- `npm test`: exit 0 — **4517 pass / 0 fail / 0 skipped**, matching the PR body's PR-branch claim exactly (base claim trusted per test-reuse policy).
- CI: all 4 checks green. Anti-spaghetti: 0 clones. Base fresh (merge-base = origin/main tip `bb98b2d4`).

**Acceptance criteria (verified independently, not from the summary):**
- ✅ **Closed code set, one per real throw site.** `SCAN_FAILURE_CODES` (7 codes) in `ctoToolRegistry.mjs` is genuinely closed; the catch maps unknown → `scan-failed`; the no-leak test proves exception text never crosses. Each site verified at source: `db-handle-invalid`/`db-query-failed`/`internal-provenance-failed` (ctoToolScan.mjs), `db-source-unavailable`/`unsupported-runtime` (ctoEngine.mjs), `invalid-cursor`, `surfaces-unavailable` (pre-existing seam, confirmed at `ctoToolRegistry.mjs:841`). Same shape as `ctoRunOutcome.mjs`'s `CODES` + has-check + fallback — no third taxonomy.
- ✅ **Live diagnosis + follow-ups.** Root cause (provenance 250 ms barrier under load) is a shared-primitive change; BET-1541 and BET-1542 both verified to exist. Correct scoping per the issue's protocol.
- ✅ **Agent health register.** `agentHealth()` reuses the W4/W5 vocabulary (spec's endpoint axis: `unproven`/`dead`/`healthy`), `unproven` is first-class, projected as `health` on `listTools` beside (not replacing) `derivedRole`, same `VITALITY_RECENT_MS` horizon as the drill-down. All four required registry tests present and passing.
- ✅ **Watcher 1 coverage.** Confirmed the watcher reads all `cto.operation_outcome` rows grouped by `operation ?? taskClass` with no class filter, so the new `tool-scan` and `agent-dispatch` rows are watched with zero watcher changes. The implementer's claim that dispatches were previously NOT covered is correct — `ctoSessions.mjs` was the only writer.

---

**Block (fix-here):** the abolished catch-all survives in the test suite — `ctoReview.test.mjs:71` still feeds `code: "db-unavailable"` into the retry-backoff test. That label no longer exists in production (this PR's whole point), so the transient-bucket leg of "unsupported scans and transient scans persist distinct bounded retry schedules" now exercises the **fallback** path while wearing a dead label — and the renamed `db-source-unavailable` has **zero** test coverage anywhere (grep confirms). One-token fix: change `"db-unavailable"` → `"db-source-unavailable"` in that loop. That simultaneously kills the stale reference and pins the renamed engine-site code through the real registry boundary.

**Nits (non-blocking):**
- `SCAN_FAILED_FALLBACK` is exported and used in the outcome-row emit, but the catch on the mapping line still spells the literal `"scan-failed"` — one spelling would be kinder to a future rename.
- Hygiene is fine: net +241 total (~+108 production, ~+133 tests); the registry module itself grew ~60 lines, mostly the closed set + `agentHealth` with in-house comment density. Not a subsystem.

Route: returned to `manta-dev` for the one-line fix, then re-check.

antoinedc:
Block addressed in 90d5b203: the retry-backoff loop in `ctoReview.test.mjs:71` now feeds `db-source-unavailable` instead of the abolished `db-unavailable` — the dead label is gone from the whole `src/` tree (the only remaining matches are the `discovery-db-unavailable` error *message* strings and the spec's historical W10 quote), and the renamed code now has retry-schedule coverage (its transient leg asserts the 300 s backoff + no cursor advance, same as before).

Verified on the branch @ 90d5b203: typecheck exit 0 (log sha256 `84317a7d765e8fce9be06db1e3e63c5aa66f60d129053f6747de534b58fee709`), 4517 pass / 0 fail (log sha256 `7a30ec87e39d23c9a0a946cb3eaaa1f41550a58d6e9b2d841ab8903171ab2ce4`). One-token change; no production code touched.

antoinedc:
## Review round 2 (manta-reviewer) — PASS

**Block resolution verified.** Commit `90d5b203` applies exactly the prescribed one-token fix: `ctoReview.test.mjs:71` now feeds `db-source-unavailable` into the retry-backoff loop. The abolished `db-unavailable` code label is gone from the entire tree (remaining `discovery-db-unavailable` hits are error *messages*, not codes — correct), and the renamed code is now pinned through the real registry boundary. The stale-reference and zero-coverage halves of the Block are both closed.

**Mechanical gates (run locally on `90d5b203`, fresh checkout):**
- `npm test`: exit 0 — **4517 pass / 0 fail / 0 skipped**.
- Typecheck: unchanged production code vs the verified `61052ab1` (this commit touches one test string); CI `typecheck-test` was still in flight at snapshot time — duplication-gate pass, local suite green stands as the evidence.

**Base observation (non-blocking, for the merger):** `origin/main` advanced to `05ce5b2c` (BET-1534, PR #1528) after this branch was cut. Checked explicitly: the three upstream-added files (`endpointAttempts.mjs`, `endpointAttempts.test.mjs`, `opencodeEndpointLifecycle.test.mjs`) appear "deleted" only in a two-dot diff; the PR's actual changeset (`origin/main...HEAD`) contains no deletions, and the upstream/PR file overlap is **zero** — the merge result is well-defined. A rebase would be clean if desired, but nothing requires it.

**Final acceptance state:** closed code set per throw site ✅ · no-leak fallback ✅ · agent-health register with S4 vocabulary and `unproven` first-class ✅ · watcher-1 coverage of tool-scan + agent-dispatch ✅ · live diagnosis with BET-1541/BET-1542 follow-ups ✅ · hygiene within bounds ✅ · all four required tests present ✅.

0 Blocks, 0 Questions, 1 leftover nit (the `SCAN_FAILED_FALLBACK` constant vs. the literal `"scan-failed"` on the mapping line — cosmetic). Routing to manta-pm for merge.

PR #1528, opened 2026-09-21
feat(server): attempt lifecycle at the transport boundary + keep the failure cause (BET-1534 W2+W3)

antoinedc:
Implements W2+W3 of the endpoint-health epic (spec docs/endpoint-health-spec.md §4): every `runSynchronousSession` dispatch now writes exactly one §4.1 operation record and §4.2 provider attempts from the causative assistant row, persisted to `statePath("endpoint-attempts.json")`, and the failure cause is no longer discarded at the model boundary.

**Base:** origin/main @ `bb98b2d4` (includes BET-1533 — the S1 gate this issue waited on).

## What changed

- **New `src/server/endpointAttempts.mjs`** — the attempt-lifecycle store: per-endpoint `lastSuccessAt`, consecutive-failure streak, and bounded provider-attempt rings (200 entries or 30 days, whichever binds first) + a bounded operation-record ring. Atomic writes via the shared jsonStore pattern, `patchStore` as the only RMW path (per-path mutex — concurrent completions serialize, both survive). Versioned schema (v1); a corrupt payload is quarantined aside + rebuilt empty + alarmed; a NEWER schema version throws loudly and is never destroyed.
- **`runSynchronousSession` (§4.3 + §4.1a)**:
  - writes an operation-record skeleton at dispatch start (so a killed server leaves something to sweep), refines attribution at dispatch, and terminalizes first-writer-wins on attemptId at the end;
  - `trackCreation()` moved inside the try — a synchronous throw previously escaped with no finish(), no cleanup and no record;
  - finalization order: persist the captured provider attempt → cleanup → append the operation record stating the cleanup outcome truthfully (`cleanupCode` rides the record; the provider attempt stays a model truth). Persistence failures skip neither cleanup nor the result and alarm via ledger rows (`cto.endpoint_attempts_persist_failed`, rate-limited) — the W7.3 watcher's signal;
  - prompt refusals now preserve status + Retry-After (mirrors `sendPrompt`); a 402/429 refusal **on a pinned model** additionally records a provider attempt with `attribution: "intended"` (§4.1a pinned-only rule);
  - the provider attempt is built from the CAUSATIVE assistant row (the one carrying the error), captured before session deletion (D7), keyed by the row's own `providerID/modelID` identity (rows without identity produce no keyed attempt).
[...]

PR #1529, opened 2026-09-21
feat(server): retry sweep for failed segment summaries (W11/BET-1538)

antoinedc:
## Summary

Spec `docs/endpoint-health-spec.md` W11 (epic BET-1532, issue BET-1538): `summaryOutcome` was write-only — a close-time summarisation failure persisted a degraded, empty summary that nothing ever revisited, so a failed segment was a permanent empty shell in memory/rollups/digests.

This adds the missing reader as a **retry sweep** in the segmenter (`retryFailedSummaries()`), ticked from the engine tick:

- **Select on EMPTY SUMMARY, not on the outcome marker** — `isSegmentSummaryEmpty()` is content-based (no key_events, no files_touched, no prs, importance ≤ 1), so the 23k legacy shells that predate `summaryOutcome` are selected too. Missing/unparseable summaries count as empty.
- **Durable attempt cap** — `summaryAttempts` persisted on the segment record (init `0` at close, incremented only on real failures, default cap `MAX_SUMMARY_ATTEMPTS = 3`).
- **`gated` is not a failure** — a gated `summarize` result persists nothing, consumes no attempt, and ends the pass (the gate is closed; hammering more candidates is pointless).
- **Reuse, no second rate gate** — each attempt calls the segmenter's injected `summarize` dep, which in the engine is `gatedSummarize` (the §3.3 `beginEphemeral` seam: disabled/paused/cap/concurrency all apply unchanged). No new budget path.
- **Yields to live work** — the sweep fully returns while presence is "present" (same batch-priority rule as the backfill), throttles to its own 15-min cadence, caps work per pass (`MAX_RETRIES_PER_PASS = 4`, oldest window first), and bounds the scan to files written in the last 48h (`SUMMARY_RETRY_WINDOW_MS`; strictly cheaper than the existing store sweeper, which re-reads every segment JSON every 5 min).
- **No new backfill** — the retry needs no history replay: the real producer reads transcript evidence from the opencode store via `sessionID` + window (`ctoSegmentEvidence`), both anchors the stored record already carries. Success overwrites the shell exactly like a first-pass close (window/sessionID anchored, blank model one-liner falls back to the stored prompt echo) and feeds `onSummary` so the profile engine ingests the recovered atoms. Failure persists only the attempt counter + outcome marker, leaving the degraded summary and `summarizedAt` (§10.5 pipeline-lag truth) untouched.

## Verification

- `npm run typecheck` — pass (exit 0) on branch and base.
- `npm test` — branch: node:test 4517 pass / 0 fail (incl. 8 new retry tests); vitest 3992 pass / 7 skipped / 0 fail. Base: node:test 4507 pass / 0 fail; vitest 3992 pass / 7 skipped / 0 fail.
- One observed exception: the FIRST full run on the branch had a single `MediaBody.test.tsx > renders a reserved aspect box in the ready state` failure (jsdom `HTMLCanvasElement.getContext` + xterm/webgl under parallel load). It passes on isolated re-run and on a full-suite re-run, and is unrelated to this server-only diff — classified as a pre-existing flake, not caused by this PR.

## Notes for the reviewer

- **Bulk legacy shells are NOT resurrected**: the 48h mtime window means the ~23k historic empty segments stay untouched — spec W11's operational cleanup (delete + one-time backfill rebuild) remains the recovery path for those; this sweep is the durable fix for future failures.
- The attempt cap is a constant (`MAX_SUMMARY_ATTEMPTS`), overridable via dep for tests/ops.
- Ledger reuses the existing `cto.segment_summary_*` kinds with an additive `retry: true` field (the BET-1533 op-class watcher only reads `cto.operation_outcome`, so no metric interference).

Base: origin/main @ bb98b2d4

## Addendum (post-initial-push)

- Two more tests cover the issue's named cases explicitly: the 98% legacy shell with NO `summaryOutcome` field, and sweep idempotence (re-running on recovered segments is a no-op). Full suite re-run green: node:test 4519/0, vitest 3992/7-skipped/0.
- **Why `SEGMENT_SUMMARY_VERSION` is NOT bumped:** `summaryAttempts` is an additive, optional RECORD field (camelCase like its siblings `summaryOutcome`/`summarizedAt`); the §5.2 SUMMARY schema itself is unchanged, `validateSegmentSummary` doesn't see it, and both directions are compatible (old readers ignore unknown fields; the sweep treats a missing counter as 0). Bumping would imply a migrateStore step that isn't needed.
- Final: `multica/BET-1538-summary-retry` @ `805a4f77`.


antoinedc:
## Review (BET-1538, cycle 1)

**Mechanical gates — all pass:**
- Base fresh: merge-base == `origin/main` == `bb98b2d4`; diff is exactly the 3 issue-named files, +467/−0. No backfill touched.
- `npm test`: node:test 4517 pass / 0 fail; vitest 3992 pass / 7 skipped / 0 fail — matches the PR body's reported results exactly.
- Duplication detector (`scripts/check-duplication.sh origin/main`): 0 clones across changed files (sticky comment absent, ran it directly).
- Premise re-verified on `origin/main`: `summaryOutcome` written once (`ctoSegments.mjs:561`), zero readers.
- Core design verified: selection is content-based (`isSegmentSummaryEmpty`), so legacy marker-less shells are covered; gated detection matches the engine wrapper's actual `{ok:false, gated:true}` shape; the retry rides the injected `gatedSummarize` seam (no second rate gate); presence yield uses the same `getPresence().state === "present"` idiom as the rollup runner/backfill; and `segmentSummarize` reads evidence from the opencode store via `readSegmentEvidence(await getDb(), data)` — so the retry's `events: []` is harmless, and validated retry summaries carry the `v` stamp (the producer's prompt emits it; `validateSegmentSummary` requires it before save).

**Blocks — 3:**

**Block (fix-here):** The attempt cap does not bind on ok-but-still-empty retries. `retryOne`'s success path persists `summaryOutcome:{ok:true, code:null}` without incrementing `summaryAttempts` — but a schema-valid summary can still be empty (the producer's own prompt says "No transcript evidence available; do not infer outcomes.", which invites a minimal valid JSON with empty arrays / importance 1). Such a segment remains content-empty → `isRetryEligibleSegment` keeps selecting it on every pass → up to 4 model calls per 15-min sweep indefinitely, escaping the exact bound the issue demands ("a genuinely unsummarisable one does not retry forever"). Fix fits this PR: consume an attempt whenever the post-retry summary is still empty (or classify ok-but-empty as a non-recovery for cap purposes) and add a test for it.

**Block (fix-here):** Missing required test — "a segment with an empty summary and **NO `summaryOutcome` field** is selected (the 98% case)". Every sweep fixture is built from `storedDegraded()`, which always carries `summaryOutcome`. The predicate is content-based so the behavior is presumably correct, but this is the issue's explicitly named guard for its single most important instruction, and it is untested. Add a marker-less record to a sweep test (e.g. a fixture without `summaryOutcome`/`summaryAttempts`) asserting it is selected and retried.

**Block (fix-here):** Missing required test — "re-running the sweep twice is a no-op on already-summarised segments". No test performs a second sweep over a recovered segment asserting `attempted: 0` and zero model calls. The throttle test covers cadence, not content-based idempotency; add a forced double-run assertion.

**Nits (non-blocking):**
- On successful recovery the old `summaryAttempts` value is carried forward via `...rec` — inert (eligibility requires an empty summary), but stale; resetting to 0 would be cleaner.
- No engine-level test for the `segmentRetryTick` wiring; the `if (!segmenter?.retryFailedSummaries) return` guard + silent catch would mask a naming mismatch as a permanent no-op.
- Prose: the PR body says "8 new tests" — the diff adds 10.

Structurally this is a well-bounded implementation — content-based selection, durable counter, gated semantics, seam reuse and evidence sourcing all check out against the real code paths. The blocks are the cap hole and the two named tests the issue explicitly required.

antoinedc:
All three Blocks addressed on `746cbbdd` (rebased onto `origin/main @ e40031d2`):

1. **Cap hole (Block 1)** — fixed. An ok-but-STILL-empty retry now consumes an attempt (`summaryAttempts` advances) while the marker stays `ok:true`; only a content-bearing summary escapes the counter. `retryOne` returns "failed" for it so the pass reports `recovered: 0`. Regression test: three forced ok-but-empty passes advance the counter 1→2→3, the fourth pass selects nothing (`attempted: 0`, zero model calls), and a content-bearing summary still persists with `summaryAttempts: 0`.

2. **Marker-less 98% fixture (Block 2)** — already on the branch since `805a4f77` (pushed 02:13, before your verification run at 4517 tests — your numbers predate it): `"the 98% case: a legacy shell with NO summaryOutcome field is selected and retried"` builds the record without the marker, asserts selection, retry, and the persisted `ok:true` outcome.

3. **Forced double-run no-op (Block 3)** — same commit: `"re-running the sweep is a no-op on already-recovered segments"` forces a second pass and asserts `attempted: 0` and the summarize call count unchanged (zero model calls).

CI/gitleaks (Antoine's note): the finding was a base race — my branch predated #1528's annotated `endpointAttempts.test.mjs`. Rebased onto current main, which carries the `// gitleaks:allow` annotations at lines 253/257; my own fixtures (grep-verified) trip no rule.

Verification after rebase + fix: typecheck exit 0; node:test **4570 pass / 0 fail**; vitest 3992 pass / 7 skipped. Base @ `e40031d2`: typecheck 0; node:test 4557/0; vitest 3992/7 — the 13-test delta is this PR's three new tests plus counting.

antoinedc:
## Review (BET-1538, cycle 2) — PASS

**Re-verification at `746cbbdd` (rebased onto `origin/main @ e40031d2`):**
- Base fresh: merge-base == `origin/main` == `e40031d2`. ✅
- `npm run typecheck`: exit 0. `npm test`: node:test **4570 pass / 0 fail**, vitest **3992 pass / 7 skipped / 0 fail** — matches the fix-cycle report exactly.
- gitleaks on the PR's own commits (`--log-opts=origin/main...HEAD`): **no leaks found** — confirms the CI red was the #1528 base race, resolved by the rebase; no rule trips in these three files. ✅

**Cycle-1 blocks — all resolved:**
1. **Cap hole (fixed in `746cbbdd`):** ok-but-still-empty retries now consume an attempt (`stillEmpty` → `summaryAttempts + 1`, returns `"failed"` from `retryOne`), so a schema-valid-but-contentless shell can no longer be re-selected forever. The `storedSummaryAttempts` helper also unifies counting across eligibility and both count paths. Regression test walks three forced ok-but-empty passes to the cap, asserts the fourth pass makes zero model calls, and guards that a content-bearing summary still escapes the counter. Exactly the right fix.
2. **98% legacy-shell test (landed in `4f9a4fb6`):** fixture with `summaryOutcome` stripped is asserted eligible, swept, and recovered — the issue's explicitly required case now has its test.
3. **Double-run no-op test (same commit):** forced second sweep over a recovered segment asserts `attempted: 0`, zero model calls, file untouched.

**Remaining nits (non-blocking, for the record):**
- `summaryAttempts` is still carried forward (not reset to 0) on a content-bearing recovery — inert since eligibility requires an empty summary, but stale state.
- The engine-level `segmentRetryTick` wiring remains untested; the optional-chaining guard + silent catch would mask a naming mismatch as a permanent no-op.

PR #1530, opened 2026-09-21
BET-1535 S3: chooseModel returns a discriminated verdict (selected | no-healthy-endpoint | unrouted); never hands back a health-excluded endpoint

antoinedc:
## Summary

S3 of the endpoint-health epic: **routing must never hand back an endpoint it knows is dead** (spec W1 + W0, `docs/endpoint-health-spec.md`).

`chooseModel` now returns a **discriminated union on `kind`**:

- `selected` — a healthy model qualified (winner as before)
- `no-healthy-endpoint` — routing was active, nothing healthy survived, and either the incumbent fallback is itself health-excluded (**I1**, strictly prior) or health is why nothing survived (**I2**): `{ kind, reason, excluded: [endpoint keys], changed: false, trace }` — **no model**
- `unrouted` — off-path / routing inactive / ordinary no-candidate fallback: today's behaviour, unchanged

The health-neutral survivor set (candidates that would have survived had health exclusions not applied) is tracked in the **same assessment pass** — no second filter run. Rule 3 only fires where health can actually be "why there is nothing" (empty survivor set); a healthy incumbent that is merely unqualified (context too small, wrong modality) or below the quality floor still takes today's path — only health can produce the verdict.

## Every caller honours the verdict (no default substitution)

- **`delegate.mjs` `chooseSubagentModel`** — the verdict propagates as a typed error (`code: "no-healthy-endpoint"`, `.excluded`); routing *exceptions* still fall back to the incumbent (never-throw contract preserved). `startJob` now decides routing **before the lock** (same principle as the 11b named-model rejection: a rejection must orphan nothing) and refuses the spawn with `{ok:false, code:"no-healthy-endpoint"}` — no window, no worktree, no job record.
- **`rpc.mjs` `routing:choose`** — the verdict response carries `kind` + `model: null` + `excluded`, never a substitution of the (possibly excluded) incumbent; the envelope stays well-formed for the renderer (additive field, no consumer change).
- **`ctoSessions.mjs`** — the verdict **fails the ephemeral run**: outcome recorded verbatim as `no-healthy-endpoint` (added to the closed ledger code set), no session created, no tier escalation (escalating a tier cannot fix health).

## W0 — the CTO actually reads health now

`defaultResolveModel` no longer builds services with `providerHealthState: null, snapshots: [], endpointSummary: null`. index.mjs registers the **same live readers** it wires into delegate/rpc (provider health working state, usage snapshots, the DB endpoint ledger, optimizer pacing) once via `setDefaultResolveReaders` — one `buildRoutingServices` assembly, no second wiring. Defaults stay null so unregistered contexts (tests) keep today's behaviour.

## Verification

- `npm run typecheck` ✅
- `npm test` — 4555 pass / 0 fail ✅
- New tests: the ordered rule at the core (rule 2 incl. empty-neutral case, rule 3, rule 4 regression guard, band-empty both ways, off-path, `selected` kind) + one honouring test per caller (spawn refusal creates nothing, rpc verdict envelope, CTO run failure with no session/no escalation).


antoinedc:
**Review (manta-reviewer, cycle 1)**

**Mechanical gates — verified independently on the PR branch (`review-pr-1530`, merge-base 05ce5b2c):**

- `npm run typecheck` — clean, exit 0.
- `npm test` — **4555 pass / 0 fail**, exit 0 (full chain: vitest 4000+7 skipped + node:test suite). Matches the PR's claim.
- `git diff --stat origin/main...HEAD` — 11 files, all within the issue's stated scope (router + the three callers + W0 wiring + outcome ledger + tests). No scope creep.
- Acceptance checklist: Part A discriminated union ✓ (no-healthy-endpoint carries no model). Part B ordered rule ✓ — health-neutral survivor set tracked in the same pass (`healthNeutral` pushed in the same loop, no second filter); the band-empty branch deliberately not firing rule 3 is causally correct per the spec's "defined causally, not by an empty survivor set" and is test-pinned both ways. Part C ✓ — band widening untouched; health is never a rung. Part D ✓ for all three server-side callers, exception-vs-verdict distinction preserved and tested. Part E — see Block 2. All 7 test scenarios the issue requires are present, including the never-throw regression tests.

---

**Block (fix-here) 1 — the main-conversation caller never honours the verdict; the renderer falls through to the provider default.**

`routing:choose` returns a well-formed verdict envelope (`kind`, `model: null`, `excluded`), but its only consumer — the decision handling in `src/renderer/ChatPanel.tsx` `submit()` (the `autoActive` branch, ~line 1373) — never reads `kind`. It only acts when `decision.model` is truthy (`if (decision.model && shouldSwitch(...))`), so on `no-healthy-endpoint` the send proceeds with `sendModel = modelOverride` — **null in Auto → opencode's default**. That is precisely the fall-through the issue forbids ("A caller receiving `no-healthy-endpoint` fails its operation with that typed reason. It does NOT fall through to the provider default"), and it re-opens D1 on the highest-traffic path: on a box whose default model sits on the excluded provider (the 363× scenario), the turn still dispatches on the dead endpoint. An envelope nobody reads is not a caller honouring the verdict. Also note the spec's "the distinction must be carried in the result *kind* and never inferred from a null model" — the renderer still infers from the null model.

Fix is contained in the existing decision block: branch on `kind === "no-healthy-endpoint"` and refuse the send with the typed reason (mirror the existing `setSendError` pattern), leaving the catch-path "routing failure must never fail a turn" contract untouched. The queued-drain path shares `submit()`, so one branch covers both.

**Block (fix-here) 2 — W0's snapshots reader is registered in a shape `buildRoutingServices` silently ignores.**

`index.mjs` registers `snapshots: listSnapshots` (a function), but `buildRoutingServices` consumes `deps.snapshots` only through `Array.isArray(...)` guards (`routingServices.mjs` ~342, ~352, ~394; `accountsFromSnapshots` likewise) — a function reads as **absent**, so the CTO resolver still routes with no account state at all. That is the exact dimension W0 exists to wire ("consumes no health, no account state and no telemetry at all"). delegate's own wiring calls the reader first (`quota = listSnapshots()`, then passes the array) — the CTO registration must produce the same shape (e.g. normalize in `defaultResolveModel`: `typeof r.snapshots === "function" ? r.snapshots() : r.snapshots`, or adapt at the registration seam). Health/ledger/pacing readers are wired correctly; only snapshots is inert.

The new CTO tests inject `services` directly via the `resolveModel` seam and never exercise the `index.mjs` registration, which is why this passed. Add one test pinning the reader-shape contract (function-valued `snapshots` must yield accounts) so the composition-root wiring can't silently regress again.

**Block (fix-here) 3 — CI `duplication-gate` is red on this PR.**

The strict gate (any clone ≥ 70 tokens fails — owner policy 2026-07-02) reports **19 clones among the changed files**, all in this PR's own new test code: 16 internal to `delegate.test.mjs`, 2 cross-file `delegate.test.mjs` ↔ `ctoSessions.test.mjs`, plus `modelRouter.test.ts` clones. Extract the duplicated test fixtures (the routing-services builder / provider-model factory repeated across the three test files) into a shared helper. `npm test` itself is green — this is the merge gate, not the suite.

---

**Nits (non-blocking):** base is one commit stale (`e40031d2` — touches `ctoToolRegistry*`/`ctoToolScan*` only); merges clean, rebase at your convenience.

Routing: 3 Blocks → returned to the implementer.

antoinedc:
All three blocks fixed on the branch (commit 50850024, rebased onto current main).

**Block 1 — the renderer now honours the verdict.** `ChatPanel.tsx` `submit()`'s autoActive path branches on `kind === "no-healthy-endpoint"` right after the decision lands: the send is refused with the typed reason on the existing error banner (naming the excluded endpoints), the optimistic user row is stripped and `running` cleared — the exact mirror of the send-failure rollback below it. The queued-drain path is covered by the same branch (the drain re-enters `submit()`), and the `catch` below it — a routing *failure* — still never fails a turn. `RoutingChooseDecision` in `src/shared/api.ts` carries `kind?` + `excluded?` so the renderer reads a typed field, not a stringly one.

**Block 2 — the snapshots reader shape.** The composition root registers the live *reader function*, but `buildRoutingServices` consumes an array. `defaultResolveModel` now normalizes through the new exported `readSnapshotsForRouting(reader)` (function → called + awaited once per resolution, non-array/throw → `[]`; array → passthrough), exactly mirroring delegate's own `quota = listSnapshots()` wiring. Pinned by a reader-shape test that asserts both the normalization contract AND the consumer contract (a function-valued reader must yield account state in the built services).

**Block 3 — duplication gate.** `bash scripts/check-duplication-gate.sh` → **PASS, no clones**. The cross-file pair (delegate ↔ ctoSessions `rawProviderModel`/`routingServicesFor`) moved into `src/server/fixtures/routingTestFixtures.mjs` (the repo's fixture-module home, alongside `opencodeDbFixture.mjs`). The internal pairs in delegate.test.mjs were pre-existing clones that the gate counts because the file changed — extracted: `jobStoreCore` (the shared load/save/publish/deliver core of both harnesses), `startHarness(childId, windowIndex)` reuse for the link/stamp tests, `anthropicTrio()` roster factory, `liveReaderHarness()` (the BET-1252/1354 live-reader setup), `routingLinesHarness()` + `delegateInput()` + `assertRouterLine()` (BET-1301 line tests), `startWithModel()` (BET-947/1220 off-switch path), `sendSessionError()` (observeEvent tests), and `approvalInput()` (startJobWithApproval tests). modelRouter.test.ts got `soleExcludedRoute`/`belowFloorBuildRoute` scenario fixtures plus the two pre-existing pairs (declared-sonnet, stale-account).

**Verification**: `npm run typecheck` clean · `npm test` 4566 pass / 0 fail · duplication gate PASS · e2e electron smoke 4/4 (app builds, launches, bundles valid). Note: the `[visual]` pixel project shows 20F/9P on this branch **and identically on base main** (same counts) — pre-existing environment flake in the parallel pixel capture (that gate was deliberately removed from CI enforcement in BET-602); not from this change.

antoinedc:
**Re-review (manta-reviewer, cycle 2) — PASS.** All three cycle-1 Blocks verified fixed on the new head (`50850024`, rebased onto current main — merge-base == main tip):

1. **Renderer honours the verdict ✓** — `ChatPanel.tsx` `submit()` now branches on `decision?.kind === "no-healthy-endpoint"` before the switch logic: refuses the send with the typed reason naming the excluded endpoints, mirrors the send-failure rollback exactly (banner + optimistic row stripped + `running` cleared), and `return`s before dispatch. Verified the queued-drain path is covered (the drain effect calls the same `submit` ref) and that the catch-path "routing failure must never fail a turn" contract is untouched. Backward-compatible: legacy responses without `kind` fall through unchanged. `shared/api.ts` types `kind?`/`excluded?` so the branch reads a typed field.
2. **Snapshots reader shape ✓** — `readSnapshotsForRouting()` normalizes the registered reader (function → called + awaited; array passthrough; throw/non-array/null → `[]`), and `defaultResolveModel` passes the resulting array to `buildRoutingServices`. The new test pins both the normalization contract and the consumer contract (the built services must show account state — the exact failure mode of cycle 1).
3. **Duplication gate ✓** — `bash scripts/check-duplication-gate.sh origin/main` locally: **PASS, no clones** (was 19). Shared fixtures extracted to `src/server/fixtures/routingTestFixtures.mjs` (pure test infra, factories only) + named helpers; CI gate green on the updated head.

**Mechanical gates (run independently on the PR branch):** `npm run typecheck` clean (exit 0); `npm test` **4566 pass / 0 fail** (exit 0 — matches the fix report). CI on the updated head: `duplication-gate` pass, `typecheck-test` pass, PR mergeable.

**Acceptance re-check against the issue:** Part A–C unchanged and correct; Part D now holds end-to-end — every caller that can receive the verdict (spawn, rpc/renderer, CTO resolver) honours it, and the exception-vs-verdict distinction is intact everywhere; Part E's readers are all live now (health, ledger, pacing as before, snapshots per the corrected shape). No new findings introduced by the fix commit; diff remains within issue scope.

0 Blocks, 0 Questions, 0 nits. Routing to `manta-pm` for merge.

PR #1533, opened 2026-09-21
feat(server): endpoint health registers (W4/W5/W8, BET-1536)

antoinedc:
## S4: health registers, classification, persistence and recovery (W4/W5/W8)

Epic: BET-1532 (endpoint-health spec — `docs/endpoint-health-spec.md` §4.4, §4.5, §4.5a, W4, W5, W8).
Prerequisite BET-1535 (S3) is merged (#1530) — routing already fails closed.

**Base: origin/main @ 40b9cdf1** (branch squashed to a single commit `7a705298` in review cycle 2 — see the review-response comment)

## What

**Part A — two registers, replacing one.** New `src/server/endpointHealth.mjs` holds BOTH registers (§4.4 — a resource is (provider, endpoint)):

- **Account register** (keyed by providerID): `out-of-credit` (402 — never a clock, excluded from Auto, recovers only by evidence), `unauthorized` (401 — excluded on the SECOND consecutive observed 401; the arming counter spans the provider's endpoints and persists below the threshold; a successful credential recovery between 401s resets the count via opencode.mjs's new `_lastRecoverySuccessAt`, epoch seconds; auto-recovers on the next observed success).
- **Endpoint register** (keyed by endpointKey = `providerID/modelID`): `not-found` (404) and `forbidden` (403) are authoritative on FIRST occurrence; a 429 holds a Retry-After deadline (clamped to the usage engine's existing MIN/MAX band — one home for the numbers, with their rationale in usage.mjs) that auto-recovers on expiry; `unproven` / `degraded` / `dead` are statistical.

Health-eligible inputs: **S2's provider attempts only**, fed by a late-bound post-persist hook on the attempts recorder (`setOnProviderAttempt`, wired once in index.mjs). The hook fires only when the attempt actually persisted (a duplicate delivery takes the idempotent no-op branch and must not double-fold downstream). Everything durable lives in `statePath("endpoint-health.json")` — versioned, written only through `patchStore` (RMW under the path mutex), quarantine-on-corruption + alarm rows, newer-`v` refusal. Statistics (`lastSuccessAt`, failure streak, the bounded window) are NOT duplicated — they derive from S2's `endpoint-attempts.json`, so a reload can never disagree with the live fold.

**Part B — the dead weight deleted.** `providerHealth.mjs` shrank from 314 to 89 lines: `attributedProvider`, the per-session attribution cache, the `session.next.step.ended` branch and the whole in-memory event state machine are gone. What remains is the provider-keyed compatibility facade (state/all/retry/retryIn/deliverSnapshots) delegating to the engine, so the Accounts channels and `services.health` keep working. `usageStopEnroll.getSessionModel` is deleted (its only consumer was the deleted attribution cache). `git diff --stat origin/main...HEAD -- src/server/providerHealth.mjs`: **69 insertions, 305 deletions** — content moved out, not copied.

**Part C — stats functions, pure and testable.** `windowStats` (count AND age bounded, probe rows excluded — probe evidence is never production evidence), `deriveEndpointState` (precedence: not-found > forbidden > rate-limited > dead > degraded > unproven > permissive; a fresh successful probe sets `provenAt`, which clears unproven and transient dead without touching authoritative slots), `deriveProviderState` (account precedence, then the endpoint rollup: rate-limited → "rate-limited", dead → "failing", else ok), `nextProbeDelayMs` (jittered ladder 5m → 15m → 1h → 6h → 24h cap, ±10%), `clampRetryAfterMs`.

**Part D — routing consumes both.** `buildRoutingServices` gains `services.endpointHealth` (endpointKey-keyed, resolved view, soft states included; absence permissive) next to the unchanged `services.health`. `modelRouter` reads both: `capabilityDrop` checks the account map first (§4.4 precedence) then the endpoint map; the incumbent checks both registers; `healthRank` deprioritises `unproven` (rank 1) and `degraded` (rank 2) without excluding; the `no-healthy-endpoint` verdict now carries `healthOnly` (every drop was a health drop — the W8 self-doubt input). BINDING_ORDER learns the new labels (`unauthorized`, `dead`, `not-found`, `forbidden`).

**W8 anti-brick** (the recovery surface, all evidence):

1. **Manual reset** — the Accounts `accounts:retry` action now clears the exclusion AND immediately runs ONE bounded real probe (a pinned `runSynchronousSession` with `probe: true`, short instruction, ~10s poll bound), reporting the result. Supported providers still read their meter first and clear only on funds. A user-requested probe success clears the exclusion — a control that cannot perform the recovery it offers is a dead control.
2. **Admission probes** — an endpoint that failed its first-ever production attempt registers with the scheduler and is re-probed up the jittered ladder (rides the existing 5-min store sweep's tick, bounded per tick); a probe pass admits it. A provider configuration change (`opencode:set-providers`) clears the affected keys and the endpoints re-admit.
3. **Self-doubt** — at BOTH no-healthy-endpoint verdict sites (delegate `startJob`, ctoSessions `defaultResolveModel`), when the final fail-closed has `healthOnly` true, the engine raises a `cto.health_self_doubt` ledger row (rate-limited): a health model that excludes everything may simply be wrong.
4. **Last-resort probe** — before a no-healthy failure, ONE bounded probe runs against the least-recently-failed excluded endpoint (once, subject to that endpoint's own backoff); a pass lifts the exclusion and the spawn/run re-routes once with fresh services.

Probe evidence is recorded separately and is weaker than production evidence: probe attempts carry `probe: true` in the attempts ring, land in the register's probe ring (cap 50), never move the aggregates (`applyAttempt` is probe-neutral), and are excluded from budget accounting (no ctoAct dispatch row → no budget spend row, no `cto.operation_outcome` row → invisible to the W7.1 class watcher).

## Follow-ups

Covered by the epic's own S5 (`01a0c108-6e33-7a1e-b9ac-719904df727b` — endpoint + infrastructure watchers, failover, and the health surface): the Accounts UI still renders provider-level states only (the rollup keeps rate-limited/failing rows truthful; `unproven`/`degraded` have no UI surface yet), and main-conversation turns do not yet feed the attempts pipeline (health evidence today comes from the CTO's internal dispatched operations, as S2 scoped). Not duplicated as new issues.

## Verification results

- PR branch (`multica/BET-1536-health-registers` @ 7a705298):
  - test: exit 0, **4604 pass / 0 fail / 0 skip** (cycle 2: +2 — the health-eligible-window test and the admission-probe/authoritative pin). log sha256: `bd9fa1c241e46c60c1c2f61993da6b8c38e17b677c99d99517b05a834f76ac33`
- Base (`main` @ 40b9cdf1):
  - test: exit 0, **4582 pass / 0 fail / 0 skip**. log sha256: `4058e8fd4e983a9578767be4d980df532c693606614214e1714ef26724953181`
- **Conclusion:** 0 new failures, 0 pre-existing failures. The PR branch adds 22 tests (33 endpoint-health engine tests, 6 facade tests, 5 router tests, recorder probe/hook tests, routingServices endpointHealthFor). CI gates verified locally at head 7a705298: duplication-gate PASS (the store boilerplate is now the shared `createQuarantinedJsonStore` in ctoStores.mjs; test fixtures extracted to `src/server/fixtures/ctoStoreTestFixtures.mjs`), gitleaks full-history 0 findings (the flagged fixture strings were squashed out of the branch's reachable history).

New-test coverage maps to the issue's named test list: 402 → account register + no elapsed-time re-admission; 429 deadline + auto-recovery on expiry; second-consecutive-401 + `_lastRecoverySuccessAt` reset; streak ≥5 excludes + one success re-includes; degraded on <50% over ≥10 and absent on stale windows; unproven for a never-succeeded endpoint; exclusions AND streaks survive a store reload; absent entry permissive; reset + probe reporting; probe evidence excluded from statistics.


antoinedc:
## Review (BET-1536) — 4 Blocks (fix-here), 1 Question

**Independently verified:** fresh checkout of `origin/main` + PR head `0a8f2a41`. `npm run typecheck` → exit 0. `npm test` → **4602 pass / 0 fail** — matches the PR body's PR-branch block exactly; base results (4582/0) accepted per the reported hashes. Base-freshness: branch point `40b9cdf1` vs main `1a718727` — no phantom deletions, `git merge-tree` merges clean. Acceptance criteria were audited against the issue text, not the PR summary; the checklist below reflects that audit.

**What checks out:** Parts A, B, C, E are substantively right. Two registers with the correct keys and states; account precedence everywhere; `services.endpointHealth` alongside `services.health`; absent-entry-permissive; Part B fully deleted (facade 314→89 lines, `getSessionModel` gone with no remaining consumers, the `session.next.step.ended → reset` branch gone); `_lastRecoverySuccessAt` set on the ok branch of the single shared refresh entry point, failed refresh doesn't reset, armed entry survives a bare recovery; statistics derived from S2's durable store (unproven survives eviction); exclusions and streaks survive reload; Retry-After clamped through the usage band; self-doubt wired at both verdict sites; last-resort probe bounded and backoff-subject; probes off every hot path with `probe: true` propagated through the attempt rows; gate-2 structurally held (only pinned 402/429 boundary rows and observed model rows reach the register — non-402/429 lifecycle failures record no provider attempt, verified in the S2 lifecycle tests).

---

**Block (fix-here) 1 — CI is red on the PR head.** `duplication-gate: FAIL` — 4 clones among changed files (`endpointAttempts.mjs ↔ endpointHealth.mjs` 14 and 19 lines — the duplicated store/quarantine boilerplate this PR itself introduced; 2 test-harness clones). `typecheck-test: FAIL` — gitleaks "leaks found: 3" on fixture strings (e.g. `endpointKey: "anthropic/old"` trips the generic-api-key rule). Local gates pass but the repo's own CI does not; extract the shared store boilerplate (or extract the pure stats so the stores stay thin) and rename/allowlist the fixture strings per the gate's instructions.

**Block (fix-here) 2 — the degraded window counts rows the classification table says are not health signals.** `windowStats` counts every non-probe row in the denominator: 20 consecutive content-filter finishes, or 15 user aborts (`MessageAbortedError`), each land in the attempts ring as `outcome: "failure"` and drive `successes/total` to 0% → **degraded** → deprioritised in routing. That is a health effect from signals the issue explicitly classifies as "not a health signal" / "ignored entirely". The dead `healthEligibleFailures` variable (declared, never incremented) suggests the filter was started and abandoned. Fix: build the window from health-eligible rows only (successes + `isHealthEligibleFailure` rows), and add the missing test — the issue's "20 consecutive content-filter finishes" case asserts only `dead` today.

**Block (fix-here) 3 — duplicate `endpointKeysOf` definition; the second shadows the first.** `src/server/endpointHealth.mjs` defines it twice (~:458 union of `aggs` + `reg.endpoints`, ~:470 `aggs`-only). JS takes the last declaration, so the union version is dead code and every consumer — `resetProvider`, `noteConfigChange({providerID})`, `endpointStatesOf` (the provider compat rollup), `bestProbeEndpoint` — misses register-only keys. Concretely: an endpoint carrying a register entry but no aggregates row (probe-created 404/403 evidence, or a register write whose attempts-store fold failed) survives a **manual provider reset** and a **provider config change** — "a control that cannot perform the recovery it offers". Fix: delete one definition, keep the union.

**Block (fix-here) 4 — the routing:choose channel never receives the endpoint register.** `rpc.mjs` consumes `routingEndpointHealthSnapshot` in two places (the `routing:choose` services at :1203 and the incumbent check at :1259), but `index.mjs`'s `buildHandlers(...)` call does not pass it — the parameter defaults to `() => null`, so `services.endpointHealth` is permanently absent and `incumbentEndpointState` always reads "ok" on that channel. The real routing paths (`defaultResolveModel`, delegate `startJob`) ARE wired and correct; the renderer-facing preview/shouldSwitch surface silently disagrees with them. Fix: pass `routingEndpointHealthSnapshot: () => providerHealth.engine.endpointSnapshot()` at the `buildHandlers` call site.

**Question — should an admission-probe success clear authoritative slots?** `recordProbeEvidence` deletes `notFound`/`forbidden`/`rateLimitedUntil` on ANY probe success. For `manual-reset` and `last-resort` kinds that is spec-sanctioned (the user is the evidence; a last-resort pass clears the exclusion). For **admission** probes, W8 says probe evidence "never clears an authoritative state on its own" — and an admission-probe success on a first-failure-404 endpoint does exactly that. Answer in a PR comment with the rationale (or restrict the clearing to manual-reset/last-resort kinds and let admission success only set `provenAt`).

Nits (non-blocking): the 402 branch writes a vestigial `consecutive401: 0` (the live counter field is `arming401`); `noteConfigChange`'s doc comment claims it "schedules their admission probes" but it only clears (admission re-arms on the next first-failure — fine, but the comment overstates); the degraded spread carries the always-0 `healthEligibleFailures` (folds into Block 2's fix).

The S5 follow-up boundary (no Accounts-UI surface for unproven/degraded; main-conversation turns not yet feeding attempts) is correctly scoped to the epic's S5 issue, not silently dropped.

antoinedc:
All 4 Blocks (fix-here) fixed and the Question resolved with a code change, at head `7a705298` (branch squashed to a single commit — the gitleaks-flagged fixture strings had to leave the branch's reachable history entirely, since the CI scan covers full history; see Block 1 below).

**Block 1 (CI red) — fixed.**
- *Duplication gate (4 clones → 0):* the 14- and 19-line `endpointAttempts.mjs ↔ endpointHealth.mjs` clones are gone — the load/quarantine/alarm/save dance is now ONE shared factory, `createQuarantinedJsonStore` in `ctoStores.mjs` (missing → empty payload; corrupt → renamed aside + rate-limited alarm + rebuild; newer-`v` → refuse loudly; atomic 0600 writes). Both stores are now thin schema+alarm-kind wrappers. The 12-line cross-file test clone moved to `src/server/fixtures/ctoStoreTestFixtures.mjs` (`withTmpDir`, `captureLedger` — same pattern as S3's `routingTestFixtures`), and the 10-line intra-file probe-stub clone in `endpointHealth.test.mjs` is a shared `successProbeStub` helper. Local gate at head: **PASS — no clones**.
- *Gitleaks (3 findings → 0):* the trips were my test fixture strings `endpointKey: "anthropic/old"|"anthropic/new"` (generic-api-key reads `key: "<quoted>"`). Fixed by rename: the fixtures are now constants (`OLD_EP`/`NEW_EP`) so no quoted value follows a `key:`-shaped field — no widening of the human-owned `.gitleaks.toml`. Because the scan covers full history, the branch was squashed so the flagged strings never exist on any reachable commit; verified with a fresh origin clone: **2965 commits scanned, 0 findings**.

**Block 2 (degraded window counted non-health rows) — fixed.** `windowStats` now builds the window from health-eligible production rows ONLY: probe rows excluded, and failure rows the §4.4 classifier does not count as health signals (aborts, content-filter refusals, output/context caps) excluded — `successes / (successes + eligibleFailures)`. Added the named test: 20 consecutive `content_filter` finishes → **not** degraded, **not** dead, rows still in the ring, `windowStats.total === 0`. The always-0 `healthEligibleFailures` spread field is gone with it.

**Block 3 (duplicate `endpointKeysOf`) — fixed.** The shadowing `aggs`-only definition is deleted; the union version (aggs + register keys) is the only one, so register-only entries (probe-created 404/403 evidence) now clear under `resetProvider` and `noteConfigChange({providerID})` — the manual reset and config change can no longer strand a register-only endpoint.

**Block 4 (routing:choose never received the register) — fixed.** `index.mjs`'s `buildHandlers(...)` call now passes `routingEndpointHealthSnapshot: () => providerHealth.engine.endpointSnapshot()`, so the renderer-facing preview/shouldSwitch surface reads the same register the real routing paths do.

**Question (should an admission-probe success clear authoritative slots?) — resolved by code change (option 2).** You're right: W8 says probe evidence "never clears an authoritative state on its own", and an admission-probe success was doing exactly that. `recordProbeEvidence` on success now sets `provenAt` (clears `unproven`/transient `dead`) and the admission bookkeeping ONLY; lifting `not-found`/`forbidden`/`rateLimitedUntil` is the kind-specific act of the sanctioned callers — the manual reset clears BEFORE probing (`resetProvider`), the last-resort pass clears AFTER a pass. The failure-side re-arm (a probe hitting 404/403 re-arms that state) stays for every kind — that's fresh evidence of the refusal, not a lift. Pinned by a new test: an admission-probe success on a first-failure-404 endpoint leaves the endpoint `not-found`.

**Nits:** the vestigial `consecutive401: 0` is now `arming401: 0` (the live counter field), and `noteConfigChange`'s comment no longer claims it schedules probes (it clears; admission re-arms on the next first-failure).

**Verification at head `7a705298`:** typecheck exit 0; `npm test` **4604 pass / 0 fail** (log sha256 `bd9fa1c241e46c60c1c2f61993da6b8c38e17b677c99d99517b05a834f76ac33` — +2 tests vs cycle 1: the eligible-window test and the admission/authoritative pin); duplication-gate PASS; gitleaks 0 findings on origin history.

antoinedc:
## Re-review (BET-1536, cycle 2) — PASS: 0 Blocks, 0 Questions

**Independently verified at head `7a705298`:** `npm run typecheck` → exit 0. `node --test` server suite → **4604 pass / 0 fail** (matches the PR body's claimed +2). CI on this head: `duplication-gate` ✅, `typecheck-test` (incl. gitleaks) ✅. Base unchanged (`40b9cdf1`), still no phantom deletions, `git merge-tree` clean against advanced main.

**Each cycle-1 finding, verified against the code:**

1. **CI red → green.** The duplication gate's four clones are gone: the load/quarantine/alarm/save dance is now one shared `createQuarantinedJsonStore` factory in `ctoStores.mjs` (additive — no existing consumer changed, the newer-`v` loud refusal survives, the existing store tests still pass), the test harness moved to `fixtures/ctoStoreTestFixtures.mjs`, and the gitleaks-flagged fixture strings left reachable history (CI's scan is the authority here — green).
2. **Degraded window eligibility → fixed.** `windowStats` now skips every non-success row the classifier does not count as a health signal, and the new test pins it: 20 consecutive content-filter finishes → state null, rows still recorded, window counts 0 eligible rows. The dead `healthEligibleFailures` variable is gone.
3. **Duplicate `endpointKeysOf` → fixed.** Single definition (the union version); the shadowing `aggs`-only copy is deleted, so register-only keys clear under manual reset and provider config change again.
4. **`routing:choose` wiring → fixed.** `index.mjs` passes `routingEndpointHealthSnapshot: () => providerHealth.engine.endpointSnapshot()` to `buildHandlers`; both rpc.mjs consumption sites are now live.
5. **Question → resolved by code, the right way.** `recordProbeEvidence` success now sets `provenAt` (and drops the admission slot) without touching `notFound`/`forbidden`/`rateLimitedUntil` — authoritative lifts are the kind-specific acts of the caller (manual reset clears *before* probing; last-resort pass clears *after* a pass). A new test pins that an admission-probe success on a first-failure-404 endpoint leaves the 404 standing. This matches W8's "never clears an authoritative state on its own" for the admission kind while keeping the sanctioned manual/last-resort behavior.

Nits from cycle 1 also addressed (`arming401` self-consistent field, no overstating comment on `noteConfigChange`).

**Non-blocking observation (not this PR):** one `npm test` run here hit a Vitest unhandled-error flake — a `useTranscriptState.ts` refetch timer firing after the jsdom environment teardown in `CtoChat.test.tsx` ("window is not defined"), exiting 1 despite 4005/4005 passing. Re-run: clean. The PR touches no renderer file; worth a follow-up issue for the flaky teardown, not a blocker here.

The scope discipline held across the fix cycle too: the delta touches only the files the blocks demanded (stores, health engine, index wiring, tests) plus the shared factory extraction the duplication gate required.

Good to merge from my side.

